PR #58, opened 2019-06-06
Rework validation to improve submit of custom components

barrymay:
Hi @bluebill1049 - please take a look at this if you have time (still in draft as I would like feedback).   This branch fixed some remaining issues I had with trying to make my custom component with custom validation work.    One notable change (you'll see in the unit tests) is that if a validation fails, I would like it to return an empty object, so that isEmptyObject handles it properly. 

Thoughts?

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -50,7 +50,7 @@ export default function useForm<Data extends DataType>(
   const reRenderForm = useState({})[1];
   const validateAndStateUpdateRef = useRef<Function>();
 
-  const renderBaseOnError = (errors, error) => {
+  const renderBaseOnError = (errors, error, name) => {
     if (errors[name] && !error[name]) {
       delete errorsRef.current[name];
       reRenderForm({});
@@ -66,7 +66,7 @@ export default function useForm<Data extends DataType>(
     const errors = errorsRef.current;
 
     errorsRef.current = { ...filterUndefinedErrors(errorsRef.current), ...error };
-    renderBaseOnError(errors, error);
+    renderBaseOnError(errors, error, name);
     return error[name];
   };
 
@@ -123,7 +123,7 @@ export default function useForm<Data extends DataType>(
 
           if (shouldUpdate || shouldUpdateValidateMode || shouldUpdateWatchMode) {
             errorsRef.current = { ...filterUndefinedErrors(errors), ...error };
-            renderBaseOnError(errorsRef.current, error);
+            renderBaseOnError(errorsRef.current, error, name);
           }
         }
 
--- a/src/logic/validateField.ts
+++ b/src/logic/validateField.ts
@@ -62,8 +62,10 @@ export default async <Data>(
       exceedMax = maxValue && valueNumber > maxValue;
       exceedMin = minValue && valueNumber < minValue;
     } else if (DATE_INPUTS.includes(type)) {
-      if (typeof maxValue === 'string') exceedMax = maxValue && new Date(value) > new Date(maxValue);
-      if (typeof minValue === 'string') exceedMin = minValue && new Date(value) < new Date(minValue);
+      if (typeof maxValue === 'string')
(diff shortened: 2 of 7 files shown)


src/logic/validateField.ts
```diff
@@ -62,8 +62,10 @@ export default async <Data>(
       exceedMax = maxValue && valueNumber > maxValue;
       exceedMin = minValue && valueNumber < minValue;
     } else if (DATE_INPUTS.includes(type)) {
-      if (typeof maxValue === 'string') exceedMax = maxValue && new Date(value) > new Date(maxValue);
```
barry4dev: Most of the changes in this file were due to prettier (set up with the new .pretterrc file added)

bluebill1049:
Thanks @barry4dev again 🙏🏻😱❤️

src/index.ts
```diff
@@ -50,7 +50,7 @@ export default function useForm<Data extends DataType>(
   const reRenderForm = useState({})[1];
   const validateAndStateUpdateRef = useRef<Function>();
 
-  const renderBaseOnError = (errors, error) => {
+  const renderBaseOnError = (errors, error, name) => {
```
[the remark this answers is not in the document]
barry4dev: Sure - but I don't see this in your latest version of renderBaseOnError - did you check it in?  (this why the PR doesn't show it either)

src/logic/validateField.ts
```diff
@@ -114,20 +124,30 @@ export default async <Data>(
 
     if (typeof validate === 'function') {
       const result = await validate(fieldValue);
-      if (typeof result !== 'boolean' || !result) {
+      if (typeof result === 'string' && !!result) {
```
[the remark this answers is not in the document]
barry4dev: yeah that change is fine - sometimes I do the !! for typescript's sake, because leaving it there makes that result a string, whereas with the !! it coerces it into a boolean.  In this case it doesn't matter so I can remove.

src/index.ts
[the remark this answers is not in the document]
bluebill1049: ur right :) I forgot to push, during rebase just take yours 👍

src/logic/validateField.ts
```diff
@@ -114,20 +124,30 @@ export default async <Data>(
...
-        return error;
+      } else if (typeof result === 'boolean' && !result) {
+        error[name] = {
+          ...error[name],
+          type: 'validate',
+          message: '',
+          ref: validateRef,
+        };
+      } else {
+        delete error[name];
```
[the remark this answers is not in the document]
barry4dev: Totally right - Nice catch!

src/logic/validateField.ts
```diff
@@ -114,20 +124,28 @@ export default async <Data>(
...
       const result = await validate(fieldValue);
-      if (typeof result !== 'boolean' || !result) {
+      if (typeof result === 'string' && result) {
         error[name] = {
           ...error[name],
           type: 'validate',
-          message: isString(result) ? result : '',
+          message: result,
+          ref: validateRef,
+        };
```
bluebill1049: `return error;`

src/logic/validateField.ts
```diff
@@ -114,20 +124,28 @@ export default async <Data>(
...
+      } else if (typeof result === 'boolean' && !result) {
+        error[name] = {
+          ...error[name],
+          type: 'validate',
+          message: '',
           ref: validateRef,
```
bluebill1049: `return error;` here too

bluebill1049:
thanks very much for the effort! merging I will fix those two comments myself ❤️

src/logic/validateField.ts
barry4dev: Yeah I think that was from prior - you can probably get rid of all the other return errors; in this method if you want to be consistent - just need to watch from side effects from other paths, or improve tests ;)

bluebill1049: yea I need to write more unit tests, that's for sure :)

PR #60, opened 2019-06-06
Improve validation type

barrymay:
One more based on the new trigger function.  This will allow the trigger to trigger a validate in a consistent way that the addEventListener does - currently the trigger always validates, whereas for other fields validates don't happen until the first submit

Also improvements to a few types so that custom fields can pass any data around


dabielf:
Yes please yes!

src/index.ts
```diff
@@ -233,15 +242,15 @@ export default function useForm<Data extends DataType>(
   }
 
   function watch(
-    filedNames?: string | string[] | undefined,
+    fieldNames?: string | string[] | undefined,
```
bluebill1049: oops 👍

src/types.ts
```diff
@@ -1,14 +1,15 @@
 import * as React from 'react';
 
-export type Validate = (data: string | number) => string | boolean;
+export type FieldValue = any;
```
bluebill1049: 👍

bluebill1049:
Thanks, @barry4dev for another great PR! ❤️  

In terms of `trigger` only triggers after clicking on Submit, if you set the `mode` to `onChange` or `onBlur` the validation trigger at change or blur level. I am curious about what others think about `trigger` behavior, whether it should `trigger` with the side effect submitted value.

In my head, I think the trigger is a manual thing, so it should trigger a validation without any other dependency.

@dabielf what're your thoughts as well since you are in this PR?

barry4dev:
No problem - the reason I did this was that I still need a mechanism for the custom component to say “I’m ready, please validate if appropriate.”  Without doing this I don’t have this.  Maybe this should be a hook off of setValue?

bluebill1049:
I think that’s more like using ‘setValue’ and ‘setError’. Any thoughts?

barry4dev:
Well it’s not a setError since that would assume I’m not using the validate check. But if you think it is better I can refactor the setValue to do this auto-validate.  Personally I think the trigger() doing this isn’t bad but I could to either way

bluebill1049:
Oh don’t change then, leave as it is 👍🏻

src/index.ts
```diff
@@ -59,12 +59,21 @@ export default function useForm<Data extends DataType>(
...
+  }
+
   const trigger = async <Name extends keyof Data>(name: Name): Promise<boolean> => {
     const field = fieldsRef.current[name]!;
     if (!field) return false;
-    const error = await validateField(field, fieldsRef.current);
+    
     const errors = errorsRef.current;
 
+    if (!isValidateEnabled()) return isEmptyObject(errors)
```
bluebill1049: @barry4dev I guess this should always return `true/false` before submit? or should just return `undefined`? any thoughts?

bluebill1049: I may add `forceValidation` as the second argument :)

barry4dev: Since trigger is technically a validate check I think return true false is best, which this code does.

Since it's still 'ok' prior to submit then I think validation doesn't fail.

That being said, I totally support the forceValidation second arg and can see where it's helpful

I could also support a name change to checkValidation() or something like that but thats a lesser concern.  I find trigger too generic a word (what are you triggering?) 😉

bluebill1049: valid point :) hows `triggerValidation`?

barry4dev: Perfect 😊

bluebill1049:
```
triggerValidation({
  name: string,
  value?: string,
  forceValidate?: boolean,
})
```

barry4dev: Cool -why the value?

barry4dev: Nvm I see it - values great - saves the extra set call.

bluebill1049: for a custom component like `react-select` where value can't be read by `ref`

```
handleChange(e) {
  triggerValidation({
    name: 'reactSelect',
    value: e.target.value,
  })
}
```

right now, at work, my colleges have to do setValue and setError manually

PR #62, opened 2019-06-07
Reduce typing overhead

barrymay:
First, apologies for the diff noise on this - majority of the diff is prettier since I added the formatonsave function for vscode so it auto-prettier's the update (hope you're using vscode! ;)) Definitely can be taken out if you like.  If you keep it in, the diffs definitely reduce over time as you keep using this flag!

The major change is related to types.ts - it's to avoid going forward the double typing you ran into with types.ts - it also makes the types in index a lot more consistent with what you want.

No functionality change here - just typing improvement.

BTW, VSCode's Organize Imports method is also a great IDE feature to remove unused imports. (used here with index.ts)

Check it out and ask any questions :)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,43 +1,39 @@
-import { useRef, useState, useEffect } from 'react';
+import { useEffect, useRef, useState } from 'react';
+import attachEventListeners from './logic/attachEventListeners';
+import combineFieldValues from './logic/combineFieldValues';
+import findRemovedFieldAndRemoveListener from './logic/findRemovedFieldAndRemoveListener';
 import getFieldsValues from './logic/getFieldsValues';
 import getFieldValue from './logic/getFieldValue';
-import validateField from './logic/validateField';
-import findRemovedFieldAndRemoveListener from './logic/findRemovedFieldAndRemoveListener';
-import attachEventListeners from './logic/attachEventListeners';
-import validateWithSchema from './logic/validateWithSchema';
-import combineFieldValues from './logic/combineFieldValues';
 import shouldUpdateWithError from './logic/shouldUpdateWithError';
-import warnMissingRef from './utils/warnMissingRef';
-import modeChecker from './utils/validationModeChecker';
-import onDomRemove from './utils/onDomRemove';
-import isRadioInput from './utils/isRadioInput';
-import isEmptyObject from './utils/isEmptyObject';
-import filterUndefinedErrors from './utils/filterUndefinedErrors';
+import validateField from './logic/validateField';
+import validateWithSchema from './logic/validateWithSchema';
 import {
-  Props,
-  Field,
-  ErrorMessages,
-  Ref,
-  SubmitPromiseResult,
-  FieldsObject,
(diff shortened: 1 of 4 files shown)


bluebill1049:
love those types @barry4dev thanks very much again! how fortunate to have you as a contributor... ❤️

src/index.ts
```diff
@@ -89,7 +92,7 @@ export default function useForm<Data extends DataType>(
     forceValidation,
   }: {
     name: Extract<keyof Data, string>;
-    value?: FieldValue;
+    value?: Data[Name];
```
bluebill1049: 👍

src/index.ts
```diff
@@ -172,7 +186,7 @@ export default function useForm<Data extends DataType>(
   );
 
   const setError = <Name extends keyof Data>(
-    name: Extract<keyof Data, string>,
+    name: Extract<Name, string>,
```
bluebill1049: 👍

src/index.ts
```diff
@@ -266,26 +297,34 @@ export default function useForm<Data extends DataType>(
...
     const values = getFieldsValues(fieldsRef.current, fieldNames);
-    const result = values === undefined || isEmptyObject(values) ? undefined : values;
+    const result =
+      values === undefined || isEmptyObject(values) ? undefined : values;
     return result === undefined ? defaultValue : result;
   }
 
-  function register(data: Ref | RegisterInput, rules?: RegisterInput): any {
-    if (!data || typeof window === 'undefined') return;
+  // TODO - can this be typed toward the React.Ref type?
```
bluebill1049: yea, I think so. That would be great. I tried that before but got some error so I moved on :(

src/index.ts
```diff
@@ -266,26 +297,34 @@ export default function useForm<Data extends DataType>(
...
+  function register(
+    refOrValidateRule: RegisterInput | Ref,
+    validateRule?: RegisterInput,
+  ): any {
+    if (!refOrValidateRule || typeof window === 'undefined') return;
 
-    if (rules && !data.name) {
-      warnMissingRef(data);
+    if (validateRule && !refOrValidateRule.name) {
```
bluebill1049: better :)

src/logic/getFieldsValues.ts
```diff
@@ -2,20 +2,20 @@ import getFieldValue from './getFieldValue';
 import isString from '../utils/isString';
 import { FieldValue, Ref } from '../types';
 
-export default function getFieldsValues(fields, filedName?: string | string[]): { [key: string]: FieldValue } | {} {
+export default function getFieldsValues(fields, fieldName?: string | string[]): { [key: string]: FieldValue } {
```
bluebill1049: oops can't type field lol

bluebill1049:
some really good stuff in here! 👍

src/index.ts
barry4dev: Yeah I tried as well, could work it out but I’m sure there’s some way 😊

PR #67, opened 2019-06-10
For Custom fields (those with no type), allow defaults to be set

barrymay:
For custom fields, the register was returning prior to the default being allowed to be set

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -228,40 +228,44 @@ export default function useForm<Data extends DataType>(
         : -1;
 
     if ((!isRadio && field) || (isRadio && existRadioOptionIndex > -1)) return;
+
     if (!type) {
       fields[name] = { ref: { name }, ...data };
-      return;
-    }
-
-    if (isRadio) {
-      if (!field)
-        fields[name] = {
-          options: [],
-          required,
-          validate,
-          ref: { type: 'radio', name },
-        };
-      if (validate) fields[name]!.validate = validate;
-
-      (fields[name]!.options || []).push({
-        ...inputData,
-        mutationWatcher: onDomRemove(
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -228,42 +228,46 @@ export default function useForm<Data extends DataType>(
...
+            validate,
+            ref: { type: 'radio', name },
+          };
+        if (validate) fields[name]!.validate = validate;
+
+        (fields[name]!.options || []).push({
+          ...inputData,
+          mutationWatcher: onDomRemove(
+            elementRef,
+            (): Function => removeEventListener(inputData, true),
```
bluebill1049: This may need to be avoided without type.

bluebill1049:
curious to know what's the reason for this change?

src/index.ts
barry4dev: this diff is just largely visual - this change largely just moves the return for !type till after the isRadio check - I'll explain more on your other question

barry4dev:
Reason for the change -

If I do this
```ts
useForm<FormData>({
    mode: 'onSubmit',
    defaultValues: {
      foo: [],
    },
  });
```

then I create a custom field
```ts
  register({name: 'foo'});
```

the foo field doesn't get the empty array because the register can't get that far

bluebill1049:
I see @barry4dev, good point :)

probably just need to omit the mutation observer bit cause it's not going to be a DOM element.

`mutationWatcher: xxx`

barry4dev:
I don't think it's going to get there - that all is in an else clause that gets triggered only if type is set

bluebill1049:
thanks, @barry4dev again 👍 hope you had a nice weekend

PR #68, opened 2019-06-10
Avoid recreation of register of register function via useCallback

barrymay:
This change allows the register to be a safe dependency of useEffect, since by using useCallback it stays the same across renders :)


bluebill1049:
This is awesome! thanks @barry4dev

PR #71, opened 2019-06-10
only update submitting references after callback

kabriel:
if you are using an async submit handler and you do something that updates the form, it will report as already submitted. This fixes that issue by delaying when those references are updated till after the submit handler has finished (or errors have been assigned)


bluebill1049:
Thanks @kabriel ❤️

PR #89, opened 2019-06-15
updated formState api

liketurbo:
added isValid property to formState, it's nice to have one instead of calculating 🤗

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -475,6 +475,7 @@ export default function useForm<Data extends DataType>(
       submitCount: submitCountRef.current,
       touched: touchedFieldsRef.current,
       isSubmitting: isSubmittingRef.current,
+      isValid: isEmptyObject(errorsRef.current),
     },
   };
 }


src/index.ts
```diff
@@ -475,6 +475,7 @@ export default function useForm<Data extends DataType>(
       submitCount: submitCountRef.current,
       touched: touchedFieldsRef.current,
       isSubmitting: isSubmittingRef.current,
+      isValid: Object.keys(errorsRef.current).length === 0
```
bluebill1049: nice, can you use `isEmptyObject`?

bluebill1049:
could you please add `invalid` type to types.ts?

src/index.ts
liketurbo: 😄, you doesn't have jquery or lodash in the project

bluebill1049:
I will update the type :) 👍 thank you contributor

liketurbo:
Not been sure what you mean by updating type in `types.ts`, if typescript can get type by ducktyping 🤔, wouldn't it be like overkill?
type

bluebill1049:
yea you are right. that was my bad :(

bluebill1049:
Screen Shot 2019-06-16 at 10 17 45 am
thank you ❤️

PR #105, opened 2019-06-27
fix useFormContext return type

### LINKED ISSUE #104: useFormContext missing return value type
I know useFormContext is a new feature, but at the moment the return type is just `{}`, which gives me type errors:

image

Here's a hacky example that resolves the issues in `useFormContext.tsx`:

```ts
import * as React from "react";
import { FormProps } from './types'

// omit fallback required unless we update to typescript 3.5
type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
type FormContextValues = Required<Omit<FormProps, "children">>;

export const FormGlobalContext = React.createContext<FormContextValues>({} as FormContextValues);

export const useFormContext = () => React.useContext<FormContextValues>(FormGlobalContext);

export function FormContext(props: FormProps){
  const { children, ...rest } = props;
  return (
    <FormGlobalContext.Provider
      value={{
        ...rest
      } as FormContextValues}
      children={children}
    />
  );
}
```

Great work on the library! ❤️

lemcii:
fixes #104

### DIFF
--- a/src/useFormContext.tsx
+++ b/src/useFormContext.tsx
@@ -1,9 +1,12 @@
 import * as React from "react";
 import { FormProps } from './types'
 
-export const FormGlobalContext = React.createContext({});
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
+type FormContextValues = Required<Omit<FormProps, "children">>;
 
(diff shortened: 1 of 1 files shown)


bluebill1049:
Thank you very much!

PR #109, opened 2019-06-29
Better get values typing

barrymay:
Fixes to improve typing for GetValues so it returns with the typed properties you expect.

Includes a couple fixes for eslint rules since they were showing up in the process.


bluebill1049:
That’s awesome Barry! Thanks as usual❤️

bluebill1049:
U have any idea how to apply generic type the the new formcontext?

PR #110, opened 2019-06-29
support validation trigger with array

bluebill1049:


src/useForm.ts
```diff
@@ -136,6 +136,23 @@ export default function useForm<Data extends DataType>(
     return isEmptyObject(error);
   };
 
+  const triggerValidation = async <Name extends keyof Data>(
+    payload:
+      | {
```
barry4dev: just a tip (based on DRY) - define the input object in an interface separate from the function and use it twice

PR #113, opened 2019-07-02
Improve types

bluebill1049:


bluebill1049:
@barry4dev I think this cleanup thing a bit. I don't think fixing the generic type tho for formContext

bluebill1049:
Thanks, @barry4dev looks like it's not breaking formContext and useForm works fine. I am going to merge it in and publish. it's an improvement for now.

PR #115, opened 2019-07-02
Consolidate types at the top of function/interface generic

barrymay:
This change seems to help quite a bit (I tested on my local, but please test on your UseFormContext case).  If you defined the generic once at the top of the type, the same keyof will be propagated to all of the functional children.  This makes sense, because if you make each individual function generic, it could be overridden, which you kinda don't want...

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -80,7 +80,10 @@ export interface SubmitPromiseResult<Data extends DataType> {
 
 export type VoidFunction = () => void;
 
-export interface FormProps<Data extends DataType = DataType> {
+export interface FormProps<
+  Data extends DataType = DataType,
+  Name extends keyof Data = keyof Data
+> {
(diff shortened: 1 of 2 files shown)


barry4dev:
Note this PR is on top of your improve-types branch

src/types.ts
```diff
@@ -80,7 +80,10 @@ export interface SubmitPromiseResult<Data extends DataType> {
 
 export type VoidFunction = () => void;
 
-export interface FormProps<Data extends DataType = DataType> {
+export interface FormProps<
+  Data extends DataType = DataType,
+  Name extends keyof Data = keyof Data
```
bluebill1049: that's cleaver

bluebill1049:
thanks, heaps @barry4dev like you suggested going to write some unit to make sure those type fails to get captured.

Issue #134, opened 2019-07-18
getValues() and watch() don't work correctly with array fields


zernie:
**Describe the bug**
`getValues()` and `watch()`  return incorrect values when working with array fields.

**To Reproduce**
```
import React from "react";
import ReactDOM from "react-dom";
import useForm from "react-hook-form";

function App() {
  const { register, getValues, watch } = useForm({
    defaultValues: { test: [0, 1] }
  });
  const test = watch("test");

  return (
    <form>
      <h3>getValues: {JSON.stringify(getValues())}</h3>
      <h3>watch: {test}</h3>
      {/* <h3>watch(): {JSON.stringify(watch())}</h3> */}

      <input name="test[0]" ref={register} type="checkbox" />
      <input name="test[1]" ref={register} type="checkbox" />
    </form>
  );
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```

https://codesandbox.io/embed/react-hook-form-basic-qgkkp
**Expected behavior**
* `getValues()` should have returned `{test: [0,1]}` (instead got `{}` on first render and `{"test[0]":"on","test[1]":false}` after clicking on the first checkbox)
* `watch('test')` should have returned `[0,1]`, instead got: `undefined`.

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser: Chrome
 - Version: 75

zernie:
Also: component doesn't rerender on checkbox uncheck. Calling `watch()` fixes that somehow

zernie:
@bluebill1049 I'm not sure if that's a bug, but `watch(anyfield)` returns undefined on first render. I've updat the example.

bluebill1049:
@zernie thanks for the issue report. I will look into `getValues()` and `watch()` issue over the weekend.

> I'm not sure if that's a bug, but watch(foo) returns undefined on the first render even though foo is set in default values. I've updated the example.

This is expected: `watch` have the second argument for a default value, because first render, all the inputs would not have been `register` just yet.

bluebill1049:
> watch('test') should have returned [0,1], instead got: undefined.

I think this is not an issue when you want to watch fields,  you probably should do the following:

`watch(['test[0]', 'test[1]']`

bluebill1049:
> Also: component doesn't rerender on checkbox uncheck. Calling watch() fixes that somehow

because `watch('test')` is not wathcing anything, it worked the first time, because it was updating the `formState` with `touched` 

[...]

PR #138, opened 2019-07-20
#134 Improve watch defaultValues and getValues with non flatten data structure

bluebill1049:
- [x] `defaultValues` inject default value into `watch`
- [x] `improve` watch with array and objects
- [x]  `getValues` with arugment to `nest`, default `nest` to `false` (similar functionality will apply to watch)
- [x] `defaultValues` support array and objects
- [x] update website with `getValues` update
- [x] update website with `defaultValues` and `defaultValue` association with `watch`

### DIFF
--- a/src/utils/isString.ts
+++ b/src/utils/isString.ts
@@ -1,1 +1,2 @@
-export default (value: any): boolean => value && typeof value === 'string';
+export default (value: any): value is string =>
+  value !== undefined && typeof value === 'string';
--- a/src/utils/isString.ts
+++ b/src/utils/isString.ts
@@ -1,2 +1,2 @@
-export default (value: any): value is string =>
+export default (value: unknown): value is string =>
   value !== undefined && typeof value === 'string';
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     },
     {
       "path": "./dist/react-hook-form.min.js",
-      "maxSize": "5 kB"
+      "maxSize": "5.5 kB"
     }
   ],
   "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -67,10 +67,6 @@
     {
       "path": "./dist/react-hook-form.min.es.js",
       "maxSize": "5 kB"
-    },
-    {
-      "path": "./dist/react-hook-form.min.js",
-      "maxSize": "5.5 kB"
     }
   ],
   "peerDependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-hook-form",
-  "version": "3.18.14",
+  "version": "3.19.0-beta.2",
   "main": "dist/react-hook-form.js",
   "module": "dist/react-hook-form.es.js",
   "types": "dist/index.d.ts",
--- a/src/logic/assignWatchFields.ts
+++ b/src/logic/assignWatchFields.ts
@@ -1,0 +1,22 @@
+import get from '../utils/get';
+import getPath from '../utils/getPath';
+import combineFieldValues from './combineFieldValues';
+import { DataType } from '../types';
+
+export default (
+  fieldValues: DataType,
+  fieldName: string,
+  watchFields: { [key: string]: boolean },
+) => {
+  if (fieldValues[fieldName]) {
+    watchFields[fieldName] = true;
+    return fieldValues[fieldName];
+  }
+
+  const combinedValues = combineFieldValues(fieldValues);
+  const values = get(combinedValues, fieldName);
+  getPath(fieldName, values).forEach(name => {
+    watchFields[name as any] = true;
+  });
+  return values;
+};
--- a/src/logic/getDefaultValue.ts
+++ b/src/logic/getDefaultValue.ts
@@ -1,0 +1,5 @@
+import get from '../utils/get';
+import { DataType } from '../types';
+
+export default (defaultValues: DataType | undefined, name: string): any =>
+  defaultValues && (defaultValues[name] || get(defaultValues, name));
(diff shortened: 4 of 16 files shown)


bluebill1049:
**`watch` API update:**

1. `watch` return default value change:
`watch(['test', 'test1']) // deafult value return [undefined, undefined] instead of undefined`
`watch() // default value return {} instead of undefined`

2. `watch` integrate with `defaultValues`
`watch` will inherit from data in `defaultValues`, but watch function arguments remain unchanged. second argument `defaultValue` will overwrite data in `defaultValues` which set up in `useForm()`

3. `watch` now support nested object look up
eg
```
  const data = watch('test');
  console.log(data);

    <form>
      <input name="test[0]" ref={register} />
      <input name="test[1]" ref={register} />
    </form>
```
this above example will watch both `test[0]` and `test[1]`

**`defaultValues` update:**

`defaultValues` now support nested object, eg:
```
const { register } = useForm({
  defaultValues: { test: {data: [1, 2]}},
})
```

**`getValues` update:**

`getValues` now support `nest` argument to return data in a nested format, eg:

```
getValues(); // {'test[0]': 1, 'test[1]': 2}
getValues({ nest: true }); // {test: [1,2]}
```

src/utils/isString.ts
```diff
@@ -1 +1,2 @@
-export default (value: any): boolean => value && typeof value === 'string';
+export default (value: any): value is string =>
```
stramel: I had made this change in one of my cleanup commits locally 😛

barrymay: Totally support the type guard!  

Btw, I might be mistaken but technically you shouldn't need the value !== undefined anymore since typeof will check that too 

And performance wise it's very similar (this is a perf check of typeof === 'undefined' but very similar
https://jsperf.com/undefined-vs-typeof

stramel: Correct, you shouldn't need the anything other than `typeof value === 'string'`. 

Just for a bit more cleanup, you could change `value` to be of type `unknown` instead. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type

Internally at my company, we have shifted to disallow the use of `any` since it is really an "escape hatch" out of Typescript.

I would end up with something like

```ts
export default function isString(value: unknown): value is string {
  return typeof value === 'string'
}
```

barrymay: Nice - I've been personally puzzling with what `unknown` is for while but after reading that handbook page again and your comment here, I sorta get it now.  

Definitely a step above any - just a side-question - do you use `unknown` as well when you're creating a custom definition for a 3rd party lib?

stramel: I believe that most 3rd party libs still use `any` for values. In our case, `FieldValue` of type `any` would be acceptable. I _think_ it could be `unknown` but I'm not 100% positive.

bluebill1049: looks like `unknown` is safer in this case, I will update cheers guys

bluebill1049:
@barrymay @stramel wondering if I can get you guys to help on `typescript` part, I still have couple `@ts-ignore ` in the repo.

bluebill1049:
I have released a beta version if you guys want to help on testing. `react-hook-form@3.19.0-beta.2`

much appreciated

stramel:
Will test in a few and will try to get rid of the ts-ignores

stramel:
@bluebill1049 Overall seems to work great!

- `getValues` currently requires a payload. We should make that optional.
- `errors` types that I updated, make the DX worse so we may want to update that eventually to make it better. Also, they still need to be updated to match the actual flattened structure.
- `watch` api: What's the benefit of the defaultValue override? If we drop that, we could clean up to accept any number of properties. `watch('attributes', 'name', 'age', 'type')`.

bluebill1049:
@stramel that's for the testing. ❤️ 

> getValues currently requires a payload. We should make that optional.
agree! my mistake

> watch api: What's the benefit of the defaultValue override? If we drop that, we could clean up to accept any number of properties. watch('attributes', 'name', 'age', 'type').
I didn't want to make a breaking change, and if people prefer to use `defaultValue` on the input itself. then you will need a `defaultValue` set at `watch`

stramel:
WRT `@ts-ignore`s I was thinking about tackling them after this PR goes through.

stramel:
@bluebill1049 	

> > getValues currently requires a payload. We should make that optional.

> agree! my mistake

All good! Just something I caught

> > errors types that I updated, make the DX worse so we may want to update that eventually to make it better. Also, they still need to be updated to match the actual flattened structure.

> :(

Yeah, I can try to fix it with some of the typing updates.

> > watch api: What's the benefit of the defaultValue override? If we drop that, we could clean up to accept any number of properties. watch('attributes', 'name', 'age', 'type').

> I didn't want to make a breaking change, and if people prefer to use defaultValue on the input itself. then you will need a defaultValue set at watch

That makes sense! I wonder if it is worth supporting the `defaultValue` on `watch` when the `input` itself could handle `defaultValue`? I suppose supporting it through `watch` allows for more generic handling and not requiring a component that has `defaultValue`.

bluebill1049:
> That makes sense! I wonder if it is worth supporting the defaultValue on watch when the input itself could handle defaultValue? I suppose supporting it through watch allows for more generic handling and not requiring a component that has defaultValue.

it's only made for the first render, otherwise `watch()` will return undefined, because input hasn't been `register`

```
const toggle = watch('test') //first render will be undefined, defaultValue can prevent this to happen

// const toggle = watch('test', 'haha') can solve this problem.

<input name="test" defaultValue="haha" />

{toggle && 'showme'}
```

Issue #135, opened 2019-07-18
`errors` structure and type mismatch


stramel:
**Describe the bug**
The `errors` structure does not match the typing when using a array value.

Our form values structure:

```ts
type FormValues = {
	emails: string[]
}
```

Our form hook looks like:

```ts
const { register, errors, handleSubmit, reset } = useForm<FormValues>)()
```

This results in `errors` structure of:

```json
{
	"emails[0]": "asdf@gmail.com",
	"emails[1]": "asdf2@gmail.com"
}
```

where the `errors` typing is expecting it to be more like:

```json
{
	"emails": ["asdf@gmail.com", "asdf2@gmail.com"]
}
```

**Expected behavior**
`errors` should match the values structure or `errors` typings should be updated to match the structure it is providing.

**Screenshots**
![image](https://user-images.githubusercontent.com/855184/61473917-1d42be00-a94d-11e9-843a-0a913c0b41b4.png)

stramel:
This also appears related to #134

bluebill1049:
Thanks, @stramel I will sort out #134 then come back to this issue. @barry4dev I may need some help over this `type` issue 🙏

bluebill1049:
There are two options:

1. fix the type for errors 
  `name: boolean | string` name as what you registered 

2. fix the errors returned from hook form 
   combine error messages for array fields into `"emails": ["asdf@gmail.com", "asdf2@gmail.com"]`

I personally prefer **option1**, because it's always `errors['inputName']` instead `errors['name'][index]` and this can be pretty long and error promting eg: `errors.xxx.xxx[0]`

stramel:
@bluebill1049 I just noticed, this also affects `defaultValues` (even for non-arrays).

I would expect this to load a nested named input,

```
{
  attributes: {
	primaryColor: '#000'
  }
}
```

However it is only working for,

```
{
	'attributes.primaryColor': '#000'
}
```

stramel:
@bluebill1049 WRT the 2 options, those are the 2 options that I came to as well when thinking about it.

When I was thinking of the solution my initial thought was option 2. As a developer of a form, I have a data structure in mind and logically it would make sense to me to refer to the values and errors using a similar structure. 

When you start to think about "Nested Fields", it really depends if you are using object or array notation as to which way is easier. 

[...]

PR #142, opened 2019-07-24
Update some types

### LINKED ISSUE #136: Wrong errors object type when using `validationSchema`
When `validationSchema` is used, the type of the `errors` object is not 
```typescript
export type ErrorMessages<Data extends DataType> = {
  [Key in keyof Data]?: Error;
};
```
but:
```typescript
export type ErrorMessages<Data extends DataType> = {
  [Key in keyof Data]?: string;
};
```

**To Reproduce**
1. Run the [`validationSchema` example](https://github.com/bluebill1049/react-hook-form/blob/master/examples/validationSchema.tsx)
2. Do not fill in anything, just click submit
3. You get the serialized errors object back:
```javascript
Errors: {
  "firstName": "firstName is a required field",
  "age": "age must be a `number` type, but the final value was: `NaN` (cast from the value `\"\"`)."
}
```

**Expected behavior**
Either the `errors` object entries should conform to the `Error` type or the `errors` object itself should be typed as something like:
```typescript
export type ObjectErrorMessages<Data extends DataType> = {
  [Key in keyof Data]?: Error;
};
export type StringErrorMessages<Data extends DataType> = {
  [Key in keyof Data]?: string;
};
export type ErrorMessages<Data extends DataType> = ObjectErrorMessages<Data> | StringErrorMessages<Data>
```

stramel:
Fixes #136 

The first item is what actually fixes #136. Please let me know if you want me to revert the other changes. 

**Changes:**
- `ErrorMessages` now can have an object using either `strings` or `Error` objects for values. (This is what fixes #136)
- Removed `Omit`, it is included in v3.5 of Typescript (https://devblogs.microsoft.com/typescript/announcing-typescript-3-5/#the-omit-helper-type)
- Anything that was manually setting up the index for an object, now is using `Record<key, value>` (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#partial-readonly-record-and-pick)
- exporting `Mode` as a type
- Re-used some already built out types (`VoidFunction` and `NumberOrString`
- Aliased `Data[Name]` to `Value` in the `FormProps` interface.
- Used `object` instead of `{}`

**NOTE:** There are some other improvements to the types that we can make but they will take longer to refactor and will probably depend on #134 and #135 resolutions.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,6 +1,5 @@
 import * as React from 'react';
 
-
 export type FieldValue = any;
 
 export type Validate = (data: FieldValue) => string | boolean;
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,25 +1,24 @@
 import * as React from 'react';
 
+
 export type FieldValue = any;
 
 export type Validate = (data: FieldValue) => string | boolean;
 
 export type NumberOrString = number | string;
 
-type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
-
-export interface DataType {
-  [key: string]: FieldValue;
-}
+export type DataType = Record<string, FieldValue>;
 
 export type OnSubmit<Data extends DataType> = (
   data: Data,
   e: React.SyntheticEvent,
-) => void;
+  ) => void;
+  
+export type Mode = 'onSubmit' | 'onBlur' | 'onChange'
 
-export interface Props<Data> {
-  mode?: 'onSubmit' | 'onBlur' | 'onChange';
-  defaultValues?: { [key: string]: any };
+export interface Props<Data extends DataType> {
+  mode?: Mode;
+  defaultValues?: Partial<Data>;
   nativeValidation?: boolean;
   validationFields?: (keyof Data)[];
   validationSchema?: any;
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -6,20 +6,18 @@ export type Validate = (data: FieldValue) => string | boolean;
 
 export type NumberOrString = number | string;
 
-type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
-
-export interface DataType {
-  [key: string]: FieldValue;
-}
+export type DataType = Record<string, FieldValue>;
```
bluebill1049: cool, i didn't even know `Record` type 👍

src/types.ts
```diff
@@ -43,9 +41,8 @@ export interface RegisterInput {
   pattern?: RegExp | { value: RegExp; message: string };
   validate?:
     | Validate
-    | { [key: string]: Validate }
-    | { value: Validate | { [key: string]: Validate }; message: string }
-    | undefined;
+    | Record<string,Validate>
+    | { value: Validate | Record<string,Validate>; message: string };
```
bluebill1049: nice

src/types.ts
```diff
@@ -69,9 +66,13 @@ export interface Error {
...
 }
 
-export type ErrorMessages<Data extends DataType> = {
+export type ObjectErrorMessages<Data extends DataType> = {
   [Key in keyof Data]?: Error;
 };
+export type StringErrorMessages<Data extends DataType> = {
+  [Key in keyof Data]?: string;
+};
+export type ErrorMessages<Data extends DataType> = ObjectErrorMessages<Data> | StringErrorMessages<Data>
```
bluebill1049: 👍

bluebill1049:
Thank you very much @stramel 👍 So good! merging

stramel:
Awesome! Once we get the errors and values resolved, I will go back through and clean up the types more.

bluebill1049:
get some sleep @stramel I am sure it's late over there now 😴

PR #147, opened 2019-07-24
Fix min and max value validation when min/max is 0

chrisparton1991:
Hi, thanks for all of your effort putting together this awesome library!

This PR fixes an issue with min and max validation not working for `min: 0` or `max: 0`, due to `0` being treated as falsy and being ignored.

The same issue exists for `minLength` and `maxLength`, but a max length of 0 doesn't sound sensible and a min length of 0 should use `required: true` instead.

That said, I'm happy to add this same fix for the length validations if you think it's worthwhile?

### DIFF
--- a/src/logic/validateField.ts
+++ b/src/logic/validateField.ts
@@ -1,4 +1,5 @@
 import getRadioValue from './getRadioValue';
+import isNullOrUndefined from '../utils/isNullOrUndefined';
 import isRadioInput from '../utils/isRadioInput';
 import getValueAndMessage from './getValueAndMessage';
 import isCheckBoxInput from '../utils/isCheckBoxInput';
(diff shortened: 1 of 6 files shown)


src/logic/validateField.ts
```diff
@@ -56,17 +57,16 @@ export default async <Data>(
...
   }
 
-  if ((min || max) && !isStringInput) {
+  if ((!isNullOrUndefined(min) || !isNullOrUndefined(max)) && !isStringInput) {
     let exceedMax;
     let exceedMin;
     const valueNumber = parseFloat(value);
     const { value: maxValue, message: maxMessage } = getValueAndMessage(max);
     const { value: minValue, message: minMessage } = getValueAndMessage(min);
-    const message = exceedMax ? maxMessage : minMessage;
```
chrisparton1991: Message was being set before `exceedMax` or `exceedMin` were initialised, so I've moved this down.

bluebill1049:
wow this is so good! thank you @chrisparton1991

bluebill1049:
could you please remove `package-lock.json`?

chrisparton1991:
Whoops! `package-lock` has been removed.

P.S. Cool to see that this library comes from Sydney :) I live in Wollongong.

bluebill1049 (approved):
Thank you very much ❤️ bug fixed 👍

bluebill1049:
@chrisparton1991 hahaha yea I saw your location. 

image

bluebill1049:
going to add you into the contributors' list 👍

chrisparton1991:
Awesome, thanks :)

PR #149, opened 2019-07-25
Cleanup ts-ignore on useForm

barrymay:
Did a bit of work around useForm's ts-ignore issues.  See my comments in the code. (@stramel  I realized late you were looking at this too, so sorry if I stepped on your toes)

BTW, A couple things I'm looking at to further improve, something I've done in my other hooks projects

- Get Eslint working properly in vscode for this project 
- Use the react-hooks ESLint plugin ( a big deal to help ensure safety)
- use UseCallback, useMemo etc to help ensure memoization of useForm's result functions and reduce redraws of the using components

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,7 @@ module.exports = {
     '@typescript-eslint/indent': 'off',
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-unused-vars': ["error", { "varsIgnorePattern": /^_/ }],
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/no-object-literal-type-assertion': 'off',
     'react/prop-types': 'off',
--- a/src/types.ts
+++ b/src/types.ts
@@ -59,15 +59,15 @@ export type FieldsObject<Data extends DataType> = {
   [Key in keyof Data]?: Field;
 };
 
-export interface Error {
+export interface ReactHookFormError {
   ref: Ref;
+  type: string;
   message?: string;
-  type?: string;
   isManual?: boolean;
 }
 
 export type ObjectErrorMessages<Data extends DataType> = {
-  [Key in keyof Data]?: Error;
+  [Key in keyof Data]?: ReactHookFormError;
 };
 export type StringErrorMessages<Data extends DataType> = {
   [Key in keyof Data]?: string;
--- a/src/useForm.test.tsx
+++ b/src/useForm.test.tsx
@@ -504,7 +504,7 @@ describe('useForm', () => {
 
     it('should remove error', () => {
       hookForm.setError('input', 'test');
-      hookForm.setError('input');
+      hookForm.clearError('input');
       expect(hookForm.errors).toEqual({});
     });
   });
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -307,7 +307,6 @@ export default function useForm<
     message?: string,
     ref?: Ref,
   ): void => {
-    // since errors can be strings or objects, and we only wa
     const errorsFromRef = errorsRef.current;
     const error = errorsFromRef[name];
     const isSameError =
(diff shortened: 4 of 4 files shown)


src/types.ts
```diff
@@ -59,15 +59,15 @@ export type FieldsObject<Data extends DataType> = {
   [Key in keyof Data]?: Field;
 };
 
-export interface Error {
+export interface ReactHookFormError {
```
barrymay: Avoid usage of the Error interface since we don't want to conflict with the generic TS one

src/useForm.test.tsx
```diff
@@ -504,7 +504,7 @@ describe('useForm', () => {
 
     it('should remove error', () => {
       hookForm.setError('input', 'test');
-      hookForm.setError('input');
+      hookForm.clearError('input');
```
barrymay: I know this is an API change but it's much cleaner, clearer if we don't allow set error to do both, feel free to disagree with me here.  (it did help clarify the ts-ignore as well)

src/useForm.ts
```diff
@@ -66,7 +67,7 @@ export default function useForm<
   const validateAndStateUpdateRef = useRef<Function>();
   const fieldsWithValidationRef = useRef(new Set());
   const validFieldsRef = useRef(new Set());
-  const reRenderForm = useState({})[1];
+  const [_unused, reRenderForm] = useState({});
```
barrymay: this extra var is on purpose (as well as the .eslintrc.js add, because the react-hooks ESLint plugin will watch this pattern and ensure it doesn't count reRenderForm as a mutating function

The _ prefix is also used by TS for similar unused vars

src/useForm.ts
```diff
@@ -134,32 +135,31 @@ export default function useForm<
...
     const fieldErrors = await validateWithSchema(validationSchema, fieldValues);
-    const isArray = Array.isArray(payload);
-    let errors;
-    let payloadName;
+    let result: boolean;
+    let errors: ErrorMessages<Data>;
 
-    if (isArray) {
-      const names = (payload as []).map(({ name }) => name);
+    if (Array.isArray(payload)) {
```
barrymay: doing it here handles the type guard behavior so payload will be an array below

src/useForm.ts
```diff
@@ -134,32 +135,31 @@ export default function useForm<
...
-            // @ts-ignore
             names.includes(key) ? { ...previous, [key]: value } : previous,
           {},
         ),
       );
+      result = isEmptyObject(fieldErrors);
     } else {
-      // @ts-ignore
-      payloadName = payload.name;
+      const payloadName = payload.name as string;
```
barrymay: I prefer avoiding as, but since fieldErrors is using strings as indexer, this is ok and locally held

src/useForm.ts
```diff
@@ -296,25 +296,29 @@ export default function useForm<
...
 
+  const clearError = (name: Name): void => {
+    delete errorsRef.current[name];
+    reRenderForm({});
+  };
+
   const setError = (
     name: Name,
-    type?: string,
+    type: string,
```
barrymay: types should always be there for setError

.eslintrc.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
     '@typescript-eslint/indent': 'off',
     '@typescript-eslint/no-non-null-assertion': 'off',
     '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-unused-vars': ["error", { "varsIgnorePattern": /^_/ }],
```
barrymay: this allows ESLint to skip warning if the var is prefixed with _ (similar to internal TS standard)

src/useForm.test.tsx
bluebill1049: agree 👍

src/useForm.ts
bluebill1049: yeah, I will update the dos for this as well.

bluebill1049:
thanks very much! ❤️ @barrymay so good 

You and @stramel have been my typescript master here. so glad to have u guys in this project.

bluebill1049:
This PR is beautiful 😍

PR #150, opened 2019-07-25
Fixed defaultValue check

kabriel:
Previously `defaultValue` was checked for truthy, but a default value of `0` would return false and not set the value in the form. changed to a `defined` check so we can handle any value, including 0.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -386,7 +386,7 @@ export default function useForm<
 
     if (defaultValues) {
       const defaultValue = defaultValues[name] || get(defaultValues, name);
-      if (defaultValue) setFieldValue(name, defaultValue);
+      if (defaultValue !== undefined) setFieldValue(name, defaultValue);
     }
 
     if (!type) return;
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -386,7 +386,7 @@ export default function useForm<
 
     if (defaultValues) {
       const defaultValue = defaultValues[name] || get(defaultValues, name);
-      if (isNullOrUndefined(defaultValue)) setFieldValue(name, defaultValue);
+      if (!isNullOrUndefined(defaultValue)) setFieldValue(name, defaultValue);
     }
 
     if (!type) return;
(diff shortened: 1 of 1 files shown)


kabriel:
@bluebill1049 do you want an issue filed for all PRs? or is this ok by itself?

src/useForm.ts
```diff
@@ -386,7 +386,7 @@ export default function useForm<
 
     if (defaultValues) {
       const defaultValue = defaultValues[name] || get(defaultValues, name);
-      if (defaultValue) setFieldValue(name, defaultValue);
+      if (defaultValue !== undefined) setFieldValue(name, defaultValue);
```
bluebill1049: can we use `isNullOrUndefined` function for this?

bluebill1049:
Thanks, @kabriel PR is great! minor comment.

src/useForm.ts
barrymay: I think this is ok - if you want it specifically null as the default, that's fine

kabriel: i didn't see the `isNullOrUndefined`, i like helpers like that a lot, great idea. fixed.

kabriel:
@bluebill1049 fixed. i like the helper!

src/useForm.ts
bluebill1049: @barrymay I think you normally want to do empty string for native input value?

barrymay: correct for typical fields, and I don't have a great use case for this, but just thinking that if this is for a custom field, you may want to allow a definitive null default.

bluebill1049: @barrymay that's a good point.

bluebill1049: @barrymay probably should remove that check completely?

barrymay: I thought @kabriel 's original (defaultValue !== undefined)  check was fine.  Having it default is undefined is a safe javascript assumption

bluebill1049: alright happy to revert it back.

PR #152, opened 2019-07-26
option to update value without triggering dirty/touched

kabriel:
i need to update the form values without triggering dirty/touched 

i need to reset the form back to the default values (not clear it) on button click, and without this the form will be dirty and my submit button will be enabled.

what do you think about this change? i'm open to other ways to get this done too if this doesn't feel right.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -211,11 +211,13 @@ export default function useForm<
     name: Name,
     value: Data[Name],
     shouldValidate: boolean = false,
+    shouldDirty: boolean = true,
   ): void => {
     setFieldValue(name, value);
-    touchedFieldsRef.current.add(name);
-    isDirtyRef.current = true;
-
+    if (shouldDirty) {
+      touchedFieldsRef.current.add(name);
+      isDirtyRef.current = true;
+    }
     reRenderForm({});
     if (shouldValidate) triggerValidation({ name });
   };


barrymay:
my two cents: (I won't override Bill here), and I get what you're doing, but why not just ensure the reset process can reset to defaults?   'dirty' is a useful concept for normal functionality and turning it off seems wrong to me.

kabriel:
Well, we would need to add a different method because `reset` will clear/empty the fields and i just want them set back to their `defaultValues`. 

it would be nice to be able to change a field value programmatically without triggering visual feedback that makes the user feel like they are the ones that did the change. but i see what you mean.

i'm open to both.

bluebill1049:
Thanks @kabriel I will take a closer look tonight.

stramel:
I agree with barry. My personal input is that we just update `reset` to take a "defaultValues-like" object, defaulting to the `defaultValues` passed in to `useForm`. If you want to completely clear the form, you could pass an empty object or null?

bluebill1049:
@stramel I like the idea when you can reset with `defaultValues`

bluebill1049:
what do you think @barrymay @kabriel `reset(values)` and `defaultValue` gets populated?

barrymay:
I'm fine with that

kabriel:
I’m down

bluebill1049:
https://github.com/bluebill1049/react-hook-form/pull/154

PR #155, opened 2019-07-27
fix error for schema validation

bluebill1049:
@barrymay sorry that was my fucked up in schema validation errors, i think this will make `type` correct as well.

### DIFF
--- a/src/logic/validateWithSchema.ts
+++ b/src/logic/validateWithSchema.ts
@@ -4,12 +4,11 @@ export function parseErrorSchema(error: DataType): ValidationReturn {
   return error.inner.reduce(
     (
       previous: DataType,
-      current: DataType,
-      index: number,
-    ): ValidationReturn => {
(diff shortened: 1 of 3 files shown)


src/logic/validateWithSchema.ts
```diff
@@ -4,12 +4,11 @@ export function parseErrorSchema(error: DataType): ValidationReturn {
...
-      current: DataType,
-      index: number,
-    ): ValidationReturn => {
-      previous[current.path] = error.errors[index];
-      return previous;
-    },
+      { path, message, type }: DataType,
+    ): ValidationReturn => ({
+      ...previous,
+      [path]: { message, ref: {}, type },
```
bluebill1049: God knows what I was thinking...

bluebill1049:
@barrymay thanks heaps for pointing out the root cause.

barrymay (commented):
Look good - thanks!

PR #156, opened 2019-07-27
Wrap setValue function with useCallback to provide stable identity

chrisparton1991:
Hi again Bill!

I found that calling `setValue` inside a `useEffect` hook causes an infinite loop. This occurs because `setValue` needs to appear in the `useEffect` dependency list, but the `setValue` function changes with each render, and that causes the `useEffect` to run in an endless loop. Reproduced in [this CodeSandbox](https://codesandbox.io/s/react-hook-form-basic-oq64n).

My proposed fix is to wrap `setValue` in a `useCallback` to give it a stable identity for the lifetime of the form.

All of the tests still pass with this change, but I don't know enough about React component testing to figure out how to reproduce the issue in a test sorry!

I'm still pretty new to hooks, so it's possible I'm breaking something here, but I _think_ what I'm doing is sensible?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,6 +5,9 @@ module.exports = {
     ecmaVersion: 2018,
     sourceType: 'module',
   },
+  plugins: [
+    'react-hooks'
+  ],
   rules: {
     '@typescript-eslint/indent': 'off',
     '@typescript-eslint/no-non-null-assertion': 'off',
@@ -12,6 +15,8 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/no-object-literal-type-assertion': 'off',
     'react/prop-types': 'off',
+    'react-hooks/rules-of-hooks': 'error',
+    'react-hooks/exhaustive-deps': 'warn'
   },
   settings: {
     react: {
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "enzyme-adapter-react-16": "^1.12.1",
     "eslint": "^5.16.0",
     "eslint-plugin-react": "^7.13.0",
+    "eslint-plugin-react-hooks": "^1.6.1",
     "husky": ">=1",
     "jest": "^24.7.1",
     "lint-staged": ">=8",
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -200,18 +200,17 @@ export default function useForm<
     }
   };
 
-  const setValue = (
-    name: Name,
-    value: Data[Name],
-    shouldValidate: boolean = false,
-  ): void => {
-    setFieldValue(name, value);
-    touchedFieldsRef.current.add(name);
-    isDirtyRef.current = true;
+  const setValue = useCallback(
+    (name: Name, value: Data[Name], shouldValidate: boolean = false): void => {
+      setFieldValue(name, value);
+      touchedFieldsRef.current.add(name);
+      isDirtyRef.current = true;
 
-    reRenderForm({});
-    if (shouldValidate) triggerValidation({ name });
-  };
+      reRenderForm({});
+      if (shouldValidate) triggerValidation({ name });
+    },
+    [],
+  );
 
   validateAndStateUpdateRef.current = validateAndStateUpdateRef.current
     ? validateAndStateUpdateRef.current


barrymay:
Hi @chrisparton1991 thanks (this is great) but I want to ensure we get the React-hooks eslint plug-in checks first so we avoid a mistake here by “over-memorizing” the callback.

If you want to try adding the React-hook plugin first and sorting out warnings pls go ahead, otherwise I’ll get to it in the next couple days.

Thanks again!

bluebill1049:
Hi @chrisparton1991,

Thanks for the PR again :) I do have a question in my head around using `useCallback` vs `useRef` for those functions, right now we have fixed usage.

Thanks @barrymay 's great suggestion.  I would love to get React-hooks `eslint` plugin-in to see warnings and potential errors. if `eslint` complain` about `useCallback` with empty dependency, I am kinda leaning towards just use `useRef` to memorize those functions. 

let me know what you guys think?

barrymay:
UseCallback is indeed better than useRef here, as useCallback will update a method if a closed-over variable gets updated.  UseRef is great for some things that need to say static but not as good for functions like this. 

The React-hooks plugin will help clarify and ensure its being used properly.

bluebill1049:
sounds great to me!

chrisparton1991:
Hi guys, good idea about adding the eslint plugin. I've just added a commit that adds it. We get a few warnings, though none of them are related to `setValue`:

Output of `npm run lint`:
```
  212:5  warning  React Hook useCallback has missing dependencies: 'reRenderForm' and 'triggerValidation'. Either include them or remove the dependency array
  441:5  warning  React Hook useCallback has a missing dependency: 'registerIntoFieldsRef'. Either include it or remove the dependency array
  588:6  warning  React Hook useCallback has a missing dependency: 'reRenderForm'. Either include it or remove the dependency array
  597:6  warning  React Hook useEffect has a missing dependency: 'unSubscribe'. Either include it or remove the dependency array. Mutable values like 'isUnMount.current' aren't valid dependencies because mutating them doesn't re-render the component
```

bluebill1049:
thanks @chrisparton1991 I probably will get your PR merged first and sort out the rest of those eslint warnings after. 🙏

barrymay:
Isn’t that first line 212 pointing to the problem caused by the new [] on setValue?

If you look at the squiggly lines in vscode you’ll see what these are pointing to.

barrymay:
On the useRef/useCallback thing Bill, i thought the same thing and that it was ok, until I ran into an issue where a variable i introduced changed and my forced static function did not pick up the change and re-create the function. 

This may sound obvious but React functional components are very different than the class components.  

From my own experience the past few months - I find it a lot safer using a useCallback with an empty array and having the eslint checker analyze it when coding rather than using useRef on a function which is not going to get checked and will be ignored by eslint as it’s just a simple mutable object red.  If you do introduce items into the dependency array and you’re confident they don’t change, that’s ok too - it’s a very helpful check IMO.

barrymay:
I'm taking a look at this ( I had wanted to work thru this the other day but ran out of time ) - will aim to build on your change, @chrisparton1991.

bluebill1049:
thanks @chrisparton1991 and @barrymay ❤️

bluebill1049:
@chrisparton1991 I will publish the patch once we sort those warnings

chrisparton1991:
Thanks guys, looking forward to the patch but no rush on my end, I've patched my `node_modules` for the time being :)

The other workaround is to wrap `setValue` in `useCallback` in the calling code.

Enjoy your Sunday, the weather is gorgeous in Wollongong today 🌞

bluebill1049:
thanks @chrisparton1991 yeah beautiful weather here! 🌞

barrymay:
this may a bit of time in the calling code, I won't in fact work this out now - as we want to be careful about sideeffects in this library.

setValue is dependent on triggerValidation which is itself dependent on setValue, so working out the dependencies is tricky

For better or worse having it recreated in the hook is sometimes safer than useCallback to ensure changes are handled, but if you're willing to use it locally go for it until we have a proper solution in the lib. (Something I want to help sort for sure if possible)

bluebill1049:
@barrymay do you see any problem if we going with `useRef` for now until we have a proper fix?

chrisparton1991:
Sounds good, the last thing I want to do is introduce a bug into the library :)

I'll use my workaround locally for the time being :+1:

barrymay:
I see useCallback with [] (ignoring triggerValidation which has its' own dependency needs)  and using useRef on functions identical (and similarly problematic functionally).    The useCallback benefit is that having ESLint complaining like that, at least the problem is acknowledged.

The issue with both is that if any dependency function should have changed (Because an inner variable changed and wasn't recreated), then the useRef/useCallback will use an older variable reference than you want (as it would be a closure on an old reference).  Hope that makes some sense.  This is the cost and the challenge of the React function component.

TLDR - if we want to sort this out (and we should, since react-hook-form is being advertised as a hooks-friendly lib!), then we ideally should sort out the ESLint deps issue.

bluebill1049:
agree ❤️  @barrymay I will spend sometimes on it today

PR #161, opened 2019-07-28
Better type hints for .watch and .unregister of useForm instance

### LINKED ISSUE #159: Better Type Hints for useForm(..).watch and .unregister's fieldName arguments
The types definitions for `.watch` and `.unregister` functions in the **useForm.d.ts** file look something like:

```
    // ...
    unregister(name: string | string[]) => void;
    // ...
    watch: (fieldNames?: string | string[], ...
    // ...
```

These two arguments with types `string | string[]` are the ones I'm requesting an improvement for.

Just like clearError, setValue or most of the other methods in this file, it is possible to add the generic `Name | Name[]` type to them.

I understand that the reasoning behind making them of a `string | string []` type is so that it could be possible to watch or unregister nested and/or array filed names. 

However, doing something like this (👇) would me a much nicer DX (as most IDEs could provide auto complete hints for most of these field names):

```
    // ...
    unregister(name: Name | string | (Name | string)[]) => void;
    // ...
    watch: (fieldNames?: Name | string | (Name | string)[], ...
    // ...
```

I also understand that this change won't add much to static type checking (because `string` is a more general type than `Name`). But it provides Hints in IDEs.

***Additional (Probably Impossible) Request:***

*I'm not exactly sure how (or if it's even possible), but if it is possible to be specific enough with the type to remove string and also support arrays and nested fields, strictly with the correct type, it would be super nice. (¯\\\_(ツ)\_/¯ I know this is a little too far fetched).*

kaleabmelkie:
Fixes #159 

## Proposed Changes

  - Enhance field `name` arguments type of `.watch` and `.unregister` (in `useForm`) from `string | string[]` to `Name | string | (Name | string)[]`. This provides better auto-complete hints in IDEs for most field names (with the exception of individual array and nested field names).

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -436,7 +436,7 @@ export default function useForm<
   );
 
   function watch(
-    fieldNames?: string | string[] | undefined,
+    fieldNames?: Name | string | (Name | string)[] | undefined,
     defaultValue?: string | Partial<Data> | undefined,
   ): FieldValue | Partial<Data> | void {
     if (isEmptyObject(fieldsRef.current)) {
@@ -446,7 +446,9 @@ export default function useForm<
       if (Array.isArray(fieldNames)) {
         return (
           defaultValue ||
-          fieldNames.map(fieldName => getDefaultValue(defaultValues, fieldName))
+          fieldNames.map(fieldName =>
+            getDefaultValue(defaultValues, fieldName as string),
+          )
         );
       }
       return defaultValue || defaultValues;
@@ -460,7 +462,7 @@ export default function useForm<
     }
     if (Array.isArray(fieldNames)) {
       return fieldNames.map(name =>
-        assignWatchFields(fieldValues, name, watchFields),
+        assignWatchFields(fieldValues, name as string, watchFields),
       );
     }
     return fieldValues;
@@ -485,8 +487,8 @@ export default function useForm<
     [registerIntoFieldsRef],
   );
 
-  const resetField = (name: string) => {
-    const field = fieldsRef.current[name];
+  const resetField = (name: Name | string) => {
+    const field = fieldsRef.current[name as string];
     if (!field) return;
     const { ref, options } = field;
     isRadioInput(ref.type) && Array.isArray(options)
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks @kaleabmelkie for the quick PR, @stramel and @barrymay you guys want to have 👁 over this?

stramel:
LGTM. I'm still planning on trying to clean up the typings. I would like to avoid the weird castings but I need to look up a few things.

bluebill1049:
cool let's merge this one in first and sort it after @stramel I think its an improvement for now?

stramel:
Yes, I should be an improvement for now

bluebill1049:
@kaleabmelkie could you please resolve the conflict? we will get this one out in the next release

barrymay:
@stramel on your comment
> LGTM. I'm still planning on trying to clean up the typings. I would like to avoid the weird castings but I need to look up a few things.

totally agree - I haven't looked into this at all, but I wonder if the strings need to be there at all, what's the case still to not use Name | Name[]?  I look forward to your PR

kaleabmelkie:
@barrymay, I believe just Name | Name[] would be insufficient for nested field names like: form.watch('foo.bar')

kaleabmelkie:
@barrymay, I wonder if something like form.watch('foo.bar') is supported?

(Sorry, I'm new to this library)

If it is not supported, then I believe the string types have no purpose, and it could be just Name | Name[]

bluebill1049:
@kaleabmelkie yes it's supported

kaleabmelkie:
Then I think that the strings will be needed.

I'll try to resolve the merge conflicts now.

kaleabmelkie:
@bluebill1049, I believe it's ready now.

bluebill1049:
thank you @kaleabmelkie

bluebill1049:
could you please do testing for the beta version too? `3.21.1-beta.1` I will test it out too.

barrymay:
@bluebill1049 thanks for the clarity ;)

PR #180, opened 2019-08-01
Cleaning up error collection when field value is valid.

cfgj:
Changes:
- Cleaning up errors when field value is valid.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -65,10 +65,15 @@ export default function useForm<
     modeChecker(mode),
   ).current;
 
-  const combineErrorsRef = (data: any) => ({
+  const combineErrorsRef = (data: ErrorMessages<Data>) => ({
     ...errorsRef.current,
     ...data,
   });
+
+  const cleanUpErrors = (errors: ErrorMessages<Data>, fields: string[]) => (
+    Object.entries(errors)
+      .reduce((result, [field, err]) => fields.some(f => f === field) ? result : { ...result, [field]: err }, {})
+  );
 
   const renderBaseOnError = useCallback(
     (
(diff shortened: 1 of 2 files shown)


src/useForm.ts
```diff
@@ -65,11 +65,16 @@ export default function useForm<
     modeChecker(mode),
   ).current;
 
-  const combineErrorsRef = (data: any) => ({
+  const combineErrorsRef = (data: ErrorMessages<Data>) => ({
```
bluebill1049: nice

src/useForm.ts
```diff
@@ -170,28 +175,17 @@ export default function useForm<
...
-        const payloadName = payload.name as string;
-        errors = combineErrorsRef(
-          fieldErrors[payloadName]
-            ? { [payloadName]: fieldErrors[payloadName] }
-            : null,
-        );
-        result = !fieldErrors[payloadName];
-      }
+
+      const names = Array.isArray(payload)
```
bluebill1049: love how you combined them

src/useForm.ts
```diff
@@ -170,28 +175,17 @@ export default function useForm<
...
+        ? payload.map(({ name }) => name as string)
+        : [payload.name as string];
+      const validFields = names.filter(name => !fieldErrors[name]);
+      const errors = cleanUpErrors(combineErrorsRef(
```
bluebill1049: `validateWithSchema` will only return errors, why we need to clean up?

src/useForm.ts
```diff
@@ -170,28 +175,17 @@ export default function useForm<
...
+        Object.entries(fieldErrors)
+          .filter(([key]) => names.includes(key))
```
bluebill1049: if my top comment is valid, then we probably don't need this filter as well.

bluebill1049:
couple minor comments, thanks very much for the PR and helping the repo ❤️

src/useForm.ts
cfgj: because `errorsRef.current` used in `combineErrorsRef` contains the old ones and when you write `const { errors } = useForm()` errors inform about past errors

src/useForm.ts
cfgj: if I removed line 185 `useForm › triggerValidation with schema › should not trigger any error when schema validation result not found` would not pass

src/useForm.ts
bluebill1049: you are right! 🙏

bluebill1049: I think we shouldn't run `combineErrorRef` here at all then.

bluebill1049: nice pick up on the issue tho @cfgj

cfgj: I'll fix it later :)

bluebill1049: Thanks heaps 🙏🏻

cfgj: After analysis the code and the docs, my code works properly because:
- `validateWithSchema` validate all fields and we want `errors` contains only errors of touched fields, hence this filter in 185.
- We mustn't remove errors of invalid fields, so the use of `combineErrorsRef` is justified.

Am I right?

bluebill1049: cool thanks @cfgj I will get it in and do a double-check, I think you are right. cheers for the contribution.

bluebill1049:
Screen Shot 2019-08-03 at 8 38 07 am
thank you :)

bluebill1049:
published a beta version if you have time to help test as well :) react-hook-form@3.21.6-beta.1

bluebill1049:
thanks, 🙏  it's published and tested.

PR #196, opened 2019-08-07
Change detached detection (and added tests) to support iFrame nesting

Nickman87:
See https://github.com/bluebill1049/react-hook-form/issues/193 for more information


bluebill1049:
@barrymay @stramel you guys have to take a look as well? 👀

src/utils/isDetached.ts
```diff
@@ -0,0 +1,10 @@
+import { Ref } from '../types';
+
+export default function isDetached(element: Ref): boolean {
+  // null elements are detached (probably a null parent)
+  if (!element) return true;
+  // If we can find our way up to a document node, we assume we are still attached!
+  if (element.nodeType === Node.DOCUMENT_NODE) return false;
```
bluebill1049: what happens when input move outside of `form`, that you no longer refer to the parent node?

barrymay: To be fair here @bluebill1049  (and I may be missing something, so pls correct me if I'm wrong!) the existing code that this replaces

```
  const isRefDeleted = !document.body.contains(ref);
```

didn't check if it was outside the form.

I can see how this simplfies the function. there were checks like
```
} else if (element.parentNode === null) {
```

in the old function (line 10 below) which are unnecessary since they were turn true on the next recursive call.

I think this change looks ok, assuming all other tests pass.

src/logic/findRemovedFieldAndRemoveListener.ts
```diff
@@ -15,13 +16,13 @@ export default function findRemovedFieldAndRemoveListener<
...
   const { name, type } = ref;
-  const isRefDeleted = !document.body.contains(ref);
+  const isRefDetached = isDetached(ref);
   touchedFieldsRef.current.delete(name);
   fieldsWithValidationRef.current.delete(name);
 
   if (isRadioInput(type) && options) {
     options.forEach(({ ref }, index): void => {
-      if (ref instanceof HTMLElement && isRefDeleted && options[index]) {
+      if (ref instanceof HTMLElement && isRefDetached && options[index]) {
```
barrymay: I think the check for `ref instanceOf HTMLElement` can be moved to line 19 for simplicity (not a blocker)

src/utils/isDetached.ts
bluebill1049: valid point! @barrymay guess I was too worried over this change, I will do some functional test with it. in the meantime, @Nickman87 if you need this for your project, you can start using it via https://github.com/bluebill1049/react-hook-form/pull/196/checks?check_run_id=187948976

src/utils/isDetached.test.ts
```diff
@@ -0,0 +1,112 @@
+import isDetached from './isDetached';
```
barrymay: Very nice adds on the testing!

bluebill1049: totally agree, lovely unit test here as well

src/logic/findRemovedFieldAndRemoveListener.ts
Nickman87: Looks cleaner to me, keeps the dom checking together

bluebill1049: what about putting this check inside `isDetachted`, you don't have to do that in line 35 as well/

bluebill1049:
FYI @Nickman87 I am going to merge this over the weekend end and test it out.

Nickman87:
@bluebill1049 I made the change and found/fixed a small bug in the process ;)

src/utils/isDetached.ts
```diff
@@ -3,6 +3,8 @@ import { Ref } from '../types';
 export default function isDetached(element: Ref): boolean {
   // null elements are detached (probably a null parent)
   if (!element) return true;
+  // None HTML Elements should never be considered detached
+  if (!(element instanceof HTMLElement)) return false;
```
bluebill1049: cool 👍

src/logic/findRemovedFieldAndRemoveListener.ts
```diff
@@ -16,13 +16,12 @@ export default function findRemovedFieldAndRemoveListener<
...
 
   const { name, type } = ref;
-  const isRefDetached = isDetached(ref);
   touchedFieldsRef.current.delete(name);
   fieldsWithValidationRef.current.delete(name);
 
   if (isRadioInput(type) && options) {
     options.forEach(({ ref }, index): void => {
-      if (ref instanceof HTMLElement && isRefDetached && options[index]) {
+      if (options[index] && isDetached(ref)) {
```
barrymay: nice find ;)

src/utils/isDetached.test.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+
+  it('should return true when the node was never attached to the main document', () => {
+    document.body.innerHTML = ''; // Make sure the body is empty
+    const node = document.createElement('div');
+
+    expect(isDetached(node)).toBeTruthy();
+  });
+
+  it('should return false when the given element is not an HTMLElement', () => {
```
barrymay: these isDetached tests make me really want to at some point set a proper type for the Ref, but it's admittedly tricky to get right

barrymay (approved):

bluebill1049:
thanks for approving @barrymay ❤️

bluebill1049:
i will fix the test.

PR #197, opened 2019-08-07
Fix triggerValidation when using schema

Nickman87:
The method `triggerValidation` returns a boolean indicating wether the requested fields were valid or not. However, when using a schema, it returns a boolean based on all the fields, even the ones that were not requested. As far as I can tell, it works fine when using basic validation, as this only loops over the requested fields to create a response.

I added two test cases for this and made a slight modification to `executeSchemaValidation` to correct this.


bluebill1049:
thanks very much for the fix! merging

Issue #211, opened 2019-08-14
[REQ] Ability to return form and field level errors from submission 


stramel:
**Is your feature request related to a problem? Please describe.**
This is a feature request for being able to pass the form errors from a submission.

**Describe the solution you'd like**
I would like to see the callback passed to `handleSubmit` be able to handle being returned an object of errors that could be merged into the existing `errors` object. Including a special case for an overall form-level error.

**Describe alternatives you've considered**
I tried manually implementing something like this and it requires a lot of work from the implementer's side.  

/cc @bluebill1049 @barrymay

bluebill1049:
@stramel can you give some code example? I am not very clear in term os requirement 🤔

stramel:
@bluebill1049 Here is some rough pseudo code.

```jsx
const { register, errors, handleSubmit } = useForm()

async function submit() {
	const response = await loginApi()
	if (response.errors) {
		return response.errors.password // 'Invalid Password'
	}
	// navigate
}


<form onSubmit={handleSubmit(submit)}>
	<input name="username" ref={ref} />
	{errors && errors.username}
	<input name="password" ref={ref} />
	{errors && errors.password} // This should display `'Invalid Password'`
</form>
```

Where the error response from the login call, would display in the `errors` object with the associated input name.

As for the form-level error, it would be similar but instead of an input name as the key in the object it would use something like `const FORM_ERROR = ''` and would be able to use it like...

```
{
	[FORM_ERROR]: 'Unexpected Error'
}
```

bluebill1049:
can you use `setError`?

stramel:
@bluebill1049 I think `setError` would work for the form-level error. Although, I like the DX of being able to return the object from the submission callback. Especially since in my case the submit function is coming from outside the function with the form.

bluebill1049:
sure i can take a look at that

bluebill1049:
@stramel basically, anything return from the `handleSubmit` will merge with errors? (am I on the right track?)

stramel:
That's what I was thinking. If we want, it could be limited to registered fields.

bluebill1049:
https://github.com/react-hook-form/react-hook-form/releases/tag/v3.21.17

stramel:
Shoot forgot to mention that I didn't test how this worked with nested naming

bluebill1049:
I think it's fine, it will just extend, right?

bluebill1049:
[...]

PR #213, opened 2019-08-15
#211 merge errors into return data from handleSubmit

bluebill1049:
right now `handleSubmit(callback)` `callback` return doesn't take any effects as it typed as `void`.

This PR will allow return data to merge into `errors` object, hence make DX better for async data submit with `errors`

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -9,10 +9,12 @@ export type NumberOrString = number | string;
 
 export type DataType = Record<string, FieldValue>;
 
+export type FieldErrors = Record<string, string>;
+
 export type OnSubmit<Data extends DataType> = (
   data: Data,
   e: React.SyntheticEvent,
-) => void;
+) => void | FieldErrors;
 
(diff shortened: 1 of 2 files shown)


stramel:
Oooooo 🧐 I will take a look here when I get home.

bluebill1049:
# NPM
npm install https://github.pika.dev/react-hook-form/react-hook-form/pr/213

# Yarn
yarn add https://github.pika.dev/react-hook-form/react-hook-form/pr/213

src/types.ts
```diff
@@ -9,10 +9,12 @@ export type NumberOrString = number | string;
 
 export type DataType = Record<string, FieldValue>;
 
+export type FieldErrors = Record<string, string>;
```
stramel: As long as we are allowing any key, this type should be fine, if we want to make it more restrictive, we will need to update this type.

stramel (approved):

stramel:
Shoot forgot to mention that I didn't test how this worked with  nested naming

PR #214, opened 2019-08-15
Form context types

### LINKED ISSUE #212: Question: FormContext Types
I am new to react-hook-form.  Doing an evaluation.  

I notice that `useForm` takes a generic parameter to shape various aspects such as `errors`.  Makes it possible to do: 
```ts
const form = useForm<Person>()
const firstNameHasError = !!form.errors.firstName
// or
const name = form.getValues().firstName
// both provide intellisense and type checking.  However,

const form = useFormContext()
const name = form.getValues().firstName
//works but no intellisense or type checking
 
```

Was hoping to use the `useFormContext` so that the form can work across child components.  However, I am unable to provide a type to `useFormContext`.  

Am I missing something?  Is this something that will be coming?  Is it even possible?  Should I be looking at a different pattern?  Such as passing as a prop the object that comes off of `useForm`?

stramel:
Fixes #212 

@barrymay @bluebill1049 Can you take a look at this? I couldn't find a good way to handle generics on context. It really bugs me that I have a `@ts-ignore`.

In this, I exposed `FormState` as a type and attempted to provide generics through the context API.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -8,5 +8,6 @@
   "prettier.eslintIntegration": true,
   "eslint.enable": true,
   "editor.formatOnSave": true,
-  "eslint.autoFixOnSave": true
+  "eslint.autoFixOnSave": true,
+  "typescript.tsdk": "node_modules/typescript/lib"
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -62,7 +62,7 @@ export type FieldsObject<Data extends DataType> = {
 
 export interface ReactHookFormError {
   ref: Ref;
-  type: string;
+  type?: string;
   message?: string;
   isManual?: boolean;
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -110,7 +110,7 @@ export interface FormProps<
   Data extends DataType = DataType,
   Name extends keyof Data = keyof Data,
   Value = Data[Name]
-> extends FormContextValues<Data,Name,Value> {
+> extends FormContextValues<Data, Name, Value> {
   children: JSX.Element[] | JSX.Element;
 }
 
@@ -125,11 +125,11 @@ export interface FormContextValues<
   ) => any;
   unregister: (name: string | string[]) => void;
   handleSubmit: (
-    callback: (data: any, e: React.SyntheticEvent) => void,
+    callback: OnSubmit<Data>,
   ) => (e: React.SyntheticEvent) => Promise<void>;
   watch: (
-    fieldNames?: string | string[] | undefined,
-    defaultValue?: string | Partial<Data> | undefined,
+    fieldNames?: string | string[],
+    defaultValue?: string | Partial<Data>,
   ) => FieldValue | Partial<Data> | void;
   unSubscribe: VoidFunction;
   reset: VoidFunction;
--- a/src/types.ts
+++ b/src/types.ts
@@ -131,7 +131,6 @@ export interface FormContextValues<
     fieldNames?: string | string[],
     defaultValue?: string | Partial<Data>,
   ) => FieldValue | Partial<Data> | void;
-  unSubscribe: VoidFunction;
   reset: VoidFunction;
   clearError: (name?: Name | Name[]) => void;
   setError: (name: Name, type?: string, message?: string, ref?: Ref) => void;
--- a/src/types.ts
+++ b/src/types.ts
@@ -97,6 +97,15 @@ export interface ValidationPayload<Name, Value> {
   value?: Value;
 }
 
+export interface FormState {
+  dirty: boolean;
+  isSubmitted: boolean;
+  submitCount: number;
+  touched: string[] | object[];
+  isSubmitting: boolean;
+  isValid: boolean;
+}
+
 export interface FormProps<
   Data extends DataType = DataType,
   Name extends keyof Data = keyof Data,
@@ -125,14 +134,7 @@ export interface FormProps<
   ) => Promise<boolean>;
   getValues: (payload?: { nest: boolean }) => DataType;
   errors: DataType;
-  formState: {
-    dirty: boolean;
-    isSubmitted: boolean;
-    submitCount: number;
-    touched: string[] | object[];
-    isSubmitting: boolean;
-    isValid: boolean;
-  };
+  formState: FormState;
 }
 
 export type FormContextValues = Omit<FormProps, 'children'>;
--- a/src/types.ts
+++ b/src/types.ts
@@ -141,7 +141,7 @@ export interface FormContextValues<
   triggerValidation: (
     payload: ValidationPayload<Name, Value> | ValidationPayload<Name, Value>[],
   ) => Promise<boolean>;
-  getValues: (payload?: { nest: boolean }) => DataType;
-  errors: DataType;
+  getValues: (payload?: { nest: boolean }) => Data;
+  errors: ObjectErrorMessages<Data>;
   formState: FormState<Data, Name>;
 }
--- a/src/types.ts
+++ b/src/types.ts
@@ -62,7 +62,7 @@ export type FieldsObject<Data extends DataType> = {
 
 export interface ReactHookFormError {
   ref: Ref;
-  type?: string;
+  type: string;
   message?: string;
   isManual?: boolean;
 }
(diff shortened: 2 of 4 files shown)


bluebill1049:
Really great stuff here ❤️  would be dream come true if context API works with `typescript `

src/useFormContext.tsx
```diff
@@ -1,23 +1,23 @@
...
+
+export const FormGlobalContext = React.createContext<FormContextType | null>(
+  null,
 );
 
-export const useFormContext = () =>
-  React.useContext<FormContextValues>(FormGlobalContext);
+export function useFormContext() {
+  const context = React.useContext(FormGlobalContext);
+  // TODO: Warn?
```
bluebill1049: curious to know the warning

src/useFormContext.tsx
```diff
@@ -1,23 +1,23 @@
...
   const { children, ...rest } = props;
   return (
-    <FormGlobalContext.Provider
-      value={
-        {
-          ...rest,
-        } as any
-      }
-    >
+    // @ts-ignore
```
bluebill1049: is it because `children` props?

src/useFormContext.tsx
```diff
@@ -1,23 +1,23 @@
 import * as React from 'react';
-import { FormContextValues, DataType } from './types';
+import { FormContextValues, FormProps, DataType } from './types';
 
-export const FormGlobalContext = React.createContext<FormContextValues>(
-  {} as FormContextValues,
+type FormContextType<T extends DataType = DataType> = FormContextValues<T>;
+
+export const FormGlobalContext = React.createContext<FormContextType | null>(
```
barrymay: this is not new code, but I wonder why this is exported?

barrymay: I see the problem (after a bit of investigation) - something to do with a conflicting defaultValue Partial type

One potential alternative to ts-ignore since it describes what you expect that rest to be at this point (not a blocker IMO since we do want to get rid of the ts-ignores and leaving it screams of the problem) 
```
    <FormGlobalContext.Provider value={rest as FormContextType}>
```

barrymay (approved):
No blockers here, just questions

barrymay (dismissed):

bluebill1049:
thanks for the code review @barrymay 😍

barrymay:
@bluebill1049 @stramel just fyi, I'll be on vacation all next week (aiming to be off the grid!), so while I'm happy to help anytime, I'm kinda off limits starting Saturday, thru Monday the 26th.

barrymay:
I'll do a bit more digging to see if I can solve this ts-ignore, it bugs me too :)

bluebill1049:
thanks @barrymay 🙏 enjoy your holiday~! 🛩

barrymay:
thanks :) BTW Bill, you'll need to mark your review as approved in GitHub in order to merge it

bluebill1049:
@barrymay yea I will sort it out.

bluebill1049 (dismissed):

src/useFormContext.tsx
stramel: I think I'm going to drop the export of the context and will test a couple different ways to address this.

src/useFormContext.tsx
stramel: No, as Barry stated, it's due to some type mismatches inside the FormContextValues.

src/useFormContext.tsx
stramel: Not really sure how to handle this from the library perspective. 

Essentially, for development purposes, it would console.log a message if the hook isn't being used inside the `FormContext` element. Typically in apps we just remove it using dead code elimination and `process.env.NODE_ENV !== 'production'` as the wrapping conditional. This should be fine for now without the warning.

src/useFormContext.tsx
stramel: Yeah... it boils down to the `watch` function's type, something with the `Partial<Data>`

src/types.ts
```diff
@@ -110,21 +110,27 @@ export interface FormProps<
...
-    callback: (data: any, e: React.SyntheticEvent) => void,
+    callback: OnSubmit<Data>,
   ) => (e: React.SyntheticEvent) => Promise<void>;
   watch: (
-    fieldNames?: string | string[] | undefined,
-    defaultValue?: string | Partial<Data> | undefined,
+    fieldNames?: string | string[],
+    defaultValue?: string | Partial<Data>,
   ) => FieldValue | Partial<Data> | void;
-  unSubscribe: VoidFunction;
```
stramel: @bluebill1049 There is no `unSubscribe` being returned from `useForm`

stramel:
BLEH, this turned out pretty gross. There doesn't seem to be a good way to handle typing of the Context API. Right now, it requires 3 casts.

```
useForm<MyType>()
// ...
<FormContext<MyType> {...methods}>
// ...
useFormContext<MyType>()
```

src/types.ts
```diff
@@ -62,7 +62,7 @@ export type FieldsObject<Data extends DataType> = {
 
 export interface ReactHookFormError {
   ref: Ref;
-  type: string;
+  type?: string;
```
stramel: @bluebill1049 `type` was being required here but was optional in the actual implementation of `setError`.

src/types.ts
```diff
@@ -90,18 +90,21 @@ export type FieldErrors = Record<string, string>;
...
-export interface FormState {
+export interface FormState<
+  Data extends DataType = DataType,
+  Name extends keyof Data = keyof Data
+> {
   dirty: boolean;
   isSubmitted: boolean;
   submitCount: number;
-  touched: string[] | object[];
+  touched: Name[];
```
stramel: I couldn't find any instance where `touched` or `touchedRef` was anything more than just `Name`.

src/types.ts
bluebill1049: oh yeah forgot to remove that. 😅

src/types.ts
bluebill1049: should be required. again my mistake not to update context api type

src/types.ts
bluebill1049: you are right 🙏

.vscode/settings.json
```diff
@@ -8,5 +8,6 @@
   "prettier.eslintIntegration": true,
   "eslint.enable": true,
   "editor.formatOnSave": true,
-  "eslint.autoFixOnSave": true
+  "eslint.autoFixOnSave": true,
+  "typescript.tsdk": "node_modules/typescript/lib"
```
bluebill1049: curious what is this?

bluebill1049 (dismissed):

.vscode/settings.json
stramel: This just tells VSCode to use the typescript version that we have installed based on our package.json. So we are linting with the same version that we are compiling with.

src/types.ts
stramel: All good!

src/types.ts
stramel: It's alright, I will update. In the refactor, we won't have separate definitions so that should simplify it.

stramel:
@bluebill1049 Mind holding off on merging and releasing a beta for this one? I'm having issues with the PikaCI after publishing updates.

bluebill1049:
Sure will do 🤙🏻

bluebill1049:
Sorry have to do that when I am at work. Getting quite late today cause was looking at the other two issues hehe

bluebill1049:
looks like pika is ready now. you still want me to release a beta version?

stramel:
Will let you know once I get home

stramel:
Okay.... soooo in reality, generic typing of context is hard.

Only need to type the hooks. If you are only passing the methods to the `FormContext`, you don't need to type the `useForm`. The `useFormContext` is required to be typed if you want intelli-sense. I think this is the best we are going to get with the current setup.

I believe we can improve this setup without the extra `useFormContext` hook and not require passing methods through to the context.

stramel:
@bluebill1049 bundlesize doesn't seem to be running anymore?

bluebill1049:
I agree with you @stramel I didn't spend too much time on ContextForm cause I know it's difficult(according to my typescript skill level)... I think we can live with it with what you got. I can document it on the website.

bundlesize is weird, it keeps hanging on branches but works in master, so I took it down. it's still in the build pipeline tho. I will look into the fix tho.

stramel:
@bluebill1049 Does it only hang if the build fails? Maybe we can post an issue on bundlesize if it is an issue on their end.

stramel:
@bluebill1049 I'm good with merging this if you're good with the changes.

bluebill1049:
yea the build pass but the tick never appear. I will investigate it and see if it's our end or theirs. if it's on bundelszie I will raise an issue there :) 👍

bluebill1049 (approved):
Thanks for your amazing work and clean up ❤️

bluebill1049:
merge ❤️

PR #218, opened 2019-08-15
Allow OnSubmit event handler to return promises

bmamouri:
Simple change to allow async event handlers in type definitions

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -12,7 +12,7 @@ export type DataType = Record<string, FieldValue>;
 export type OnSubmit<Data extends DataType> = (
   data: Data,
   e: React.SyntheticEvent,
-) => void;
+) => Promise<void>;
 
 export type Mode = keyof typeof VALIDATION_MODE;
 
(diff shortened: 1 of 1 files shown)


bluebill1049:
Thanks for the PR @bmamouri looks like this change break the build (automation).

Screen Shot 2019-08-15 at 6 59 50 pm

bluebill1049:
probably should be void | Promise<void>

src/types.ts
```diff
@@ -12,7 +12,7 @@ export type DataType = Record<string, FieldValue>;
 export type OnSubmit<Data extends DataType> = (
   data: Data,
   e: React.SyntheticEvent,
-) => void;
+) => Promise<void>;
```
stramel: `void | Promise<void>`

stramel (changes requested):

bluebill1049:
@stramel those automation tests is pretty useful here 😺

src/types.ts
stramel: @bmamouri You should be able to just accept this suggested change now :)

bluebill1049:
@bmamouri can you please accept the change and we can merge it in :)?

bmamouri:
@bluebill1049, Great library mate! Good to see that you are from Sydney too.

@stramel, Thanks for the fix

bluebill1049 (approved):

bluebill1049:
thanks mate @bmamouri and your contribution ❤️

PR #225, opened 2019-08-17
Provide Register Overloads

stramel:
Just trying to figure out how to get the function overloads from typescript working with react. Hoping to do some more cleanup on it after the fact.

- Switched to `useLayoutEffect` instead of `useEffect` for the `isUnmounted` check, since it closer follows what you expect.
- Starting to clamp down on typings which caused some issues without fully reworking. Hope to cover any `@ts-ignore` or `any` that I added in future PRs.

Function Overloads are fixed for `register` and _shouldn't_ cause additional renders.


bluebill1049:
Took off barry, he can enjoy his lovely holiday :)

bluebill1049:
bloody great effort, rest... please 🙏

package.json
```diff
@@ -10,7 +10,7 @@
     "build:min": "rollup -c rollup.min.config.js",
     "watch": "tsc --watch",
     "release": "npm version",
-    "lint": "eslint ./src --ext .jsx,.ts --ignore-pattern *.test.ts",
+    "lint": "eslint ./src --ext .jsx,.ts --ignore-pattern *.test.ts --report-unused-disable-directives",
```
bluebill1049: curious if we can have this in the eslint config

src/logic/shouldUpdateWithError.ts
```diff
@@ -1,17 +1,20 @@
 import isEmptyObject from '../utils/isEmptyObject';
-import { DataType } from '../types';
+import { TFormValues, ObjectErrorMessages } from '../types';
```
bluebill1049: TFormValues why we need this `T` ? generic?

package.json
stramel: > This is only available as a command-line flag at the moment, and can't be enabled in a configuration file. (Since this is not a rule, it was decided in #6190 that it shouldn't be configured in the same way as a rule.)

https://github.com/eslint/eslint/issues/9249#issuecomment-333240295

Doesn't look like it.

src/logic/shouldUpdateWithError.ts
stramel: We don't need the `T`, I was using it to denote that as a difference between what we use for the base vs. what the user passes in. In reality, I should probably drop `T` since it isn't a generic type. I want to keep `FormValues`, `FieldName`, and `FieldValue` as the types that we use for the user's types.

Any thoughts on other names? `BaseFormValues` or `DefaultFormValues`?

stramel:
I figured out the string issue that we were having. Hoping to have that fixed in here too.

src/logic/shouldUpdateWithError.ts
stramel: `TFieldName` is going to be generically typed so maybe better for consistency to just leave it as `TFormValues`?

stramel:
Tried to cover too much in this PR. Going to try to cover some of the type enforcement later.

bluebill1049:
Yea sure. I do agree.

PR #229, opened 2019-08-19
Fix getMultipleSelectValue for HTMLOptionsCollection

tizz98:
I've been getting the following error when using this library with `react-bootstrap` (https://react-bootstrap.github.io/)

```
options.filter is not a function
```

This is because the ref being passed to `register` is actually `HTMLOptionsCollection` not `HTMLOptionElement[]`. I fixed this temporarily by hacking `HTMLOptionsCollection` but it would be nice not to have to do that.


src/logic/getMultipleSelectValue.ts
```diff
@@ -1,2 +1,5 @@
-export default (options: HTMLOptionElement[]): string[] =>
+const filterOptions = (options: HTMLOptionElement[]): string[] =>
```
bluebill1049: export default (options: HTMLOptionElement[] | HTMLOptionsCollection): string[] =>	const filterOptions = (options: HTMLOptionElement[]): string[] =>
  [...options].filter(({ selected }): boolean => selected).map(({ value }): string => value);

bluebill1049 (changes requested):

bluebill1049:
thanks very much for the PR ❤️

src/logic/getMultipleSelectValue.ts
bluebill1049: I think we can simplify the above to what I suggested :)

stramel: agreed

bluebill1049:
thank you :)

bluebill1049:
@tizz98 I will publish the patch tonight. if you need it now. you can use https://cdn.pika.dev/react-hook-form/master

bluebill1049:
published :) https://github.com/react-hook-form/react-hook-form/releases/tag/v3.22.0

PR #232, opened 2019-08-19
Function overloads

stramel:
This should just cover the `register` function overloads primarily.

I also swapped `useEffect` to `useLayoutEffect`.


src/useForm.ts
```diff
@@ -653,16 +666,16 @@ export default function useForm<
     return isEmptyObject(output) ? defaultValues : output;
   };
 
-  useEffect(
-    (): VoidFunction => () => {
+  useLayoutEffect(
+    () => () => {
```
bluebill1049: What’s the reason of using layout effect here?

src/useForm.ts
```diff
@@ -483,24 +395,125 @@ export default function useForm<
     );
   }
 
-  const register = useCallback(
-    (refOrValidateRule: RegisterInput | Ref, validateRule?: RegisterInput) => {
-      if (!refOrValidateRule || typeof window === 'undefined') return;
+  function __registerIntoFieldsRef(
```
bluebill1049: Can we not using __ for function name?

bluebill1049:
Thanks @stramel for the awesome PR I will take a closer look tonight

src/useForm.ts
stramel: > If you’re migrating code from a class component, note useLayoutEffect fires in the same phase as componentDidMount and componentDidUpdate.

https://reactjs.org/docs/hooks-reference.html#uselayouteffect

It is recommended if you're doing mount/unmount checks to use `useLayoutEffect` instead.

src/useForm.ts
stramel: We can. I was using it for creating a distinction between the public and private methods. I will leave it up to you.

src/useForm.ts
bluebill1049: nice! 

> However, we recommend starting with useEffect first and only trying useLayoutEffect if that causes a problem.

I think we are fine to use `useLayout` here. 

I also read this article: https://kentcdodds.com/blog/useeffect-vs-uselayouteffect

> useLayoutEffect: If you need to mutate the DOM and/or DO need to perform measurements
> useEffect: If you don't need to interact with the DOM at all or your DOM changes are unobservable (seriously, most of the time you should use this). 

sounds like `useEffect` is what we want.

src/useForm.ts
bluebill1049: cool! I get why you doing it here, but we are not writing a class in here. if we do __, then rest of them should all be __ hehe prefer to keep it clean without the __ (my personal opinion here)

stramel: Right, I planned on doing it to the rest, but you mentioned you wanted me to only update a bit at a time. So it would be a progressive shift.

src/types.ts
```diff
@@ -122,10 +122,8 @@ export interface FormContextValues<
   Name extends keyof Data = keyof Data,
   Value = Data[Name]
 > {
-  register: (
-    refOrValidateRule: RegisterInput | Ref,
-    validateRule?: RegisterInput,
-  ) => any;
+  register(validateRule: RegisterInput): (ref: Ref) => void;
+  register(input: Ref, validationOptions?: RegisterInput): void;
```
bluebill1049: nice ❤️

src/useForm.ts
```diff
@@ -483,24 +395,125 @@ export default function useForm<
...
+      });
+    }
+  }
+
+  function register(validateRule: RegisterInput): (ref: Ref) => void;
+  function register(input: Ref, validationOptions?: RegisterInput): void;
+  function register(
+    refOrValidateRule: RegisterInput | Ref,
+    validationOptions?: RegisterInput,
+  ): void | ((ref: Ref) => void) {
```
bluebill1049: 😍

src/types.ts
stramel: Actually hoping to drop this in the long run. It is creating a double type definition. But for now, this is necessary.

src/useForm.ts
```diff
@@ -483,24 +395,125 @@ export default function useForm<
...
+    if (typeof window === 'undefined' || !refOrValidateRule) {
+      if (process.env.NODE_ENV !== 'production') {
```
bluebill1049: 👍

src/useForm.ts
stramel: Alright, I can revert it.

stramel:
Typings seem to be working now. Let me know what's left you want to clean up @bluebill1049

src/useForm.ts
stramel: Also, with `mangle` on prefix `__` will allow us to trim a bit more off the top.

bluebill1049: not sure what do you mean by that.

src/useForm.ts
```diff
@@ -483,24 +395,125 @@ export default function useForm<
...
+        if (typeof window === 'undefined') {
+          console.warn('⚠ Currently, only browser usage is supported.');
```
bluebill1049: can we remove this warning for the server-side?

src/useForm.ts
```diff
@@ -483,24 +395,125 @@ export default function useForm<
...
-    [registerIntoFieldsRef],
-  );
+    if (
+      isObject(refOrValidateRule) &&
+      (validationOptions || 'name' in refOrValidateRule)
+    ) {
+      return __registerIntoFieldsRef(refOrValidateRule, validationOptions);
+    }
+
+    return (ref?: Ref | null) => {
```
bluebill1049: why we need this to be optional, it should always register with object, then you can remove the if check below.

bluebill1049:
minor comments, if you could address or reply.

src/useForm.ts
stramel: We can remove it. It _should_ get removed from production code in most build setups anyways.

src/useForm.ts
stramel: You had a check for if it was defined previously, a test that explicitly passed undefined, and the typing didn't reflect that. Maybe we need to change the logic then

src/useForm.ts
bluebill1049: sure, maybe annoying when ppl under dev mode.

src/useForm.ts
bluebill1049: oh, I see, that's my bad. `register` should always have an argument.

src/useForm.ts
bluebill1049: I am googling `mangle`, i didn't know what it was hehe

src/useForm.ts
stramel: Okay, I will make those changes.

bluebill1049: thanks @stramel

src/useForm.ts
stramel: Right, so the `mangle` option will allow the minifier to change the name of the function. That's why it's important to be able to use `__` prefix on private functions.

src/useForm.ts
stramel: Ended up having to allow `null` but it is required now.

PR #234, opened 2019-08-20
Replace enzyme

stramel:
Figured this might be a good improvement.

`@testing-library/react-hooks` is a nice library for testing the hooks. I just directly ported what you previously had to the new library. We can improve it in the future.

Also, `@testing-library` has a cypress lib.


bluebill1049 (approved):

bluebill1049:
Too good @stramel bye enzyme

PR #235, opened 2019-08-20
Add IE11 bundle

stramel:
Builds out a `react-hook-form.ie11.js` bundled file along with the CJS and ESM builds.

Adds minification to the build as well.


bluebill1049 (approved):
Thanks! this saves me a lot of time to generate IE 11 lib. I will update the documentation site soon as well.

`import useFrom from 'react-hook-form/dist/react-hook-form.ie11'`

PR #237, opened 2019-08-20
Clean up config rollup config

stramel:
Keep a single copy of the config and utilize the package name for the files.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -22,7 +22,6 @@ export function getConfig({
   return {
     input: 'src/index.ts',
     external: ['react', 'react-dom'],
-    // export: 'named',
     plugins: [
       typescript({
         tsconfig,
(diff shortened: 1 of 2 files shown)


rollup.config.js
```diff
@@ -1,29 +1,42 @@
...
+  output = outputConfig,
+} = {}) {
+  return {
+    input: 'src/index.ts',
+    external: ['react', 'react-dom'],
+    plugins: [
+      typescript({
+        tsconfig,
+      }),
+      terser({
```
bluebill1049: We only want this for ie11 build right?

stramel: Nah, we should do it for all. It won't hurt anything to minify the code. Most libraries do.

bluebill1049: interesting :) let's see if it exceed 5kb hehe

bluebill1049: think you are right hehe

PR #238, opened 2019-08-20
Enhance CircleCI Config

stramel:
Cleaning up the CircleCI config.

Re-add minification to builds.


package.json
```diff
@@ -5,10 +5,10 @@
   "module": "dist/react-hook-form.es.js",
   "types": "dist/index.d.ts",
   "scripts": {
-    "clean": "rm -rf dist/* && rm -rf .rpt2_cache",
-    "build": "yarn clean && yarn build:modern && yarn build:ie11",
+    "clean": "rimraf dist",
```
bluebill1049: do we have `rimraf`? or we should just use `rm`?

stramel: we have `rimraf`, `rm` doesn't function well on Windows

PR #241, opened 2019-08-21
Fix register return type

stramel:
`ref` typings didn't match what we were returning.

Going to test this change locally post-build.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -405,8 +405,8 @@ export default function useForm<
     );
   }
 
-  function __registerIntoFieldsRef(
-    elementRef: Ref,
+  function __registerIntoFieldsRef<Element extends ElementLike>(
+    elementRef: Element,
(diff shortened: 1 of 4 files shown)


src/useForm.ts
```diff
@@ -404,8 +405,8 @@ export default function useForm<
     );
   }
 
-  function __registerIntoFieldsRef(
-    elementRef: Ref,
+  function __registerIntoFieldsRef<Element extends ElementLike>(
```
bluebill1049: didn't know this `ElementLike`,pretty cool

stramel: I made it 😝 to better match that we allow `{ name: string}` instead of full element objects.

bluebill1049:
Please do testing on this :)

stramel:
Yeah definitely needs more work still.

PR #246, opened 2019-08-22
improve on getting value

bluebill1049:
- make sure getting value always go through `value[name]` before using `get` function

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-hook-form",
-  "version": "3.22.1",
+  "version": "3.22.2-beta.1",
   "main": "dist/react-hook-form.js",
   "module": "dist/react-hook-form.es.js",
   "types": "dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
   },
   "bundlesize": [
     {
-      "path": "./dist/react-hook-form.es.js",
+      "path": "./dist/react-hook-form.min.es.js",
       "maxSize": "5 kB"
     }
   ],
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-hook-form",
-  "version": "3.22.2-beta.1",
+  "version": "3.22.2-beta.4",
   "main": "dist/react-hook-form.js",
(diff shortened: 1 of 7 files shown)


src/logic/getValueWithDefaults.ts
```diff
@@ -0,0 +1,9 @@
+import get from '../utils/get';
+import { FieldValue } from '../types';
+
+export default function getValueWithDefault<
+  Data extends FieldValue,
```
bluebill1049: @stramel can you please double-check the type in here?

src/logic/getDefaultValue.ts
```diff
@@ -1,8 +1,9 @@
...
-import { DataType } from '../types';
+import { FieldValue } from '../types';
 
-export default (
-  defaultValues: DataType | undefined,
-  name: string,
-): DataType | string | number | boolean =>
-  defaultValues && (defaultValues[name] || get(defaultValues, name));
+export default function getDefaultValue<
+  Data extends FieldValue,
```
bluebill1049: @stramel please double-check the type 🙏

rollup.config.js
```diff
@@ -23,14 +23,6 @@ export function getConfig({
         tsconfig,
         clean: true,
       }),
-      terser({
```
bluebill1049: this uglify plugin is the root cause of the last release. I am removing it, I think it screws up the import and export for some reason, I didn't look too deep into it.

package.json
```diff
@@ -9,6 +9,7 @@
     "prebuild": "yarn clean",
     "build": "yarn build:modern && yarn build:ie11",
     "build:modern": "rollup -c",
+    "build:min": "rollup -c rollup.min.config.js",
```
bluebill1049: keep min bundle as separate to prevent screwing up the default export

rollup.config.js
bluebill1049: 3.22.2-beta.2 is the version with terser, try it in codesandbox or any env you will see the same error about default export.

bluebill1049:
@stramel looks like you are busy, I will get this one in, you can add comments on the commit in master.

PR #257, opened 2019-08-25
fix register type

bluebill1049:
can you take a look if this fix the type?

### DIFF
--- a/src/contextTypes.ts
+++ b/src/contextTypes.ts
@@ -26,8 +26,8 @@ export interface FormContextValues<
   Value = Data[Name]
 > {
   register<Element extends ElementLike = ElementLike>(
-    ref: Element | null,
-  ): (validateRule: ValidationOptions) => void;
+    validateRule: ValidationOptions,
+  ): (ref: Element | null) => void;
   register<Element extends ElementLike = ElementLike>(
     ref: Element | null,
     validationOptions?: ValidationOptions,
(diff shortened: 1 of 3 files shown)


bluebill1049:
if it's the problem with return, maybe we just return `any` for now?

bluebill1049:
also how do I reproduce the return issue? it passes the unit test :(

src/contextTypes.ts
```diff
@@ -27,11 +27,15 @@ export interface FormContextValues<
...
   register<Element extends ElementLike = ElementLike>(
     validateRule: ValidationOptions,
-  ): (instance: Element | null) => void;
+  ): (ref: Element | null) => void;
   register<Element extends ElementLike = ElementLike>(
-    instance: Element,
+    ref: Element | null,
     validationOptions?: ValidationOptions,
   ): void;
+  register<Element extends ElementLike = ElementLike>(
```
stramel: This declaration isn't needed I believe

src/useForm.ts
```diff
@@ -524,13 +524,13 @@ export default function useForm<
...
+    ref: Element | null,
     validationOptions?: ValidationOptions,
-  ): undefined;
+  ): void;
   function register<Element extends ElementLike = ElementLike>(
-    refOrValidateRule: ValidationOptions | Element,
+    refOrValidateRule: ValidationOptions | Element | null,
     validationOptions?: ValidationOptions,
-  ): ((ref: Element | null) => void) | undefined {
+  ): ((ref: Element | null) => void) | void {
```
stramel: I thought this gave me an error when I had it like this

PR #275, opened 2019-08-30
fix validation fields

bluebill1049:


src/useForm.ts
```diff
@@ -260,7 +262,11 @@ export default function useForm<
   validateAndStateUpdateRef.current = validateAndStateUpdateRef.current
     ? validateAndStateUpdateRef.current
     : async ({ target: { name }, type }: Ref): Promise<void> => {
-        if (isArray(validationFields) && !validationFields.includes(name))
+        console.log(validationFieldsRef.current);
```
vincentdesmares: I don't think this debug should have been commited.

vincentdesmares (changes requested):

src/useForm.ts
bluebill1049: haha, yea thanks for picking up this :)

PR #281, opened 2019-09-01
fix fieldValidate and clean up

bluebill1049:


src/useForm.ts
```diff
@@ -567,12 +566,12 @@ export default function useForm<
       isObject(refOrValidateRule) &&
       (validationOptions || 'name' in refOrValidateRule)
     ) {
-      __registerIntoFieldsRef(refOrValidateRule as Element, validationOptions);
+      registerIntoFieldsRef(refOrValidateRule as Element, validationOptions);
```
bluebill1049: FYI @stramel I am changing back this `__`, cause we are not leveraging `mangle` just yet.

bluebill1049: we can roll that back once we have a working version with mangle :)

PR #282, opened 2019-09-01
Change parameter naming

JeromeDeLeon:
To avoid conflicting parameter name with function name and for consistency, since you use `validationSchemaOption` in the `useForm`.


bluebill1049:
Oh my god, too good :) haha finally got a PR from you @JeromeDeLeon Thank you ❤️

JeromeDeLeon:
It would be better to assign another reviewer for every commit or changes so that we don't regret it later. What do you think?

bluebill1049:
It's good :) merged

PR #289, opened 2019-09-03
Add package description for better npm tooling display

barrymay:
Create this PR because when looking at react-hook-form via the npm extension in VSCode, it gives a poor tooltip response that is likely fixed via a proper package.json description


bluebill1049:
thanks @barrymay plan to release a patch with all the update this Friday :)

PR #290, opened 2019-09-04
close #283 fix performance issue

### LINKED ISSUE #283: Is it possible to validate fields only on `blur` and `submit` and not `change` events?
I need to validate fields only on `blur` and `submit` events. Even when I choose `mode = onBlur`, still it validates the field both on `change` and `blur` events. The validation in `change` event is no needed for me. As far as I understand, `mode = onBlur` should validate only on `blur`. Am I right?

Make validation only on `blur` and `submit` possible.

bluebill1049:
- patch the performance issue with `onBlur` mode and non `onBlur` mode

### DIFF
--- a/src/logic/findRemovedFieldAndRemoveListener.test.ts
+++ b/src/logic/findRemovedFieldAndRemoveListener.test.ts
@@ -8,8 +8,7 @@ describe('findMissDomAndClean', () => {
   beforeEach(() => {
     jest.resetAllMocks();
 
-    // @ts-ignore
-    isDetached.mockImplementation(() => {
+    (isDetached as any).mockImplementation(() => {
       return true;
     });
   });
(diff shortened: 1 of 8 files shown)


src/logic/validateField.test.ts
```diff
@@ -447,8 +446,7 @@ describe('validateField', () => {
       },
     });
 
-    // @ts-ignore
-    getRadioValue.mockImplementation(() => {
```
barrymay: you might as well leave the ts-ignore so we don't forget it ;)

src/logic/findRemovedFieldAndRemoveListener.test.ts
```diff
@@ -21,8 +20,7 @@ describe('findMissDomAndClean', () => {
     };
     expect(
       findRemovedFieldAndRemoveListener(
-        // @ts-ignore
```
barrymay: see my other comment of ts-ignore vs any

bluebill1049: I agree @barrymay will leave those `ts-ignore` :)

PR #299, opened 2019-09-08
[WIP] Trying different context approach

stramel:
**Do not merge**

Ideally, this would result in a more minimal API change when using the `FormContext`.

**Before:**
```jsx
import React from "react"
import useForm, { FormContext, useFormContext } from "react-hook-form"

function YourForm() {
  const methods = useForm()
  const onSubmit = data => { console.log(data) }

  return (
    <FormContext {...methods} > // pass all methods into the context
      <form onSubmit={methods.handleSubmit(onSubmit)}>
        <NestedInput />
        <input type="submit" />
      </form>
    </FormContext>
  )
}

function NestedInput() {
  const { register } = useFormContext() // retrieve all hook methods
  return <input name="test" ref={register} />
}
```

**After:**
```jsx
import React from "react"
import useForm, { FormContext } from "react-hook-form"

function YourForm() {
  const { handleSubmit } = useForm()  // retrieve all hook methods
  const onSubmit = data => { console.log(data) }

  return (
      <NestedInput />
      <input type="submit" />
    </form>
  )
}

function NestedInput() {
  const { register } = useForm() // retrieve all hook methods
  return <input name="test" ref={register} />
}

function MyPage() {
	return (
		<FormContext mode="onSubmit" defaultValues={{ test: 'foobar' }} >
			<YourForm />
		</FormContext>
	)
}
```

This means the change necessary to start using the `FormContext` is to add the options to the `FormContext` instead of the `useForm` and wrap your form with the `FormContext`. This drops the need for `useFormContext` hook.

Looking for input.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,2 +1,2 @@
 export { default } from './useForm';
-export { FormContext, useFormContext } from './useFormContext';
+export { FormContext } from './useFormContext';
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -38,8 +38,9 @@ import {
   ValidationPayload,
   ElementLike,
 } from './types';
+import { useFormContext } from './useFormContext';
 
-export default function useForm<
+export default function useCreateForm<
   FormValues extends FieldValues,
   FieldName extends keyof FormValues = keyof FormValues
 >({
@@ -783,4 +784,13 @@ export default function useForm<
           }),
     },
   };
+}
+
+export function useForm<FormValues extends FieldValues>(
+  options: Options<FormValues>,
+) {
+  const formContext = useFormContext();
+  const form = useCreateForm<FormValues>(options);
+
+  return formContext || form;
 }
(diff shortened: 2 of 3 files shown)


barrymay:
Interesting, so in this change, the idea is that useForm on an inner component can pull in the nearest form parent's context?

IMO, the useForm vs useFormContext difference has a clearer meaning, giving that the useFormContext has a distinct parent/child relationship to it (hence the use of context in the name)

Have people been complaining about it?

barrymay:
I feel If FormContext is staying, then it makes sense for useFormContext to pull from it.  If we didn't need FormContext as a component somehow, then I agree it's not necessary.  (just my two cents, I haven't honestly use this model yet, but I understand the benefit for sure!)

bluebill1049:
I think I agree with @barrymay's point of view. I think this lib need more love on the typescript at the moment. maybe we can focus on that? I made a PR today and @barrymay already left few comments and i agree one of the concern is I am adding more `as` when I am trying to fix the type. would be great to get your guys' help on that since you guys r better at TS than I do ❤️

PR #300, opened 2019-09-08
improve on setValue performance

bluebill1049:
trigger less re-render on setValue


stramel (dismissed):
Looks alright to me since `setFieldValue` isn't public. Although, still feels strange that it is returning a boolean

bluebill1049:
@stramel valid concern, it's a flag to trigger re-render or skip. let me know if you have better suggestion.

PR #303, opened 2019-09-10
Improve/types

bluebill1049:
try to clean up some of the types. please review :) more of progressively improve on it. bit by bit

### DIFF
--- a/src/contextTypes.ts
+++ b/src/contextTypes.ts
@@ -8,7 +8,6 @@ import {
   Ref,
   ValidationOptions,
   ValidationPayload,
-  VoidFunction,
 } from './types';
 
 export interface FormProps<
@@ -42,7 +41,7 @@ export interface FormContextValues<
     fields: (Name | string)[],
     defaultValues?: Partial<Data>,
   ): Partial<Data>;
-  reset: VoidFunction;
+  reset: () => void;
   clearError: (name?: Name | Name[]) => void;
   setError: (name: Name, type: string, message?: string, ref?: Ref) => void;
   setValue: (name: Name, value: Value, shouldValidate?: boolean) => void;
--- a/src/contextTypes.ts
+++ b/src/contextTypes.ts
@@ -3,11 +3,12 @@ import {
   FieldValues,
   ElementLike,
   FormState,
-  Errors,
+  FieldErrors,
   OnSubmit,
   Ref,
   ValidationOptions,
   ValidationPayload,
+  FieldValue,
 } from './types';
 
 export interface FormProps<
@@ -36,7 +37,7 @@ export interface FormContextValues<
     callback: OnSubmit<Data>,
   ) => (e: React.SyntheticEvent) => Promise<void>;
   watch(): Data;
-  watch(field: Name | string, defaultValue?: string): any | void;
+  watch(field: Name | string, defaultValue?: string): FieldValue | void;
   watch(
     fields: (Name | string)[],
     defaultValues?: Partial<Data>,
@@ -49,6 +50,6 @@ export interface FormContextValues<
     payload: ValidationPayload<Name, Value> | ValidationPayload<Name, Value>[],
   ) => Promise<boolean>;
   getValues: (payload?: { nest: boolean }) => Data;
-  errors: Errors<Data>;
+  errors: FieldErrors<Data>;
   formState: FormState<Data, Name>;
 }
--- a/src/contextTypes.ts
+++ b/src/contextTypes.ts
@@ -2,9 +2,8 @@ import * as React from 'react';
 import {
   FieldValues,
   ElementLike,
-  FieldValue,
   FormState,
-  ObjectErrorMessages,
+  Errors,
   OnSubmit,
   Ref,
   ValidationOptions,
@@ -38,7 +37,7 @@ export interface FormContextValues<
     callback: OnSubmit<Data>,
   ) => (e: React.SyntheticEvent) => Promise<void>;
   watch(): Data;
-  watch(field: Name | string, defaultValue?: string): FieldValue | void;
+  watch(field: Name | string, defaultValue?: string): any | void;
   watch(
     fields: (Name | string)[],
     defaultValues?: Partial<Data>,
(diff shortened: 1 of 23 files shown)


bluebill1049:
Feel free to apply suggestions :)

src/useForm.ts
```diff
@@ -190,23 +191,24 @@ export default function useForm<
...
     ): Promise<boolean> => {
       const { fieldErrors } = await validateWithSchemaCurry(
         combineFieldValues(getFieldsValues(fieldsRef.current)),
       );
       const names = isArray(payload)
-        ? payload.map(({ name }) => name as string)
-        : [payload.name as string];
+        ? payload.map(({ name }) => name)
+        : [payload.name];
+      // @ts-ignore
```
bluebill1049: I couldn't find a way to not having ts-ignore here, help is much appreciated :)

src/contextTypes.ts
```diff
@@ -38,19 +36,19 @@ export interface FormContextValues<
     callback: OnSubmit<Data>,
   ) => (e: React.SyntheticEvent) => Promise<void>;
   watch(): Data;
-  watch(field: Name | string, defaultValue?: string): FieldValue | void;
+  watch(field: Name | string, defaultValue?: string): any | void;
```
barrymay: why the any change? (where we need an any it would be good to explain why in a comment if possible)

src/contextTypes.ts
```diff
@@ -38,19 +36,19 @@ export interface FormContextValues<
...
   watch(
     fields: (Name | string)[],
     defaultValues?: Partial<Data>,
   ): Partial<Data>;
-  reset: VoidFunction;
+  reset: () => void;
```
barrymay: not a blocker, but a VoidFunction is a TS-defined alias for () => void so it is the same thing ;)

src/logic/getDefaultValue.ts
```diff
@@ -1,11 +1,7 @@
 import get from '../utils/get';
-import { FieldValue } from '../types';
 import isUndefined from '../utils/isUndefined';
 
-export default <
-  Data extends FieldValue,
-  FieldName extends keyof Data = keyof Data
->(
+export default <Data extends any, FieldName extends keyof Data = keyof Data>(
```
barrymay: I appreciate that FieldValue was an any so it did the same thing, but having that name gave it a semantic meaning (I'm fine to be overruled here ;) )

src/types.ts
```diff
@@ -83,35 +90,24 @@ export type FieldsObject<Data extends FieldValues> = {
   [Key in keyof Data]?: Field;
 };
 
-export interface ReactHookFormError {
+export interface Error {
```
barrymay: I would just be careful here for high level name collision with a JS Error

barrymay (dismissed):
Overall seems like a step in the right direction.  You can put a beta release out and we can check it.  
My major concern is it seems we're adding more 'as' cases than removing them.

src/types.ts
bluebill1049: yes, you are right thanks @barrymay

src/contextTypes.ts
bluebill1049: I think I will revert this back like you suggested it's more semantic as FieldValue.

bluebill1049:
@barrymay is it better to leave more ts-ignore?

bluebill1049:
`3.23.6-beta.1` please help on testing with this beta ❤️

barrymay:
> @barrymay is it better to leave more ts-ignore? (for some of the `as`)

If it's leaving it to "as any" to make it work, then I would say yes.  There are some valid cases for as, but leaving the ts-ignore says "this is not great, we should come back to try to work it out)

bluebill1049:
thanks @barrymay i will update accordingly

bluebill1049:
`3.23.6-beta.2`

src/types.ts
```diff
@@ -1,10 +1,14 @@
 import * as React from 'react';
 
+export type DefaultFieldValues = Record<string, unknown>;
+
 export type FieldValue = any;
 
 export type FieldValues = Record<string, FieldValue>;
 
-export type Validate = (data: FieldValue) => string | boolean | void;
```
barrymay: there's a problem with this (based on my testing of the beta).  Previously my code worked fine with the below code (taking inputs, giving string outputs), but in the new code this isn't allowed.   Any idea why?

```
validate: (inputValue: string[]) => {
          if (!inputValue || !Array.isArray(inputValue)) {
            return 'Invalid result from player list';
          }
          if (!inputValue.length) {
            return 'At least one player must be selected';
          } else if (inputValue.length > 4) {
            return 'Only 4 players are allowed';
          }
          return '';
```

barrymay (changes requested):
See my error - I don't think this change can go forward, we need to ensure the validation functions continue to work as expected.  Were we expecting a breaking change?

bluebill1049:
no we shouldn't expect a breaking changing. I am investigating why

src/types.ts
bluebill1049: this was moved to line 58, so it's not been deleted or changed. I am looking into why

bluebill1049: dame! find the issue `FieldValue` thanks heaps @barrymay

bluebill1049:
going to add an automation check for this usage too, so it will report that issue earlier

barrymay (dismissed):
I tested with 3.23.6-beta.3, looks good to go from my test!

bluebill1049:
@stramel if you got sometime please take a review on this 🙏, if you are busy then I will merge this at the end of the day.

stramel:
If you modify the FieldName to extend a string you won't have to do FieldName | string.

src/useForm.ts
```diff
@@ -25,23 +25,23 @@ import modeChecker from './utils/validationModeChecker';
...
   ValidationPayload,
   ElementLike,
+  DefaultFieldValues,
 } from './types';
 
 export default function useForm<
-  FormValues extends FieldValues,
-  FieldName extends keyof FormValues = keyof FormValues
+  FormValues extends FieldValues = DefaultFieldValues,
+  FieldName extends keyof FormValues = keyof FormValues,
```
stramel: This will need changed to `FieldName extends Extract<keyof FormValues, string> = Extract<keyof FormValues, string>,` in order for it to be assignable to strings

stramel:
Lots of casts in here :/

bluebill1049:
@stramel we can improve on the next PR with your help ❤️

src/useForm.ts
bluebill1049: hmmm more complicated typescript for me hehe

PR #311, opened 2019-09-15
support multiple error messages

bluebill1049:
- allow `validate` function to return multiple error messages

```
validate: (text) => text !== 'test' && ['message1', 'message2']
```


stramel:
Does this support multiple error messages in validate's object form?

bluebill1049:
> Does this support multiple error messages in validate's object form?

Yes

stramel:
Personally, I'm not a fan of the array of messages return value. While this supports multiple error messages on a single key, I was asking more of if multiple could be on the object.

What is the use-case for returning an array of strings for a field?

bluebill1049:
@stramel valid point I will close this PR for now.

PR #317, opened 2019-09-18
Fix/set value with undefined

bluebill1049:


The-Code-Monkey (dismissed):
Looks alright to me no issues should be cause by this

bluebill1049:
thanks @The-Code-Monkey 🙏

bluebill1049:
FYI @The-Code-Monkey I forgot to remove the default value, which set undefined/null to ''

PR #329, opened 2019-09-23
More verbose warning when name attribute is missing on referenced element

wouterds:
Found "Miss ref" not clear while debugging this.


bluebill1049:
Thank you :)

PR #340, opened 2019-09-30
Remove Instance HTMLElement Value

tadeumx1:
I found this error when using this amazing library on React Native
Image HTMLElement Error

And I realized that on line 115 the UseForm try to like cast the ref forcing to be a HTML Element, and when I have a TextInput this is a React Native Element

When I start typing in the input this error appears when the UseForm try to change the value

So i remove the instanceof HTMLElement and this library back to work on React Native

And I also run the tests and they are running normal

Cheers

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -111,8 +111,7 @@ export default function useForm<
     const ref = field.ref;
     const { type } = ref;
     const options = field.options;
-    const value =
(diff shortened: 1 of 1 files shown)


tadeumx1:
I solved the conflict with branch master too

bluebill1049:
thanks for your effort, I will get your PR in and then merge the other PR which is improved version.

PR #341, opened 2019-09-30
Allow react-native projects that use a document polyfill to use library

cthelander:
Fixed up the check on value to allow projects that use a document polyfill to use the library by checking for `window` and referencing `HTMLElement` from there; tested working on react-native.


bluebill1049:
thanks for helping on this issue, i will get it and clean few things up with the undefined check :)

PR #358, opened 2019-10-08
Fixing key

### LINKED ISSUE #351: Help: Typescript errors support for field arrays
**Describe the bug**
I have a form in which I dynamically create field arrays (as in the examples). 
Therefore I have the following Typescript interface:

```typescript
export interface ICostInformationFormValues {
  rentalCost: number;
  additionalCost: Array<{ name: string; value: number }>;
}
```

I want to display validation errors for the dynamically generated fields and that's the problem. The field arrays are handled differently in the errors than in for example getValues({ nested: true }). 
Instead of using my Typescript interface, the Field Array error messages are stored with the index of the field name (e.g.: "additionalCost[1][value]"). 

When I try to read the errors for this field:
```typescript
errors["additionalCost[1][value]"]
```

 I get the following Typescript error message:
```typescript
Element implicitly has an 'any' type because expression of type '"additionalCost[1][value]"' can't be used to index type 'Partial<Record<"rentalCost" | "additionalCost", FieldError>>'.
  Property 'additionalCost[1][value]' does not exist on type 'Partial<Record<"rentalCost" | "additionalCost", FieldError>>'.
```

When I simply log the errors value to my console I can see the error messages for the array fields. I just can't read them with Typescript support.

**Expected behavior**
I would expect to be able to read the error messages based on my Typescript interface like I read the errors for every other field. 

```typescript
errors.additionalCost[1]["value"].message
```

**Additional context**
I'm using [Material-UI ](https://material-ui.com) to create my textfields.
If you need more context please let me know.

Thanks for your help :)
Daniel

stramel:
Abstracts the `FieldName` and `FieldValue` out of generics into types.

This also utilizes the `Extract` helper that pulls out only the `string` keys.

There is still a bit of clean up left in here. Tried to tackle some things as I saw them.

Should resolve #351

### DIFF
--- a/src/contextTypes.ts
+++ b/src/contextTypes.ts
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import {
   FieldValues,
+  FieldName,
+  FieldValue,
   ElementLike,
   FormState,
   FieldErrors,
@@ -8,21 +10,15 @@ import {
   Ref,
   ValidationOptions,
   ValidationPayload,
-  DefaultFieldValues,
 } from './types';
 
-export interface FormProps<
-  FormValues extends FieldValues = DefaultFieldValues,
-  FieldName extends keyof FormValues = keyof FormValues,
-  FieldValue = FormValues[FieldName]
-> extends FormContextValues<FormValues, FieldName, FieldValue> {
+export interface FormProps<FormValues extends FieldValues = FieldValues>
+  extends FormContextValues<FormValues> {
   children: JSX.Element[] | JSX.Element;
 }
 
 export interface FormContextValues<
-  FormValues extends FieldValues = DefaultFieldValues,
-  FieldName extends keyof FormValues = keyof FormValues,
-  FieldValue = FormValues[FieldName]
+  FormValues extends FieldValues = FieldValues
 > {
   register<Element extends ElementLike = ElementLike>(
     validateRule: ValidationOptions,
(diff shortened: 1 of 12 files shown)


src/types.ts
```diff
@@ -93,11 +99,11 @@ export interface Field extends ValidationOptions {
 }
 
 export type FieldsRefs<Data extends FieldValues> = Partial<
```
barrymay: (sorry I know I've been lacking on my assistance lately, but I do want to help ;) )

since you removed Data in 23, does FormValues fit here too?  Would love to get rid of Data as a type name since it's kinda meaningless

src/contextTypes.ts
```diff
@@ -1,28 +1,24 @@
 import * as React from 'react';
 import {
   FieldValues,
+  FieldName,
```
barrymay: I love the complexity reduction be importing this in :)

barrymay (approved):
since this is such a huge change, I'll take a test run of this once it's in a beta build - otherwise I would helpfully trust the unit tests :)

bluebill1049:
thanks very much @stramel and @barrymay 's code review 👍

bluebill1049:
This Pika CI is no longer working for some reason...

barrymay:
I test from latest master and let you know

bluebill1049:
thanks @barrymay i will merge this and publish a beta version soon.

bluebill1049:
`react-hook-form@3.23.19-beta.3` here is the beta version @barrymay @stramel

barrymay:
Tested the beta locally - looks good :)

bluebill1049:
awesome thanks @barrymay

bluebill1049:
tested on my end as well. looks good. going to release the patch this weekend.

PR #362, opened 2019-10-10
feature on revalidate mode

bluebill1049:
New Feature: `reValidateMode`

**Reason for the API:**

There are few users who request having validation onBlur instead of onChagne, while this library is restricted to the onChagne, to work on this feature which requires some users' effect. I think it makes sense to make it as optional as the library should not dictate how re-validate works on their behalf.

**reValidateMode**
- `onChange` defualt
- `onBlur`
- `onSubmit`

allow the user to config how re-validation works in react-hook-form, right now the default behavior is reValidate `onChange`. This new option will allow the user to set re-validate to be `blur` or `submit`

eg:
```
useForm({
  mode: 'onBlur',
  reValidateMode: 'onBlur',
})
```

### DIFF
--- a/cypress/integration/reValidateMode.ts
+++ b/cypress/integration/reValidateMode.ts
@@ -21,6 +21,12 @@ context('re-validate mode', () => {
 
   it('should re-validate the form only onBlur with mode onSubmit and reValidateMode onBlur', () => {
     cy.visit('http://localhost:3000/re-validate-mode/onSubmit/onBlur');
+    cy.get('input[name="firstName"]').focus();
+    cy.get('input[name="firstName"]').blur();
+
+    cy.get('input[name="lastName"]').focus();
+    cy.get('input[name="lastName"]').blur();
+    cy.get('p').should('have.length', 0);
 
     cy.get('button#submit').click();
 
@@ -35,7 +41,7 @@ context('re-validate mode', () => {
     cy.get('input[name="lastName"]').blur();
 
     cy.get('p').should('have.length', 0);
-    cy.get('#renderCount').contains('7');
+    cy.get('#renderCount').contains('9');
   });
 
   it('should re-validate the form only onSubmit with mode onBlur and reValidateMode onSubmit', () => {
--- a/cypress/integration/reValidateMode.ts
+++ b/cypress/integration/reValidateMode.ts
@@ -1,0 +1,40 @@
+context('re validate mode', () => {
+  it('should re validate the form only onSubmit with mode onSubmit and reValidateMode onSubmit', () => {
+    cy.visit('http://localhost:3000/re-validate-mode/onSubmit/onSubmit');
+
+    cy.get('button#submit').click();
+
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+
+    cy.get('input[name="firstName"]').type('luo123456');
+    cy.get('input[name="lastName"]').type('luo12');
+
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+
+    cy.get('button#submit').click();
+
+    cy.get('p').should('have.length', 0);
+    cy.get('#renderCount').contains('5');
+  });
+
+  it('should re validate the form only onBlur with mode onSubmit and reValidateMode onBlur', () => {
+    cy.visit('http://localhost:3000/re-validate-mode/onSubmit/onBlur');
+
+    cy.get('button#submit').click();
+
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+
+    cy.get('input[name="firstName"]').type('luo123456');
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="firstName"]').blur();
+    cy.get('input[name="lastName"]').type('luo12');
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+    cy.get('input[name="lastName"]').blur();
+
+    cy.get('p').should('have.length', 0);
+    cy.get('#renderCount').contains('7');
+  });
+});
--- a/cypress/integration/reValidateMode.ts
+++ b/cypress/integration/reValidateMode.ts
@@ -1,5 +1,5 @@
-context('re validate mode', () => {
-  it('should re validate the form only onSubmit with mode onSubmit and reValidateMode onSubmit', () => {
+context('re-validate mode', () => {
+  it('should re-validate the form only onSubmit with mode onSubmit and reValidateMode onSubmit', () => {
     cy.visit('http://localhost:3000/re-validate-mode/onSubmit/onSubmit');
 
     cy.get('button#submit').click();
@@ -19,7 +19,7 @@ context('re validate mode', () => {
     cy.get('#renderCount').contains('5');
   });
 
-  it('should re validate the form only onBlur with mode onSubmit and reValidateMode onBlur', () => {
+  it('should re-validate the form only onBlur with mode onSubmit and reValidateMode onBlur', () => {
     cy.visit('http://localhost:3000/re-validate-mode/onSubmit/onBlur');
 
     cy.get('button#submit').click();
@@ -36,5 +36,47 @@ context('re validate mode', () => {
 
     cy.get('p').should('have.length', 0);
     cy.get('#renderCount').contains('7');
+  });
+
+  it("should re-validate the form onBlur only with mode onBlur and reValidateMode onBlur", () => {
+    cy.visit('http://localhost:3000/re-validate-mode/onBlur/onBlur');
+
+    cy.get('p').should('have.length', 0);
+
+    cy.get('input[name="firstName"]').focus();
+    cy.get('input[name="firstName"]').blur();
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="lastName"]').focus();
+    cy.get('input[name="lastName"]').blur();
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+
+    cy.get('input[name="firstName"]').type('luo123456');
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="firstName"]').blur();
+    cy.get('input[name="lastName"]').type('luo12');
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+    cy.get('input[name="lastName"]').blur();
+
+    cy.get('p').should('have.length', 0);
+    cy.get('#renderCount').contains('7');
+  });
+
+  it("should re-validate the form onChange with mode onBlur and reValidateMode onChange", () => {
+    cy.visit('http://localhost:3000/re-validate-mode/onBlur/onChange');
+
+    cy.get('p').should('have.length', 0);
+
+    cy.get('input[name="firstName"]').focus();
+    cy.get('input[name="firstName"]').blur();
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="lastName"]').focus();
+    cy.get('input[name="lastName"]').blur();
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+
+    cy.get('input[name="firstName"]').type('luo123456');
+    cy.get('input[name="lastName"]').type('luo12');
+
+    cy.get('p').should('have.length', 0);
+    cy.get('#renderCount').contains('5');
   });
 });
--- a/cypress/integration/reValidateMode.ts
+++ b/cypress/integration/reValidateMode.ts
@@ -38,7 +38,53 @@ context('re-validate mode', () => {
     cy.get('#renderCount').contains('7');
   });
 
-  it("should re-validate the form onBlur only with mode onBlur and reValidateMode onBlur", () => {
+  it('should re-validate the form only onSubmit with mode onBlur and reValidateMode onSubmit', () => {
+    cy.visit('http://localhost:3000/re-validate-mode/onBlur/onSubmit');
+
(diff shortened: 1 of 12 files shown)


src/constants.ts
```diff
@@ -13,3 +13,9 @@ export const REQUIRED_ATTRIBUTE = 'required';
 export const PATTERN_ATTRIBUTE = 'pattern';
 
 export const UNDEFINED = 'undefined';
+
```
barrymay: Im fine with the constant but just for clarity I would create a wrapper object called Events and have these three properties in it.  Gives them a namespace and better meaning.

If you want greater safety on that you can call object.freeze on the Object so they can't get changed

cypress/integration/reValidateMode.ts
```diff
@@ -0,0 +1,82 @@
+context('re-validate mode', () => {
+  it('should re-validate the form only onSubmit with mode onSubmit and reValidateMode onSubmit', () => {
```
barrymay: These are great but are we testing that if onSubmit is on and you just blur, nothing will change?

barrymay (dismissed):

barrymay:
Besides the tests issue this seems sound. Let's do another beta before releasing.

src/constants.ts
bluebill1049: sounds good to me :)

cypress/integration/reValidateMode.ts
bluebill1049: it's in the automation test 👍

bluebill1049:
`react-hook-form@3.24.0-beta.1`

src/constants.ts
barrymay: Perfect 😊

rkurbatov:
Oh, ok, that's about revalidation (in addition to validation onChange, as I understand). And what about making form touched onBlur instead of onChange?

src/types.ts
```diff
@@ -45,6 +45,7 @@ export interface Schema<Data> {
 
 export type Options<FormValues extends FieldValues = FieldValues> = Partial<{
   mode: Mode;
+  reValidateMode: Mode;
```
stramel: This typing is incorrect. This results in `'onSubmit' | 'onChange' | 'onBlur'`.

In order for this to be correct and use the `keyof EVENTS`, you will have to change the constant to 

```
const EVENTS = {
  blur: 'blur',
  change: 'change',
  input: 'input'
}
```

Then this will have to be 

```suggestion
  reValidateMode: keyof EVENTS;
```

src/useForm.ts
```diff
@@ -298,10 +303,11 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
 
         if (!ref) return;
 
-        const isBlurEvent = type === 'blur';
+        const isBlurEvent = type === EVENTS.BLUR;
```
stramel: Pretty sure that the typing on the params is incorrect and the fact that `Ref` is `any` is killing any error here. 

The param appears to be an `Event`. I was very confused as to why a `Ref`'s `type` would be equal to `'blur'`.

bluebill1049: what would you recommend to improve this? it's for the event type when the ref is registered.

bluebill1049:
> Oh, ok, that's about revalidation (in addition to validation onChange, as I understand). And what about making form touched onBlur instead of onChange?
> 
> Something like: `touchMode: 'onChange' (default) / 'onBlur'`. That will allow to make error messaging much more flexible.

why not just 
```
useForm({
  revalidation: isTouch ? 'xx' : 'xx'
})
```

src/useForm.ts
stramel: Switch `Ref` to the correct typing. As far as I can tell, it would be something like 

`type Ref = HTMLInputElement | HTMLSelectElement | HTMLTextAreaElement | HTMLMeterElement`

rkurbatov:
@bluebill1049 I mean I need to keep validation 'onChange' while setting input touched only 'onBlur', not on change. For example, the user inputs the email address and I don't want to disturb him with an error if the address is invalid until he leaves the input for the first time. 

That discussion is beyond the scope of the task, sorry for that. I'll provide PR with changes after the merge of that PR.

bluebill1049:
@rkurbatov I think I get what you mean now, I think it kind makes sense to leave such logic at your app level, but happy to see a PR and let others involve on this topic. To be honest, I don't want to introduce too many APIs (if you can achieve that at your code level, i would consider avoiding) because react hook form's APIs are mostly primitive. I did hesitate to introduce this feature for a while, however, it convinced me how difficult is to achieve `onBlur` mode with `onBlur` revalidate and hence the reason to introduce this API.

rkurbatov:
I am afraid I can't do that thing programmatically, I tried several approaches. And the change fits in line with other APIs:
- make fields touched on every user's input or
- make fields touched just on blur.

Validity is one thing, clearness is another one, they can be treated differently in complex solutions.

bluebill1049:
```
useForm({
  mode: 'submit',
  reValidateMode: 'submit',
})
```

if your validation is customized, you probably can skip validation on react hook form and attach your own `onChange` onBlur`.

bluebill1049:
> I am afraid I can't do that thing programmatically, I tried several approaches. And the change fits in line with other APIs:
> 
> * make fields touched on every user's input or
> * make fields touched just on blur.
> 
> Validity is one thing, clearness is another one, they can be treated differently in complex solutions.

```
useForm({
  mode: 'submit',
  reValidateMode: 'submit',
})
```

if your validation rules (when is triggering) is customized, you probably can skip validation on react hook form and attach your own `onChange` onBlur`.

rkurbatov:
@bluebill1049 I'm ok with validation rules :) 'onChange' is just fine in most cases. I'm about 'touched' rules. Sometimes I need to have input touched only after blur, not after any input and I cannot change that. It's not about validation, it's mostly about error display. I mean a separate mechanism of having touched behavior controllable just like we do with validation.

bluebill1049:
> @bluebill1049 I'm ok with validation rules :) 'onChange' is just fine in most cases. I'm about 'touched' rules. Sometimes I need to have input touched only after blur, not after any input and I cannot change that. It's not about validation, it's mostly about error display. I mean a separate mechanism of having touched behavior controllable just like we do with validation.

I guess you can leverage what's in the `formState.touched` and use it as condition to show errors?

rkurbatov:
But touched becomes true right after any change and I cannot change it. Sometimes I want assume input as 'touched' after the blur, not after any 'change'. That allows, for example, treat submit button enabled state based on form validity and error state based on dirtiness. Thus in best case user doesn't see any error if he enters all data correctly and only sees error if the field loses focus and validation returns false for that field.

src/useForm.ts
bluebill1049: oh ok i see. Thanks :)

bluebill1049: Screen Shot 2019-10-12 at 5 34 37 pm
any idea why this is happening?

bluebill1049:
Screen Shot 2019-10-12 at 9 27 59 pm
@stramel you have any insight on how to fix this one?

bluebill1049:
please help on testing: `react-hook-form@3.24.0-beta.2` 👍

cypress/integration/reValidateMode.ts
JeromeDeLeon: I tested the `mode = "onBlur"` and `reValidateMode = "onSubmit"` and field is still triggering re-validation `onChange`. Is it the right behavior?

cypress/integration/reValidateMode.ts
```diff
@@ -0,0 +1,82 @@
...
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+
+    cy.get('input[name="firstName"]').type('luo123456');
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="firstName"]').blur();
+    cy.get('input[name="lastName"]').type('luo12');
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+    cy.get('input[name="lastName"]').blur();
+
```
JeromeDeLeon: It is not checking if you try to empty the field after blurring the field. It is triggering the validation.

```
cy.get('input="firstName"]').clear();
cy.get('input[name="firstName"] + p').contains('firstName error');
```

barrymay:
As far as backward compatibility, I'm fine with this change - I made one change (pushed) to fix the ts-ignore ;). I'll defer to @JeromeDeLeon for the confirmation that this change helps

src/useForm.ts
```diff
@@ -593,6 +597,8 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
         fields[typedName].options[fields[typedName].options.length - 1]
```
barrymay: @bluebill1049 @stramel (I'm trying to fix the other ts-ignores and stuck here) I know this isn't part of this change, but I'm confused here.. What is this line doing?  How do we know what was stored in the options on this?

src/useForm.ts
```diff
@@ -520,10 +520,12 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
       );
 
       if (isRadio) {
+        let currentField: Field | undefined = fields[typedName];
```
JeromeDeLeon: Other uses of fields[typeName] is having the same issue with `undefined`.
I think it's better if we do something like below to avoid checking if value is undefined.

```
let currentField = fields[typedName] || {};

(currentField.options ? currentField.options : [])
```

barrymay: While I agree with you it's a little clearer, the TS Compiler had problems with it.

Since it's not that far off from what I'm doing, I'll leave it.

barrymay: I'm separately working on the other ts-ignores, and I'm stuck on one change which I don't understand (see my other comment to @bluebill1049  and @stramel )

src/useForm.ts
JeromeDeLeon: That line is mostly to get the last value of options and that is the object containing the `ref` and `mutationWatcher` but the only concern is `ref` that is use to attach event listener

src/useForm.ts
JeromeDeLeon: We can do this.
```js
let currentField = (fields[typedName] || {}) as Field;
```

barrymay: true, but if I'm going to be honest, the TS as is a forced cast IMO, and it'll make the {} you're putting in ignore a check for .ref later.  I'm generally trying to avoid as's in the code I write.  

Putting a higher let `let resultField: Field` variable at the top is in the code I'm writing for the resulting field, so for that I totally agree, but not for this specific case.

JeromeDeLeon: I push some code so feel free to review and tell me what you think. I still trying to wrap some things up and you're right about the TS as feature.

JeromeDeLeon: The `validateField` is forcing us because of it's `Field | undefined` scenario where `undefined` is not acceptable as argument. I was very reluctant to use it as first since it somehow forces me to hack things around.

barrymay: Saw your last change - very nice!   Made a slight couple adjustments (outside of what you did) for clarity to remove as's

cypress/integration/reValidateMode.ts
bluebill1049: thanks @JeromeDeLeon i think i know the fix. thanks for testing

bluebill1049:
thanks @JeromeDeLeon for spotting that onSubmit validation issue, I have patched above. 

if you have time, please re-verify: 3.24.0-beta.3

bluebill1049:
@barrymay thanks for helping review and try to remove those ts-ignore

cypress/integration/reValidateMode.ts
JeromeDeLeon: @bluebill1049, for this case, it's still triggering the validation.

bluebill1049:
Nice job on removing the ignore @JeromeDeLeon 🙏🏻👍🏻

JeromeDeLeon:
@bluebill1049 Have you seen my comment for mode `onSubmit` and reValidateMode `onBlur`?

cypress/integration/reValidateMode.ts
bluebill1049: I think it’s correct because I have triggered submit on the beginning of the test, but good point on this tho. I will test out the validation without submit first 👍🏻

bluebill1049:
> @bluebill1049 Have you seen my comment for mode `onSubmit` and reValidateMode `onBlur`?

oops, I did miss the common, replied and improved automation test.

bluebill1049:
I think we r getting close to release this feature 🚀

cypress/integration/reValidateMode.ts
```diff
@@ -0,0 +1,134 @@
...
+    cy.get('button#submit').click();
+
+    cy.get('p').should('have.length', 0);
+    cy.get('#renderCount').contains('7');
+  });
+
+  it('should re-validate the form onBlur only with mode onBlur and reValidateMode onBlur', () => {
+    cy.visit('http://localhost:3000/re-validate-mode/onBlur/onBlur');
+
+    cy.get('p').should('have.length', 0);
```
JeromeDeLeon: @bluebill1049 Gotta ask why you check the error message at the beginning of the test?

bluebill1049: oops, cause it's a mistake! thanks

JeromeDeLeon: No worries! 😄

JeromeDeLeon:
> I think we r getting close to release this feature 🚀

Are we still missing something?

bluebill1049:
> > I think we r getting close to release this feature 🚀
> 
> Are we still missing something?

looks like I broked the test...

bluebill1049:
looks like a false alarm, let's merge it. I think it's good to go now :)

JeromeDeLeon:
> looks like a false alarm, let's merge it. I think it's good to go now :)

Great!

ArvindNikam76:
reValidateMode :"onChange" should update formState.isValid?

PR #378, opened 2019-10-16
Feature/touched mode

rkurbatov:
Here is crude and least invasive implementation I'd wanted to discuss.

Redux-forms that we currently use have a nice option controlling touched state behavior. It allows marking form element as touched on blur by default, providing much better UX. Here is an example of the form:

Input: Email
Input: Password
Submit button

Let's assume the email field has an email validator, password field has length validator and submit button is inactive while invalid. 

The user enters email (which is for sure invalid while he enters) and then presses Tab or selects another field causing blur event. There is no need to show error on change. Button remains disabled - the form is invalid.

Then the user enters a password and when password length is enough, the submit button becomes active so the user can press it and submit. There was also no error during the input. The user sees the error only after the blur occurred (if any).

There is no way to imitate that behavior in react-hook-form. If I set validation mode to onBlur I cannot manage button disabled state - user needs to click it to make it active and click it once more to submit data. If I set validation mode to onChange the user sees an error while he enters email and password - that's quite distractive.

The new option syntax is in line with other options and allows us to manage error visibility in a more flexible way improving UX. If you ok with it, I will add tests.

There is probably no value in 'onSubmit' option...


bluebill1049:
First of all, thanks for your PR.

>  If I set validation mode to onBlur I cannot manage button disabled state

`isValid` is not working for you?

rkurbatov:
@bluebill1049 ehm... I mean it works, but the user needs to unfocus to perform validation. Compare:

typical scenario for most users:
enter valid email - no errors during enter
enter long enough password - no errors duing enter
data are validated on every change - button becomes active (isValid check) - the user submits form
if something gets wrong, user presses the inactive button and sees the error.

current scenario:
enter valid email - no errors during enter (onBlur) - errors during enter (onChange)
enter long enough password - no errors during enter (onBlur) - errors during enter (onChange)
button is still disabled even if the user thinks data are valid - the user needs to unfocus form and only then press the button or press the button twice.

DemoDevelops (dismissed):

CoreyTrombley (dismissed):
LGTM

wshen001 (dismissed):
Nice feature!

bluebill1049:
I am not 100% sold on this feature, I still believe you can achieve this without introduce extra pros. I will take a deeper look into this soon. in the meantime other contributors please take a look and leave your feedback as well. hopefully we don't have to add more API and only when necessary, in the end I still prefer to keep react-hook-form simple and small set of APIs.

bluebill1049:
looking close to this issue, I think I start to make more sense now.

src/useForm.ts
```diff
@@ -79,6 +80,10 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
     isOnBlur: isReValidateOnBlur,
     isOnSubmit: isReValidateOnSubmit,
   } = useRef(modeChecker(reValidateMode)).current;
+  const {
+    isOnBlur: isTouchedOnBlur,
+    // isOnSubmit: isTouchedOnSubmit,
```
bluebill1049: I think we can remove this.

src/useForm.ts
```diff
@@ -46,6 +46,7 @@ import {
 export default function useForm<FormValues extends FieldValues = FieldValues>({
   mode = VALIDATION_MODE.onSubmit,
   reValidateMode = VALIDATION_MODE.onChange,
+  touchedMode = VALIDATION_MODE.onChange,
```
bluebill1049: I guess we only need to support `onChange` and `onBlur`

src/types.ts
```diff
@@ -50,6 +50,7 @@ export interface Schema<Data> {
 export type Options<FormValues extends FieldValues = FieldValues> = Partial<{
   mode: Mode;
   reValidateMode: Mode;
+  touchedMode: Mode;
```
bluebill1049: Type needs to be updated.

bluebill1049:
would be nice to add a unit test, if you can :)

src/useForm.ts
rkurbatov: Sure, it's just a PoC. Seems like touched on submit is not needed, so I will update types and provide tests.

bluebill1049: Sweet. Thanks

JeromeDeLeon (dismissed):
LGTM! Seems like 3 more persons have been wanting this feature as well and will improve the UX.

bluebill1049:
@rkurbatov any progress on this PR?

bluebill1049:
I am happy to take over if you are busy too.

rkurbatov:
Sorry, it was long weekend. :) I am investigating tests to understand how simulate touched behavior and events reaction. There is no tests for touched at all :) If it's not a problem to you as you are the main contributor, I would appreciate that, really.

bluebill1049:
if you can fix the Type for `touchMode`, let's get it in and I will sort out the rest (unite and automation test). In fact, I can even release an early beta version for you to work on with your project.

rkurbatov:
@bluebill1049 early beta would be nice :) @JeromeDeLeon wow, thanks!

bluebill1049:
thanks @JeromeDeLeon ❤️

bluebill1049:
@rkurbatov can you please test out this beta version? `3.24.2-beta.5` I have removed `touchMode`, instead touched is triggered onBlur instead onChange. so you don't need to change your code and it should just work out for you. let me know how did it go.

hernanG89:
Hi everyone ! I'd like to know how to test this feature. Do i have any fn like setValue=>"setTouched" to control by myself that prop?

Issue #369, opened 2019-10-14
Typescript build failing because of 'react-native' import


kobber:
**Describe the bug**
After upgrading to `3.24.0` our Typescript build is failing. This is caused by the typings importing `'react-native'`, which we're not using in our project.

```
ERROR in [at-loader] ./node_modules/react-hook-form/dist/types.d.ts:2:30
    TS2307: Cannot find module 'react-native'.

ERROR in [at-loader] ./node_modules/react-hook-form/dist/useForm.d.ts:2:30
    TS2307: Cannot find module 'react-native'.
```

**To Reproduce**
I'm assuming this happens for every Typescript based project that doesn't use `'react-native'`.

**Expected behavior**
The package should not have a dependency on `'react-native'`.

bluebill1049:
oops, that's my bad. I will patch that.

bluebill1049:
can you please test out this beta version? `3.24.1-beta.1`

kobber:
> can you please test out this beta version? `3.24.1-beta.1`

Ran `npm install react-hook-form@3.24.1-beta.1`, but it didn't seem to make a difference.

I'm still seeing `import * as ReactNative from 'react-native';` on line `2` of `dist/useForm.d.ts`.

bluebill1049:
thanks, strange I have already include react native peer dependency

bluebill1049:
also, I have tried on my own ts project, it's working fine with the latest version as well.

bluebill1049:
what's this `at-loader`?

bluebill1049:
can you make your at-loader not resolving node_module folder?

kobber:
I tried solving the problem by manually installing `react-native` and `@types/react-native`, and while this fixed the `Cannot find module 'react-native'` bug, it introduced other compile errors with duplicated global definitions.

I don't believe a web based project should have a dependency on `react-native` at all. Could the things required for react-native support be moved out somehow? Either to a separate package or a subfolder like so:
```ts
import useForm from "react-hook-form/react-native";
```

bluebill1049:
because react-hook-form is universal it supports type for both react and react-native. did you try to exclude node_module for at-loader?

bluebill1049:
let me know before I investigate other options for you

bluebill1049:
@barrymay could you please double-check on your end as well? I couldn't reproduce this issue.

kobber:
[...]

PR #381, opened 2019-10-17
Fix react native TS types for register - related to #369

barrymay:
As a follow-up to my investigation on #369, I don't see a problem using react-native import-wise with react-native.  In terms of the actual implementation, it's only using types from react-native, not any implementation.

That said, in doing a local test, I did see a case where using react-native type elements in TS causes a problem, because react-native's elements don't support a name property.  (I tried following the demo from https://react-hook-form.com in TS and ran into this case)

So I added an extra name property model to the register call
- Allows passing in a ref that has no name, but allows the name to be specified in the register call
- In the same, way, allows a register call to be called outside of JSX with the name properly specified on the separate object properly

Here's an example of usage with the method - the other model using jsx is equivalent to what's on on react-hook-form.com
```
const testRef = useRef(null);
useEffect(() => {
 register(test2Ref.current, { name: "foo2" });
}, []);

...

return (
      <TextInput
        ref={testRef}
        onChangeText={text => setValue("foo", text)}
      />
);
```

(technically, it may be possible to get rid of the ref here altogether since we're manually setting the value in the onChangeText call, but it appears sufficient for now)

@bluebill1049 @stramel feel free to give any feedback!


src/useForm.ts
```diff
@@ -598,19 +598,45 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
     }
   }
 
+  // For JSX registry
+  // React-Native (Element has no name prop, so it must be passed in on teh validateRule)
+  function register<Element>(
+    validateRule: ValidationOptions & { name: string },
```
bluebill1049: maybe it's worth to turn `{ name: string }` into a type?

src/useForm.ts
```diff
@@ -598,19 +598,45 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
...
   ): void;
+
   function register<Element extends ElementLike = ElementLike>(
     refOrValidateRule: ValidationOptions | Element | null,
-    validationOptions?: ValidationOptions,
+    validationOptions?: ValidationOptions & { name?: string },
   ): ((ref: Element | null) => void) | void {
     if (typeof window === UNDEFINED || !refOrValidateRule) return;
 
+    if (validationOptions && typeof validationOptions.name === 'string') {
```
bluebill1049: I think we only support

`register({ name: 'xxx' }, { required: true })`

is this try to support 

`register({ name: 'xxx', required: true })`

bluebill1049:
thanks very much @barrymay for investigating on this issue :)

src/useForm.ts
barrymay: exactly - because when you pass in a ref to register from react native, that ref will not have a name on it (because react-native doesn't support the name property on those text elements)

bluebill1049: sounds good 👍

src/useForm.ts
barrymay: fair and done (albeit maybe a bit overdone, I won't fight it :) )

bluebill1049 (approved):

src/useForm.ts
barrymay: yes - with this exact change (and I noted it in the comment) if you put the name in the second argument, it will override the name put on the element.  You likely wouldn't do it, but it maybe helpful in a rare case.

bluebill1049:
@barrymay merge whenever you are ready :)

PR #383, opened 2019-10-17
close #369 remove react native type

### LINKED ISSUE #369: Typescript build failing because of 'react-native' import

bluebill1049:
Just a place holder for now, if it is an issue we will remove react-native types for now


src/types.ts
```diff
@@ -26,7 +25,7 @@ type Mode = keyof ValidationMode;
 
 export type OnSubmit<FormValues extends FieldValues> = (
   data: FormValues,
-  e: React.SyntheticEvent | ReactNative.GestureResponderEvent,
+  e: React.SyntheticEvent | any,
```
JeromeDeLeon: Not a fun of `any` because basically telling that you don't care what type it receives. If you put `any` there, might as well remove the `React.SyntheticEvent`. Is there any reason why you remove this types? 😄

JeromeDeLeon: Sorry, I just read the PR and seems like it was more like an issue regarding the use case of typescript on the web requiring to install its type dependency on react-native.

src/types.ts
```diff
@@ -26,7 +25,7 @@ type Mode = keyof ValidationMode;
 
 export type OnSubmit<FormValues extends FieldValues> = (
   data: FormValues,
-  e: React.SyntheticEvent | ReactNative.GestureResponderEvent,
```
barrymay: Let's put a //TODO comment in both places here to explain that the any is purposely there to replace ReactNative.GestureResponderEvent - that way this is easy to find

barrymay (dismissed):
I strongly recommend the TODO comments there to make followup easy, but otherwise this is good to go.

src/types.ts
stramel: `| any` 😨

bluebill1049: sad face :(

bluebill1049:
I will leave a todo there and release this for now so it impacts fewer projects.

src/types.ts
bluebill1049: it's probably too harsh to ask people to turn on `skipLibCheck`

src/types.ts
bluebill1049: will do thanks.

src/types.ts
bluebill1049: due to this issue: https://github.com/react-hook-form/react-hook-form/issues/369

bluebill1049:
let me know it's ready to be merged :) and we will release it <3

bluebill1049:
@barrymay you want to merge your fix as well for the release?

barrymay:
@bluebill1049 can we create a beta with this my last commit as a basis?  This should make it work just fine for both cases, since both frameworks inherit from BaseSyntheticEvent

src/types.ts
```diff
@@ -25,8 +25,7 @@ type Mode = keyof ValidationMode;
 
 export type OnSubmit<FormValues extends FieldValues> = (
   data: FormValues,
-  // todo: #369 reported error in ts build while skipLibCheck isn't set to true
-  e: React.SyntheticEvent | any,
+  e: React.BaseSyntheticEvent,
```
bluebill1049: 👀

bluebill1049: I guess this will work for react native as well 👍

barrymay:
> @barrymay you want to merge your fix as well for the release?

I do, since it allow for consistent usage with react-native and doesn't cause further issues.

If I have time, I want to investigate create a react-native target package, but this case should be fine for both scenarios for now.

bluebill1049:
`3.24.1-beta.3`

bluebill1049:
thanks @barrymay

barrymay (approved):
I'm fine with this change if we can leverage BaseSyntheticEvent to avoid the any.  A beta test for this is worthwhile

bluebill1049:
> I'm fine with this change if we can leverage BaseSyntheticEvent to avoid the any. A beta test for this is worthwhile

sounds good to me!

barrymay:
ok - so two things (@stramel I'd like your feedback on this also)

1) Bill, Can we get another beta with my other PR in it (that should be fine regardless of what we decide here)
2) If we go with BaseSyntheticEvent, it causes a slight typing problem if you use the event property in TS

if you call ```handleSubmit(onSubmit)``` as well

ANY you want to use the e for the event with specific properties of SyntheticEvent, then you need to do the following

```
 const onSubmit = (target: any, e: BaseSyntheticEvent) => {
    const x: SyntheticEvent = e as SyntheticEvent;
}
```
because SyntheticEvent is a subtype of BaseSyntheticEvent, you can just assign it to e, but an as is perfectly safe here.

We can accept this or go with the `| any` case, which makes it easier to use, but it less type safe.

I'll go with whatever you guys decide. (personally the BaseSyntheticEvent has more meaning IMO)

src/types.ts
barrymay: exactly - the GestureResponderEvent also inherits from BaseSyntheticEvent

bluebill1049:
@barrymay happy to go with what u suggested.

bluebill1049:
@stramel you happy with this?

stramel:
Seems okay, haven't had much time to look into it.

stramel (approved):

bluebill1049:
ok thanks @stramel let's take a look this again if other users report issue

PR #386, opened 2019-10-18
Improve typing of `setValue`

ldanet:
With the latest versions of `react-hook-form` it seems that `setValue` (amongst others) doesn't do type checking properly anymore, as you can see on [this sandbox](https://codesandbox.io/s/react-hook-form-ts-playground-7v3gj?fontsize=14&module=%2Fsrc%2Fapp.tsx): 

This PR aims to fix the type checking for `setValue` so that it 
1. will only accept field names defined in form data type when provided
1. will only accept the correct type of value for a given field name.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -271,12 +271,14 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
     [executeSchemaValidation, executeValidation, validationSchema],
   );
 
-  const setValue = useCallback(
-    (
-      name: FieldName<FormValues>,
-      value: FieldValue<FormValues>,
-      shouldValidate: boolean = false,
-    ): void | Promise<boolean> => {
+  const setValue = useCallback<
+    <Name extends FieldName<FormValues>>(
+      name: Name,
+      value: FormValues[Name],
+      shouldValidate?: boolean,
+    ) => void | Promise<boolean>
+  >(
+    (name, value, shouldValidate = false) => {
       setValueInternal(name, value);
       const shouldRender =
         isWatchAllRef.current || watchFieldsRef.current[name];
@@ -284,6 +286,7 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
         return triggerValidation({ name }, shouldRender);
       }
       if (shouldRender) render({});
+      return;
     },
     [setValueInternal, triggerValidation],
   );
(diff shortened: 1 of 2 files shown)


src/useForm.ts
```diff
@@ -271,19 +271,22 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
...
+  >(
+    (name, value, shouldValidate = false) => {
       setValueInternal(name, value);
       const shouldRender =
         isWatchAllRef.current || watchFieldsRef.current[name];
       if (shouldValidate) {
         return triggerValidation({ name }, shouldRender);
       }
       if (shouldRender) render({});
+      return;
```
bluebill1049: do we need this return?

ldanet: My code editor complains about `Not all code paths return a value. ts(7030)` if I remove it, but none of the `yarn test` `yarn lint` `yarn build` commands complain about it.
I can remove it.

ldanet: Actually it does fail 
``` 
FAIL  src/useForm.test.tsx
  ● Test suite failed to run

    TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    src/useForm.ts:281:5 - error TS7030: Not all code paths return a value.

    281     (name, value, shouldValidate = false) => {
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

bluebill1049: All good

barrymay: I feel that this weirdness is due to the fact that setValue is returning both a promise and a non-promise value, which feels wrong to me, but since it doesn't change what's happening already we can let this go.

barrymay: to fix the weirdness, I might suggest something like this, but it would need further testing:
```js
  const setValue = useCallback<
    <Name extends FieldName<FormValues>>(
      name: Name,
      value: FormValues[Name],
      shouldValidate?: boolean,
    ) => Promise<boolean>
  >(
    (name, value, shouldValidate = false) => {
      setValueInternal(name, value);
      const shouldRender =
        isWatchAllRef.current || watchFieldsRef.current[name];
      if (shouldValidate) {
        return triggerValidation({ name }, shouldRender);
      }
      return Promise.resolve(true);
    },
    [setValueInternal, triggerValidation],
  );
```

barrymay (approved):

Issue #298, opened 2019-09-08
formState.isValid is not being updated when fields are registered in useEffect


lnmunhoz:
**Describe the bug**
The `formState.isValid` property doesn't update when using `useEffect` to `register` the fields.

**To Reproduce**
The codesandbox below shows the behaviour.
https://codesandbox.io/embed/bug-react-hook-form-formstateisvalid-wont-change-if-fields-are-registered-on-useeffect-segvo

**Expected behavior**
The formState.isValid should change accordingly.

**Additional context**
I just started using this library, so I hopefully I didn't miss anything. But I am trying for more than 1 hour and I can't find any issues with the example code. It suppose to work according to the docs.

bluebill1049:
hey, because you are using custom `register` and `ref` is not attached during the registration. https://react-hook-form.com/api#setValue during setValue, you should trigger validation.

setValue('name', 'value', true) // third argument will trigger validation.

lnmunhoz:
Oh all right! I didn't see that. Makes sense.

In my opinion this validate should be `true` by default, being false is so counterintuitive.
I will give it a try. Thank you!

bluebill1049:
Sure we will improve on it in the following patch.

bluebill1049:
However it need to be default to false. Because you aren’t certain it’s valid just yet

PR #389, opened 2019-10-18
fix isValid issue with validationFields and consistency with isValid

bluebill1049:
This PR is about: 

- `isValid` will return false consistently through different validation `mode`
- fix a bug which not includes `validationFields` for `isValid` form state


src/useForm.ts
```diff
@@ -547,7 +547,8 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
     }
 
     if (validateOptions && !isEmptyObject(validateOptions)) {
-      fieldsWithValidationRef.current.add(name);
+      if (!validationFields || validationFields.includes(name))
```
barrymay: minor code style issue (we can agree to disagree) all if statements should have brackets, event single line ones (this prevents accidental bugs if someone adds another line

bluebill1049: agree :) fixing

barrymay: cool, pretty sure there's an eslint rule for that if you IDE isn't reporting that

barrymay:
general question - does this change still allow the case for onSubmit where you can click submit and validate regardless of errors for the first time?  I can confirm it with a beta cut, but if you've tested this, then all good.

src/useForm.ts
bluebill1049: I think my IDE doesn't report that at the moment (webstorm)

barrymay: ok, I'll check into it

bluebill1049:
> general question - does this change still allow the case for onSubmit where you can click submit and validate regardless of errors for the first time? I can confirm it with a beta cut, but if you've tested this, then all good.

This change only affect `isValid` form state, it should consistently return false at the initial render. I did test it with example and automation.

barrymay:
cool - playing devil's advocate here... what if the starting state of the form was actually valid.  So my general question is ( and it might need more work to fully fix ) .  Should isValid be undefined on a form until the first validation turn is taken?

barrymay (approved):
TBH, I'm not sure what exactly this effects, but I don't see a reason to block it (it's more likely false than true to be sure)

bluebill1049:
> cool - playing devil's advocate here... what if the starting state of the form was actually valid. So my general question is ( and it might need more work to fully fix ) . Should isValid be undefined on a form until the first validation turn is taken?

That's perfect! we can improve on this if users report such use case.

src/useForm.ts
JeromeDeLeon: [eslint curly](https://eslint.org/docs/rules/curly)

JeromeDeLeon (approved):
LGTM! I haven't been able to catch up lately. Sorry, guys. 😞

JeromeDeLeon:
> cool - playing devil's advocate here... what if the starting state of the form was actually valid. So my general question is ( and it might need more work to fully fix ) . Should isValid be undefined on a form until the first validation turn is taken?

How do we affirm this? Do you have any example in mind? @barrymay 
I'm curious as well how do we handle such cases, but I think there will be rare cases.

barrymay:
@JeromeDeLeon to be clear, I'm not blocking this change, but I'm curious how this came up and what problems it's causing, since isValid is not used as an exposed value of useForm.  

@bluebill1049 do you have a use case where this caused an issue?  I see the Cypress changes (great) but I don't understand why this suddenly is an issue.

src/useForm.ts
barrymay: thanks - please see #391 ;)

bluebill1049:
@barrymay one of the issues, people reported for inconsistency for `isValid` which i agree

current behaviour

|  mode | isValid  |
|---|---|
|  submit | true  |
|  change | false  |
| blur  |  false  |

after this fix

|  mode | isValid  |
|---|---|
|  submit | false  |
|  change | false  |
| blur  |  false  |

also validationFields is not correctly reflected on `isValid` attribute as well, it shouldn't be included during valid check when it excludes such fields

barrymay:
well... it depends what isValid is used for (if you can point to one of those issues raised or an email you got about it, I'd be happy to see the relationship)  ;) In my case I'm not using it so this change doesn't bother me much. 

I still feel that the undefined status of isValid would be better until the initial validation run is performed, but that will admittedly take some work to get right.  

Feel free to merge :)

barrymay:
btw, is there a beta with this in it?

bluebill1049:
there is a beta: `3.24.2-beta.2`

the first one: 
https://github.com/react-hook-form/react-hook-form/issues/298#issuecomment-543571586

the second issue: is reported in spectrum with validationFields  
https://spectrum.chat/react-hook-form/help/best-way-to-register-unregister-conditional-form-fields~a86951dd-3a86-4d00-8c99-b72a2e37c7b4

barrymay:
Thanks - I’ll test and revert back with results

barrymay:
all good - let's go with it.  thanks!

bluebill1049:
thanks @barrymay for verifying.

PR #391, opened 2019-10-18
Add Curly Eslint Rule

barrymay:
- Add curly branch requirement for if statements
- Add lint:fix run call to both lint and prettier the changed files

(this change includes a run of lint:fix)


package.json
```diff
@@ -14,6 +14,7 @@
     "build:ie11": "rollup -c rollup.ie11.config.js",
     "prettier": "prettier --write './src/**/*.ts' './src/**/*.tsx' --config ./.prettierrc",
     "lint": "eslint ./src --ext .jsx,.ts --ignore-pattern *.test.ts --report-unused-disable-directives",
+    "lint:fix": "eslint ./src --ext .jsx,.ts --ignore-pattern *.test.ts --report-unused-disable-directives --fix && prettier --check 'src/**/*.ts' --write ",
```
bluebill1049: `prettier --check 'src/**/*.ts' --write` is this make above `prettier` script redudant?

bluebill1049: also is it make sense to put it part of the lint: stage?

barrymay: Agree on both - totally missed that the prettier script was right there..
I'll rename the old one to lint:check just in case you don't want to write it yet.

barrymay: see the latest - I think the link call should auto prettier in case linting changed code ;)

package.json
```diff
@@ -48,8 +48,8 @@
     "@types/react": "^16.9.0",
     "@types/react-dom": "^16.9.0",
     "@types/react-test-renderer": "^16.9.0",
-    "@typescript-eslint/eslint-plugin": "^1.9.0",
-    "@typescript-eslint/parser": "^1.9.0",
+    "@typescript-eslint/eslint-plugin": "^2.4.0",
```
barrymay: update to support latest TS (was throwing errors on running previously)

barrymay:
FYI most of this change was from an automated prettier run - see package.json and eslintrc for the actual changes

bluebill1049:
thanks a lot @barrymay great PR!

PR #400, opened 2019-10-22
skip early on validate function

bluebill1049:


JeromeDeLeon:
It would now be inconsistent IMO if this would not be the same as `abortEarly` using `Yup` schema.
I'm not saying this is bad, but if this approach is going to be push, might as well update `abortEarly` to support the same behavior for consistency. :smile

JeromeDeLeon:
Similar to my question before, about the difference between `abortEarly` because I thought that would also stop from validating the rest if one emits an error.

bluebill1049:
Yea true! Maybe we should one option for both?

JeromeDeLeon:
We would break the current API if we flatten the option for `abortEarly` but would allow us to be consistent.

bluebill1049:
Maybe we should skip early to true by default?

JeromeDeLeon:
Better if we set it to `false`for backward compatibility.

bluebill1049:
This is one of the mistake that i have made.... for validation (required, max, min) they actually skip early, but then valdiation schema is abortEarly false...

bluebill1049:
I am thinking of correcting them all in one go by set abortEarly to true for all...

bluebill1049:
or I can just fix in this PR and leave schema validation as it is right now

JeromeDeLeon:
I see, then apply this PR first then we can have discussion with other members later

bluebill1049:
personally I am towards fix validation to be consistent and ignore validation schema for now.
 your thoughts?

bluebill1049:
ok @JeromeDeLeon thanks

JeromeDeLeon:
I opened a discussion for this or should I opened it publicly to know opinion of others?

bluebill1049:
sounds good @JeromeDeLeon

JeromeDeLeon (approved):
LGTM! 👍

PR #401, opened 2019-10-22
remove touch mode and made touched by blur

bluebill1049:
after reading a few more docs and questions/answers. I realized touched should be only triggered by onBlur which means when control has lost focus. 

so I removed `touchedMode` and made default behavior to `onBlur` for touched state

### DIFF
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -21,7 +21,7 @@
 
 <div align="center"><p align="center"><a href="https://react-hook-form.com/zh" title="React Hook Form - Simple React forms validation"><img src="https://raw.githubusercontent.com/bluebill1049/react-hook-form/master/website/example.gif" alt="React Hook Form Logo - React hook form validation" width="750px" /></a></p></div>
 
-<a href="https://github.com/react-hook-form/react-hook-form">🇦🇺English</a> | 🇨🇳 简体中文 | <a href="./README.ja-JP.md">🇯🇵Japanese</a>
+<a href="https://github.com/react-hook-form/react-hook-form">🇦🇺English</a> | 🇨🇳简体中文 | <a href="./README.ja-JP.md">🇯🇵Japanese</a>
 
 ## 特性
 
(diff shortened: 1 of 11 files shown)


bluebill1049:
i will fix the automation tonight.

JeromeDeLeon (dismissed):

JeromeDeLeon:
The build seems to be failing for cypress

bluebill1049:
yea thanks @JeromeDeLeon I will fix it all tonight, just want to get your thoughts on it. looks like you agree on such change. 👍 I will sort out the rest.

JeromeDeLeon (dismissed):

bluebill1049:
thanks @JeromeDeLeon for reviewing 👍

rkurbatov:
@bluebill1049 it works perfectly for us, just as it was in redux-forms so now we can switch smoothly. But it's a breaking change, won't it affect somebody's else projects?

bluebill1049:
@rkurbatov I think it was a mistake on the very get-go, now the behavior is correct. touched was meant to only trigger onBlur, not onChange. I will make this as a minor change and info people in the release note.

leventeszmolka:
Hi,

It seems that on controlled inputs `touched` still gets triggered `onChange` and not `onBlur`.

bluebill1049:
@leventeszmolka provide a codeandbox for us to reproduce

leventeszmolka:
@bluebill1049 please see the codesandbox example here: https://codesandbox.io/s/wispy-smoke-50v9s?file=/src/App.js

bluebill1049:
@leventeszmolka you are right on that, can you raise an issue on this?

leventeszmolka:
@bluebill1049 I have raised the issue #1588. Many thanks for looking into it.

jtomaszewski:
Would you consider adding `touchedMode = "onFocus" | "onBlur"` ? Depending on how you looked at it, I might consider a field as touched if the user has interacted with, that is they focused with it at least once. Why would we delay that until the blur event happens?

Our use case: I want to show a warning on a field "no 2", wheneveruser focuses on the field "no 3 or later", while not filling the field no 2 first.

PR #413, opened 2019-10-26
Proxy form state typefix

barrymay:
Simple type improvements


bluebill1049:
thanks @barrymay looks good!

PR #430, opened 2019-10-29
Feature: support multiple errors for single field

bluebill1049:
Allow a single field to return multiple errors.

current behavior:
```
validateAllFieldCriteria: false, // default 

errors: {
  firstName: {
    type: 'xxx',
    message: 'xxx'
  }
}
```

we will expect:
```
validateAllFieldCriteria: true,

errors: {
  firstName: {
    types: {
      required: 'xxx',
      minLength: 'xxx',
    },
  }
}
```

Yet to be complete: 

- [x] allow multiple errors
- [x] setErrors

```
setErrors({
  name: 'xxx',
  types: {},
})
```


src/types.ts
```diff
@@ -55,6 +55,7 @@ export type Options<FormValues extends FieldValues = FieldValues> = Partial<{
   validationSchema: any;
   nativeValidation: boolean;
   submitFocusError: boolean;
+  displayAllErrors: boolean;
```
bluebill1049: Help: let me know if you guys have a better option name. @barrymay @JeromeDeLeon

bluebill1049: For allowing a single field to return multiple errors.

barrymay: suggested name validateAllFieldCriteria
or: 
validateCriteriaMode:
- firstErrorDetected (default for backward compat and performance)
- all

The library already validates multiple fields at the same time, but t

barrymay:
general note for tasks - if this is a PR based on a Spectrum request, would be great to have a link on the PR to point to it to avoid any confusion

src/types.ts
bluebill1049: thanks @barrymay great suggestion.

bluebill1049:
> general note for tasks - if this is a PR based on a Spectrum request, would be great to have a link on the PR to point to it to avoid any confusion

https://github.com/react-hook-form/react-hook-form/projects/1 (it is on one of the todo there)

it's triggered by spectrum issue yesterday, and it's been asked multiple times in the github issue and spectrum. I did some research yesterday with other form library seem like people do request quite a bit for such a feature. looks like implementation is not so difficult as well.

src/types.ts
bluebill1049: i love `validateCriteriaMode`

bluebill1049: any thoughts @JeromeDeLeon ?

bluebill1049: hmm i love `validateAllFieldCriteria` as well :)

barrymay: the only reason I suggest validateCriteriaMode is so that you can easily add more options as needed.    You can also create a string literal type in TypeScript to ensure the write text is provided...

barrymay: I'm fine with validateAllFieldCriteria for now

JeromeDeLeon: I'd like to know the difference between the use case of `validateSchemaOptions = { abortEarly: false }` and `validateAllFieldCriteria` aside from different use case such as with yup validation and native validation?

Because in the future, we might support returning all errors from field in schema validation and creating another config that does the same thing is not good DX at least IMO. :smiley: 

But if we can differentiate them, I'll go with what @barrymay suggested :+1:

src/useForm.ts
```diff
@@ -56,6 +56,7 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
   nativeValidation,
   submitFocusError = true,
   validationSchemaOption = { abortEarly: false },
+  validateAllFieldCriteria = false,
```
JeromeDeLeon: I'd like to know the difference between the use case of `validateSchemaOptions = { abortEarly: false }` and `validateAllFieldCriteria` aside from different use case such as with yup validation and native validation?

Because in the future, we might support returning all errors from field in schema validation and creating another config that does the same thing is not good DX at least IMO. :smiley: 

But if we can differentiate them, I'll go with what @barrymay suggested :+1:

bluebill1049: This feature will return all errors from a field.

JeromeDeLeon: Yes, definitely go with `validateAllFieldCriteria`

JeromeDeLeon: Is this available in beta? I'd like to test to sort my mind on it. :dizzy_face:

bluebill1049: I will release a beta 🙏🏻

JeromeDeLeon: Okay. I sorted it out. :smile: Seems like we have different use case with Yup that's why I was confused. `validateAllFieldCriteria` seems better name :grin:

bluebill1049: nice ~

bluebill1049: `react-hook-form@3.27.0-beta.1` here is the beta version. if you want to do some test, which i will do as well :)

JeromeDeLeon: I tested it with the new config using the latest beta your provided but I was still getting one error per field.

bluebill1049: thanks @JeromeDeLeon i will test it soon as well

bluebill1049:
FYI: I will be working over automation over the weekend.

bluebill1049:
still WIP...

src/types.ts
bluebill1049: oh sorry, @barrymay I just saw the message now. I am fine with both, whatever you guys prefer to use and make more sense 👍

bluebill1049: I like `validateCriteriaMode` as well, wondering what sort of options you have in your mind. Because it sounds like it's more future proof. It's better to get it right so we don't have to introduce breaking change.

bluebill1049: also not in rush to deliver this feature as well, I am taking my time with this one.

barrymay: cool - I think it's unlikely to introduce much more but at least you have the flexibility to do so without adding more flags ;)

bluebill1049: sounds good to me, will adjust accordingly.

bluebill1049: btw @barrymay I think you forgot to turn on your badge as a react hook form team member :) 
https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/publicizing-or-hiding-organization-membership

PR #442, opened 2019-11-02
close #440 use validation schema error instead of picking individual error

### LINKED ISSUE #440: Yup schema re-validation does not remove error on dependent fields
**Describe the bug**
When using .when() with yup, and reValidateMode is "onChange", only the field that is changed has its error removed, even if changing that field affects the validity of other fields

**To Reproduce**
1. Go to [Codesandbox](https://codesandbox.io/s/react-hook-form-validation-onchange-70qtf)
2. Click on 'Submit', both validation errors will appear
3. Change one of the inputs, only one validation error will disappear, but the form is now valid

**Codesandbox link**
[Codesandbox](https://codesandbox.io/s/react-hook-form-validation-onchange-70qtf)

**Expected behavior**
Both fields which are now valid would have their errors cleared

bluebill1049:
@JeromeDeLeon this should fix the issue you raised in spectrum

### DIFF
--- a/cypress/integration/basicSchemaValidation.ts
+++ b/cypress/integration/basicSchemaValidation.ts
@@ -108,7 +108,7 @@ context('basicSchemaValidation form validation', () => {
     cy.get('input[name="checkbox"]').check();
 
     cy.get('p').should('have.length', 0);
-    cy.get('#renderCount').contains('28');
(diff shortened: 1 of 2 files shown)


JeromeDeLeon (approved):
Thanks! :+1:

PR #464, opened 2019-11-06
Fixes #440. Prevents field names from hanging around in validFieldsRe…

### LINKED ISSUE #440: Yup schema re-validation does not remove error on dependent fields

MikeAmelung:
…f even if they have been invalidated indirectly.

### DIFF
--- a/app/src/basicSchemaValidation.tsx
+++ b/app/src/basicSchemaValidation.tsx
@@ -4,26 +4,39 @@ import * as yup from 'yup';
 
 let renderCounter = 0;
 
-const validationSchema = yup.object().shape({
-  firstName: yup.string().required(),
-  lastName: yup
-    .string()
-    .max(5)
-    .required(),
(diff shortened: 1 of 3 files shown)


bluebill1049:
omg too good, you even updated automation script! thank you

bluebill1049:
I will patch it in the next release :)

PR #467, opened 2019-11-06
WIP Feature: setErrors support multiple error at once with single render

### LINKED ISSUE #466: setErrors for multiple fields
I think that it's necessary to have this api: `setErrors`
When the server return the response of errors. I don't have to set error for all fields one by one by using `setError`. The form will only render one time.

bluebill1049:
close https://github.com/react-hook-form/react-hook-form/issues/466

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -95,6 +95,12 @@ export interface FieldError {
   isManual?: boolean;
 }
 
+export interface ManualFieldError<FormValues> {
+  name: FieldName<FormValues>;
+  type: string;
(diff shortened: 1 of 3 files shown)


ericlifs:
Has this change been reverted? I'm not seeing the `setErrors` function getting exported and I need it at the moment. Not sure if it has been replaced tho.

arunpt:
@bluebill1049 @ericlifs I'm also looking for the same. This would be a very useful feature.

PR #474, opened 2019-11-07
fix pattern/required issue

bluebill1049:
if required is not defined, the `pattern` should be skipped


rtalvarez:
I was just about to open an issue with this bug, only to find all I had to do was bump to latest

Thanks for all your hard work @bluebill1049 !

bluebill1049:
@rtalvarez my pleasure! leave a little ⭐️ if you find it useful 🙏 too. thank you ❤️

PR #478, opened 2019-11-07
Wrap handleSubmit with try-finally

krnlde:
Since an failing API call will destroy the form right now (e.g. `isSubmitting` is not being reset to `false`), I wrapped the `callback(...)` in a try-finally statement to ensure the variables are set in any case.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -807,6 +807,7 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
       isSubmittingRef.current = true;
       render({});
     }
+
     try {
       if (validationSchema) {
         fieldValues = getFieldsValues(fields);
@@ -889,12 +890,11 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
 
         errorsRef.current = fieldErrors;
       }
-
+    } finally {
       if (isUnMount.current) {
         return;
       }
 
-    } finally {
       isSubmittedRef.current = true;
       isSubmittingRef.current = false;
       submitCountRef.current = submitCountRef.current + 1;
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks for the PR @krnlde I think it will make more sense to try and catch in your App-level rather in the lib itself.

krnlde:
Of course, but then the library is not sure to be consistent. If I choose to catch the error on a higher level and not in the `onSubmit` the form will never get out of the `isSubmitting = true`. And with the approach above, no error is caught, since I only do try-finally and not try-catch-finally. I hope this clarifies the usecase a bit. Cheers :)

src/useForm.ts
```diff
@@ -807,97 +807,99 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
...
-            return false;
-          }
-          return previous;
-        }, true);
+        errorsRef.current = fieldErrors;
       }
 
-      errorsRef.current = fieldErrors;
-    }
+      if (isUnMount.current) {
```
bluebill1049: could you move this into finally as well.

bluebill1049:
you have a valid point. I will do some testing around it and let you know.

bluebill1049:
thank you merged and will be published in the next patch

src/useForm.ts
krnlde: Hey man is this still relevant? You also pushed something, and now I'm confused :D

bluebill1049: I did the fix myself. it's all good.

bluebill1049: thanks for the PR.

PR #482, opened 2019-11-08
Reset accepts partial FormValues

krnlde:
`reset()` accepts partial FormValues but this is not reflected in the method signature.


bluebill1049:
thanks @krnlde looks like build failed.

bluebill1049:
you are awesome! @JeromeDeLeon ❤️  thanks @kabriel too

JeromeDeLeon:
I was about to pull request because I noticed it too then I saw this :laughing:

PR #484, opened 2019-11-08
remove validation fields

bluebill1049:
This API is pretty useless to consider we have `unregister` API which is more declarative. 

Please use `unregister` in your app instead.


JeromeDeLeon:
Can we include it in the documentation or something in the release that this API is deprecated?

bluebill1049:
Thanks, @JeromeDeLeon hmmm,  good point but I don't think anyone is using it... But I will make sure to include that in the release note. (not a breaking change, I will do a migrate guide to use unregister instead)

JeromeDeLeon:
No worries, more like another section like `Removed API` even if no one uses it :smiley:

bluebill1049:
sounds good

PR #486, opened 2019-11-09
Remove remaining `validationFields`

JeromeDeLeon:
Might as well warn users that this API will be deprecated before removing in the next release :smiley:


bluebill1049:
thanks @JeromeDeLeon!! i missed them lol

PR #500, opened 2019-11-13
Feature/array of checkboxes

ldanet:
Adds support for groups of checkboxes as a single field as in the first example there:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Using_checkbox_inputs

### DIFF
--- a/src/logic/__snapshots__/findRemovedFieldAndRemoveListener.test.ts.snap
+++ b/src/logic/__snapshots__/findRemovedFieldAndRemoveListener.test.ts.snap
@@ -29,4 +29,6 @@ Object {
 }
 `;
 
+exports[`findMissDomAndClean should work for checkbox type input 1`] = `undefined`;
+
 exports[`findMissDomAndClean should work for radio type input 1`] = `undefined`;
--- a/src/logic/findRemovedFieldAndRemoveListener.test.ts
+++ b/src/logic/findRemovedFieldAndRemoveListener.test.ts
@@ -124,6 +124,7 @@ describe('findMissDomAndClean', () => {
       }),
     ).toMatchSnapshot();
   });
+
   it('should work for checkbox type input', () => {
     const ref = document.createElement('input');
     ref.setAttribute('name', 'test');
--- a/src/logic/findRemovedFieldAndRemoveListener.test.ts
+++ b/src/logic/findRemovedFieldAndRemoveListener.test.ts
@@ -124,6 +124,38 @@ describe('findMissDomAndClean', () => {
       }),
     ).toMatchSnapshot();
   });
+  it('should work for checkbox type input', () => {
+    const ref = document.createElement('input');
+    ref.setAttribute('name', 'test');
+    ref.setAttribute('type', 'checkbox');
+    document.body.contains = jest.fn(() => false);
+    const disconnect = jest.fn();
+    const fields = {
+      test: {
+        name: 'test',
+        ref: {},
+        mutationWatcher: {
+          disconnect,
+        },
+      },
+      test1: {
+        name: 'test',
+        ref: {
+          type: 'checkbox',
+        },
+      },
+    };
+
+    expect(
+      findRemovedFieldAndRemoveListener(fields, () => {}, {
+        ref: { name: 'test', type: 'checkbox' },
+        options: [{ ref }],
+        mutationWatcher: {
+          disconnect,
+        },
+      }),
+    ).toMatchSnapshot();
+  });
 
   it('should not remove event listener when type is not Element', () => {
     (isDetached as any).mockImplementation(() => {
--- a/src/logic/getCheckboxValue.test.ts
+++ b/src/logic/getCheckboxValue.test.ts
@@ -93,6 +93,7 @@ describe('getCheckboxValue', () => {
       ]),
     ).toEqual({ value: ['2', '3'], isValid: true });
   });
+
   it('should return values for checked boxes only', () => {
     expect(
       getCheckboxValue([
@@ -123,6 +124,7 @@ describe('getCheckboxValue', () => {
       ]),
     ).toEqual({ value: ['3'], isValid: true });
   });
+
   it('should return empty array for multi checkbox with no checked box', () => {
     expect(
       getCheckboxValue([
--- a/src/logic/getCheckboxValue.test.ts
+++ b/src/logic/getCheckboxValue.test.ts
@@ -10,44 +10,138 @@ describe('getCheckboxValue', () => {
 
   it('should return checked value if single checkbox is checked', () => {
     expect(
-      getCheckboxValue([{ ref: { name: 'bill', checked: true, value: '3' } }]),
+      getCheckboxValue([
+        {
+          ref: {
+            name: 'bill',
+            checked: true,
+            value: '3',
+            attributes: { value: '3' },
+          },
+        },
+      ]),
     ).toEqual({ value: '3', isValid: true });
   });
 
   it('should return true if single checkbox is checked and has no value', () => {
     expect(
-      getCheckboxValue([{ ref: { name: 'bill', checked: true } }]),
+      getCheckboxValue([
+        { ref: { name: 'bill', checked: true, attributes: {} } },
+      ]),
+    ).toEqual({ value: true, isValid: true });
+  });
+
+  it('should return true if single checkbox is checked and has empty value', () => {
+    expect(
+      getCheckboxValue([
+        {
+          ref: {
+            name: 'bill',
+            checked: true,
+            value: '',
+            attributes: { value: 'test' },
+          },
+        },
+      ]),
+    ).toEqual({ value: true, isValid: true });
+    expect(
+      getCheckboxValue([
+        {
+          ref: {
+            name: 'bill',
+            checked: true,
+            attributes: { value: 'test' },
+          },
+        },
+      ]),
     ).toEqual({ value: true, isValid: true });
   });
 
   it('should return false if single checkbox is un-checked', () => {
     expect(
-      getCheckboxValue([{ ref: { name: 'bill', checked: false, value: '1' } }]),
+      getCheckboxValue([
+        {
+          ref: {
+            name: 'bill',
+            checked: false,
+            attributes: {},
+          },
+        },
+      ]),
     ).toEqual({ value: false, isValid: false });
   });
 
   it('should return multiple selected values', () => {
     expect(
       getCheckboxValue([
-        { ref: { name: 'bill', checked: true, value: '2' } },
-        { ref: { name: 'bill', checked: true, value: '3' } },
+        {
+          ref: {
+            name: 'bill',
+            checked: true,
+            value: '2',
+            attributes: { value: '2' },
+          },
+        },
+        {
+          ref: {
+            name: 'bill',
+            checked: true,
+            value: '3',
+            attributes: { value: '3' },
+          },
+        },
       ]),
     ).toEqual({ value: ['2', '3'], isValid: true });
   });
   it('should return values for checked boxes only', () => {
     expect(
       getCheckboxValue([
-        { ref: { name: 'bill', checked: false, value: '2' } },
-        { ref: { name: 'bill', checked: true, value: '3' } },
-        { ref: { name: 'bill', checked: false, value: '4' } },
+        {
+          ref: {
+            name: 'bill',
+            checked: false,
+            value: '2',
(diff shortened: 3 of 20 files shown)


ldanet:
I tried to not have breaking changes and have single checkboxes behave the same as they already do.

bluebill1049:
looks like two automation test suits failed.

src/useForm.ts
```diff
@@ -187,8 +187,17 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
               selectRef.value,
             )),
         );
+      } else if (isCheckBoxInput(type) && options) {
```
bluebill1049: The tricky part is when options get populated, for the `radio` button. you can assume options will always be there. That why I was suggesting to use PHP like syntax to indicate that this checkbox is actually an array.

`<input name="test[]" />`

ldanet: the `&& options` here is just to make Typescript happy. Since the checkbox is registered the same as the radio, it shares the same assumptions most of the code, apart from the places where you get the value. 
This line of code reflects the code for the radio input a few lines above:
```ts
if (isRadioInput(type) && options) {
  options.forEach(
     ({ ref: radioRef }) => (radioRef.checked = radioRef.value === value),
   );
);
```

ldanet:
> looks like two automation test suits failed.

I forgot to update the cypress tests! Doing that.

bluebill1049:
> > looks like two automation test suits failed.
> 
> I forgot to update the cypress tests! Doing that.

awesome :) thanks

src/useForm.ts
bluebill1049: oh I see, I haven't looked it close enough :) I will take a look and help to test as well.

src/logic/findRemovedFieldAndRemoveListener.test.ts
```diff
@@ -124,6 +124,38 @@ describe('findMissDomAndClean', () => {
       }),
     ).toMatchSnapshot();
   });
+  it('should work for checkbox type input', () => {
```
bluebill1049:
```suggestion

  it('should work for checkbox type input', () => {
```

src/logic/getCheckboxValue.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  it('should return multiple selected values', () => {
+    expect(
+      getCheckboxValue([
+        { ref: { name: 'bill', checked: true, value: '2' } },
+        { ref: { name: 'bill', checked: true, value: '3' } },
+      ]),
+    ).toEqual({ value: ['2', '3'], isValid: true });
+  });
+  it('should return values for checked boxes only', () => {
```
bluebill1049:
```suggestion

  it('should return values for checked boxes only', () => {
```

src/logic/getCheckboxValue.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    expect(
+      getCheckboxValue([
+        { ref: { name: 'bill', checked: false, value: '2' } },
+        { ref: { name: 'bill', checked: true, value: '3' } },
+        { ref: { name: 'bill', checked: false, value: '4' } },
+      ]),
+    ).toEqual({ value: ['3'], isValid: true });
+  });
+  it('should return empty array for multi checkbox with no checked box', () => {
```
bluebill1049:
```suggestion

  it('should return empty array for multi checkbox with no checked box', () => {
```

src/logic/getCheckboxValue.ts
```diff
@@ -0,0 +1,35 @@
+import isArray from '../utils/isArray';
+import { RadioOrCheckboxOption } from '../types';
```
bluebill1049: minor: could you please move type to the bottom of the `import`? (consistency)

src/logic/getCheckboxValue.ts
```diff
@@ -0,0 +1,35 @@
...
+import { isUndefined } from 'util';
```
bluebill1049:
```suggestion
import isUndefined from '../utils/isUndefined';
```

src/logic/getCheckboxValue.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+const defaultResult: CheckboxFieldResult = {
+  value: false,
+  isValid: false,
+};
+
+export default (options?: RadioOrCheckboxOption[]): CheckboxFieldResult => {
+  if (isArray(options)) {
+    if (options.length > 1) {
```
bluebill1049:
```suggestion
    if (options.length) {
```

src/logic/getCheckboxValue.ts
```diff
@@ -0,0 +1,35 @@
...
+      const values = options
+        .filter(({ ref: { checked } }) => checked)
+        .map(({ ref: { value } }) => value);
+      return { value: values, isValid: !!values.length };
+    } else {
+      if (options[0].ref.checked) {
```
bluebill1049: `const { checked, value } = options[0].ref;`

src/logic/getCheckboxValue.ts
```diff
@@ -0,0 +1,35 @@
...
+          isEmptyString(options[0].ref.value)
+          ? { value: true, isValid: true }
+          : { value: options[0].ref.value, isValid: true };
+      }
+      return { value: false, isValid: false };
+    }
+  } else {
+    return defaultResult;
+  }
+};
```
bluebill1049: This is really smart! @ldanet your solution is great. ❤️

src/logic/getFieldValue.test.ts
```diff
@@ -113,24 +96,17 @@ describe('getFieldValue', () => {
       ),
     ).toEqual('');
   });
-
-  it('should return false when checkbox is not checked', () => {
+  it('should return false when checkbox input value is not found', () => {
```
bluebill1049:
```suggestion

  it('should return false when checkbox input value is not found', () => {
```

src/logic/getFieldValue.test.ts
```diff
@@ -113,24 +96,17 @@ describe('getFieldValue', () => {
...
-        },
+        {},
         {
           type: 'checkbox',
           value: 'value',
           name: 'test',
         },
       ),
-    ).toBeFalsy();
+    ).toEqual(false);
```
bluebill1049:
```suggestion
    ).toBeFalsy();
```

src/logic/getFieldValue.test.ts
```diff
@@ -144,10 +120,10 @@ describe('getFieldValue', () => {
           },
         },
         {
-          type: 'files',
+          type: 'file',
```
bluebill1049: Great pick up! thank you

src/useForm.test.tsx
```diff
@@ -307,6 +353,62 @@ describe('useForm', () => {
         } as React.SyntheticEvent);
       });
     });
+    it('should set value of multiple checkbox input correctly', async () => {
```
bluebill1049:
```suggestion

    it('should set value of multiple checkbox input correctly', async () => {
```

src/useForm.ts
```diff
@@ -187,8 +187,17 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
...
+        options.length > 1
```
```suggestion
        options.length
```

src/useForm.ts
```diff
@@ -630,10 +639,11 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
     };
     const fields = fieldsRef.current;
     const isRadio = isRadioInput(type);
+    const isCheckBox = isCheckBoxInput(type);
     let currentField = fields[name] as Field;
 
     if (
-      isRadio
+      isRadio || isCheckBox
```
bluebill1049: maybe we can change this into a variable?

src/logic/__snapshots__/findRemovedFieldAndRemoveListener.test.ts.snap
```diff
@@ -29,4 +29,6 @@ Object {
 }
 `;
 
+exports[`findMissDomAndClean should work for checkbox type input 1`] = `undefined`;
```
bluebill1049: hmmm, this is for me to fix, looks like those unit tests r unless...

src/logic/getCheckboxValue.test.ts
```diff
@@ -0,0 +1,54 @@
+import getCheckboxValue from './getCheckboxValue';
```
bluebill1049: nice unit test :)

src/logic/getCheckboxValue.ts
```diff
@@ -0,0 +1,35 @@
...
+        .map(({ ref: { value } }) => value);
+      return { value: values, isValid: !!values.length };
+    } else {
+      if (options[0].ref.checked) {
+        return isUndefined(options[0].ref.value) ||
+          isEmptyString(options[0].ref.value)
+          ? { value: true, isValid: true }
+          : { value: options[0].ref.value, isValid: true };
+      }
+      return { value: false, isValid: false };
```
bluebill1049:
```suggestion
      return defaultResult;
```

src/logic/getFieldValue.test.ts
```diff
@@ -44,42 +51,18 @@ describe('getFieldValue', () => {
...
-        {
-          type: 'checkbox',
-          checked: 'test',
-        },
-      ),
-    ).toBeTruthy();
-  });
-
-  it('should return checked if type is checked without value', () => {
+  it('should return the correct value when type is checkbox', () => {
```
bluebill1049:
```suggestion

  it('should return the correct value when type is checkbox', () => {
```

src/logic/getFieldValue.ts
```diff
@@ -27,14 +26,8 @@ export default function getFieldValue<FormValues extends FieldValues>(
...
-    if (checked) {
-      return ref.attributes && ref.attributes.value
-        ? isUndefined(value) || isEmptyString(value)
-          ? true
-          : value
-        : true;
-    }
-    return false;
+    const field = fields[name];
+    return field ? getCheckboxValue(field.options).value : false;
```
bluebill1049: I think we can bring this logic inside `getCheckboxValue`

`return getCheckboxValue(fields[name]).value;`

src/logic/validateField.ts
```diff
@@ -56,17 +57,17 @@ export default async <FormValues extends FieldValues>(
...
+      (isCheckBox && !getCheckboxValue(options).isValid) ||
+      (isRadio && !getRadioValue(options).isValid))
   ) {
     const message = isString(required) ? required : '';
 
     error[typedName] = {
       type: INPUT_VALIDATION_RULES.required,
       message,
-      ref: isRadio ? fields[typedName].options[0].ref : ref,
+      ref: isRadio || isCheckBox ? fields[typedName].options[0].ref : ref,
```
bluebill1049: maybe we can hoist it into a variable as well :)

bluebill1049 (changes requested):
All minor comments :) really great stuff here. Thanks for your solution, it is a lot better than mine.

src/logic/getCheckboxValue.ts
ldanet: It's `> 1` not ` > 0`, we still need to compare :smile:

bluebill1049: you are right!

src/useForm.ts
ldanet: Indeed they're never used alone now.

src/logic/getFieldValue.ts
ldanet: The issue is that if `fields[name]` is undefined (and Typescript has no reason to think it couldn't be), then trying to access `fields[name].options` is an error.

bluebill1049: that's fine :) i will remove the suggestion.

bluebill1049 (approved):
THANK YOU!

src/logic/getCheckboxValue.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const defaultResult: CheckboxFieldResult = {
+  value: false,
+  isValid: false,
+};
+
+export default (options?: RadioOrCheckboxOption[]): CheckboxFieldResult => {
+  if (isArray(options)) {
+    if (options.length > 1) {
```
JeromeDeLeon: What about `options[1]` ?

bluebill1049: yeah! love it less code the better :)

PR #526, opened 2019-11-16
close #524 patch unmount with validation

### LINKED ISSUE #524: setState on unmounted component when closing modal form on keypress
**Describe the bug**

- Form is rendered in a modal.
- escape key press closes modal (unmounts form)
- while in `mode: 'onBlur'` this key triggers onBlur validation which is likely setting state
after form is unmounted

**To Reproduce**
- show modal form
- touch inside input
- press escape key to close modal
- see console error

**Codesandbox link**
https://codesandbox.io/s/react-hook-form-errors-gfmtj

**Expected behavior**
library checks mounted state before setting state onBlur (or elsewhere)

 - OS: Linux
 - Browser: Brave

bluebill1049:


src/useForm.ts
```diff
@@ -390,7 +390,7 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
         const currentError = errors[name];
         let error;
 
-        if (!field) {
+        if (!field || isUnMount.current) {
```
cdock1029: I tested this locally, logged out a few places in the code. At the point of this check, `isUnmount` has not been updated yet and is still `false`. The `useEffect` callback runs after this code above runs subsequent to the onBlur.

src/useForm.ts
```diff
@@ -433,7 +433,11 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
           error = await validateFieldCurry(field);
         }
 
-        if (!renderBaseOnError(name, error) && shouldUpdateState) {
+        if (
```
cdock1029: same as above, `isUnmount` is false, `render` executes, then `useEffect` runs

src/useForm.ts
cdock1029: I think it is worth trying to solve this further so you can short-circuit parts of the code if unmounted, but in the short term the core of the bug can be solved with this:

```js
const [, /*render*/_render] = useState();
const render = useCallback(obj => {
  if (!isUnmount.current) {
    _render(obj)
  }  
}, [])

```

bluebill1049: > I tested this locally, logged out a few places in the code. At the point of this check, `isUnmount` has not been updated yet and is still `false`. The `useEffect` callback runs after this code above runs subsequent to the onBlur.

thanks for testing @cdock1029 do you have a codesandbox where I can reproduce this?

src/useForm.ts
bluebill1049: Actually, that was my bad, the above unmount check should be removed. This check should prevent the warning (memory leak).

src/useForm.ts
bluebill1049: > I think it is worth trying to solve this further so you can short-circuit parts of the code if unmounted, but in the short term the core of the bug can be solved with this:
> 
> ```js
> const [, /*render*/_render] = useState();
> const render = useCallback(obj => {
>   if (!isUnmount.current) {
>     _render(obj)
>   }  
> }, [])
> ```

I like this idea! do you want to submit a PR, since you are the one figure this out?

src/useForm.ts
bluebill1049: Is this patched version still causing you with issue?

src/useForm.ts
bluebill1049: I have removed this line: https://github.com/react-hook-form/react-hook-form/commit/c96901a3660b9de0b9333124a8258287d4a265c6

cdock1029: Sorry I updated my comment trying to be a little more clear of what I found, and what I wasn't quite certain about. Still seeing the issue, the problem I see is that even after unmount, `resetFieldRef` and `renderBaseOnError` execute, and since the call `render`, that's why we see the error.

I think since your only `state`update is `render`, you might as well wrap that function with it's own internal guard. This should avoid current and future mistakes of setting state while unmounted.

To **further** optimize not running additional code while unmounted, you could apply `isUnmounted` checks to `resetFieldRef` and `renderBaseOnError` where appropriate.

bluebill1049: sounds good to me. interested to send a PR since you know the solution?

PR #527, opened 2019-11-17
wrap render state update with isUnmount guard

cdock1029:


bluebill1049:
thanks @cdock1029 I think we can safely remove other `render` checks for mounted as well.

bluebill1049 (approved):
Never mind, i fixed it :)

bluebill1049:
@cdock1029 let's test this before we merge it in. you probably have some test code already.

cdock1029:
> Never mind, i fixed it :)

beat me too it by a minute

bluebill1049:
haha @cdock1029 I tested it and it's working for me :) merge when you are ready 👍

cdock1029:
I'm having trouble getting the error to happen with Modal -> Form in testing vs running component in live example or app.

I've isolate the situation to confirm this bug only happens when form rendered inside a `createPortal` component, and this fix solves it but can't reproduce with testing-library. Maybe it handles portals differently.

Don't have any time right now to keep messing with it if you want to take a stab at it yourself.
This is a minimal example reproducing it, if you can translate to testing-library:

https://codesandbox.io/s/react-hook-form-defaultvalues-t9xoc

bluebill1049:
no worries i will take it over :) and investigate further

bluebill1049:
I can confirm this PR fix the issue, however, i am a bit confused about why it only triggers error the first time as well.

bluebill1049:
I couldn't reproduce in testing-library as well... I tried with `blur` event, which is even working without this patch, then I did with press `esc` key and it's not even trigger 'blur' event. I will merge this PR first, looks like this issue is resolved by this patch from testing in the browser environment. strange it's not even a problem at all with testing lib. Feel free to leave your thoughts and future discovery here. I won't release this patch in a rush anyhow :)

PR #538, opened 2019-11-20
useFormContext optimize

yi-tuan:
A friendlier error is given when using useFormContext outside of the FormContext

### DIFF
--- a/src/useFormContext.tsx
+++ b/src/useFormContext.tsx
@@ -7,13 +7,7 @@ const FormGlobalContext = React.createContext<FormContextValues<
 > | null>(null);
 
 export function useFormContext<T extends FieldValues>(): FormContextValues<T> {
-  const formGlobalContext = React.useContext(FormGlobalContext);
-
-  if (formGlobalContext === null) {
-    throw new Error('useFormContext must under FormContext');
-  }
-
-  return formGlobalContext as FormContextValues<T>;
+  return React.useContext(FormGlobalContext) as FormContextValues<T>;
 }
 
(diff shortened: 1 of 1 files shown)


src/useFormContext.tsx
```diff
@@ -7,7 +7,13 @@ const FormGlobalContext = React.createContext<FormContextValues<
...
 
 export function useFormContext<T extends FieldValues>(): FormContextValues<T> {
-  return React.useContext(FormGlobalContext as any);
+  const formGlobalContext = React.useContext(FormGlobalContext);
+
+  if (formGlobalContext === null) {
+    throw new Error('useFormContext must under FormContext');
+  }
+
+  return formGlobalContext as FormContextValues<T>;
```
bluebill1049: Hey @yi-tuan thanks for the PR! 🙏 To be honest, I probably can live without this. because when you calling `useFormContext` the object will be empty anyhow, but I like 

`return React.useContext(FormGlobalContext) as FormContextValues<T>;`

it's better than that `as any`.

can you update that and let's merge it in?

yi-tuan: yes, the code works fine without this, but i think we need to tell developers that `useFormContext` is not being used correctly.If I happen to use it in the wrong place, I can quickly locate error code.

bluebill1049: Thanks @yi-tuan I definitely see your point. at the same time, I try to reduce the amount of 'unnecessary(can be avoided)' code inside the lib, every byte matters. Hope you understand. thanks 🙏

PR #560, opened 2019-11-25
fix #555 support all primitive types for custom fields

### LINKED ISSUE #555: Problem watching complex custom fields
I'm trying to implement a single source of truth via react-hook-form for my dynamic form.
Let's say there is a dynamic array of objects (`animals`) with props `name` and `type`, so the fields will have the following names: `animals[index].name` and `animals[index].type`. And `type` is a custom field.
I watch all the animals using `watch('animals')`, animals can be added and removed dynamically and I have a few animals defined initially.

The problem appears when I try to register a custom field that has nested props, i.e. `register({name: 'animals[0].type'})`. It doesn't matter if I have `defaultValues` suppled to the `watch` method or not I have the following error:
```
Uncaught TypeError: Cannot convert undefined or null to object
```

One possible workaround that I figured out is to manually set initial value right after the field was registered, i.e. `setValue('animals[0].type', 'Cat')`. But this makes the form dirty and it's not the behavior I want for the initial fields.

The second issue is that even if I put `type: 'custom'` it won't re-render component after `setValue` is called. I believe this is only when it's a complex field (with nested props) and when you watch not all fields.

**Expected behavior**
1. `watch` method should either use `defaultValues` provided as the second argument when trying to resolve values for complex custom fields or just use `undefined` if `defaultValues` wasn't provided
2. `setValue` should trigger re-render if a field was registered with `type: 'custom'` prop even if it's a complex custom field.

P.S. Thank you for your great work on this library!

GProst:
`getPath` method threw an error in certain cases when `values` contained some primitive types which are not strings. You can execute updated `getPath.test.ts` on the current `master` branch to see how it fails.

I created a new `isPrimitive` util and refactored `getPath` method

### DIFF
--- a/src/utils/getPath.ts
+++ b/src/utils/getPath.ts
@@ -1,32 +1,23 @@
 import flatArray from './flatArray';
-import isString from './isString';
-import isObject from './isObject';
+import isPrimitive from './isPrimitive';
 import { FieldValues, FieldName } from '../types';
 import isArray from './isArray';
 
 const getPath = <FormValues extends FieldValues = FieldValues>(
   path: FieldName<FormValues>,
-  values: FormValues | string[] | string,
+  values: FormValues | any[],
 ): any[] =>
   isArray(values)
     ? values.map((item, index) => {
         const pathWithIndex = `${path}[${index}]`;
-
-        if (isArray(item)) {
-          return getPath(pathWithIndex, item);
-        } else if (isObject(item)) {
-          return Object.entries(item).map(([key, objectValue]: [string, any]) =>
-            isString(objectValue)
-              ? `${pathWithIndex}.${key}`
-              : getPath(`${pathWithIndex}.${key}`, objectValue),
-          );
-        }
-
-        return pathWithIndex;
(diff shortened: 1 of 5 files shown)


bluebill1049:
amazing @GProst thanks for the PR! I will test it out soon.

src/utils/getPath.ts
```diff
@@ -1,32 +1,23 @@
...
-import isString from './isString';
-import isObject from './isObject';
+import isPrimitive from './isPrimitive';
 import { FieldValues, FieldName } from '../types';
 import isArray from './isArray';
 
 const getPath = <FormValues extends FieldValues = FieldValues>(
   path: FieldName<FormValues>,
-  values: FormValues | string[] | string,
+  values: FormValues | any[],
```
GProst: I believe we only expect either an object here (`FormValues`) or an array of any type.
I rewrote this method in this fact in mind, hopefully it looks simpler now and works properly 😄
@bluebill1049  Looking forward to your feedback!

bluebill1049:
looks really good man!

bluebill1049 (approved):

PR #602, opened 2019-12-01
fix: fix async validate type

yi-tuan:


src/types.ts
```diff
@@ -68,7 +68,7 @@ export type ValidationValue = boolean | number | string | RegExp;
 
 export type ValidateResult = string | boolean | undefined;
 
-export type Validate = (data: any) => ValidateResult;
+export type Validate = (data: any) => ValidateResult | Promise<ValidateResult>;
```
bluebill1049: thanks for your PR, why do you return promise?

bluebill1049: even async function should return `string` or `boolean` right? (do i miss another use case)

yi-tuan: image
async function should return `Promise`.

bluebill1049: oh, I see. thanks! my bad.

yi-tuan: This library is very useful, I should thank you.

bluebill1049: my pleasure @yi-tuan thanks for your help too. ❤️

Issue #579, opened 2019-11-27
It is not a typescript error to access an incorrect field on the errors object.


mkellyclare:
**Describe the bug**
It is not a typescript error to access an incorrect field on the errors object. In the [typescript section of the docs](https://react-hook-form.com/api#TypeScript) - there is a line that should demonstrate this behaviour and a comment indicating that a typescript error should be there.
```javascript
errors.bill; // ❌: property bill does not exist
```
However, when testing locally on in the linked sandbox, there is no typescript error on this line.

**To Reproduce**
1. Go to the [typescript section of the docs](https://react-hook-form.com/api#TypeScript)
2. Click on the code sandbox link
3. Expect to see typescript error on line 29

**Codesandbox link**
https://codesandbox.io/s/react-hook-form-typescript-mm

**Expected behavior**
If I try and access a field in the errors object that is not defined, this should be a typescript error.

bluebill1049:
your codesandbox is empty: https://codesandbox.io/s/react-hook-form-typescript-mm can you supply one?

mkellyclare:
Strange, it looks like the typescript section of the docs has been removed. The code sandbox was the one linked directly from that.

I had copied the example from the docs locally, here is a codesandbox with the code that was in that example before it was removed that demonstrates the issue - https://codesandbox.io/s/red-dawn-xgn9l

bluebill1049:
@mkellyclare not removed, migrated to get started page

bluebill1049:
hey @stramel would you be able to help on this?

bluebill1049:
@stramel any update on this issue?

iangregsondev:
Probably not helpful, but locally I am also seeing the same issue. As per the docs I create a type 

```
type FormData = {
  email: string
}
```

And using 

```
  const { register, handleSubmit, watch, errors } = useForm<FormData>()
```

And this does not highlight any issue, but it should as its not on the type.

```
console.log(errors.somethingNotThere)
```

This was created the latest version and I am using vscode.

bluebill1049:
i am looking into this over the weekend @iangregsondev. let me know @stramel if you have worked on it already i can branch off what you left off.

PR #630, opened 2019-12-06
WIP on issue #579

bluebill1049:


alias-mac:
@bluebill1049, this merge broke the types on errors that have nested fields, like `errors['address.street']`.

I wasn't able to trace the exact line, but the problem is that this makes `errors` defined as `Partial<Record<"firstName" | "lastName" | ...` and `address` in my case is a nested field, like `address.street`.

bluebill1049:
hey, @alias-mac sorry to break your app... looks like I have miss that use case. I will try to patch that soon.

bluebill1049:
hey @alias-mac  what about giving it a type for now until we release the next major version for the nested errors object. since optional chaining is getting better coverage, we can start doing `errors?.address?.street`

for now: is this going to fix your problem?
`['address.street']: string`?

alias-mac:
@bluebill1049 thanks for the fast reply!
I upgraded to react-scripts 3.3.0 recently to get optional chaining 😄, so I agree with you on doing that. I believe that was your goal anyway by reading https://react-hook-form.com/api#errors

I added hack for now to keep using the latest version that contains this fix. It also helped me catch some inputs that weren't registered but were using the `errors` property (which would never be defined).

Thanks for keeping this library awesome 🎉 !
Looking forward on seeing the nested support working.

bluebill1049:
thanks @alias-mac it should be pretty simple to switch to `errors` object, I did flat errors purely because optional chaining was not that popular at the time when I was designing the API. Cheers for the support! if you find it useful please consider to support us by leaving a  start too :) (no pressure tho)

PR #635, opened 2019-12-08
improve controlled component onchange with setValue

bluebill1049:
Original issue: https://github.com/react-hook-form/react-hook-form/issues/477

This fix was inspired by MUI's author & related to the following GitHub issue. 
It was originally written by Kent ❤️ : 
https://github.com/facebook/react/issues/10135#issuecomment-401496776

@JeromeDeLeon I will keep the reference in this PR instead of the code. 

This PR will improve the following API

`reset()`
`setValue()`

with controlled components.

PR #666, opened 2019-12-14
💪V4

bluebill1049:
## 🔨 These are the breaking changes

- [x] ✨improve module **exports**:

(we will export all types as well)

```
``` 

🚨`import useForm from 'react-hook-form';`  will no longer work

---

- [x] ✨nested `errors` object and better typescript support

```
type form = {
  yourDetail: {
    firstName: string,
  }
}

errors?.yourDetail?.firstName
```

🚨no more flat errors :-) better TS support

---

- [x] ✨triggerValidation argument change from `Object`, `Object[]` to `String`, `String[]`

```
triggerValidation('firstName');
triggerValidation(['firstName', 'lastName']);
```
🚨`triggerValidation({ name: 'test', value: 'test'  });` no longer work

---

- [x] ✨watch support `{ nest: boolean }`

```
watch(); // { 'test.firstName': 'bill' }
watch({ nest: true }); // { test: { firstName: 'bill' } }
```

--- 

- [x] ✨improve custom `register`

```
register('test', { required: true });
```

🚨`register({ name: 'test'})` => `register('test')`

---
- [x] ✨`setError` support nested object

```
setError('yourDetail.firstName', 'test');
errors.yourDetails.firstName;
```

🚨`setError('yourDetail.firstName')` no longer output flat errors object 

---

## 🥂 New feature

- [x] `control`:  this will be the prop pass to RHFInput

we probably can skip this hack too: https://github.com/react-hook-form/react-hook-form/pull/635

```
<RHF control={control} name="test" />
```

https://github.com/react-hook-form/react-hook-form-input/pull/45

---

- [x] move `RHFInput` into the main repo and rename it to `Controller`

https://github.com/react-hook-form/react-hook-form-input

---

## ✂️ API to be removed list:

- [x] `validationSchemaOption`: hardly anyone want to use validation with abort early, or change the config.

---

- [x] native validation: hardly anyone used this feature. https://react-hook-form.com/api/#Browserbuiltinvalidation

FYI @JeromeDeLeon @kotarella1110 @adhrinae @krnlde 

Feedback welcome guys


src/useForm.ts
```diff
@@ -635,6 +615,7 @@ export default function useForm<FormValues extends FieldValues = FieldValues>({
   }
 
   function watch(): FormValues;
+  function watch(option: { nest: boolean }): FormValues;
```
JeromeDeLeon: so if we specify `nest` to `true`, we aren't allowed to specify fields?

bluebill1049: not with `nest`, because you are getting an object and you can deconstruct it.

bluebill1049: or perhaps we can :)  but it will increase the complexity of the API. We can go there when there are people requesting it.

JeromeDeLeon: I haven't use `watch` that much and therefore, I thought that the return would automatically  be constructed like `getValues({ nest: true })`

bluebill1049: yes you are right. it's basically `getValues`.

bluebill1049:
Screen Shot 2019-12-14 at 9 41 08 pm

OMG! 666

giphy

JeromeDeLeon:
> Screen Shot 2019-12-14 at 9 41 08 pm
> 
> OMG! 666
> 
> giphy

We're doing breaking changes, hence the number lol

bopfer:
I like all these changes and I think they are nice quality of life improvements.

One other potential change I would like to see, if possible, is related to Yup schema validation.  Currently, the validation returns the last error from the Yup validations. See this sandbox: https://codesandbox.io/s/react-hook-form-reset-form-2wg09

But, I think it would make more sense to return the first one, `required` in this case.  With the current `react-hook-form`, I need to invert all my Yup schemas to put the first things I want validated at the end.  However, I would prefer to have them at the start.  It's a minor issue and would be a breaking change for current Yup users, but it would be nice.

PS: I have tried using `validationSchemaOption: { abortEarly: true }`, but that doesn't seem to return any errors.  So, maybe there is just a bug there that needs to be fixed.  I am not really sure.  Either way, I think returning the first error would still be a better option.

bluebill1049:
> But, I think it would make more sense to return the first one, `required` in this case. With the current `react-hook-form`, I need to invert all my Yup schemas to put the first things I want validated at the end. However, I would prefer to have them at the start. It's a minor issue and would be a breaking change for current Yup users, but it would be nice.

I think it totally make sense to return the first error it encounter, is this coming from Yup tho? is there any configure we can change to let Yup return the first encounter error.

JeromeDeLeon:
@bopfer Upon looking your concern, it actually return an error when using `validate` of `Yup`. The problem is when [parsingErrorSchema](https://github.com/react-hook-form/react-hook-form/blob/1bb6e543237c9b0f9a8e0e657bdd600912980061/src/logic/validateWithSchema.ts#L21). I'll try to look more into it.

bluebill1049:
> @bopfer Upon looking your concern, it actually return an error when using `validate` of `Yup`. The problem is when [parsingErrorSchema](https://github.com/react-hook-form/react-hook-form/blob/1bb6e543237c9b0f9a8e0e657bdd600912980061/src/logic/validateWithSchema.ts#L21). I'll try to look more into it.

oh cool @JeromeDeLeon feel free to branch off from this PR to get this one right 👍

JeromeDeLeon:
Oops. I re-added it again because of `abortEarly`

bluebill1049:
> Oops. I re-added it again because of `abortEarly`

no worries, i think this validation schema option is pretty useless, hence removing it.

bluebill1049:
just branch off from v4 branch if you want to change something @JeromeDeLeon

JeromeDeLeon:
@bopfer , Seems like we're removing the `abortEarly` for the next version. I tried searching for use case of `abortEarly` and based on this [issue](https://github.com/jquense/yup/issues/718), the default value of `abortEaly` is `true` according to [Yup](https://github.com/jquense/yup#mixedvalidatevalue-any-options-object-promiseany-validationerror), they run in parallel and the first error the emit will be returned to you. It's not the first validation you set per field but the first error the comes out when you validate.

bluebill1049:
> @bopfer , Seems like we're removing the `abortEarly` for the next version. I tried searching for use case of `abortEarly` and based on this [issue](https://github.com/jquense/yup/issues/718), the default value of `abortEaly` is `true` according to [Yup](https://github.com/jquense/yup#mixedvalidatevalue-any-options-object-promiseany-validationerror), they run in parallel and the first error the emit will be returned to you. It's not the first validation you set per field but the first error the comes out when you validate.

my understanding `abortEarly: true` is whenever the first error encounter, and stop rest of the validation.

what @bopfer is after is order of the validation result.

```
const validationSchema = object().shape({
  firstName: string()
    .required()
    .min(3)
});
```

this will result `firstName must be at least 3 characters` first appear than required, i think it's our error parsing or the order of yup validation. (mostly due to us, i haven't look close enough).

bopfer:
> @bopfer , Seems like we're removing the `abortEarly` for the next version. I tried searching for use case of `abortEarly` and based on this [issue](https://github.com/jquense/yup/issues/718), the default value of `abortEaly` is `true` according to [Yup](https://github.com/jquense/yup#mixedvalidatevalue-any-options-object-promiseany-validationerror), they run in parallel and the first error that emitted will be returned to you. It's not the first validation you set per field like `.required().min(3)` but the first error from the form the comes out when you validate.

I think removing `abortEarly` is fine.  It doesn't appear to work right now anyway.  `false` appears to be the default and `true` doesn't return any errors.

> I think it's our error parsing or the order of yup validation. (mostly due to us, i haven't look close enough).

It would be awesome if something can just be flipped in the parsing to return the first error.  Assuming Yup gives the errors an in order that makes sense.

JeromeDeLeon:
@bopfer let me know your thoughts about this #668

cypress/integration/formState.ts
```diff
@@ -3,38 +3,38 @@ context('form state', () => {
     cy.visit('http://localhost:3000/formState/onSubmit');
 
     cy.get('#state').contains(
-      '{"dirty":false,"isSubmitted":false,"submitCount":0,"touched":[],"isSubmitting":false,"isValid":false}',
+      '{"dirty":false,"isSubmitted":false,"submitCount":0,"touched":{},"isSubmitting":false,"isValid":false}',
```
JeromeDeLeon: This is really clever! So this is how you get a hold of component state by printing them on the dom and track them. Nice!

JeromeDeLeon: This is what I wanted. I mean, I really don't know I to monitor my state in testing. So this is one way of doing it. :)

bopfer:
> @bopfer let me know your thoughts about this #668

That looks like it will take care of what I wanted.  Thanks!

bluebill1049:
> > @bopfer let me know your thoughts about this #668
> 
> That looks like it will take care of what I wanted. Thanks!

by the way, we are not fixing it for @bopfer. We think that's the right and consistent way for the validation report. It's more of a correction.

krnlde:
Oh wow, I just come out of the weekend and have so much to read :D Sorry guys I won't manage to pick up everything.

bluebill1049:
> Oh wow, I just come out of the weekend and have so much to read :D Sorry guys I won't manage to pick up everything.

thanks! that's all good, just take a look at the summary on the top. let us know your feedback. much appreciated.

src/useForm.ts
```diff
@@ -323,9 +304,7 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
         combineFieldValues(getFieldsValues(fieldsRef.current)),
       );
       const isMultipleFields = isArray(payload);
-      const names = isArray(payload)
-        ? payload.map(({ name }) => name)
-        : [payload.name];
+      const names = isArray(payload) ? payload.map(name => name) : [name];
```
kotarella1110: `map` is unnecessary.
Isn't it a payload `instead` of a `name`?

the following is correct:

```ts
const names = isArray(payload) ? payload : [payload];
```

src/useForm.ts
```diff
@@ -363,13 +342,11 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
...
     async (
-      payload?:
-        | ValidationPayload<FieldName<FormValues>, FieldValue<FormValues>>
-        | ValidationPayload<FieldName<FormValues>, FieldValue<FormValues>>[],
+      payload?: FieldName<FormValues> | FieldName<FormValues>[],
       shouldRender?: boolean,
     ): Promise<boolean> => {
       const fields =
-        payload || Object.keys(fieldsRef.current).map(name => ({ name }));
+        payload || Object.keys(fieldsRef.current).map(name => name);
```
kotarella1110: `map` is unnecessary.

the following is correct:
```ts
const fields = payload || Object.keys(fieldsRef.current);
```

src/useForm.ts
bluebill1049: Awesome pick up. Send PR to this branch if u want ❤️

src/useForm.ts
bluebill1049: Correct!

src/useForm.ts
kotarella1110: Thanks. I will!

kotarella1110: Sorry, this is already fixed :scream:

src/types.ts
```diff
@@ -103,9 +103,15 @@ export type FieldRefs<FormValues extends FieldValues> = Partial<
   Record<FieldName<FormValues>, Field>
 >;
 
-export type FieldErrors<FormValues extends FieldValues> = Partial<
-  Record<FieldName<FormValues>, FieldError>
->;
+export type FieldErrors<FormValues> = {
+  [Key in keyof FormValues]?: FormValues[Key] extends any[]
```
bluebill1049: @stramel review here plz

krnlde:
I read the summary and I really like the new stuff. My only bit of concern would be the Set instead of Array regarding browser compatibility. But if the pipeline already transpiles/polyfills that everything's cool :) This'll be a major version bump, right?

bluebill1049:
> I read the summary and I really like the new stuff. My only bit of concern would be the Set instead of Array regarding browser compatibility. But if the pipeline already transpiles/polyfills that everything's cool :)

good point. alternative we can convert into `Object`? 

1. `touched.test`

2. `touched.includes('test')`

3. `touched.has('test')`

@kotarella1110 @JeromeDeLeon @bopfer @krnlde which one is you guys reckon is easier for developers.

bluebill1049:
Early beta version: `4.0.0-beta.1`

Pratikkakkad:
when can we expect this to release? I am just really excited 😄

bluebill1049:
> when can we expect this to release? I am just really excited 😄

Thanks @Pratikkakkad I am still working on it. I will ping you once we are ready or have release candidate version.

src/types.ts
stramel: This looks fine to me. Looks pretty similar to the other lib's typing that we were discussing.

stramel: **NOTE:** I didn't test this

bluebill1049: awesome thanks @stramel i did test it.

app/src/basic.tsx
```diff
@@ -20,15 +20,35 @@ const Basic: React.FC = (props: any) => {
...
+    };
+    arrayItem: { test1: string; test2: string }[];
   }>({
     mode: props.match.params.mode,
   });
   const onSubmit = () => {};
 
   renderCounter++;
 
+  console.log(errors);
```
JeromeDeLeon: Are you doing your testing here? @bluebill1049

bluebill1049: oops i will remove that!

bluebill1049:
`react-hook-form@4.0.0-beta.2`

bluebill1049:
Hey guys, if you have free sometimes. Please helping on testing the new RC: `react-hook-form@4.0.0-rc.1`

bluebill1049:
`react-hook-form@4.0.0-rc.2`

bluebill1049:
`react-hook-form@4.0.0-rc.3`

bluebill1049:
Looks pretty good at my end:

https://codesandbox.io/s/react-hook-form-v4-quick-test-9e18f

bluebill1049:
Thanks again for your help @umidbekkarimov 🙏

bluebill1049:
Thanks 💪@JeromeDeLeon 💪@kotarella1110 and 💪@umidbekkarimov 

merged ❤️

src/logic/skipValidation.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  it('when re-validate mode is blur, not blur event and has error ', () => {
+    expect(
+      skipValidation({
+        hasError: true,
+        isBlurEvent: false,
+        isOnSubmit: false,
+        isReValidateOnSubmit: true,
+        isOnBlur: false,
+        isReValidateOnBlur: false,
```
AhmedAymanM: what I get from the test message is that `isReValidateOnSubmit` should be `false` and `isReValidateOnBlur` should be `true`

bluebill1049: you are right! feel free to send a PR 🙏

AhmedAymanM: opened a PR :) 
https://github.com/react-hook-form/react-hook-form/pull/702
best of luck 🙏

bluebill1049: merged thank you :)

AhmedAymanM: awesome! there is no need! :)

babramczyk:
@bluebill1049 Awesome work! Thanks for the nested `errors` object, it's a huge help!

bluebill1049:
> @bluebill1049 Awesome work! Thanks for the nested `errors` object, it's a huge help!

nice! i am glad you find it useful 👍 now that optional chaining is adopted by most of the transpilers.

PR #677, opened 2019-12-16
add editedFields to formState

francois-pasquier:
This PR allows one to check for an initial edition on a field before showing errors like `required` on `mode="onBlur"`

@bluebill1049 tell me if there is any problem


bluebill1049:
hey @MeisterTea,

first of all, thanks for submitting a PR. However, I prefer this logic to leave at your app-level by leveraging watch() and combined with default values. I think that's quite easily achievable. I am happy to produce a codesandbox for this behavior.

PR #694, opened 2019-12-23
Add control to FormContextValues, return FormContextValues in useForm

avocadowastaken:
Make `useForm` and `FormContextProvider` APIs consistent.


bluebill1049:
thanks a lot for looking into this! @umidbekkarimov really appreciated it.

PR #705, opened 2019-12-24
fix #704 only run validate schema when reading isValid state

### LINKED ISSUE #704: Why component is rendering twice when using validationSchema
1. Go to https://codesandbox.io/s/react-hook-form-rhfinput-iov6f
2. Check number of renders when validationSchema is used
3. Check number of renders when validationSchema is unused

I'm trying to understand why when i'm using a validationSchema in useForm, my component is rendering twice. And when i'm not using it, component is rendering once.
However, when i'm using rules in TextField register, my component is rendering once.

bluebill1049:


stephoune:
Thank's !!! :D

Issue #698, opened 2019-12-23
Inconsistent `watch()` behaviour on complex field names


KATT:
**Describe the bug**

When having default values with complex fields the initial `watch()`-call gives the default values rather than the flattened version.

**To Reproduce**
1. Go to https://codesandbox.io/s/react-hook-form-watch-issue-hbgwl
2. Observe _"Data from `watch()`:"_ below the Submit-button
3. Edit _Complex field_

**Codesandbox link**

https://codesandbox.io/s/react-hook-form-watch-issue-hbgwl

**Expected behavior**

Both initial `watch()`-call and following `watch()`-calls should produce `Record<string, any>` rather than whatever default form values you put in.

bluebill1049:
This is expected behaviour for V3 as it described in the Doc, however in V4 which will release soon will introduce `watch({ nest: true })` which produce nested object.

bluebill1049:
V4: watchAPI: https://react-hook-form-website-git-v4.bluebill1049.now.sh/api/#watch

KATT:
@bluebill1049 I also tried calling it with `watch(undefined, defaultValues)` but it didn't flatten it either.

I finally got to this workaround -- 

```ts
const { watch } = useFormContext();
const [values, setValues] = useState(watch());

useEffect(() => {
  setValues(watch());
}, [watch]);
```

But it ain't very intuitive.

KATT:
Actually this is what I landed on to get a consistent format of `Record<string, any>` 

```ts
const { watch } = useFormContext();
const [values, setValues] = useState<Record<string, any>>({});
useEffect(() => {
  const t = setTimeout(() => {
    setValues(watch());
  }, 1);

  return () => clearTimeout(t);
}, [watch]);
```

Got any nicer way? If the internal flatten method is exposed I could just call that rather than this mess

KATT:
Thanks for the insanely quick response. I'll wait for V4.

bluebill1049:
actually @KATT you want to give a try on the RC version, then we can go from there: 

`react-hook-form@4.0.0-rc3`

KATT:
Getting same results on V4 on first render when doing `{ nest: false}`

Another issue with complex default values is that I kinda have to use 

```
useForm<any>({
  defaultValues: initialValues,
  reValidateMode: 'onBlur',
});
```

With complex fields as it complains about missing field names when doing reffing of for instance `name="complex.field"`

KATT:
[...]

PR #707, opened 2019-12-25
Failing tests 

KATT:
Failing tests for #698

Unsure why `useForm.test.ts` is using so many mocks as it kinda puts a spanner in actually testing the real behaviour of the hook, so I created a new file called  "integration".

I can try to do a patch for this.

### DIFF
--- a/src/useForm-integration.test.ts
+++ b/src/useForm-integration.test.ts
@@ -1,0 +1,52 @@
+import { renderHook, act } from '@testing-library/react-hooks';
+import { useForm } from './useForm';
+
+describe('getValues', () => {
+  describe('nest: false', () => {
+    it('should returned a flattened version of defaultValues', () => {
+      const { result } = renderHook(() =>
+        useForm({
+          defaultValues: {
+            foo: 'bar',
+            complex: {
+              field: 'value',
+            },
+          },
+        }),
+      );
+      act(() => {
+        const values = result.current.getValues({ nest: false });
+
+        expect(values).toEqual({
+          foo: 'bar',
+          'complex.field': 'value',
+        });
+      });
+    });
+  });
+});
+
+describe('watch', () => {
+  describe('nest: false', () => {
+    it('should returned a flattened version of defaultValues', () => {
+      const { result } = renderHook(() =>
+        useForm({
+          defaultValues: {
+            foo: 'bar',
+            complex: {
+              field: 'value',
+            },
+          },
+        }),
+      );
+      act(() => {
+        const values = result.current.watch({ nest: false });
+
+        expect(values).toEqual({
+          foo: 'bar',
+          'complex.field': 'value',
+        });
+      });
+    });
+  });
+});


bluebill1049:
There is no such feature as `nest: false` in the project at the moment, and I don't think we need it for now.

> Unsure why useForm.test.ts is using so many mocks as it kinda puts a spanner in actually testing the real behaviour of the hook, so I created a new file called "integration".

Yes, we going to improve on that.

as for now: you can supply flat default value or write a utility function at your app level.

KATT:
@bluebill1049 if I supply a patch that make the `nest: false` behave as expected (i.e. making above pass) - would you be interested in it?

bluebill1049:
hey @KATT maybe start with a utility function to flat an object (object contain array and object) into a flat object (to start with).  my only concern is `watch()` by default is flat, the problem is only on the default value, so it's more like `defaultFlatValue`.

KATT:
Cool. 

After using it for a few days now it seems like complex objects doesn't feel like a first class citizen, for instance the above issue + TS complains when you do do `ref={register}` on a `name="complex.obj"` as it won't be in the FieldNames.

I just hacked a bit using a MobX-observable representation of values to get performance deep down in a long table.


```ts
function flattenObj(
  obj: Record<string, any> | any[],
  map: Record<string, any> = {},
  path: string | null = null,
) {
  const isArr = Array.isArray(obj);

  for (const key in obj) {
    const suffix = isArr ? `[${key}]` : path ? `.${key}` : key;
    const p = `${path ?? ''}${suffix}`;
    const value = obj[key];
    if (typeof value === 'object' && value !== null) {
      flattenObj(value, map, p);
    } else {
      map[p] = value;
    }
  }

  return map;
}

function useMyForm<T = any>(initialValues: T) {
  const form = useForm<any>({
    defaultValues: initialValues,
    reValidateMode: 'onBlur',
  });
  const { getValues, watch } = form;

  // Memoized MobX observable flat map of `values`
  const values = useMemo(
    () => observable(flattenObj(getValues({ nest: false }))),
    [getValues],
  );

  // Update MobX state of values
  useEffect(() => {
    const raw = watch({ nest: false });
    const newValues = flattenObj(raw);
    for (const [key, value] of Object.entries(newValues)) {
      values[key] = value;
    }
  }, [values, form, watch]);

  return {
    ...form,
    values,
  };
}

/**
 * Only exists for getting strict typing on `getValues()`
 */
function useBulkForm(initialValues: InitialValues) {
  return useMyForm(initialValues);
}

type UseAnyForm = ReturnType<typeof useMyForm>;
type UseBulkForm = ReturnType<typeof useBulkForm>;



const ResultInputs: React.FC<{
  disabled: boolean;
  namePrefix;
  onChange: () => void;
  label?: string;
  register: UseAnyForm['register'];
  values: UseBulkForm['values'];
}> = ({ disabled, namePrefix, onChange, label, register, values }) => {
  const enabledKey = `${namePrefix}.enabled`;

  return useObserver(() => {
    const enabled = values[enabledKey];
    console.log('render', enabledKey);
    // console.log('rerender', namePrefix, enabled);
    const checkbox = false ? (
      // This doesn't work for some reason
      <FormControlLabel
        control={<Checkbox />}
        label={`${namePrefix}.enabled`}
        disabled={disabled}
        inputRef={register}
        name="enabled"
      />
    ) : (
      // tmp workaround
      <label>
        <input
          type="checkbox"
          disabled={disabled}
          ref={register}
          name={`${namePrefix}.enabled`}
          onChange={onChange}
        />{' '}
        Played
      </label>
    );

    return (
      <table>
        <tr>
          {label && <th>{label}</th>}
          <td>{checkbox}</td>
          <td>
            <div style={{ display: enabled ? 'inline-block' : 'none' }}>
              <TextField
                name={`${namePrefix}.value`}
                type="text"
                inputRef={register}
                disabled={disabled}
                label="Value"
                onChange={onChange}
              />
            </div>
          </td>
        </tr>
      </table>
    );
  });
};
```

The hook causes some lag as `getValues()` takes > 100 ms on my form, but that's probably a side-effect of the design using refs over controlled inputs, haven't dug into how values are actually fetched in `watch()`.



Kinda makes me have a stab to write a form library using MobX observables (but I def won't).

bluebill1049:
> TS complains when you do do `ref={register}` on a `name="complex.obj"` as it won't be in the FieldNames.

can you show me an example of this? i dont think i will complain

PR #715, opened 2019-12-26
add ErrorMessage component

kotarella1110:
Make [RHFError component](https://github.com/react-hook-form/react-hook-form-error) part of the React Hook Form, and rename to ErrorMessage component.

See here: https://github.com/react-hook-form/react-hook-form-error/pull/4

### DIFF
--- a/src/errorMessage.test.tsx
+++ b/src/errorMessage.test.tsx
@@ -1,0 +1,17 @@
+import * as React from 'react';
+import { render } from '@testing-library/react';
+import { ErrorMessage } from './index';
+
+describe('React Hook Form Error Message', () => {
+  it('should render correctly', () => {
+    const { asFragment } = render(
+      <ErrorMessage
+        errors={{ test: { type: 'test', message: 'test' } }}
+        as={<span />}
+        name="test"
+      />,
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+});
--- a/src/errorMessage.test.tsx
+++ b/src/errorMessage.test.tsx
@@ -4,11 +4,98 @@ import { ErrorMessage } from './index';
 
 describe('React Hook Form Error Message', () => {
   it('should render correctly', () => {
+    const { asFragment } = render(<ErrorMessage name="test" />);
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should render correctly with flat errors', () => {
     const { asFragment } = render(
       <ErrorMessage
-        errors={{ test: { type: 'test', message: 'test' } }}
+        errors={{ flat: { type: 'flat', message: 'flat' } }}
+        name="flat"
+      />,
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should render correctly with flat errors and as with string', () => {
+    const { asFragment } = render(
+      <ErrorMessage
+        as="span"
+        errors={{
+          nested: {
+            object: { type: 'object', message: 'object' },
+          },
+        }}
+        name="nested.object"
+      />,
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should render correctly with nested errors object', () => {
+    const { asFragment } = render(
+      <ErrorMessage
+        errors={{
+          nested: {
+            object: { type: 'object', message: 'object' },
+          },
+        }}
+        name="nested.object"
+      />,
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should render correctly with nested errors object and as with component', () => {
+    const { asFragment } = render(
+      <ErrorMessage
         as={<span />}
-        name="test"
+        errors={{
+          nested: {
+            object: { type: 'object', message: 'object' },
+          },
+        }}
+        name="nested.object"
+      />,
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should render correctly with nested errors array', () => {
+    const { asFragment } = render(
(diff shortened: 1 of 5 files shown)


bluebill1049:
awesome stuff! Let's not worry about multiple errors for now, leave to the next minor version release.

bluebill1049 (approved):

bluebill1049:
awesome work!

src/errorMessage.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+      <ErrorMessage
+        errors={{ test: { type: 'test', message: 'test' } }}
+        as={<span />}
+        name="test"
+      />,
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+});
```
bluebill1049: shall we write a unit test for nested object and nested array?

bluebill1049 (dismissed):

src/errorMessage.test.tsx
kotarella1110: completed!

bluebill1049 (approved):

kotarella1110:
@bluebill1049 I will support about multiple errors messages in the next minor version!
Let's consider it together 👍

mnesarco:
Hi Guys, thanks for this amazing lib.
I am trying to use the ErrorMessage component as shown in docs, but it does render nothing when I submit the form:

```
import React from 'react';
import { PageHeader } from '../PageHeader';
import { Card, Form, Button } from 'react-bootstrap';
import { useForm, Controller, ErrorMessage } from 'react-hook-form'

export function CreatePollForm() {

    const { register, handleSubmit, errors, control } = useForm()
    const onSubmit = (data: any) => {
        console.log(JSON.stringify(data))
    }

    return (
        <>
            <PageHeader title="Create Poll" />
            <Card>
                <Card.Body>
                    <Form noValidate onSubmit={handleSubmit(onSubmit)}>
                        <Form.Group controlId="name">
                            <Form.Label>Name</Form.Label>
                            <Controller
                                as={
                                    <Form.Control
                                        isInvalid={!!errors.name}
                                    />
                                }
                                control={control}
                                name="name"
                                rules={{ required: true, minLength: 10 }}
                            />
                            <ErrorMessage errors={errors} name="name" />
                        </Form.Group>
                        <Button type="submit">Save</Button>
                    </Form>
                </Card.Body>
            </Card>
        </>
    )

}

```

bluebill1049:
@mnesarco I think you forgot to set the error message

`rules={{ required: 'This is required' }}`

mnesarco:
Thank you @bluebill1049 , I was expecting some kind of default validation messages like in JSF, but later I realized that any validation message must be provided every time. Maybe a future enhancement could be adding default validations messages ;)

src/errorMessage.tsx
```diff
@@ -0,0 +1,46 @@
...
+  errors: errorsFromProps,
+  name,
+}: Props<Errors, Name>) => {
+  const methods = useFormContext();
+  const errors = errorsFromProps || (methods?.errors as Errors);
+  const error = get(errors, name) as FieldError | undefined;
+  const message = error?.message;
+  if (!message) {
+    return null;
+  }
```
indigoviolet: I think this is not an exhaustive way of getting errors, for ex:

if you use a yup schema like
```
emails: yup.array().of(yup.string().email("Invalid email"))
```
The errors object is like:
```
emails: [{message: "Invalid email"....}]
```
which will make it look like there is no error if you try to access `get(errors, "email")?.message`

bluebill1049: you mean this right: `get(errors, "email[0]")?.message`

indigoviolet: Perhaps, but wouldn't that break with
```
emails: yup.array().of(yup.string().email("Invalid email")).required()
```
?

bluebill1049: is your input name: `<input name="emails[0]" />`? if that's the name it should work

indigoviolet: In short, I was expecting `ErrorMessage` to display a message for any error for that field -- both for "required" emails, and for any invalid email (perhaps only the first error is sufficient). It would be difficult to enumerate all the paths at which error messages could exist.

bluebill1049: @kotarella1110 is already planing for the next minor release, which will give you a render prop and with message and messages as arguments. maybe that's the solution?

indigoviolet: > is your input name: `<input name="emails[0]" />`? if that's the name it should work

I should have explained this better, I have a "tag" input where multiple emails can be entered, and at least one is required. 

> @kotarella1110 is already planing for the next minor release, which will give you a render prop and with `message` and `messages` as arguments. maybe that's the solution?

That is definitely a great idea. It might also be nice to have the default behavior handle `message || messages[0]`. Thanks for this library!

bluebill1049: > I should have explained this better, I have a "tag" input where multiple emails can be entered, and at least one is required.

pretty sure that's documented under `register` section on the website. 
https://react-hook-form.com/api#register

but do let us know how we can improve on this, PR welcome as well.

indigoviolet: It's possible I've misunderstood, but I don't think I have a field array as described in the documentation there. I have a single (custom) input field, whose value is an array of emails similar to 

image

I want to validate that this data is required, and that each of the emails is valid, which is what my yup schema above does.

bluebill1049: oh i see. I thought was field array. you want to provide a codesandbox for your use-case, it's easier to take a look that way.

PR #718, opened 2019-12-27
:beetle: fix Controller component props and event names

dru89:
This adds two small fixes to the `<Controller />` component:

- `eventWrapper` expected an event name like `blur` but was being passed
  a method name like `onBlur`. Now the two will match.

- The component's `onBlurName` and `onChangeName` props weren't being
  used if a custom handler wasn't provided.


bluebill1049:
aha! thank you 🙏

PR #727, opened 2019-12-27
Support custom value name

JeromeDeLeon:
Basically, it allows to override the `value | checked` prop to support other UI lib like `office-ui-fabric-react` that doesn't make use of `value`.

I only tried `ChoiceGroup` therefore, I assumed that some form components works the same.

### DIFF
--- a/src/controller.test.tsx
+++ b/src/controller.test.tsx
@@ -2,52 +2,41 @@ import * as React from 'react';
 import { render, fireEvent } from '@testing-library/react';
 import { Controller } from './index';
 
+function reconfigureControl(changedControl = {}) {
+  const defaultControl = {
+    defaultValues: {},
+    fields: {},
+    setValue: jest.fn(),
+    register: jest.fn(),
+    unregister: jest.fn(),
+    errors: {},
+    mode: { isOnSubmit: false, isOnBlur: false },
+    reValidateMode: {
+      isReValidateOnBlur: false,
+      isReValidateOnSubmit: false,
+    },
+    formState: { isSubmitted: false },
+  };
+
+  return Object.assign({}, defaultControl, changedControl);
+}
+
 describe('React Hook Form Input', () => {
   it('should render correctly with as with string', () => {
+    const control = reconfigureControl();
+
     const { asFragment } = render(
-      <Controller
-        name="test"
-        as="input"
-        control={{
-          defaultValues: {},
-          fields: {},
-          setValue: jest.fn(),
-          register: jest.fn(),
-          unregister: jest.fn(),
-          errors: {},
-          mode: { isOnSubmit: false, isOnBlur: false },
-          reValidateMode: {
-            isReValidateOnBlur: false,
-            isReValidateOnSubmit: false,
-          },
-          formState: { isSubmitted: false },
-        }}
-      />,
+      <Controller name="test" as="input" control={control} />,
     );
 
     expect(asFragment()).toMatchSnapshot();
   });
 
   it('should render correctly with as with component', () => {
+    const control = reconfigureControl();
+
     const { asFragment } = render(
-      <Controller
-        name="test"
-        as={<input />}
-        control={{
-          defaultValues: {},
-          fields: {},
-          setValue: jest.fn(),
-          register: jest.fn(),
-          unregister: jest.fn(),
-          errors: {},
-          mode: { isOnSubmit: false, isOnBlur: false },
-          reValidateMode: {
-            isReValidateOnBlur: false,
-            isReValidateOnSubmit: false,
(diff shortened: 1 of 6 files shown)


src/controller.tsx
```diff
@@ -113,6 +115,7 @@ const Controller = ({
     registerField();
   }
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
   React.useEffect(() => () => unregister(name), []);
```
JeromeDeLeon: Should we leave it as it is? I'm disabling the linting because my editor cries for it.

bluebill1049: i think just add unregister as dep, should resolve the problem.

bluebill1049:
@JeromeDeLeon if you set defaultValue as boolean, it should detect it as checkbox right?

JeromeDeLeon:
Yes, I supposed so. The way you checked the `value` at the beginning should resolve as checkbox as long as it is a boolean.

bluebill1049:
> Yes, I supposed so. The way you check the `value` at the beginning should resolve as checkbox as long as it is a boolean.

Yea that's why i am not sure if we still need `valueName`, but let's keep this PR open, we can merge it when people run into such issue.

bluebill1049:
I am even thinking about remove `onChangeName` and `onBlurName`. `onChange` is pretty standard for all components. (Keep API simple)

JeromeDeLeon:
The problem is that, it's not connected to the UI, in this [Codesandbox](https://codesandbox.io/s/hopeful-chebyshev-5rqgc?fontsize=14&hidenavigation=1&theme=dark) the prop to use in `ChoiceGroup` must be `selectedKey` and not `value`. Therefore, even if we set the `value` to "D", it is not going to reflect in the UI.

bluebill1049:
god dame! why can't they just follow the bloody standard...

JeromeDeLeon:
I think it's fair to support other UI lib as well as long as it's not going to have drastic changes to this lib. This component is generic after all.

JeromeDeLeon:
That's what I thought too when I was trying that UI lib.

bluebill1049:
> I think it's fair to support other UI lib as well as long as it's not going to have drastic changes to this lib. This component is generic after all.

yea it's fair, i just haven't encounter such component doesn't use value prop.

src/controller.test.tsx
```diff
@@ -2,78 +2,58 @@ import * as React from 'react';
 import { render, fireEvent } from '@testing-library/react';
 import { Controller } from './index';
 
+function reconfigureControl(changedControl = {}) {
```
bluebill1049: nice

JeromeDeLeon:
Upon looking to their components, ComboBox (Select), and Dropdown are making use of `selectedKey` instead of just `value`. I must say that thought, it's correct name, it doesn't follow the norm or how HTML uses it. :laughing:

src/controller.test.tsx
```diff
@@ -192,27 +145,24 @@ describe('React Hook Form Input', () => {
...
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('should support custom value name', () => {
+    const control = reconfigureControl();
+
+    const { asFragment } = render(
+      <Controller name="test" as="input" valueName="value" control={control} />,
```
bluebill1049: probably better to change "value" to "valueSelected" so snapshot can reflect that.

JeromeDeLeon: Why do you think so?

JeromeDeLeon: If you think that's better, I'm good :smile:

bluebill1049: otherwise the snapshot looks as usual: 

`<Select value="" />`

I think you want your snapshot to look like the following right?

`<Select valueSelected ="" />`

bluebill1049:
looks like controller.ts   automation failed? you want to check it out? 

`yarn start:app && yarn cypress:open`

I can sort out it but it's better to get you up to speed as a core team member.

JeromeDeLeon:
Yes, I'll sort this one out.

src/controller.test.tsx
JeromeDeLeon: Oops.. sorry.

src/controller.tsx
JeromeDeLeon: It is causing the error, and I think we should leave it for now.

JeromeDeLeon: when `unregister` changes reference, the field is unregisted in the form.

bluebill1049: oh let's leave it out then.

bluebill1049 (approved):

PR #731, opened 2019-12-28
close #730 to support min and max for custom register

### LINKED ISSUE #730: Controller can't use number validators
**Describe the bug**
Number validators like `min` and `max` don't work when you're using a `<Controller />` to wrap a component.

**Codesandbox link**
https://codesandbox.io/s/amazing-lichterman-ocm5j

**Expected behavior**
`min` and `max` should still function properly whenever you're using `<Controller />` if the value is a number.

**Additional context**
It looks like the `min` and `max` validators only work if the `<input />`'s type is `"number"`. https://github.com/react-hook-form/react-hook-form/blob/60b5af5197acfa4dcffdd9e034c4fa1e34b6c089/src/logic/validateField.ts#L86-L102

And when you use `<Controller />`, you're sort of foregoing the whole `ref` altogether: https://github.com/react-hook-form/react-hook-form/blob/60b5af5197acfa4dcffdd9e034c4fa1e34b6c089/src/controller.tsx#L93-L110

I thought about submitting a pull request, but I wasn't sure what the best option was for fixing this. Some options I was thinking about:

1. The `type="number"` check in `validateField` could be removed and instead just check to see if the value is "number-like". Or if `typeof value === "number"` as a possible check.
2. Alternatively, instead of using `min` and `max` for dates, too, there could be a separate rule like `minDate` and `maxDate`, but those would be breaking changes.
3. There could be a `type` property on `<Controller />`. This feels like a bit of a hack, though, and I can't think of any other reasons to declare a "type" on controller.

bluebill1049:


src/controller.tsx
```diff
@@ -128,7 +128,13 @@ const Controller = ({
         ? { [onBlurName]: eventWrapper(onBlur, EVENTS.BLUR) }
         : { [onBlurName]: handleBlur }
       : {}),
-    ...{ [valueName || (isCheckboxInput ? 'checked' : 'value')]: value },
```
JeromeDeLeon: Were you not fun of this style of code? Just asking :wink:

bluebill1049: no i am revert it back 👍 i was testing the package size, because one is string and the others is object. looks the same

JeromeDeLeon: Oh, is that so. I'm not aware to this kind of thing. Good point tho. :+1:

bluebill1049: yea every bytes counts :) the less we can ship the better <3

bluebill1049: let me know your thoughts on the code. i am going to add some unit tests, if you are cool with it. we can merge it and release with your update :)

JeromeDeLeon (dismissed):
lgtm! :+1:

PR #756, opened 2020-01-02
add runtime field name check and add unit test

bluebill1049:
This PR is related to an issue around TS runtime name check:

because you can `setValue('test.data[0]')` there is no way we can do check for type at build time. so this PR will address this issue during runtime and only fire during development mode.

### DIFF
--- a/src/logic/reportFieldNotFound.test.ts
+++ b/src/logic/reportFieldNotFound.test.ts
@@ -5,6 +5,15 @@ jest.mock('../constants', () => ({
 }));
 
 describe('reportFieldNotFound', () => {
+  const env = process.env.NODE_ENV;
+  beforeEach(() => {
+    process.env.NODE_ENV = 'test';
+  });
+
+  afterEach(() => {
+    process.env.NODE_ENV = env;
+  });
+
   it('should report when field is not found', () => {
     expect(() => reportFieldNotFound('test')).toThrowError(
       'test field not found.',
--- a/src/logic/reportFieldNotFound.test.ts
+++ b/src/logic/reportFieldNotFound.test.ts
@@ -5,24 +5,15 @@ jest.mock('../constants', () => ({
 }));
 
 describe('reportFieldNotFound', () => {
-  const original = console.error;
-  const errorLog = jest.fn();
-
-  beforeEach(() => {
-    console.error = errorLog;
-  });
-
-  afterEach(() => {
-    console.error = original;
-  });
-
   it('should report when field is not found', () => {
-    reportFieldNotFound('test');
-    expect(errorLog).toBeCalledWith('test field not found.');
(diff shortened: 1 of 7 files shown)


src/logic/reportFieldNotFound.test.ts
```diff
@@ -0,0 +1,28 @@
...
+
+jest.mock('../constants', () => ({
+  IS_DEV_ENV: true,
+}));
+
+describe('reportFieldNotFound', () => {
+  const original = console.error;
+  const errorLog = jest.fn();
+
+  beforeEach(() => {
```
JeromeDeLeon: What's the purpose of this one?

src/logic/reportFieldNotFound.ts
```diff
@@ -0,0 +1,8 @@
+import isUndefined from '../utils/isUndefined';
+import { IS_DEV_ENV } from '../constants';
+
+export default (name: any, fields?: any): void => {
+  if (IS_DEV_ENV && ((fields && isUndefined(fields[name])) || !fields)) {
+    console.error(`${name} field not found.`);
```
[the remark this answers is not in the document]
bluebill1049: agree

src/logic/reportFieldNotFound.test.ts
bluebill1049: so we dont populate the real console.error

src/logic/reportFieldNotFound.ts
```diff
@@ -0,0 +1,10 @@
+import isUndefined from '../utils/isUndefined';
+
+export default (name: any, fields?: any): void => {
+  if (
+    process.env.NODE_ENV !== 'production' &&
```
JeromeDeLeon: Why revert this one? I prefer the constant value tho, but if it's blowing up the size then forget about it :smile:

src/logic/reportFieldNotFound.test.ts
```diff
@@ -0,0 +1,28 @@
+import reportFieldNotFound from './reportFieldNotFound';
+
+jest.mock('../constants', () => ({
```
JeromeDeLeon: We can now remove this one if you opt out the constant value.

src/logic/reportFieldNotFound.ts
bluebill1049: compiler like rollup will not pick it up, we need to have it inline

JeromeDeLeon: I see. Too bad.

JeromeDeLeon (approved):
lgtm! :+1:

bluebill1049:
thanks review @JeromeDeLeon

JeromeDeLeon:
No worries :smile:

Issue #725, opened 2019-12-27
FieldError's message property


kusmierz:
**Describe the bug**
I think there is a small issue in TS typings introduced somewhere in 4.x release. I can't read `.message` property on `errors`.

It applies only to `FieldErrors<any>`... `FieldErrors<FooBar>` with `type FooBar = { foobar: string; }` works as expected...

**To Reproduce**

try to use `message` property on FieldError:

```ts
const {
  errors,
} = useFormContext();

const name = 'username';
console.log(errors[name]?.message);
```

**TS Play link**
https://www.typescriptlang.org/play/#code/KYDwDg9gTgLgBASwHY2FAZgQwMbDgMQWABsATAUSimjgG8AoOOGATzGAC44BnGKZAOYBuRkzgB6cTwQBbMMQToWcUsHZJVSbEW6imkqMHQB+LgCUjIsRPGt23U3ACyAV2IwE84IRIUq0bisxGWBubkwBYEdefiRhPURuJ0wkF0xiRwAjCAhiYBSRAF96UEhYFWBsYkxDZjY8HzJKaihuAB58aBkANXSXUIA+OABeOgSAbQBpYGVkOABrGYh0Ai7e4n7uAF1HTqgevtCpma24UFQNbjgUlnHT41X99c3jli3x1JlMtFPz4Eu4BBMgArSrwB6NPwtdp7A4bI7TN4fFxfH4DO5wLiQ5o0LFrQ7cV6-EAXUhXIGg7DgghEJr+VodfHwwmIrZDLG0qHQIoiejYCBIXhnencDm+HEMm5DUa0Qq8-mC+BITAhEZwADk6BymRq6pEaGh4wADDsAHQhMIRYAlEXjTXa3Vmi3hSI26GmrVAmrGc2hF3Wg0BD4q4BOv1WoA

**Expected behavior**

It would be nice to have an ability to use `.message` as in 3.x

**Screenshots**
```
Property 'message' does not exist on type 'FieldError | FieldErrors<any> | FieldErrors<any>[]'.
Property 'message' does not exist on type 'FieldErrors<any>[]'.  TS2339
```

bluebill1049:
can you provide a codesandbox for this issue?

bluebill1049:
This is probably due to the nested errors object which introduced at v4

bluebill1049:
tested: https://codesandbox.io/s/nostalgic-driscoll-p5ize-delete-p5ize type is working for me

[...]

PR #762, opened 2020-01-02
Typescript DeepPartial for default values

### LINKED ISSUE #759: [TS] Default values for nested fields are required
**Describe the bug**
Having nested fields in our FormSchema require to fill all nested fields in defaultValues.

**To Reproduce**
1. Introduce nested form schema, ie.:

```ts
type FormData = {
  user: {
    firstName: string;
    age: number;
  }
  website?: string;
}
```

2. define default values for nested field:

```ts
useForm<FormData>({
    defaultValues: {
      user: {
        age: 18,
      },
    },
  });
```

ERROR!
```
Type '{ age: number; }' is not assignable to type 'Shape<object, { firstName: string; age: number; }>'.
  Property 'firstName' is missing in type '{ age: number; }' but required in type '{ firstName: string; age: number; }'.ts(2322)
index.tsx(10, 5): 'firstName' is declared here.
index.tsx(9, 3): The expected type comes from property 'user' which is declared here on type 'Partial<Id<Partial<Pick<Shape<object, { user: Shape<object, { firstName: ...; age: ...; }>; website: string; }>, "user" | "website">> & Pick<Shape<object, { user: Shape<object, { firstName: ...; age: ...; }>; website: string; }>, never>>>'
```

**Codesandbox link**
https://codesandbox.io/s/react-hook-form-typescript-14qt0

**Expected behavior**
No TS error at all.

**Additional context**
The problem is related with `UseFormOptions`' `defaultValues: Partial<FormValues>;`, it should be a `DeepPartial` instead:

```ts
// ...
export type DeepPartial<T> = {
  [P in keyof T]?:
  T[P] extends Array<infer U> ? Array<DeepPartial<U>> :
    T[P] extends ReadonlyArray<infer U> ? ReadonlyArray<DeepPartial<U>> :
      DeepPartial<T[P]>
};
// ...
export declare type UseFormOptions<FormValues extends FieldValues = FieldValues> = Partial<{
// ...
    defaultValues: DeepPartial<FormValues>;
// ...
```

kusmierz:
This should fix #759.

Also it fails...

```
FAIL  src/useForm.test.tsx
  ● useForm › when defaultValues is supplied › should populate the input with them

    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

      Object {
    -   "deep": Object {
    -     "values": "5",
    -   },
    +   "deep.nested": undefined,
    +   "deep.values": "5",
        "test": "data",
      }

      1148 |       await act(async () => {
      1149 |         await result.current.handleSubmit((data: any) => {
    > 1150 |           expect(data).toEqual({
           |                        ^
      1151 |             test: 'data',
      1152 |             deep: {
      1153 |               values: '5',

      at src/useForm.test.tsx:1150:24
      at src/useForm.ts:914:17
      at src/useForm.test.tsx:1149:9
```

But to be honest, I'm not sure if it's a bug or I'm doing something wrong...


bluebill1049:
Thanks a look @kusmierz looks like the build failed. I can take a look why during lunch break as well :)

src/useForm.test.tsx
```diff
@@ -1135,12 +1138,16 @@ describe('useForm', () => {
...
         result.current.register({ type: 'text', name: 'test' });
+        result.current.register({ type: 'text', name: 'deep.nested' });
+        result.current.register({ type: 'text', name: 'deep.values' });
       });
 
       await act(async () => {
         await result.current.handleSubmit((data: any) => {
           expect(data).toEqual({
             test: 'data',
+            'deep.nested': undefined,
```
kusmierz: is it proper behaviour? I thought it should be deep nested as well here

bluebill1049: it's not the correct behavior because I have stubbed the function.

bluebill1049: thanks a lot for helping on this one @kusmierz

kusmierz: Ah, I can see now, thanks :)

bluebill1049:
hey @kusmierz looks like this DeepPartial do cause some issues: https://spectrum.chat/react-hook-form/general/new-deeppartial-for-defaultvalues-is-causing-issues-with-reset~a9b886b6-9692-4682-9bd3-9161d3f85591 We may have to roll that DeepPartial back

bluebill1049:
After googling around, i saw the `DeepPartial` type in one of the Gist and people did mention it not working property for some types. So maybe we need to wait for offical `DeepPartial` type.

kusmierz:
Maybe we could use [DeepPartial](https://github.com/piotrwitek/utility-types/blob/2ae7412a9edf12f34fedbf594facf43cf04f7e32/src/mapped-types.ts#L504) from [piotrwitek/utility-types](https://github.com/piotrwitek/utility-types) instead?

`reset()` with DeepPartial arg should solve half of the @bopfer issue.

The second half of the issue seems to not be a `react-hook-form` fault. `useLocalFormContext` uses generic `any` `FieldValues`, which is why it loses types... This is why TS reports "bad" types. See fixed version https://codesandbox.io/s/affectionate-sun-nmb4p, which defines `useLocalFormContext` similar as RHF does [useFormContext](https://github.com/react-hook-form/react-hook-form/blob/master/src/useFormContext.tsx#L9).

BTW, this is more or less connected with #725.

bluebill1049:
hey @kusmierz i think this one looks better:

```
export type DeepPartial<T> = T extends Function
  ? T
  : T extends Array<infer U>
  ? _DeepPartialArray<U>
  : T extends object
  ? _DeepPartialObject<T>
  : T | undefined;
```

maybe we should give it a try, see if that improve the type?

bopfer:
The `reset` is good now.  So, that part is taken care of.

The issue with `AnotherComponent` in my sandbox, is that in my real app, components like that are part of a shared library of form components that are used by many different forms throughout the app.  The data for each form can vary widely.

For example, I have an `ImageUpload` component that displays the currently uploaded image or allows a new one to be uploaded.  If the name of that field on the form is `imageURL`, then I want to grab the default value via `defaultValues[name]`, where `name` is a prop with the value `imageURL` in one form.  But, it might be something like `primaryImageURL` in another form.  Either way, I would like `defaultValues[name]` to give me a string. 

It's possible I have a flaw in my component structure, but I am not seeing it at the moment if there is one.  I see the value for `DeepPartial` in simple form components.  But, It doesn't seem to hold up for abstracted components that are meant to be re-used by many different forms.

Prior to the `DeepPartial` my shared components would get `any` for `defaultValues[name]`, which is not ideal either, but it works.  My TypeScript knowledge is not deep enough to know if there is a way to make the `DeepPartial` work better.

If the `DeepPartial` sticks around, then I will need to do something like this in my components:

```javascript
const { defaultValues } = useLocalFormContext<Record<string, string>>();
```

Which I am fine with, since I know what my app is doing and it is slightly better than getting `any`.  But, I can see the `DeepPartial` causing similar issues for many others.

bluebill1049:
Thanks @bopfer i will give a try tonight with what @kusmierz suggest and produce a beta version. Hopefully that will resolve most of your issue (from your codesandbox).However if this doesn't work out well, maybe we should consider go back to `Partial`? at least `Partial` is officially supported by TS and probably lead less issue. Let me know your thoughts as well @kusmierz. I guess the goal is make dev's life easier.

kusmierz:
In my opinion it is not `DeepPartial` fault itself, but stricter typing (@bopfer example loses typings in `useLocalFormContext`, so it basically has `any` type, see the example I've fixed; what's more react-hook-form's `userFormContext` works properly with `DeepPartial`, at least I haven't found any issue). So either we stick with `Partial` with `any` (as @bopfer mentioned above), or more strict typing with more "problems" (as seen in #725, which basically is `DeepPartial` implementation as well :P). IMHO it would be nice to have it very strict, because it would be easier to find potential errors (I know having `any` is easier to code, but also it would be harder to maintain). And in case somebody would like to ignore types, he could cast it to `any`... but somebody who would like to type it strict... well... he can't do anything...

Or, at least have it consistent: either strict `FieldErrors` (#725) and defaultValues, or less strict (`any`?!) *errors* and *defaultValues*.  What do you think, @bluebill1049?

bluebill1049:
> In my opinion it is not `DeepPartial` fault itself, but stricter typing (@bopfer example loses typings in `useLocalFormContext`, so it basically has `any` type, see the example I've fixed; what's more react-hook-form's `userFormContext` works properly with `DeepPartial`, at least I haven't found any issue). So either we stick with `Partial` with `any` (as @bopfer mentioned above), or more strict typing with more "problems" (as seen in #725, which basically is `DeepPartial` implementation as well :P). IMHO it would be nice to have it very strict, because it would be easier to find potential errors (I know having `any` is easier to code, but also it would be harder to maintain). And in case somebody would like to ignore types, he could cast it to `any`... but somebody who would like to type it strict... well... he can't do anything...
> 
> Or, at least have it consistent: either strict `FieldErrors` (#725) and defaultValues, or less strict (`any`?!) _errors_ and _defaultValues_. What do you think, @bluebill1049?

You have valid point as well... it's a hard decision to make, or we can leave as DeepPartial and see if any users encounter issues like what @bopfer did.

bopfer:
I wanted to drop a new sandbox here just to show a more concrete example of the potential "issues" with `DeepPartial`:

https://codesandbox.io/s/angry-haibt-mdfym

There are 2 distinct forms, each with an `imageUrl`.  They use a shared `ImageUpload` component to show the current image and allow upload of a new image.  See `forms/ImageUpload.tsx` to see the issue the `DeepPartial` causes.  I have **many** of these shared components.  Like I said above, I am fine with using `Record<string, string>` as the generic to my context since it will give me what I need. It is probably equivalent to `Partial` giving `any`, as far as type safety goes.

I just think that abstracted form components are more common than not, and @bluebill1049 will have to field all the issues that come in related to the `DeepPartial` strictness :smile: Using `any` is pretty common in 3rd party packages for a reason.

I think it's fine to leave it in for now and see how it all shakes out.

bluebill1049:
> I wanted to drop a new sandbox here just to show a more concrete example of the potential "issues" with `DeepPartial`:
> 
> https://codesandbox.io/s/angry-haibt-mdfym
> 
> There are 2 distinct forms, each with an `imageUrl`. They use a shared `ImageUpload` component to show the current image and allow upload of a new image. See `forms/ImageUpload.tsx` to see the issue the `DeepPartial` causes. I have **many** of these shared components. Like I said above, I am fine with using `Record<string, string>` as the generic to my context since it will give me what I need. It is probably equivalent to `Partial` giving `any`, as far as type safety goes.
> 
> I just think that abstracted form components are more common than not, and @bluebill1049 will have to field all the issues that come in related to the `DeepPartial` strictness 😄 Using `any` is pretty common in 3rd party packages for a reason.
> 
> I think it's fine to leave it in for now and see how it all shakes out.

Thanks @bopfer for the example. Let's leave it for now and see how it goes 👍

bluebill1049:
Is this related to DeepPartial? https://spectrum.chat/react-hook-form/help/type-for-the-error-object~ceafb2a9-da6d-4651-9aac-c1d615519a65 i haven't look into it yet.

bopfer:
I don't think it's related directly to `DeepPartial`, but it is a similar issue.  I have an error component that is almost identical to what's in that Spectrum chat and I had to pass this generic to make it work:

```javascript
const { errors } = useFormContext<Record<string, string>>();
```

I don't think that is the "right" way to handle it, but it works :smile:

bluebill1049:
Oh thanks @bopfer

bopfer:
I guess the proper solution is to pass the generic for the specific form.  But, they are shared components in this case.  So, that doesn't really work.

Just for more history... I had to make that change with the upgrade from version 4.0.0 to 4.1.0.

kusmierz:
@bopfer, is it same issue as in #725?

bopfer:
@kusmierz, yes, it's the same issue.

PR #767, opened 2020-01-03
Fix name of prettier config file in vscode settings

kyle221b:
The prettier settings are correct in the package.json file, but in the vscode settings, the file referenced is wrong


bluebill1049:
oops thanks a lot for that!

Issue #777, opened 2020-01-05
Typescript x Controlled components


ScreamZ:
Hello,

I'm using Elastic UI, which is a great framework for back-offices. I starting being used to with react-forms-hook, but I still have some questions.

I'm just facing a controlled component which is `EuiSwitch`, a simple switch.

I succeed in doing such a thing now:

```tsx
  <Controller as={<EuiSwitch label="Actif ?" />} control={control} name="enabled" />
```
The thing is that now Typescript is yielding that there are some missing props on `EuiSwitch`.
Before I was doing a custom component with:

```tsx
import { EuiSwitch, EuiSwitchEvent } from "@elastic/eui";
import React, { useEffect, useState } from "react";
import { FormContextValues } from "react-hook-form";

interface Props {
  name: string;
  label: string;
  register: FormContextValues["register"];
  unregister: FormContextValues["unregister"];
  setValueFn: FormContextValues["setValue"];
  defaultChecked?: boolean;
}

export const Switch: React.FC<Props> = props => {
  const [checked, setInternalValue] = useState(props.defaultChecked ?? true);

  const onChange = (data: EuiSwitchEvent) => {
    props.setValueFn(props.name!, data.target.checked);
    setInternalValue(data.target.checked);
  };

  useEffect(() => {
    if (props.name) {
      props.register({ name: props.name });
      props.setValueFn(props.name!, props.defaultChecked);
    }

    return () => props.unregister(props.name);
  }, [props.name, props.defaultChecked]);

  return (
    <EuiSwitch
      name={props.name}
      defaultChecked={props.defaultChecked}
      label={props.label}
      checked={checked}
      onChange={onChange}
    />
  );
};
```

I the second case, nothing is yielding, because I'm filling props.
I could have also used `FormContext`

## Question
Do you think we could maybe find another pattern that would make this work without warning (currently I'm just doing a `@ts-ignore`)

I guess we could pass a `React.FunctionComponent<EuiSwitchProps>` instead of a `JSX.Element` and imagine some kind of genericity using things described here: https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/master/ADVANCED.md

create another prop which allows passing some of the props, as optional

```tsx
<Controller<Partial<EuiSwitchProps>>
  as={EuiSwitch}
  withProps={label: "Actif?"}
  name="test"
/>
```

Kind of thing? What do you thing about that?

bluebill1049:
[...]

PR #779, opened 2020-01-05
New controller with types supports implementation suggestion

ScreamZ:
# Suggestion: Another controller implementation

Thanks for your great work again, I really enjoy my developping experience with that lib. I had a bit of time so I spent my day about thinking how to make that lib even greater as it is :) 

As mentioned in #777 here is my first draft of an implementation that should work for controlled components and typescript requirements in kind of props.

**This is a work in progress suggestion but I would like to get your feedback.**

Currently, despite the fact they works (I had to check deep in the code to understand their behavior) the controllers are subject to error because the API is not really easy to understand when you're new. You have props like `onChangeName` which are not currently typed, this can be confusing.

Also, there are some kinds of things that seems strange to me, I made an issue so we can discuss it, it's mostly related to documentation: https://github.com/react-hook-form/react-hook-form/issues/780

In my opinion, it's not really clear on what args are passed down to InnerComponent, or even the type of the even component (JSX element, ReactElement, string ?)

What's following is a proposition that might be totally irrelevant, but imagine we would implement that, would it be better for some cases?

# PROS
- Clarity, this doesn't magically inject props using `valueName` or `onChangeName` things, which make complicated to take in hand and are currently not typed enough. If we solve typings this argument is irrelevant and it can be just easy as moving Controller component to the useForm args as I did for this new controller.
- Effortless, at the moment only two props are given for the child function, everything else can be accessed outside of this scope, like errors and etc…
- There are no 3 ways to do, more choice, more bugs? :p (can also be solved in current implementation)

# LACK
- For now, we are forced to provide `<keyof FormValues, boolean>` or something else but I guess we could improve and generate values of typings directly from form typings as it's exposed from form hook.
- Currently, it only supports `useFormContext` but can be easily decoupled?

## Usage

```tsx
type FormValues = { a: boolean; b: string }

const Dummy: React.FC = () => {
  const { Controller } = useForm<FormValues>();

  return (
    <Controller<keyof FormValues, boolean> name="a" defaultValue={false}>
      {arg => (
        <EuiSwitch
          name="a" // It's just for the DOM and Accessibility, we could completly generate it if none is provided.
          label="Some label"
          checked={arg.value}
          onChange={(val: any) => {arg.onValueChange(Boolean(val.target.value ?? val))}}
        />
      )}
      </Controller>
  );
};
```

### DIFF
--- a/src/contextTypes.ts
+++ b/src/contextTypes.ts
@@ -12,6 +12,7 @@ import {
   ManualFieldError,
   MultipleFieldErrors,
 } from './types';
+import { TypedController } from './typedController';
 
 export interface FormProps<FormValues extends FieldValues = FieldValues>
   extends FormContextValues<FormValues> {
@@ -21,6 +22,7 @@ export interface FormProps<FormValues extends FieldValues = FieldValues>
 export interface FormContextValues<
   FormValues extends FieldValues = FieldValues
 > {
+  Controller: typeof TypedController;
   register<Element extends ElementLike = ElementLike>(
     validationOptions: ValidationOptions,
   ): (ref: Element | null) => void;
--- a/src/typedController.tsx
+++ b/src/typedController.tsx
@@ -1,0 +1,55 @@
+import React, { useEffect, useState } from 'react';
+import { useFormContext } from './useFormContext';
+
+interface ControllerProps<FormValuesKeys, ValueType> {
+  /**
+   * The name that will be used for register.
+   *
+   * @type {FormValuesKeys}
+   * @memberof Props
+   */
+  name: FormValuesKeys;
+
+  /**
+   * The default value to provide to the internal state.
+   * Take priority over useForm default value argument.
+   */
+  defaultValue?: ValueType;
+
+  children(params: ControllerCommand<ValueType>): React.ReactElement;
+}
+
+interface ControllerCommand<ValueType> {
+  value?: ValueType;
+  onValueChange(data: ValueType): void;
+}
+
+/**
+ *
+ * @param props
+ */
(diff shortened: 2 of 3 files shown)


bluebill1049:
Thanks a lot for this detailed PR! 🙏 is any chance we can avoid introducing another controlled component? sorry i just woke up i probably need all more comments more carefully :)

ScreamZ:
Yes, I wrote I lot today about that repo :p And In the same time I learned a lot about the library, some things might seem weird :p

You have both my feeling before diving in the code and after.

I guess we can avoid creating a new controller component, we just need some doc and we will be fine :)

One cool thing to be to move the Controller component to useForm variable, so we can leverage typings from useform, let's think about it? :)

bluebill1049:
> Yes, I wrote I lot today about that repo :p And In the same time I learned a lot about the library, some things might seem weird :p
> 
> You have both my feeling before diving in the code and after.
> 
> I guess we can avoid creating a new controller component, we just need some doc and we will be fine :)
> 
> One cool thing to be to move the Controller component to useForm variable, so we can leverage typings from useform, let's think about it? :)

By the way, I really appreciated your effort in looking into this! 🙏

JeromeDeLeon:
This is a lot to take in all at once. If we could avoid changing too much about the API, that would be great. 👌

Is it really matters to have type of let's say `name`? I'm not really against with it or anything. 

I'll try to take a look at it more later. Thanks for everything and all :)

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

bluebill1049:
Thanks for contributing. going to close this one off for now. I think if we can improve the current Controller component, it will be more ideal :)

PR #787, opened 2020-01-06
Sync up 'reset' type

bopfer:
Update 'reset' type in the context to be DeepPartial and sync up with the useForm type.


bluebill1049:
thanks @bopfer

PR #790, opened 2020-01-07
Fix setValue for file type input

brennedith:
**Description**

setValue would try to assign the FileList object to _value_ instead of _files_.

**Motivation**

Inputs type file would break after assigning the FileList object as defaultValue or setValue since _value_ only accepts strings. FileList object can be assigned to inputs type file though _files_, which automatically sets the _value_ of the DOM element.

MDN: [input file](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#files)

See [this spectrum chat for more detail](https://spectrum.chat/react-hook-form/help/defaultvalue-of-an-input-type-file~23c450c7-a0c3-46c8-ba8d-889c1bc45430)

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -8,6 +8,8 @@ export const VALIDATION_MODE: ValidationMode = {
 
 export const RADIO_INPUT = 'radio';
 
+export const FILE_INPUT = 'file';
+
 export const VALUE = 'value';
 
 export const UNDEFINED = 'undefined';
--- a/src/logic/getFieldValue.ts
+++ b/src/logic/getFieldValue.ts
@@ -1,6 +1,7 @@
 import getRadioValue from './getRadioValue';
 import getMultipleSelectValue from './getMultipleSelectValue';
 import isRadioInput from '../utils/isRadioInput';
+import isFileInput from '../utils/isFileInput';
 import isCheckBox from '../utils/isCheckBoxInput';
 import isMultipleSelect from '../utils/isMultipleSelect';
 import getCheckboxValue from './getCheckboxValue';
@@ -13,7 +14,7 @@ export default function getFieldValue<FormValues extends FieldValues>(
   const { type, name, options, value, files } = ref;
   const field = fields[name];
 
-  if (type === 'file') {
+  if (isFileInput(type)) {
     return files;
   }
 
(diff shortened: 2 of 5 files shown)


bluebill1049:
This is awesome @brennedith PR, thank you!

brennedith:
Thanks @bluebill1049, please let me know if there is anything else I can do.

bluebill1049 (approved):

morales-franco:
Hi @bluebill1049 @brennedith Do you know if it is possible to assign a default value when you have a Field Array ? In my case, I receive a file list from an external popup and then I need to keep their values in files inputs, like that:  
image

I use typescript in my code, and receive the next error:
_Type 'File | undefined' is not assignable to type 'string | number | readonly string[] | undefined'.
  Type 'File' is not assignable to type 'string | number | readonly string[] | undefined'._

Because content field is type: File and DefaultValue expects a string :/

morales-franco:
> Hi @bluebill1049 @brennedith Do you know if it is possible to assign a default value when you have a Field Array ? In my case, I receive a file list from an external popup and then I need to keep their values in files inputs, like that:
> image
> 
> I use typescript in my code, and receive the next error:
> _Type 'File | undefined' is not assignable to type 'string | number | readonly string[] | undefined'. Type 'File' is not assignable to type 'string | number | readonly string[] | undefined'._
> 
> Because content field is type: File and DefaultValue expects a string :/
> 
> where attachments are the fields collection:
> image

I tested the following:
I didn't assign a default value to any file inputs.

Then I submited the form:

I can see that On submit the form object lost the file values (and change the type of content from File to FileList however the fileList is empty)
image

In other hand, attachment fields (from useFieldArray)
image

Ok, useFieldArray hook keep alive the fields but not in the form object

matheussbarros:
> **Description**
> 
> setValue would try to assign the FileList object to _value_ instead of _files_.
> 
> **Motivation**
> 
> Inputs type file would break after assigning the FileList object as defaultValue or setValue since _value_ only accepts strings. FileList object can be assigned to inputs type file though _files_, which automatically sets the _value_ of the DOM element.
> 
> MDN: [input file](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#files)
> 
> See [this spectrum chat for more detail](https://spectrum.chat/react-hook-form/help/defaultvalue-of-an-input-type-file~23c450c7-a0c3-46c8-ba8d-889c1bc45430)

Codesandbox with the solution is off, can you do it again?

PR #817, opened 2020-01-12
Feature/support custom schema validation

bluebill1049:
This feature request is waiting for more upvote. 

related PR: https://github.com/react-hook-form/react-hook-form-website/pull/139

if you want such feature, you can thumb up this comment. 👍 

------

example for `@hapi/joi`, this will apply for any other schema/custom validation.

```
import React from "react";
import useForm from "react-hook-form";
import joi from "@hapi/joi";

const schema = Joi.object({
  username: Joi.string().alphanum().min(3).max(30).required()
});

// Build your resolver to display errors and pass valid values
const resolver = (data: any) => {
  const { error, value: values } = validationSchema.validate(data, {
    abortEarly: false,
  });
  return {
    values: error ? {} : values,
    errors: error
      ? error.details.reduce((previous, currentError) => {
          return {
            ...previous,
            [currentError.path[0]]: currentError,
          };
        }, {})
      : {},
  };
};

function App() {
  const { register, handleSubmit, errors } = useForm({
    schemaResolver: resolver
   });
  
  return (
      <input type="text" name="username" ref={register} />
    </form>
  );
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@
 - [Tiny size](https://bundlephobia.com/result?p=react-hook-form@latest) without any dependency
 - Follows HTML standard for validation
 - Compatible with React Native
-- Support [Yup](https://github.com/jquense/yup) schema-based validation
+- Support [Yup](https://github.com/jquense/yup), [Joi](https://github.com/hapijs/joi) or custom schema-based validation
 - Support browser native validation
 - Build forms quickly with the [form builder](https://react-hook-form.com/form-builder)
 
--- a/app/src/app.tsx
+++ b/app/src/app.tsx
@@ -18,6 +18,7 @@ import ConditionalField from './conditionalField';
 import FormStateWithSchema from './formStateWithSchema';
 import SetValueWithSchema from './setValueWithSchema';
 import SetValueWithTrigger from './setValueWithTrigger';
+import CustomSchemaValidation from './customSchemaValidation';
 
 const App: React.FC = () => {
   return (
@@ -44,6 +45,11 @@ const App: React.FC = () => {
         exact
         component={BasicSchemaValidation}
       />
+        <Route
+          path="/customSchemaValidation/:mode"
+          exact
+          component={CustomSchemaValidation}
+        />
       <Route path="/setError" exact component={SetError} />
       <Route
         path="/setValueWithTrigger"
--- a/app/src/customSchemaValidation.tsx
+++ b/app/src/customSchemaValidation.tsx
@@ -1,0 +1,117 @@
+import React from 'react';
+import useForm from 'react-hook-form';
+import Joi from '@hapi/joi';
+
+let renderCounter = 0;
+
+const validationSchema = Joi.object({
+  firstName: Joi.string().required(),
+  lastName: Joi.string()
+    .max(5)
+    .required(),
+  min: Joi.number()
+    .min(10)
+    .required(),
+  max: Joi.number()
+    .max(20)
+    .required(),
+  minDate: Joi.date().min('2019-08-01'),
+  maxDate: Joi.date().max('2019-08-01'),
+  minLength: Joi.string().min(2),
+  minRequiredLength: Joi.string().required(),
+  selectNumber: Joi.string().required(),
+  pattern: Joi.string().required(),
+  radio: Joi.string().required(),
+  checkbox: Joi.required(),
+});
+
+const resolver = (data: any) => {
+  const { error, value: values } = validationSchema.validate(data, {
+    abortEarly: false,
+  });
+
+  return {
+    values: error ? {} : values,
+    errors: error
+      ? error.details.reduce((previous, currentError) => {
+          return {
+            ...previous,
+            [currentError.path[0]]: currentError,
+          };
+        }, {})
+      : {},
+  };
+};
+
+const BasicSchemaValidation: React.FC = (props: any) => {
+  const { register, handleSubmit, errors } = useForm<{
+    firstName: string;
+    lastName: string;
+    min: string;
+    max: string;
+    minDate: string;
+    maxDate: string;
+    minLength: string;
+    minRequiredLength: string;
+    selectNumber: string;
+    pattern: string;
+    radio: string;
+    checkbox: string;
+    multiple: string;
+    validate: string;
+  }>({
+    validationSchema: {
+      resolver,
+    },
+    mode: props.match.params.mode,
+  });
+  const onSubmit = () => {};
+
+  renderCounter++;
+
+  return (
+    <form onSubmit={handleSubmit(onSubmit)}>
+      <input name="firstName" ref={register} placeholder="firstName" />
+      {errors.firstName && <p>firstName error</p>}
+      <input name="lastName" ref={register} placeholder="lastName" />
+      {errors.lastName && <p>lastName error</p>}
+      <input type="number" name="min" ref={register} placeholder="min" />
+      {errors.min && <p>min error</p>}
+      <input type="number" name="max" ref={register} placeholder="max" />
+      {errors.max && <p>max error</p>}
+      <input type="date" name="minDate" ref={register} placeholder="minDate" />
+      {errors.minDate && <p>minDate error</p>}
+      <input type="date" name="maxDate" ref={register} placeholder="maxDate" />
+      {errors.maxDate && <p>maxDate error</p>}
+      <input name="minLength" ref={register} placeholder="minLength" />
+      {errors.minLength && <p>minLength error</p>}
+      <input
+        name="minRequiredLength"
+        ref={register}
+        placeholder="minRequiredLength"
+      />
+      {errors.minRequiredLength && <p>minRequiredLength error</p>}
+      <select name="selectNumber" ref={register}>
+        <option value="">Select</option>
+        <option value={1}>1</option>
+        <option value={2}>1</option>
+      </select>
+      {errors.selectNumber && <p>selectNumber error</p>}
+      <input name="pattern" ref={register} placeholder="pattern" />
+      {errors.pattern && <p>pattern error</p>}
+      Radio1
+      <input type="radio" name="radio" ref={register} value="1" />
+      Radio2
+      <input type="radio" name="radio" ref={register} value="2" />
+      Radio3
+      <input type="radio" name="radio" ref={register} value="3" />
+      {errors.radio && <p>radio error</p>}
+      <input type="checkbox" name="checkbox" ref={register} />
+      {errors.checkbox && <p>checkbox error</p>}
+      <button>Submit</button>
+      <div id="renderCount">{renderCounter}</div>
+    </form>
+  );
+};
+
+export default BasicSchemaValidation;
--- a/cypress/integration/customSchemaValidation.ts
+++ b/cypress/integration/customSchemaValidation.ts
@@ -1,0 +1,161 @@
+context('basicSchemaValidation form validation', () => {
+  it('should validate the form with onSubmit mode', () => {
+    cy.visit('http://localhost:3000/customSchemaValidation/onSubmit');
+    cy.get('button').click();
+
+    cy.focused().should('have.attr', 'name', 'firstName');
+
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+    cy.get('select[name="selectNumber"] + p').contains('selectNumber error');
+    cy.get('input[name="minRequiredLength"] + p').contains(
+      'minRequiredLength error',
+    );
+    cy.get('input[name="radio"] + p').contains('radio error');
+
+    cy.get('input[name="firstName"]').type('bill');
+    cy.get('input[name="lastName"]').type('luo123456');
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+    cy.get('select[name="selectNumber"]').select('1');
+    cy.get('input[name="pattern"]').type('luo');
+    cy.get('input[name="min"]').type('1');
+    cy.get('input[name="max"]').type('21');
+    cy.get('input[name="minDate"]').type('2019-07-30');
+    cy.get('input[name="maxDate"]').type('2019-08-02');
+    cy.get('input[name="lastName"]')
+      .clear()
+      .type('luo');
+    cy.get('input[name="minLength"]').type('2');
+    cy.get('input[name="minLength"] + p').contains('minLength error');
+    cy.get('input[name="min"] + p').contains('min error');
+    cy.get('input[name="max"] + p').contains('max error');
+    cy.get('input[name="minDate"] + p').contains('minDate error');
+    cy.get('input[name="maxDate"] + p').contains('maxDate error');
+
+    cy.get('input[name="pattern"]').type('23');
+    cy.get('input[name="minLength"]').type('bi');
+    cy.get('input[name="minRequiredLength"]').type('bi');
+    cy.get('input[name="radio"]').check('1');
+    cy.get('input[name="min"]')
+      .clear()
+      .type('11');
+    cy.get('input[name="max"]')
+      .clear()
+      .type('19');
+    cy.get('input[name="minDate"]').type('2019-08-01');
+    cy.get('input[name="maxDate"]').type('2019-08-01');
+    cy.get('input[name="checkbox"]').check();
+
+    cy.get('p').should('have.length', 0);
+    cy.get('#renderCount').contains('24');
+  });
+
+  it('should validate the form with onBlur mode', () => {
+    cy.visit('http://localhost:3000/customSchemaValidation/onBlur');
+
+    cy.get('input[name="firstName"]').focus();
+    cy.get('input[name="firstName"]').blur();
+    cy.get('input[name="firstName"] + p').contains('firstName error');
+    cy.get('input[name="firstName"]').type('bill');
+    cy.get('input[name="lastName"]').focus();
+    cy.get('input[name="lastName"]').blur();
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+    cy.get('input[name="lastName"]').type('luo123456');
+    cy.get('input[name="lastName"]').blur();
+    cy.get('input[name="lastName"] + p').contains('lastName error');
+    cy.get('select[name="selectNumber"]').focus();
+    cy.get('select[name="selectNumber"]').blur();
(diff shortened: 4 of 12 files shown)


hemedani:
the yup lib is very heavy and Superstruct is so lighter.
please merge this pr and make the hook-form more useful

JeromeDeLeon:
@bluebill1049 ?

bluebill1049:
> the yup lib is very heavy and Superstruct is so lighter.
> please merge this pr and make the hook-form more useful

Thanks for the follow up.

1. i personally hope to see more up votes for such a feature.

I am currently busy working on other issues and bug, which is more urgency to get it right for exciting users. Also on a note in terms of merging the PR.

1. it's not going to just a merge (the branch itself is quite outdated with current code), so i will be cherry pick the fix and make it work with current code base
2. will need to update the docs as well.

bluebill1049:
FYI i am not forgetting about this PR, currently busy working over improvements with `useFieldArray`

bluebill1049:
Thanks for your patience and votes, I will be working on it in the next weeks.

fubhy:
This is also required to allow form validation to use external / contextual values to compare against reliably. Since the `useForm` hook has no notion of `deps` (as other hooks), once it's initialized, the validationSchema, etc. cannot be updated anymore. Therefore, the only way to pass in external values for validation in custom `test` callbacks is through refs (`useRef`). This is a bit unwieldy in large forms. A simple fix would be to allow passing an updatable `context` object that, on every change, would trigger re-validation and also would be accessible to Yup (and other validation libraries). Yup supports such a `context` object natively.

bluebill1049:
What's this `context` got to do with support custom validation schema? my understanding is that the user will be using the `resolver` callback and sort out your validation there. 

Provide some code example or proposal for your suggestion above.

fubhy:
> What's this `context` got to do with support custom validation schema? my understanding is that the user will be using the `resolver` callback and sort out your validation there.
> 
> Provide some code example or proposal for your suggestion above.

Passing external values (e.g. from props / state / whatever) to the validation is currently impossible:

```

export interface MyComponentProps {
  changingPropValue: string;
}

const validationSchema = Yup.object().shape({
  myField: Yup.string().test({
    test: function(value) {
      // Yup passes the current 'context' as `this` to the test function. That way, it normally allows consumers to
      // provide a custom `options` object that can hold a changing context e.g. for comparing values.
      const context = this.options.context as MyComponentProps;

      if (value !== context.changingPropValue) {
        // Validation failed.
        return false;
      }

      // Validation successful.
      return true;
    },
  }),
});

const MyComponent: React.FC<MyComponentProps> = props => {
  const form = useForm({
    validationSchema,
    // NOTE: You removed this in v4 for some reason. This is currently NOT possible.
    validationSchemaOptions: {
      context: props,
    },
  });
  
  ...
};
```

The only way to allow this functionality is to use a `useRef` hook and inline the validation schema which is quite dirty:

```

const MyComponent: React.FC<MyComponentProps> = props => {
  const contextRef = useRef(props.changingPropValue);
  useEffect(() => {
    contextRef.current = props.changingPropValue;
  }, [props.changingPropValue]);    
  
  const form = useForm({
    validationSchema: Yup.object().shape({
      myField: Yup.string().test({
        test: (value) => {
          if (value !== contextRef.current) {
            // Validation failed.
            return false;
          }

          // Validation successful.
          return true;
        },
      }),
    }),
  });

  ...
};
```

If a custom validation callback that allows for other validation providers (e.g. Joi, etc.) to be passed in also solves this problem, that's nice. However, since `useForm()` does NOT have any notion of `deps` it has to ensure that contextual values can be updated for the validation to consider them.

hemedani:
> > What's this `context` got to do with support custom validation schema? my understanding is that the user will be using the `resolver` callback and sort out your validation there.
> > Provide some code example or proposal for your suggestion above.
> 
> Passing external values (e.g. from props / state / whatever) to the validation is currently impossible:
> 
> ```
> 
> export interface MyComponentProps {
>   changingPropValue: string;
> }
> 
> const validationSchema = Yup.object().shape({
>   myField: Yup.string().test({
>     test: function(value) {
>       // Yup passes the current 'context' as `this` to the test function. That way, it normally allows consumers to
>       // provide a custom `options` object that can hold a changing context e.g. for comparing values.
>       const context = this.options.context as MyComponentProps;
> 
>       if (value !== context.changingPropValue) {
>         // Validation failed.
>         return false;
>       }
> 
>       // Validation successful.
>       return true;
>     },
>   }),
> });
> 
> const MyComponent: React.FC<MyComponentProps> = props => {
>   const form = useForm({
>     validationSchema,
>     // NOTE: You removed this in v4 for some reason. This is currently NOT possible.
>     validationSchemaOptions: {
>       context: props,
>     },
>   });
>   
>   ...
> };
> ```
> 
> The only way to allow this functionality is to use a `useRef` hook and inline the validation schema which is quite dirty:
> 
> ```
> 
> const MyComponent: React.FC<MyComponentProps> = props => {
>   const contextRef = useRef(props.changingPropValue);
>   useEffect(() => {
>     contextRef.current = props.changingPropValue;
>   }, [props.changingPropValue]);    
>   
>   const form = useForm({
>     validationSchema: Yup.object().shape({
>       myField: Yup.string().test({
>         test: (value) => {
>           if (value !== contextRef.current) {
>             // Validation failed.
>             return false;
>           }
> 
>           // Validation successful.
>           return true;
>         },
>       }),
>     }),
>   });
> 
>   ...
> };
> ```
> 
> If a custom validation callback that allows for other validation providers (e.g. Joi, etc.) to be passed in also solves this problem, that's nice. However, since `useForm()` does NOT have any notion of `deps` it has to ensure that contextual values can be updated for the validation to consider them. So, since `useForm()` does not react to changes of the passed values (`validationSchema`, `defaultValues`, etc.) and simply continues to use the values that were originally passed, there is no way to provide contextual, changing parameters for e.g. validation.

you shuld use hidden input and set that value to the value of props / state / whatever

fubhy:
> you shuld use hidden input and set that value to the value of props / state / whatever

I actually did that before but that makes it even more problematic because it adds these values to the form payload although they do not belong there. Furthermore, the example I used is simplified quite a bit. Some of the values I need for comparison as the values I actually need are not serializable. I don't think that it would be correct to put them in hidden form values if they are not part of the form.

bluebill1049:
```
const resolver = (props, context) => {
  //pass your props here
}


const { register, handleSubmit, errors } = useForm({
  validationSchema: { resolver: resolver.bind(null, props) }
});
```

bluebill1049:
I don't get why people love to uppercase particular words in issues or PR, not only in this thread but some of the other issues as well. man... you try to screaming at me? hahah i can read lowercase and uppercase just fine. I DONT need that....

fubhy:
> I don't get why people love to uppercase particular words in issues or PR,

Umh what? That's just a normal way to emphasize certain words?! You use bold, cool, I use uppercase. What's the problem there?

> There is nothing required here buddy, I don't do required work, I do it all for free for everyone as open source. so next time please consider asking nicely.

You are being quite condescending. Why exactly? All I stated was that this was required for sth. to work in a particular way that would make it function better. Not that YOU are REQUIRED to implement it. 

I do open source work too, what's your point?

I came here to contribute to this issue stating another use-case. You are suddenly acting weird, calling me "buddy" out of nowhere. What exactly triggered you?

JeromeDeLeon:
We're actually playing nice here. He's just expressing his thoughts. As you can read, he called you "buddy". You know that's something someone called you when trying to be friendly and all. If that offended you, he should apologize for being friendly but I don't think so. Don't get so triggered because maybe he just misunderstood when you said **required**. That's all.

Also, for use case of uppercase title, you are right, but he just shared his thoughts, is there anything wrong with that? He was even laughing in his context of message.

Thanks for trying to help solving the problem, but patient and understanding are the key. Thanks again. Have a nice day everyone.

fubhy:
Well I guess that's a problem with the interpretation of the english language then. Because in this context, calling someone "buddy" like that is like talking someone down. So absolutely not friendly and quite condescending. I'd suggest not doing that in such a discussion. If that was not how you meant it, however, then it's all cool.

Same goes for what I said before by the way though... That was me trying to be purely helpful and explaing an absolutely valid (and imho indeed common) use-case that the current solution for validation doesn't cover.

bluebill1049:
First of all, thanks for your contribution towards the issue. If u read my message twice, I said it’s in other issues as well when people use uppercase for certain words, which we don’t use at work at least for me. So the issue stuck with me for awhile and here is my expression and my take on it. To be honest I rarely use bold as well. It’s important to remember there is another real person that are you dealing with and potential I am dealing many more issues than you are, and honestly respect is important for you and myself. When someone jump out the context with required work to do, it triggers me. I don’t get paid for what I do and I don’t get told so. If I misunderstood that probably because I am tired of working, but my expression remained the same, I don’t like reading uppercase words, I feel like someone 😱 and when u asking for improvement asking nicely. We all play nice here and I expect and hope everyone else do the same.

fubhy:
> First of all, thanks for your contribution towards the issue. If u read my message twice, I said it’s in other issues as well when people use uppercase for certain words, which we don’t use at work at least for me. So the issue stuck with me for awhile and here is my expression and my take on it. To be honest I rarely use bold as well. It’s important to remember there is another real person that are you dealing with and potential I am dealing many more issues than you are, and honestly respect is important for you and myself. When someone jump out the context with required work to do, it triggers me. I don’t get paid for what I do and I don’t get told so. If I misunderstood that probably because I am tired of working, but my expression remained the same, I don’t like reading uppercase words, I feel like someone 😱 and when u asking for improvement asking nicely. We all play nice here and I expect and hope everyone else do the same.

Alright. No harm done. Let's put it down as a misinterpretation of the used wording on both sides then. I was really just trying to explain my issue and emphasizing the parts that I thought were important to mention. The word "required" was in no way directed at you in the sense of "this is what I require from you". Same with bold or uppercase words by the way, those are never meant as shouting and always simply emphasize the words, that's all.

I spend countless hours on open source projects myself, speaking at conferences for free and helping people on stackoverflow many hours every week. I would never phrase a comment on github in this way.

bluebill1049:
> > First of all, thanks for your contribution towards the issue. If u read my message twice, I said it’s in other issues as well when people use uppercase for certain words, which we don’t use at work at least for me. So the issue stuck with me for awhile and here is my expression and my take on it. To be honest I rarely use bold as well. It’s important to remember there is another real person that are you dealing with and potential I am dealing many more issues than you are, and honestly respect is important for you and myself. When someone jump out the context with required work to do, it triggers me. I don’t get paid for what I do and I don’t get told so. If I misunderstood that probably because I am tired of working, but my expression remained the same, I don’t like reading uppercase words, I feel like someone 😱 and when u asking for improvement asking nicely. We all play nice here and I expect and hope everyone else do the same.
> 
> Alright. No harm done. Let's put it down as a misinterpretation of the used wording on both sides then. I was really just trying to explain my issue and emphasizing the parts that I thought were important to mention. The word "required" was in no way directed at you in the sense of "this is what I require from you". Same with bold or uppercase words by the way, those are never meant as shouting and always simply emphasize the words, that's all.
> 
> I spend countless hours on open source projects myself, speaking at conferences for free and helping people on stackoverflow many hours every week. I would never phrase a comment on github in this way.

Respect on your hrs and work towards the community. I think we do have misunderstanding  between each others. Let’s move on with this problem and working toward a good solution for context, so what’s your thoughts on my solution above?

fubhy:
> Let’s move on with this problem and working toward a good solution for context, so what’s your thoughts on my solution above?

A function callback for validation would be flexible enough to manually plug in any Yup / Joi / Superstruct validation including custom contexts. At that point, I wonder however is `validationSchema` is still the right word for the configuration option. Maybe `validationSchema` could simply remain in place for backwards compatibility and a new option `validationCallback` could be added that would overrule `validationSchema` and simply allow a user to pass in a function as you suggested.

bluebill1049:
What about a new prop called

schemeContext?

Which pass the props as second argument.

bluebill1049:
Yea I think we will deprecate validation scheme eventually.

For now in my head:

SchemeResolver
schemeContext

(Sorry i try to type while standing on the bus)

fubhy:
> What about a new prop called
> 
> schemeContext?
> 
> Which pass the props as second argument.

You mean the props of the component that calls `useForm`? I think that would not be flexible enough because in some use-cases the user might want to further transform values coming from `props` and then use the resulting values as the validation context.

I think by simply allowing users to pass a function callback for validation it would allow for full control and flexibility. The only thing that has to be ensured is that the returned value from this callback (or promise) conforms to the expected type / error shape. So at that point the function would basically be an adapter between `reacy-hook-form` and the different validation providers (Yup, Joi, etc.).

I think I am in favor for just a callback in this case and to then eventually deprecate the direct support for validationSchema eventually.

bluebill1049:
Can u plz give an example of your callback implantation? Because resolve in my eyes is a callback function

fubhy:
I was thinking of something like this:

```typescript

const myStaticValidationSchema = Yup.object().shape({
  myContextAwareField: Yup.number().test({
    test: function (value) {
      const context = this.options.context as MyChangingValidationContext;
      return context.someNumber.isGreaterThan(value);
    },
  }),
});

const MyComponent: React.FC = () => {
  const context = useMemo(() => {
    return computeExpensiveContextForValidation(props.foo, props.bar);
  }, [props.foo, props.bar]);

   const form = useForm({
     // For this to work, it's important that useForm internally allows the reference of `validationCallback` to change 
     // instead of locking it to the value at initialization / first render. That way, it would
     // allow the value of `context` to change over time if required.
     validationCallback: (values) => myStaticValidationSchema.validate(values, context),
   });
};
```

bluebill1049:
I think that’s the same as the resolver right? Correct me if I am wrong.

bluebill1049:
Actually looks like we can remove context then....😅

fubhy:
> Actually looks like we can remove context then....

Yes, a function callback seems like the better option.

> I think that’s the same as the resolver right? Basically bring resolver online, Correct me if I am wrong.

I think I don't understand what exactly the `resolver` part you mentioned entails. Why is it necessary to provide the configuration options like this:

```
{
  validationSchema: { resovler }
}
```

To be that seems a bit unwieldy. Is it required due to how validation is invoked internally?

bluebill1049:
@fubhy that's the same as your callback, except i named it as resolver, because it resolve errors and values

```
import React from "react";
import useForm from "react-hook-form";
import joi from "@hapi/joi";

const schema = Joi.object({
  username: Joi.string().alphanum().min(3).max(30).required()
});

// Build your resolver to display errors and pass valid values
const resolver = (data: any) => {
  const { error, value: values } = validationSchema.validate(data, {
    abortEarly: false,
  });
  return {
    values: error ? {} : values,
    errors: error
      ? error.details.reduce((previous, currentError) => {
          return {
            ...previous,
            [currentError.path[0]]: currentError,
          };
        }, {})
      : {},
  };
};

function App() {
  const { register, handleSubmit, errors } = useForm({
    schemaResolver: resolver
   });
  
  return (
      <input type="text" name="username" ref={register} />
    </form>
  );
}
```

fubhy:
Ah, that makes sense. Because yes, it also has to convert the returned values (error objects) from the underlying validation handler (Joi vs. Yup, etc.) into a format that react-hook-form understands. In this case, I would argue, however, that `schemaResolver` is not a good name because whether or not there is a validation schema underneath is an implementation detail. There might not be a schema at all (user doesn't use Yup or Joi and simply uses the callback to run some custom validation). Hence why I proposed to ambiguously call it `validationCallback` and then let the user decide whether there's a schema or not. If that makes sense?

bluebill1049:
> Ah, that makes sense. Because yes, it also has to convert the returned values (error objects) from the underlying validation handler (Joi vs. Yup, etc.) into a format that react-hook-form understands. In this case, I would argue, however, that `schemaResolver` is not a good name because whether or not there is a validation schema underneath is an implementation detail. There might not be a schema at all (user doesn't use Yup or Joi and simply uses the callback to run some custom validation). Hence why I proposed to ambiguously call it `validationCallback` and then let the user decide whether there's a schema or not. If that makes sense?

valid point above, how about `validationResolver`?

fubhy:
Yes, validationResolver would also work. That it's a callback is a bit redundant information anyways (it's a function after all). +1 for `validationResolver` or `validationHandler`.

fubhy:
Actually... There is some value in the `context` object. By adding a way for the form / validation to react to changes of the `context`, it would make it very convenient to automatically re-run validation whenever necessary. Of course, a user could always simply do this:

```
const context = useMemo(() => {
  const expensivelyComputedContext = computeContext(props.myDependency);
  return expensiveComputedContext;
}, [props.myDependency]);

const form = useForm({
  validationResolver: (values) => myYupSchema.validate(values, { context }),
});

useEffect(() => {
  form.triggerValidation();
}, [form, context]);
```

But it would of course be nicer to do just this:

```
const context = useMemo(() => {
  const expensivelyComputedContext = computeContext(props.myDependency);
  return expensiveComputedContext;
}, [props.myDependency]);

const form = useForm({
  // This context is passed as a second argument to the validationResolver and also to the
  // submit handler (!). Internally, useForm observes updates to this object just like other hooks
  // do with `deps`.
  formContext: context,
  validationResolver: (values, context) => myYupSchema.validate(values, { context }),
});
```

bluebill1049:
i think we can go with this then.

1. really it's a form level validation
2. resolver will sort out the errors and value
3. context to support external value
4. group all form level validate config in a single object

```
validate: {
  resolver,
  context
}
```

fubhy:
> i think we can go with this then.
> 
> 1. really it's a form level validation
> 2. resolver will sort out the errors and value
> 3. context to support external value
> 4. group all form level validate config in a single object
> 
> ```
> validate: {
>   resolver,
>   context
> }
> ```

In my last comment I also mentioned that potentially, the `context` values might also be good as an extra argument for the submit handler. This is not absolutely necessary because one can of course load this into the submit handler manually but it might be nice for the API. So if the context is a FORM level context (not just validation) it should probably be separately configured (not as part of a `validate` sub-object). What do you think? I would personally be in favor of keeping the configuration options of the `useForm` hook flat anyways (no nested objects), but don't have a strong opinion on that if you decide that the `context` should only be used for validation exclusively.

bluebill1049:
Thanks @fubhy 👍

bluebill1049:
new PR here: https://github.com/react-hook-form/react-hook-form/pull/974

PR #818, opened 2020-01-12
chore(eslintrc): change react version to `detect`

neighborhood999:
The `package.json` react version is `16.12.0`, but in the eslintrc is `16.9`, so I think it can adjust to `detect` automatically picks the version you have installed.


bluebill1049:
wow, that's cool. I didn't even know that! something to learn each day. thanks @neighborhood999

bluebill1049 (approved):

PR #820, opened 2020-01-14
`isSubmitted` and `submitCount` should only update on success

SimplyComplexable:
**Problem**
`isSubmitted` and `submitCount` update whenever the submit handler is triggered, regardless of whether or not the submit was successful. This should only update if the form is valid and there were no issues during the submit process.

**Solution**
Move these updates out of the `finally` block and into the `try` block, so that they only happen if there are no issues with submit.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -915,6 +915,8 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
         if (isEmptyObject(fieldErrors)) {
           errorsRef.current = {};
           await callback(transformToNestObject(fieldValues), e);
+          isSubmittedRef.current = true;
+          submitCountRef.current = submitCountRef.current + 1;
         } else {
           if (submitFocusError) {
             for (const key in fieldsRef.current) {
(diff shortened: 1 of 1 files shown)


bluebill1049:
Hey @SimplyComplexable thanks for the PR, however i think submit count is irrelevant to validation and so does isSubmitted.

SimplyComplexable:
Hi @bluebill1049 with that in mind, is there a way that `react-hook-form` tracks whether or not the form has actually submitted? Currently I'm doing this using normal react state that's updated inside my own submit handler, but it makes more sense to me for this state to be part of the the hook. 
Is there any other state I can utilize for this?

bluebill1049:
> Hi @bluebill1049 with that in mind, is there a way that `react-hook-form` tracks whether or not the form has actually submitted? Currently I'm doing this using normal react state that's updated inside my own submit handler, but it makes more sense to me for this state to be part of the the hook.
> Is there any other state I can utilize for this?

hmmm good question. i think you may have to track yourself at handleSubmit level.

bluebill1049:
i am going to close this PR now, let me know if you still have question. i am more than happy to help for a solution :)

Issue #827, opened 2020-01-14
Can allow 'empty form' append in 'useFieldArray'?


tanmen:
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
I want to create an empty form, as in previous versions.
When useFieldArray is used, the contents are required in append, so the behavior like the previous version cannot be performed.

**Describe alternatives you've considered**
Allows empty objects to be append, prepend, and insert.

**Additional context**
Not unique key problem may occur.

bluebill1049:
can you share a codesandbox with this issue @tanmen ?

tanmen:
https://codesandbox.io/s/react-hook-form-usefieldarray-7vdfv
↑Current version issues

https://codesandbox.io/s/react-hook-form-array-fields-s1kfh
↑'Add friend' function I expect

bluebill1049:
can you pass an empty object?

tanmen:
ok! i try it :)

PR #830, opened 2020-01-14
allow empty object in prepend, append and insert.

tanmen:
allow undefined form object.

issue: #827

before
```js
const { control } = useForm();
const { append } = useFieldArray({ control, name: "test" });
append() // <- error
```

after
```js
const { control } = useForm();
const { append } = useFieldArray({ control, name: "test" });
append() // <- arrow undfined or null
```

### DIFF
--- a/src/useFieldArray.test.ts
+++ b/src/useFieldArray.test.ts
@@ -53,6 +53,16 @@ describe('useFieldArray', () => {
       { id: '1', test: 'test' },
       { id: '1', test: 'test1' },
     ]);
+
+    act(() => {
+      result.current.append({});
+    });
+
+    expect(result.current.fields).toEqual([
+      { id: '1', test: 'test' },
+      { id: '1', test: 'test1' },
+      { id: '1' },
+    ]);
   });
 
   it('should pre-append data into the fields', () => {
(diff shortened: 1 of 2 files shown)


bluebill1049:
hey @tanmen thanks for the PR, maybe it's easier to make sure the user put down an argument for the append() we can improve the doc.  if you support empty, then you need to deal with null and undefined as well. what do you think?

tanmen:
@bluebill1049
maybe it's easier to make sure the user put down an argument for the append() we can improve the doc.
-> thank you :)

if you support empty, then you need to deal with null and undefined as well. what do you think?
-> i think too :) i try add type null.

If my perception is wrong, feel free to say.

tanmen:
this sample code.
https://codesandbox.io/s/hopeful-dawn-wpeqr

I have question.
Can I check type in codesandbox?

bluebill1049:
Sorry I meant, we should always ask dev to put an argument and not support an empty argument.

follow the `push`, eg:

`[1,2,3].push()` This will push nothing into the array

So to work around with the `append` issue you are having: I think you can simply do `append({})` and translate to js 

`[1,2,3].push({})` This will push something into the array.

tanmen:
If always wanted an argument, I felt that only empty object was good, so I chose only empty object.

bluebill1049:
> If always wanted an argument, I felt that only empty object was good, so I chose only empty object.
> what do you think?

sounds great 👍

tanmen:
@bluebill1049 thank you :)
Can i add test?

bluebill1049 (dismissed):

bluebill1049:
> @bluebill1049 thank you :)
> Can i add test?

why not! ❤️

bluebill1049 (approved):
TY very much 👍

tanmen:
Thank you always so much :)

PR #839, opened 2020-01-15
close #833 dirty value with useFieldArray

### LINKED ISSUE #833: Calling useFieldArray hook methods does not update formState.dirty value
**Describe the bug**

Calling useFieldArray hook methods does not update formState.dirty value.

**To Reproduce**
1. Go to https://codesandbox.io/s/react-hook-form-usefieldarray-popep
2. Click on APPEND
3. Observe that formState hasn't changed

**Codesandbox link**
https://codesandbox.io/s/react-hook-form-usefieldarray-popep

**Expected behavior**
append, prepend, remove, swap, move, insert should all update dirty state to true

 - OS: Windows
 - Browser Chrome Version 79.0.3945.117 (Official Build) (64-bit)
 - Version 4.5.2

**Additional context**

bluebill1049:


tomas-c:
Found one interesting case:

1. Go to https://codesandbox.io/s/react-hook-form-usefieldarray-9u12n
2. Append a field and observe the state becomes dirty=true
3. Remove the appended field and observe the sate is still dirty=true
4. Click on the remaining field and then outside it so that onBlur event is trigerred
5. Observe that the state is now dirty=false

It looks like after the blur event react-hook-form compares all the fields to the default values in order to determine whether dirty state should change. Should we trigger the same for useFieldArray actions?

bluebill1049:
> Found one interesting case:
> 
> 1. Go to https://codesandbox.io/s/react-hook-form-usefieldarray-9u12n
> 2. Append a field and observe the state becomes dirty=true
> 3. Remove the appended field and observe the sate is still dirty=true
> 4. Click on the remaining field and then outside it so that onBlur event is trigerred
> 5. Observe that the state is now dirty=false
> 
> It looks like after the blur event react-hook-form compares all the fields to the default values in order to determine whether dirty state should change. Should we trigger the same for useFieldArray actions?

make sense, will sort out tmr.

bluebill1049:
> Found one interesting case:
> 
> 1. Go to https://codesandbox.io/s/react-hook-form-usefieldarray-9u12n
> 2. Append a field and observe the state becomes dirty=true
> 3. Remove the appended field and observe the sate is still dirty=true
> 4. Click on the remaining field and then outside it so that onBlur event is trigerred
> 5. Observe that the state is now dirty=false
> 
> It looks like after the blur event react-hook-form compares all the fields to the default values in order to determine whether dirty state should change. Should we trigger the same for useFieldArray actions?

try this beta version: 4.5.4-beta.1

PR #842, opened 2020-01-15
fix issue around reset with fieldArray

bluebill1049:


src/useFieldArray.ts
```diff
@@ -25,7 +25,7 @@ export function useFieldArray<
   const resetFields = () => {
     for (const key in globalFields) {
       if (isMatchFieldArrayName(key, name)) {
-        unregister(key);
+        unregister(key, true);
```
kotarella1110: @bluebill1049 Isn't this change unnecessary?

bluebill1049: hey @kotarella1110 its necessary, for `useFieldArray`, i am letting that custom hook to take care of `unregister` instead of `useForm`. Because mutationObserver is always going to be behind of react reconciler. so useFieldArray will take care `unregister` inputs.

kotarella1110: @bluebill1049 OK! thanks for your reply.

PR #845, opened 2020-01-15
Feature: Value transformation for string fields (type="text"/type="pa…

AntoineEsteve:
**Issue:**
- If you trim an input value using `onChange`, it's not possible to let the user type spaces (eg: between words).
- If you trim once the form is submitted, it's too late because you can't use the form validation (eg: a required field filled with spaces will be valid).

**Feature:** The `register` function can take 2 new optional parameters:
- `trim?: boolean`
- `transform?: (value: string) => string`
Both options change the value returned with `getValues`/`watch`/`handleSubmit`... but won't affect the input value directly.

### DIFF
--- a/app/src/app.tsx
+++ b/app/src/app.tsx
@@ -21,6 +21,7 @@ import SetValueWithTrigger from './setValueWithTrigger';
 import IsValid from './isValid';
 import Controller from './controller';
 import UseFieldArray from './useFieldArray';
+import Transform from './transform';
 
 const App: React.FC = () => {
   return (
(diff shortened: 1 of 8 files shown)


bluebill1049:
Thanks very much for this detailed PR! ❤️ Let's have a quick discussion next time, i am always on Spectrum :) 

I love the idea `transform`, however i prefer to keep `register` purely for validation purpose, this is mostly coming from DX perspective, otherwise it will open the gate for any other attributes, which make the functional of register complicated. To be honest, i never want RHF to do everything (swiss knife) happy to keep it small and simple.

So what's an alternative solution in my head: i think i would be duty for a mask input component, and wrapped with `Controller` to just collect data.

let me know what do you think. again thanks very much for this detailed PR, i can see a lot of good effort and time spend on it.

cheers
bill

bluebill1049:
Thanks for your great effort ❤️ closing this PR for now.

PR #846, opened 2020-01-16
clean up code for useFieldArray

kotarella1110:
minor fix this PR: https://github.com/react-hook-form/react-hook-form/pull/842

* ~~removed the second argument passed by calling the `unregister` function.~~
* instead of fixing if statement in `removeEventListenerAndRef` function, fixed position of ` resetFieldArray ` function called in `reset` function. this one is more simple.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -1052,7 +1052,7 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
       isReValidateOnSubmit,
     },
     errors: errorsRef.current,
-    fields: fieldsRef.current,
+    fieldsRef,
     resetFieldArrayFunctionRef,
     fieldArrayNamesRef,
   };
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -495,12 +495,7 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
 
   const removeEventListenerAndRef = useCallback(
     (field: Field | undefined, forceDelete?: boolean) => {
-      if (
-        !field ||
-        (field &&
-          isNameInFieldArray(fieldArrayNamesRef.current, field.ref.name) &&
-          !forceDelete)
-      ) {
+      if (!field) {
         return;
       }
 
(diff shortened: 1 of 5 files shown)


src/useForm.ts
```diff
@@ -495,12 +495,7 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
 
   const removeEventListenerAndRef = useCallback(
     (field: Field | undefined, forceDelete?: boolean) => {
-      if (
```
bluebill1049: i think we still need these code :)

kotarella1110: Could you explain a little more?
I don't think there is a case that causes problems.

bluebill1049: i will chat with you in spectrum :)

src/useForm.ts
```diff
@@ -1052,7 +1052,7 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
       isReValidateOnSubmit,
     },
     errors: errorsRef.current,
-    fields: fieldsRef.current,
+    fieldsRef,
```
bluebill1049: what are we changing here?

kotarella1110: After calling `reset ()` of useForm, `fieldsRef.current` of useForm and `globalFields` of useFieldArray are not synchronized. 
If we call `reset ()` at least once, `globalFields` will be an empty object.

スクリーンショット 2020-01-16 17 55 32

kotarella1110: The fixed code will sync correctly.

スクリーンショット 2020-01-16 17 56 14

codesandbox: https://codesandbox.io/s/react-hook-form-usefieldarray-issue-846-w75rv

kotarella1110: `MutationObserver` will not be kicked because `uregister` has been fixed to be called correctly.

bluebill1049 (approved):

src/useForm.ts
bluebill1049: let's merge it in and i will test out with a beta version.

bluebill1049:
please test out this beta version: react-hook-form@4.5.4-beta.2

bluebill1049:
Tested on my end looking good! Fantastic job on this fix 👍 素晴らしい仕事!

kotarella1110:
Thanks @bluebill1049 ! どうもありがとう!

PR #849, opened 2020-01-16
Assign empty filelist

hekkim:
First of all, I'm really sorry about that I made a mistake about handling `<input type="file" />` in the previous PR (#835).

So, when you try to clean up files through `setValue`, you can see the following error because input's files only accept `FileList`.
```
Uncaught TypeError: Failed to set the 'files' property on 'HTMLInputElement': The provided value is not of type 'FileList'.
```

I just changed `input.files` assigned empty `FileList` when user pass `null` or `undefined`.

I'm sorry again for the previous mistake.

ps. I would like to add test case for sure, but it looks like there's no way to create `FileList` manually in node environment. If I find the way, I'll add test cases.
ps2. If you think user needs to pass empty `FileList` manually to `setValue` for removing files, I'm going to change the codes to accept only `FileList`.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -195,7 +195,7 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
         if (value instanceof FileList) {
           ref.files = value;
         } else {
-          ref.files = new DataTransfer().files;
+          ref.value = value;
         }
       } else if (isMultipleSelect(type)) {
         [...ref.options].forEach(
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -194,6 +194,8 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
       } else if (isFileInput(type)) {
         if (value instanceof FileList) {
           ref.files = value;
+        } else {
+          ref.files = new DataTransfer().files;
         }
       } else if (isMultipleSelect(type)) {
         [...ref.options].forEach(


bluebill1049:
that's all good, perhaps follow with a quick codesandbox with your solution above and i can do a quick test as well. it's my bad as well which i didn't test out your solution last time. all good tho :) we can patch it soon.

bluebill1049:
Looking at stack overflow looks like set empty string is correct https://stackoverflow.com/questions/20549241/how-to-reset-input-type-file

hekkim:
Here's a codesandbox example.

https://codesandbox.io/s/sleepy-currying-duzgg

Reproduce scenario.

1. Attach an image.
2. Click Delete button.
3. Show an error.

The difference with codes in stackoverflow that you found is we're assign FileList into `files` instead of `value`.
So, the error is occurred because we're trying to assign invalid type of value.

bluebill1049:
awesome i will do a quick test with your code tonight and verify the fix 👍

src/useForm.ts
```diff
@@ -194,6 +194,8 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
       } else if (isFileInput(type)) {
         if (value instanceof FileList) {
           ref.files = value;
+        } else {
+          ref.files = new DataTransfer().files;
```
```suggestion
          ref.value = value;
```

bluebill1049: let's do that, it's probably better if user just want to set the value of the input.

hekkimmt: Cool. It works very well.
I'm gonna push changes very soon.

bluebill1049: you can click above with commit suggestion :)

bluebill1049 (approved):

src/useForm.ts
hekkimmt: Oops. I just updated it again.
It's my first time to contribute codes in open source, so I'm a little awkward to handle it. 😅 
From the next time, I'll follow the direction like you said.

PR #850, opened 2020-01-16
refactor(useForm): Replace `Proxy` with property getters.

avocadowastaken:
Currently, `useForm` has different behaviours if `Proxy` is supported or not, these changes will make it similar for both cases.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -94,13 +94,14 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
     typeof document !== UNDEFINED &&
     !isWindowUndefined &&
     !isUndefined(window.HTMLElement);
+  const isProxyEnabled = isWeb && 'Proxy' in window;
   const readFormState = useRef<ReadFormState>({
-    dirty: false,
+    dirty: !isProxyEnabled,
     isSubmitted: isOnSubmit,
-    submitCount: false,
-    touched: false,
-    isSubmitting: false,
-    isValid: false,
+    submitCount: !isProxyEnabled,
+    touched: !isProxyEnabled,
+    isSubmitting: !isProxyEnabled,
+    isValid: !isProxyEnabled,
   });
   const {
     isOnBlur: isReValidateOnBlur,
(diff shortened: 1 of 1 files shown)


src/useForm.ts
```diff
@@ -94,14 +94,13 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
     typeof document !== UNDEFINED &&
     !isWindowUndefined &&
     !isUndefined(window.HTMLElement);
-  const isProxyEnabled = isWeb && 'Proxy' in window;
```
bluebill1049: you will need this for react native :)

bluebill1049:
thanks for the PR 👍 but i think we do need feature detect for proxy. (react-native)

umidbekkarimov:
@bluebill1049 Oh, it had never occurred to me. Is it because React Native will always have controlled components?  At least we should add some comments regarding that.

The reason why I wanted to replace Proxy is that we lint our built file with `eslint-plugin-compat` to ensure compatibility with some old browsers and there are no good Proxy polyfill that covers all edge cases.

src/useForm.ts
```diff
@@ -1076,17 +1100,6 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
...
-            if (prop in obj) {
-              readFormState.current[prop] = true;
-              return obj[prop];
-            }
-
-            return {};
-          },
-        })
-      : formState,
+    formState,
```
bluebill1049: To be honest i think let's stay for what we having right now. This is introduce quite bit extra code for this lib (size perspective), but thanks for the PR tho.

bluebill1049:
> @bluebill1049 Oh, it had never occurred to me. Is it because React Native will always have controlled components? At least we should add some comments regarding that.
> 
> The reason why I wanted to replace Proxy is that we lint our built file with `eslint-plugin-compat` to ensure compatibility with some old browsers and there are no good Proxy polyfill that covers all edge cases.

we do feature detect for `Proxy` in the lib (Proxy is a web only API). That's fine with old browser as well, only diff is more re-renders.

PR #855, opened 2020-01-17
fix #833 field array dirty formState

### LINKED ISSUE #833: Calling useFieldArray hook methods does not update formState.dirty value

bluebill1049:


bluebill1049:
https://codesandbox.io/s/react-hook-form-usefieldarray-yikiz

pretty stable now :)

bluebill1049:
@JeromeDeLeon @kotarella1110 please take a look at this PR if you guys have sometimes. We can get this boy out this weekend.

JeromeDeLeon:
When I moved the `console.log` in between `useForm` and `useFieldArray`, `dirty` is working and set to `true` when I appended a row; however, when the `log` is below the custom hooks, `dirty` was not setting correctly on first append scenario and will only be set to `true` after the second one.

bluebill1049:
> When I moved the `console.log` in between `useForm` and `useFieldArray`, `dirty` is working and set to `true` when I appended a row; however, when the `log` is below the custom hooks, `dirty` was not setting correctly on first append scenario and will only be set to `true` after the second one.

thanks @JeromeDeLeon you have a screenshot of the actual error.

JeromeDeLeon:
Not Working
image

Working
image

bluebill1049:
hmm looks like it's broken now, i must have broke it.

bluebill1049:
should be better now: https://github.com/react-hook-form/react-hook-form/pull/855

still need to fix onBlur

src/useFieldArray.ts
```diff
@@ -28,13 +33,16 @@ export function useFieldArray<
   const [fields, setField] = React.useState<
     WithFieldId<Partial<FormArrayValues>>[]
   >(mapIds(memoizedDefaultValues));
+  const shouldCheckDirty = readFormStateRef.current.dirty;
```
JeromeDeLeon: Instead of doing this, why not read it real-time?

```js
if (readFormStateRef.current.dirty) {
  // logic
}
```

JeromeDeLeon (changes requested):

src/useFieldArray.ts
bluebill1049: yea i am fixing it. you r right.

JeromeDeLeon:
Yes, it's working now. What was the cause of it? @bluebill1049

bluebill1049:
> Yes, it's working now. What was the cause of it? @bluebill1049

your comment above :)

src/useForm.ts
```diff
@@ -223,7 +223,7 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
   );
 
   const setDirty = (name: FieldName<FormValues>): boolean => {
```
JeromeDeLeon: Are we fine doing this? I mean, I just thought we're just setting a value here therefore, do not expect to return anything here. Better if we could rename to something like `verifyDirty` or something like that.

bluebill1049: you are right. the return is for deciding re-render or not. feel free to come up with a better name for it.

Issue #836, opened 2020-01-14
onBlur mode, revalidateMode "onChange": revalidation is not triggered with nested inputs


VGLoic:
**Describe the bug**
When mode is "onBlur", reValidateMode is "onChange" and have an input field with name "nested" like, e.g. "nested[0]". The revalidation mode onChange is not triggered.

**To Reproduce**
Use the codesandbox or
1. Set the mode to "onBlur", revalidateMode to "onChange", create a input with name "nested[0]" and validation rule (`required: true` for example)
2. trigger the error with an `onBlur` trigger
3. change the value with a valid one
4. error is still here (it should vanish at the `onBlur` though)

**Codesandbox link**
https://codesandbox.io/s/react-hook-form-rhfinput-dsd5f

**Expected behavior**
Error should disappear when changing the value (with a correct one with respect to the validation rule)

**Desktop (please complete the following information):**
 OS: iOS
 Browser Chrome
 Version 78.0.3904.97

**Additional context**
Keep on the great work :)

bluebill1049:
thanks for the issue report: react-hook-form@4.5.3-beta.1. let me know if this fix your issue.

VGLoic:
Looks great to me :) thanks a lot!

PR #863, opened 2020-01-19
onBlur mode, revalidateMode "onChange" revalidation fix for nested fields

cdagli:
Had the same issue with https://github.com/react-hook-form/react-hook-form/issues/836. I didn't have much time to debug but in my scenario I am hitting the one in `controller.tsx`, so just applied  the same solution (https://github.com/react-hook-form/react-hook-form/commit/f8699083a9ef478c67165d20405db349119abd6b) to the one in `controller.tsx` 

Search for `errors[name]` returns couple of other occurrences, not sure if they need the same thing though.


bluebill1049:
thanks ;)

PR #882, opened 2020-01-21
clean up test code for Controller and useFieldArray

kotarella1110:


src/controller.test.tsx
```diff
@@ -1,57 +1,7 @@
...
-        touched: false,
-        isSubmitting: false,
-        isValid: false,
-      },
-    },
-  };
-
-  return Object.assign({}, defaultControl, changedControl);
-}
+import reconfigureControl from './utils/test/reconfigureControl';
```
bluebill1049: nice clean up @kotarella1110 can we place this file inside useform.test.tsx and export it out? because we will export all `utils` folder functions. It's cleaner to keep utils folder just for the lib itself not the test.

kotarella1110: OK! I will.

bluebill1049 (approved):

PR #896, opened 2020-01-23
😭 Fix use field array

bluebill1049:
Sorry guys for not produce high quality custom hook with `useFieldArray`. I am going to take my time over the weekend to polish those issues around `useFieldArray`. If you are using it, i would recommend for this PR to be merged and published. Thanks for your patient, and feel free to help me and us on testing. (🙏🏻 Building field array uncontrolled is completely new to me as well, hence made few mistakes here and there)

- [x] remove mapID of each render
  - for native input you will have to use `register({})` to trigger ref callback
- [x] fix errors got removed after remove item
- [x] fix delete issue
- [x] dirty
- [x] async reset dirty

### DIFF
--- a/app/src/useFieldArray.tsx
+++ b/app/src/useFieldArray.tsx
@@ -1,5 +1,10 @@
 import React from 'react';
-import { useForm, useFieldArray, Controller } from 'react-hook-form';
+import {
+  useForm,
+  useFieldArray,
+  Controller,
+  ErrorMessage,
+} from 'react-hook-form';
 
 let renderCount = 0;
 
@@ -10,6 +15,7 @@ const UseFieldArray: React.FC = (props: any) => {
     register,
     formState: { dirty },
     reset,
+    errors,
   } = useForm<{
     data: { name: string }[];
   }>({
@@ -45,10 +51,12 @@ const UseFieldArray: React.FC = (props: any) => {
           data: [{ name: 'test' }, { name: 'test1' }, { name: 'test2' }],
         });
       }
-    }, 100);
+    }, 10);
   }, []);
 
   renderCount++;
+
+  console.log(errors);
 
   return (
     <form onSubmit={handleSubmit(onSubmit)}>
@@ -61,17 +69,23 @@ const UseFieldArray: React.FC = (props: any) => {
                 name={`data[${index}].name`}
                 defaultValue={data.name}
                 data-order={index}
-                ref={register({ required: true })}
+                ref={register({ required: 'This is required' })}
               />
             ) : (
               <Controller
                 as={<input id={`field${index}`} />}
                 control={control}
+                rules={{
+                  required: 'This is required',
+                }}
                 name={`data[${index}].name`}
                 defaultValue={data.name}
                 data-order={index}
               />
             )}
+            <ErrorMessage errors={errors} name={`data[${index}].name`}>
+              {({ message }) => <p id={`error${index}`}>{message}</p>}
+            </ErrorMessage>
             <button id={`delete${index}`} onClick={() => remove(index)}>
               Delete
             </button>
--- a/app/src/useFieldArray.tsx
+++ b/app/src/useFieldArray.tsx
@@ -44,7 +44,7 @@ const UseFieldArray: React.FC = (props: any) => {
                 name={`data[${index}].name`}
                 defaultValue={data.name}
                 data-order={index}
-                ref={register}
+                ref={register({})}
               />
             ) : (
               <Controller
--- a/app/src/useFieldArray.tsx
+++ b/app/src/useFieldArray.tsx
@@ -4,7 +4,13 @@ import { useForm, useFieldArray, Controller } from 'react-hook-form';
 let renderCount = 0;
 
 const UseFieldArray: React.FC = (props: any) => {
-  const { control, handleSubmit, register } = useForm<{
+  const {
+    control,
+    handleSubmit,
+    register,
+    formState: { dirty },
+    reset,
+  } = useForm<{
     data: { name: string }[];
   }>({
     ...(props.match.params.mode === 'default'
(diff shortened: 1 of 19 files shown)


bluebill1049:
Early beta version: 4.6.3-beta.2 (Please help testing) ❤️

app/src/useFieldArray.tsx
```diff
@@ -44,7 +44,7 @@ const UseFieldArray: React.FC = (props: any) => {
                 name={`data[${index}].name`}
                 defaultValue={data.name}
                 data-order={index}
-                ref={register}
+                ref={register({})}
```
kotarella1110: Why do you need an empty object?

src/logic/getFieldValueByName.ts
```diff
@@ -0,0 +1,7 @@
+import transformToNestObject from './transformToNestObject';
+import getFieldsValues from './getFieldsValues';
+
+export default <T, K extends keyof T>(fields: T, name?: K) => {
+  const results = transformToNestObject(getFieldsValues(fields));
+  return name ? results[name] : results;
```
kotarella1110: dot syntax not supported for name.

kotarella1110 (commented):
@bluebill1049 Can I commit to this branch to support dot syntax for name?

app/src/useFieldArray.tsx
bluebill1049: for native input you will have to use register({}) to trigger ref callback during map.

```
eg: ref={(e) => { console.log('e') }}
```
will alway print the ref

```
const test= (e) => console.log(e);
ref={test}
```

will print the ref, and unmount will return null

src/logic/getFieldValueByName.ts
bluebill1049: i guess it's safer to do `get(results, name)`?

app/src/useFieldArray.tsx
bluebill1049: also i will update the doc for this, and this only apply to `useFieldArray`

src/logic/getFieldValueByName.ts
kotarella1110: yes! that's what I think.

app/src/useFieldArray.tsx
kotarella1110: I'll look into this!

src/logic/getFieldValueByName.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  },
+  'nested.test[2].what[1].test': {
+    ref: {
+      name: 'nested.test[2].what[1].test',
+      value: 'nested.test[2].what[1].test',
+    },
+  },
+};
+
+describe('getFieldValueByName', () => {
```
bluebill1049: ❤️

bluebill1049:
latest beta: 4.7.0-beta.1

app/src/useFieldArray.tsx
kotarella1110: @bluebill1049 What is the bug caused by `ref={register}`?

bluebill1049: each time you remove, prepend, swap, move requires re-render, hence each ref needs to be updated. 

do an example below:
```
const test = (e) => console.log(e);
{[1,2,3].slice(0, index).map(() => <input ref={test} />)}
```
make a button to remove array index at a time, check the log

bluebill1049: ideally if we can support `register()` is better than asking people to put `register({})`

kotarella1110: This problem can be solved by removing `useCallback`.

https://github.com/react-hook-form/react-hook-form/blob/5494803a59a1432ab2dd6ccbd6d574c95da833b0/src/useForm.ts#L1106

bluebill1049: yea good point, if we remove useCallback, then during useEffect user will get infinite loop

kotarella1110: @bluebill1049 Sorry. Can you give an example where an infinite loop is triggered?
The following is an example of removing useCallback.

https://codesandbox.io/s/react-hook-form-usefieldarray-612i2

bluebill1049: try this: 

```
useEffect(() => {
 register(xxx)
}, [register])
```

kotarella1110: Tried but no infinite loop :cry:

bluebill1049: really? even with useCallback?? wow

bluebill1049: it should run into infinite loop without useCallback with register

bluebill1049: FYI, it will lead to an issue like this one: https://github.com/react-hook-form/react-hook-form/issues/899

stramel: @bluebill1049 My developer just ran into this. Do we have an issue to track and fix this?

bluebill1049: @stramel did you try to use just `register()` instead of just `register`

stramel: Yeah, that's a fine workaround but is very unintuitive from a DX perspective

bluebill1049: yea true that, i have spend quite bit time on it. The problem is how ref is gets invoked when item mount and unmount. 

The following shows the diff

```
ref={() => {
  console.log('always called');
}}

const test = () => {
    console.log('only mount and unmount');
}

ref={test}
```

I haven't found a solution for it jus yet, that's why documented on the website.

stramel: Can we create an issue to track this? I might take a look into it further

bluebill1049: Yes sure 👍

Issue #854, opened 2020-01-16
Controller.onBlur should not overwrite value when in onBlur validation mode


IronSean:
**Describe the bug**
I'm running into an issue where an input I'm attempting to wrap (which is a DatePicker) returns a Date from OnChange, but a formatted String from OnBlur. react-hook-form is seemingly saving this String, then passing it back as the `value` which blows up the app as the PropType check fails extecting a Date.

I'm not sure who is truly at fault here. They shouldn't be returning a String on their OnBlur event (and I opened [this issue](https://github.com/telerik/kendo-react/issues/447) about it). But It also seems unintuitive that react-hook-form would be overwriting the saved values in it's controller on a OnBlur, even if it's attempting to re-run the validations then. OnChange seems like the more appropriate method to update the saved value.

**To Reproduce**
1. Wrap a @progress/kendo-react-dateinput DatePicker in a react-hook-library Controller
2. Set validation to OnBlur
3. Interact with the DateInput to trigger a blur event

**Codesandbox link**
https://codesandbox.io/s/broken-thunder-9f4m3

**Expected behavior**
I would expect OnChange to update the values in the Controller, and OnBlur to not modify those values, but just trigger the validations on the already saved values if validations are in OnBlur mode.

**Desktop (please complete the following information):**
 - OS: Windows (and CodeSandbox)
 - Browser Chrome
 - Version 79

**Additional context**
As mentioned, I opened an issue there as well, and the reproduction in the code sandbox is very straightforward.

bluebill1049:
Your codesandbox is broken, also have u tried overwrite on blur?

Hendrafoo:
@bluebill1049 take a look at my sandbox
https://codesandbox.io/s/react-hook-form-get-started-596q6
but this one is with Material UI Autocomplete

bluebill1049:
thanks @Hendrafoo ❤️

bluebill1049:
i have append validation to your example as well @Hendrafoo https://codesandbox.io/s/react-hook-form-get-started-47mhr

IronSean:
Yes, the codesandbox is broken, It wouldn't be reproducing the bug if it worked. The validation itself is not the issue. The issue is that during OnBlur, you are reading the text value of the input instead of the Date value of the input that OnChange returns, then passing it back. The issue is just that in OnBlur mode, the OnBlur action is updating the saved values in the controller then writing them back to the input. 

[...]

PR #902, opened 2020-01-24
fix(Controller): Invoke `onBlur` method when `mode` is `onBlur`.

avocadowastaken:
After #862 behavior of `onBlur` passed to `Controller` become inconsistent, it get called when `mode` is `onSubmit` or `onChange`, but stop working when it's `onBlur`.

Here is the example with broken behavior https://codesandbox.io/s/react-hook-form-rhfinput-rice1

### DIFF
--- a/src/controller.test.tsx
+++ b/src/controller.test.tsx
@@ -137,9 +137,7 @@ describe('Controller', () => {
 
   it('should invoke custom onBlur method', () => {
     const onBlur = jest.fn();
-    const setValue = jest.fn();
     const control = reconfigureControl({
-      setValue,
       mode: { isOnSubmit: false, isOnBlur: true },
     });
 
@@ -159,7 +157,6 @@ describe('Controller', () => {
       },
     });
 
-    expect(setValue).not.toBeCalled();
     expect(onBlur).toBeCalled();
   });
 
--- a/src/controller.test.tsx
+++ b/src/controller.test.tsx
@@ -135,6 +135,34 @@ describe('Controller', () => {
     expect(onChange).toBeCalled();
   });
 
+  it('should invoke custom onBlur method', () => {
+    const onBlur = jest.fn();
+    const setValue = jest.fn();
+    const control = reconfigureControl({
+      setValue,
+      mode: { isOnSubmit: false, isOnBlur: true },
+    });
+
+    const { getByPlaceholderText } = render(
+      <Controller
+        defaultValue=""
+        name="test"
+        as={<input placeholder="test" />}
+        onBlur={onBlur}
+        control={control}
+      />,
+    );
+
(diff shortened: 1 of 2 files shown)


bluebill1049:
hey @umidbekkarimov we actually removed onBlur by reason. is there any reason why you need `onBlur`? you should be able to read data from onChange instead of onBlur

umidbekkarimov:
@bluebill1049 I need `onBlur` sometimes on `input` itself, but now it become inconsistent, I've updated PR description.

bluebill1049:
> @bluebill1049 I need `onBlur` sometimes on `input` itself, but now it become inconsistent, I've updated PR description.

can you give some more detail? if you want to use `onBlur` you can attach on your external component.

bluebill1049:
original issue: https://github.com/react-hook-form/react-hook-form/issues/854

umidbekkarimov:
> can you give some more detail? if you want to use onBlur you can attach on your external component.

I've updated codesandbox:  https://codesandbox.io/s/react-hook-form-rhfinput-rice1
It's broken for external component `onBlur` too. It's happening because of this line https://github.com/react-hook-form/react-hook-form/blob/master/src/controller.tsx#L115. Controller overrides `onBlur` for an external component and it never gets called.

So there are to ways to fix it:
1. (A hacky one) Call `InnerComponent.props[onBlurName]` here https://github.com/react-hook-form/react-hook-form/blob/master/src/controller.tsx#L115, which will work only if user passes `JSXElement`.
2. Call `onBlur` passed to `Controller` (as I did it in pr.)


> original issue: #854

I had these problems too and fixed that on my side https://github.com/superdispatch/ui/commit/9a75a880b36bd2e4e1429fd80be5ed47515dbb1e

bluebill1049:
my question was why not put onBlur inside your actual component

`<Controller as={<Test onBlur={xxx}} />`

umidbekkarimov:
Please check the code sandbox example once more, it should call alert on blur, there are two alerts passed one to controller and second to text field, none of them are called when the mode is onBlur.

bluebill1049:
> Please check the code sandbox example once more, it should call alert on blur, there are two alerts passed one to controller and second to text field, none of them are called when the mode is onBlur.
> […](#)
> On Jan 25, 2020, at 10:15, Bill ***@***.***> wrote: ﻿ my question was why not put onBlur inside your actual component <Controller as={<Test onBlur={xxx}} /> — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

you are right, i will take a look on this soon.

bluebill1049:
you are right. this is due to our onBlur overwrite children props. thanks for the fix. merging.

PR #916, opened 2020-01-27
support register method with optional first argument for useFieldArray hook

kotarella1110:
Support not only `ref={register({})}` but also `ref={register()}`.

related: https://github.com/react-hook-form/react-hook-form/pull/896


bluebill1049:
thanks @kotarella1110 awesome stuff! i will give a test tonight on this.

PR #939, opened 2020-01-31
close #938 performance enhancement for watch

### LINKED ISSUE #938: Improve watch performance
I've noticed that when we use `watch('field')` it gets the values for all fields in the form before it can actually watch for it.
This is problematic when using `useFieldArray` which could potentially have lots of rows that can be watched.

image

here is what the current code looks like: 
```
    function watch(fieldNames, defaultValue) {
        const combinedDefaultValues = isUndefined(defaultValue)
            ? isUndefined(defaultValuesRef.current)
                ? {}
                : defaultValuesRef.current
            : defaultValue;
        // get values of every single field in the form.
        const fieldValues = getFieldsValues(fieldsRef.current);
```
one solution could be to only get the values of the fields that needs to watched.
```
        const fieldValues = getFieldsValues(fieldsRef.current, fieldNames);
```
and then update `getFieldsValues` to something like:

```
var getFieldsValues = (fields, filter) => {
    const values = {};
    Object.values(fields).forEach(({ ref, ref: { name } }) => {
        if (!filter || filter.indexOf(name) > -1) {
            values[name] = getFieldValue(fields, ref);
        }
    });
    return values;
}
```
I'm not sure what other things this could impact but I've seem rendering time dropping from 114ms to 42ms with this change.

bluebill1049:


bluebill1049:
@kotarella1110 if you have sometimes, please help on improve TS :)

kotarella1110:
I'll take a look later!

src/logic/getFieldsValues.ts
```diff
@@ -1,13 +1,19 @@
...
-  (Object.values(fields) as Field[]).reduce(
-    (previous, { ref, ref: { name } }) => ({
-      ...previous,
-      ...{ [name]: getFieldValue(fields, ref) },
-    }),
-    {} as FormValues,
-  );
+  search?: any,
+) => {
+  const output: Record<string, FormValues> = {};
```
kotarella1110: how is this?

```ts
const output: Partial<Record<FieldName<FormValues>, FormValues>> = {};
```

src/logic/getFieldsValues.ts
```diff
@@ -1,13 +1,19 @@
...
+  const isSearchString = isString(search);
+
+  for (const name in fields) {
+    if (!isSearchString || (isSearchString && name.startsWith(search))) {
+      output[name] = getFieldValue(fields, (fields[name] || {}).ref);
```
kotarella1110:
```ts
output[name as FieldName<FormValues>] = getFieldValue(fields, (fields[name] || {}).ref);
```

kotarella1110:
By the way, the `FieldName` type is effectively a `string` type.

```ts
export type FieldName<FormValues extends FieldValues> =
  | (keyof FormValues & string)
  | string
// => string
```

Is this a problem if fixed as follows?

```ts
export type FieldName = string
```

bluebill1049:
> By the way, the `FieldName` type is effectively a `string` type.
> 
> ```ts
> export type FieldName<FormValues extends FieldValues> =
>   | (keyof FormValues & string)
>   | string
> // => string
> ```
> 
> Is this a problem if fixed as follows?
> 
> ```ts
> export type FieldName = string
> ```

This is more for a reference point of view.

src/logic/getFieldsValues.ts
bluebill1049: sounds good :)

src/logic/getFieldsValues.ts
bluebill1049: sounds good :)

bluebill1049:
please commit your suggested change :)

kotarella1110:
Changing the `FieldName` type to `string` simplifies the type.

`type FieldName<FormValues extends FieldValues> = (keyof FormValues & string) | string`:

スクリーンショット 2020-01-31 16 34 39

`type FieldName = string`:

スクリーンショット 2020-01-31 16 35 27

bluebill1049:
> Changing the `FieldName` type to `string` simplifies the type.
> 
> `type FieldName<FormValues extends FieldValues> = (keyof FormValues & string) | string`:
> 
> スクリーンショット 2020-01-31 16 34 39
> 
> `type FieldName = string`:
> 
> スクリーンショット 2020-01-31 16 35 27

ok if that's the case, let's go with it then.

kotarella1110 (approved):

bluebill1049:
@kotarella1110 you want to make the change to `string`? then we can merge this PR

bluebill1049:
actually let's merge this PR, you can create a separate PR.

kotarella1110:
@bluebill1049 I want to it! But changing the `FieldName` type requires us to fix it in large scope.
Let's merge this once. After that, I create a new PR.

bluebill1049:
yea take your time. love to have some TS upgrade from you ❤️

Issue #924, opened 2020-01-28
Unexpected fields values with useFieldArray when watch fields


ryanwalters:
**Describe the bug**
I'm not 100% sure that this is a bug, but I know that what is happening is not what I expected.

When setting `defaultValues` on an array with `useForm`, I would expect the result `fields` array using `useFieldArray` to contain those values as-is rather than becoming mangled.

**Example:**
```js
// In our case, the default values match a GraphQL schema so
// we can't necessarily change the shape

const { control } = useForm({
  defaultValues: {
    sources: ['abc.com', 'def.com']
  }
});
const { fields } = useFieldArray({ control, name: 'sources' });

console.log('fields', fields); 

// Expected output
/**
[{
  "value": "abc.com",
  "id": "5f0137bf-e201-425e-a7b0-f6cf76e0b3ca"
},
{
  "value": "def.com",
  "id": "5a93295c-7d9e-4762-a003-52406a611882"
}]
*/

// Actual output:
/**
[{
  "0": "a",
  "1": "b",
  "2": "c",
  "3": ".",
  "4": "c",
  "5": "o",
  "6": "m",
  "id": "5f0137bf-e201-425e-a7b0-f6cf76e0b3ca"
},
{
  "0": "d",
  "1": "e",
  "2": "f",
  "3": ".",
  "4": "c",
  "5": "o",
  "6": "m",
  "id": "5a93295c-7d9e-4762-a003-52406a611882"
}]
*/ 
```

This issue then snowballs into other issues when trying to display, append, delete, etc the values of the fields, but I figured I'd start here and step through the other issues only if needed.

**To Reproduce**
1. Go to https://codesandbox.io/s/upbeat-bose-pt26m
1. Look at console
1. See issue described above

**Codesandbox link**
https://codesandbox.io/s/upbeat-bose-pt26m

**Expected behavior**
See `Expected output` in the code snippet above.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 79.0.3945.130

bluebill1049:
I think a lot of people having this concern and question: https://github.com/react-hook-form/react-hook-form/issues/919

i will update the website soon, i think you are expecting the form to be controlled, but indeed it's not, watch() API is not same as state, watch is for watching input change after input rendered.

ryanwalters:
[...]

PR #940, opened 2020-01-31
assignId: Assign non-object values to value property

ryanwalters:
Possible solution to #924.

### DIFF
--- a/src/logic/mapIds.ts
+++ b/src/logic/mapIds.ts
@@ -10,7 +10,7 @@ export const appendId = <
 >(
   value: FormArrayValues,
 ) => ({
-  ...(isObject(value) ? { ...value } : { value }),
+  ...(isObject(value) ? value : { value }),
   id: generateId(),
 });
 
--- a/src/logic/mapIds.ts
+++ b/src/logic/mapIds.ts
@@ -1,4 +1,5 @@
 import isArray from '../utils/isArray';
+import isObject from '../utils/isObject';
 import generateId from './generateId';
 import { FieldValues } from '../types';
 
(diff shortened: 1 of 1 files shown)


src/logic/mapIds.ts
```diff
@@ -9,7 +10,7 @@ export const appendId = <
 >(
   value: FormArrayValues,
 ) => ({
-  ...value,
+  ...(isObject(value) ? { ...value } : { value }),
```
bluebill1049: i guess this will still return
```
{
 id: 'xxx',
 value: 'xxx'
}
```
you cool with that?

bluebill1049:
```suggestion
  ...(isObject(value) ? value : { value }),
```

bluebill1049: i mean with `fields` object

ryanwalters: Yep! It keeps the shape consistent for non-object values, which is perfectly acceptable to me.

bluebill1049: the improvement is you can do append('test') instead of  append({ value: 'test' }), also hard code the data as `value` probably not the best thing as well.

bluebill1049: i guess documentation probably can help on this.

ryanwalters: The issue is not with appending, it's with setting `defaultValues`. Currently, string values get spread, which creates a crazy looking object like:
```js
{
  "0": "a",
  "1": "b",
  "2": "c",
  "3": ".",
  "4": "c",
  "5": "o",
  "6": "m",
  "id": "5f0137bf-e201-425e-a7b0-f6cf76e0b3ca"
}
```

bluebill1049: oh crap.

bluebill1049: merge

PR #946, opened 2020-01-31
improve #938 watch perf with search array

bluebill1049:


src/logic/getFieldsValues.ts
```diff
@@ -1,19 +1,29 @@
...
-import { FieldName, FieldValues, FieldRefs } from '../types';
 import isString from '../utils/isString';
+import isArray from '../utils/isArray';
+import isUndefined from '../utils/isUndefined';
+import { FieldName, FieldValues, FieldRefs } from '../types';
 
 export default <FormValues extends FieldValues>(
   fields: FieldRefs<FormValues>,
-  search?: any,
+  search?: unknown,
```
cdagli: @bluebill1049 Possibly we have a bug here. When you use `watch({nest: true})` search becomes an object here: https://github.com/react-hook-form/react-hook-form/blob/master/src/useForm.ts#L650

bluebill1049: good pick up, i am doing type check on fieldNames on array and string only. let me know if you actually encounter issues with it.

cdagli: Please check these 2: 
4.7.3 (works) => https://codesandbox.io/s/react-hook-form-473-next0-watchnesttrue-iw6xs
4.8.1 (fails) => https://codesandbox.io/s/react-hook-form-481-watchnesttrue-iw6xs 

image

bluebill1049: oh. you are right thanks a lot

bluebill1049: i will do a patch right away, you want to take a quick look the hot fix?

bluebill1049: https://github.com/react-hook-form/react-hook-form/commit/171ab4fe88f5d088a59cf3bc0d18b2be4a8c40a1
i will follow up with unit test and autoamtion afterward. hopefully, fewer ppl get affected by this issue.

cdagli: Sorry, I missed your previous message but now it works. Thanks!

PR #947, opened 2020-02-01
include dirty fields form state

bluebill1049:


maximeyuxihuang:
Hello, I don't really know where to ask this.
My question is: Is the dirtyFields field in the formState ready to be used ?
I see that it's missing in the doc but still shown  in the codesandbox example provided (but the field is not updated apparently) : https://codesandbox.io/s/7o2wrp86k6
Is there any main thread about this ?

PR #952, opened 2020-02-02
Remove deprecated field from options type.

fubhy:


fubhy:
The `SchemaValidateOptions` type is still used statically in the validate() call so I didn't remove that.

bluebill1049:
Thanks mate :)

PR #959, opened 2020-02-03
Add a dependency for rules on registration field

hekkim:
# What I want
I'm trying to create some input fields conditionally required, but `registerField()` is not called when changing `rules` props in `<Controller />`.

## Solution
I just added a dependency for `rules` on `useEffect` in `Controller`.
In that case, it re-register input field again when `rules` is changed.

## Codesandbox
https://codesandbox.io/s/patient-snow-s0dcl?fontsize=14&hidenavigation=1&theme=dark

## Notes
Also, updates a bit of tests in cypress, but it looks like there're some severe problems in `onBlur` and `onChange` mode while I'm working on it.
Could you guys give me any advice how to resolve it?

### DIFF
--- a/app/src/controller.tsx
+++ b/app/src/controller.tsx
@@ -148,6 +148,7 @@ export default function Field(props: any) {
       <span id="renderCount">{renderCount}</span>
 
       <button
+        id="reset"
         type="button"
         onClick={() => {
           reset({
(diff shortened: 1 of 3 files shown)


hekkim:
Let me take a look why it fails to pass cypress test. 😢

bluebill1049:
Hi there, thanks for the PR, I did leave out rules by purpose. because rules is likely to be an object, with `Object.is` check in the dep, it will likely to running into a loop without memo. so so alternative, it's probably easier to just run `unregister` and change your `rules`

hekkim:
> Hi there, thanks for the PR, I did leave out rules by purpose. because rules is likely to be an object, with `Object.is` check in the dep, it will likely to running into a loop without memo. so so alternative, it's probably easier to just run `unregister` and change your `rules`

Gotcha. Thanks for your advice.

PR #965, opened 2020-02-04
Example: form using getValues()

Geenzo:
### Reason for PR
Noticed there was no usecase of the `getValues()` method in the examples folder as far as i can see when looking for one. Created one myself incase anyone in future could use this, for e,g password confirmation fields etc.

### Live Example
https://codesandbox.io/s/react-hook-form-getvalues-compare-field-values-qrtvx (added to examples readme)


bluebill1049:
awesome! thanks very much.

PR #972, opened 2020-02-05
 Change `onSubmit` event to` onChange`

ghost:
The example doesn't work because the wrong event is applied.

### DIFF
--- a/examples/validationOnFieldChange.tsx
+++ b/examples/validationOnFieldChange.tsx
@@ -12,7 +12,7 @@ export default function App() {
 
   return (
     <div className="App">
-      <form onSubmit={handleSubmit(onSubmit)}>
+      <form onChange={handleSubmit(onSubmit)}>
         <div>
           <label htmlFor="firstName">First Name</label>
           <input


bluebill1049:
thanks for the PR, however, i think onSubmit is correct. do you have a codesandbox to produce that it's not working.

siuxco:
Hi Bill! how are you?.

I understand that the event should trigger as soon as the input value is modified and not when the form is submitted (enter).

**Here is the example**
Try changing onSubmit to onChange
[CodeSanbox](https://codesandbox.io/s/validationonfieldchange-il1wh)

Thanks!

bluebill1049:
Hey @siuxco :)

I am good. thanks ❤️ 

if we change to `onChange`, the form will submit on each keyStroke. is that the behavior you want?

bill

siuxco:
> if we change to `onChange`, the form will submit on each keyStroke. is that the behavior you want?

I was trying to do an autocomplete and get the current value of the input. I interpreted that the demo was about that. Now that you mention it, I understand that the approach is only in the validation and not in obtaining the value as the user writes.

This is what I was looking for:
https://react-hook-form.com/api/#watch

Thank you very much for your time and your patience Bill and keep up the good work!

bluebill1049:
aha no worries @siuxco thanks for contributing tho! looking forwards for more PRs from you.

cheers
bill

PR #974, opened 2020-02-05
Feature/support custom schema validation (new)

bluebill1049:
This feature request is waiting for more upvote. 

related PR: https://github.com/react-hook-form/react-hook-form-website/pull/139

if you want such feature, you can thumb up this comment. 👍 

------

example for `@hapi/joi`, this will apply for any other schema/custom validation.

```
import React from "react";
import useForm from "react-hook-form";
import joi from "@hapi/joi";

const schema = Joi.object({
  username: Joi.string().alphanum().min(3).max(30).required()
});

// Build your resolver to display errors and pass valid values
const resolver = (data: any) => {
  const { error, value: values } = validationSchema.validate(data, {
    abortEarly: false,
  });
  return {
    values: error ? {} : values,
    errors: error
      ? error.details.reduce((previous, currentError) => {
          return {
            ...previous,
            [currentError.path[0]]: currentError,
          };
        }, {})
      : {},
  };
};

function App() {
  const { register, handleSubmit, errors } = useForm({
    validationResolver: resolver
   });
  
  return (
      <input type="text" name="username" ref={register} />
    </form>
  );
}
```

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -44,23 +44,24 @@ export type SchemaValidateOptions = Partial<{
   abortEarly: boolean;
   stripUnknown: boolean;
   recursive: boolean;
-  context: object;
+  context: any;
 }>;
 
-export type ValidationResolver = <FormValues>(
+export type ValidationResolver = <FormValues, ValidationContext>(
   values: FieldValues,
-  validationContext: object,
+  validationContext: ValidationContext,
 ) => { values: FieldValues; errors: FieldErrors<FormValues> };
 
 export type UseFormOptions<
-  FormValues extends FieldValues = FieldValues
+  FormValues extends FieldValues = FieldValues,
+  ValidationContext = undefined
 > = Partial<{
   mode: Mode;
   reValidateMode: Mode;
   defaultValues: DeepPartial<FormValues>;
   validationSchema: any;
   validationResolver: ValidationResolver;
-  validationContext: object;
+  validationContext: ValidationContext;
   submitFocusError: boolean;
   validateCriteriaMode: 'firstError' | 'all';
 }>;
--- a/src/types.ts
+++ b/src/types.ts
@@ -50,7 +50,7 @@ export type SchemaValidateOptions = Partial<{
 export type ValidationResolver = <FormValues, ValidationContext>(
   values: FieldValues,
   validationContext: ValidationContext,
-) => { values: FieldValues; errors: FieldErrors<FormValues> };
+) => { values: FieldValues | {}; errors: FieldErrors<FormValues> | {} };
 
 export type UseFormOptions<
   FormValues extends FieldValues = FieldValues,
--- a/src/types.ts
+++ b/src/types.ts
@@ -54,6 +54,11 @@ export type UseFormOptions<
   reValidateMode: Mode;
   defaultValues: DeepPartial<FormValues>;
   validationSchema: any;
+  validationResolver: (
+    values: FieldValues,
+    context: any,
+  ) => { values: FieldValues; errors: FieldErrors<FormValues> };
+  validationContext: object;
   submitFocusError: boolean;
   validateCriteriaMode: 'firstError' | 'all';
 }>;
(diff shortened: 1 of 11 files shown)


src/types.ts
```diff
@@ -54,6 +54,11 @@ export type UseFormOptions<
   reValidateMode: Mode;
   defaultValues: DeepPartial<FormValues>;
   validationSchema: any;
+  validationResolver: (
+    values: FieldValues,
+    context: any,
+  ) => { values: FieldValues; errors: FieldErrors<FormValues> };
+  validationContext: object;
```
fubhy: The context could really be anything. Even a string or boolean. Maybe we need another generic here?

```
useForm<TFormValue, TFormContext>()
```

fubhy: Oh and do you plan to also forward the `context` value to the submit handler? Or rather not?

src/useForm.ts
```diff
@@ -436,6 +442,8 @@ export function useForm<FormValues extends FieldValues = FieldValues>({
         if (validationSchema) {
           const { errors } = await validateWithSchema<FormValues>(
```
fubhy: You could log a note here in development mode saying that `validationSchema` is deprecated so you can more easily remove it in the future. I think the `validationResolver` pattern is more flexible and you might want to eventually just ship with a single pattern, no?

fubhy:
Really looking forward to this. It's going to make it a whole lot easier to build for more sophisticated validation requirements.

src/types.ts
bluebill1049: > The context could really be anything. Even a string or boolean. Maybe we need another generic here?
> 
> ```
> useForm<TFormValue, TFormContext>()
> ```

prefer to keep it as `object` and align with what normally `context` type is

bluebill1049: > Oh and do you plan to also forward the `context` value to the submit handler? Or rather not?

probably no, but will pass to yup context for now.

src/useForm.ts
bluebill1049: it can be annoying to the user, I will update doc to reflect that, but yes next major we probably going to remove it.

src/types.ts
fubhy: > > The context could really be anything. Even a string or boolean. Maybe we need another generic here?
> > ```
> > useForm<TFormValue, TFormContext>()
> > ```
> 
> prefer to keep it as `object` and align with what normally `context` type is

In that case, maybe use a generic and default it to `object`? `useForm<TFormValue, TFormContext = {}>`

That way users can be more strict about their types if they want to.

> probably no, but will pass to yup context for now.

I was just thinking that maybe if there are asynchronous handlers in the context (like a rest api handler / websocket connection / web3 / ethereum client / etc. it might be equally important for the validation and the submission. Is there a technical argument that would prevent it from being passed to the submission handler? I am considering this `context` as moire of a `form context` (instead of just for validation). Sure, in most cases it would probably be most useful for validation but it is not limited to it.

bluebill1049: > In that case, maybe use a generic and default it to `object`? `useForm<TFormValue, TFormContext = {}>`

sounds good to me, will adjust tmr

src/useForm.ts
hemedani: I think it's better to use [superstruct](https://github.com/ianstormtaylor/superstruct) instead of [joi](https://github.com/hapijs/joi) in your test and example.
superstruct wrote in typescript and it's very light lib. see bundlephobia details :
[joi](https://bundlephobia.com/result?p=joi@14.3.1)
[yup](https://bundlephobia.com/result?p=yup@0.28.1)
[superstruct](https://bundlephobia.com/result?p=superstruct@0.8.3)

bluebill1049: once we release a beta version, can you please supply a superstruct version? we can include in the code example & website.

bluebill1049:
guys if you watching this PR, please help out testing: 

`react-hook-form@4.8.3-next.0`

src/useForm.ts
hemedani: yes sure

PR #976, opened 2020-02-05
Allow the field value to be transformed before passing it to the controlled component

fubhy:
I was trying to integrate a react-select component when I ran across the problem with the different `value` format (`{ value: "foo", label: "bar" }`) for react-select.

I thought it would make sense to allow the selected value to be transformed before passing it to the controlled component instead of forcing the structure of the `form values` to be changed to accommodate react-select.

That way, one can continue using the form values / fields however one wants to structure them and not have it dictated by react-select.

### DIFF
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -17,6 +17,7 @@ const Controller = <ControlProp extends Control = Control>({
   onChange,
   onChangeName = VALIDATION_MODE.onChange,
   onBlurName = VALIDATION_MODE.onBlur,
+  valueTransformer,
   valueName,
   defaultValue,
   control,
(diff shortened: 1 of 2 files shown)


src/controller.tsx
```diff
@@ -132,7 +133,11 @@ const Controller = <ControlProp extends Control = Control>({
           },
         }
       : {}),
-    ...{ [valueName || (isCheckboxInput ? 'checked' : VALUE)]: value },
+    ...{
+      [valueName || (isCheckboxInput ? 'checked' : VALUE)]: valueTransformer
+        ? valueTransformer(value)
```
JeromeDeLeon: It would be better to name this as `transformValue`

bluebill1049:
First of all, thanks a lot for this PR 🙏 

I think we can resolve this at `onChange`.  Prefer to keep the component itself simple and fewer props.

bluebill1049:
closing this PR for now.

fubhy:
Solving this on onChange(), however, means that it affects the structure of the form values which is not desirable. Doing it when passing it to the underlying component is inexpensive and allows us to control our form values structure without weird data structures from `react-select` or others leaking into it.

PR #983, opened 2020-02-06
improve type keyName for useFieldArray

kotarella1110:
```ts
useFieldArray<{ name: string }>()({
  control,
  name: "data"
});
// const fields: Partial<{
//     name: string;
//     id: string
// }>[]
```

```ts
const { fields } = useFieldArray<{ name: string }, "customId">()({
  keyName: "customId",
  control,
  name: "data"
});
// const fields: Partial<{
//     name: string;
//     customId: string
// }>[]
```

related: https://github.com/react-hook-form/react-hook-form/pull/957

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -13,14 +13,25 @@ import prependAt from './utils/prepend';
 import isArray from './utils/isArray';
 import insertAt from './utils/insert';
 import fillEmptyArray from './utils/fillEmptyArray';
-import { FieldValues, Control, UseFieldArrayProps, Field } from './types';
+import {
+  Field,
+  FieldValues,
+  UseFieldArrayProps,
+  Control,
+  ArrayField,
+} from './types';
 
 const { useEffect, useRef, useState } = React;
 
-export function useFieldArray<
+export const useFieldArray = <
   FormArrayValues extends FieldValues = FieldValues,
+  KeyName extends string = 'id',
   ControlProp extends Control = Control
->({ control, name, keyName = 'id' }: UseFieldArrayProps<ControlProp>) {
+>({
+  control,
+  name,
+  keyName = 'id' as KeyName,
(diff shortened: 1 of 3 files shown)


src/useFieldArray.ts
```diff
@@ -13,14 +13,25 @@ import prependAt from './utils/prepend';
...
+export const useFieldArray = <
   FormArrayValues extends FieldValues = FieldValues,
+  KeyName extends string = 'id',
   ControlProp extends Control = Control
->({ control, name, keyName = 'id' }: UseFieldArrayProps<ControlProp>) {
+>({
+  control,
+  name,
+  keyName = 'id' as KeyName,
+}: UseFieldArrayProps<KeyName, ControlProp>) => {
```
bluebill1049: what about `ControlProp` as first argument?

kotarella1110: Is this correct for this?

```ts
UseFieldArrayProps<ControlProp, KeyName>
```

bluebill1049: so it's not a breaking change for ts

kotarella1110: > so it's not a breaking change for ts

https://github.com/react-hook-form/react-hook-form/blob/ce68ee457a2fd7252c8099eef797cc85cdb232e3/src/useFieldArray.ts#L26-L30

Is it necessary to change the order `ControlProp` and `KeyName` of `useFieldArray` as follows?

```
 export const useFieldArray = < 
   FormArrayValues extends FieldValues = FieldValues, 
   ControlProp extends Control = Control 
   KeyName extends string = 'id', 
 >({
```

bluebill1049: yea, i think keep KeyName as last one would be nicer.

bluebill1049: actually... probably what you had is correct...

bluebill1049 (approved):

PR #991, opened 2020-02-07
fix #988 isValid form state with useFieldArray schema

### LINKED ISSUE #988: useFieldArray hook Yup array validation not invalidating Form
**Describe the bug**
When using the useFieldArray hook alongside with Yup array validation the form is always valid although the validation specifies otherwise.

**To Reproduce**
1. Go to the codesandbox link (https://codesandbox.io/s/react-hook-form-usefieldarray-uhoxs)
2. Toggle the console
3. Delete the initial Field
4. The form is now valid when Yup validation specifies min of 1 record in the array

**Codesandbox link**
https://codesandbox.io/s/react-hook-form-usefieldarray-uhoxs

**Expected behavior**
uppon building a validation such as 
```
const formSchema = {
  name: yup.string().required("form.required_message")
};

const fieldsSchema = yup.object().shape({
  test: yup
    .array()
    .of(yup.object().shape(formSchema))
    .required("Must have fields")
    .min(1, "Minimum of 1 field")
});
```
If the form doesn't have at least one object inside test array it should return `formState.isValid = false`

**Screenshots**
image
image

 - OS: MacOs
 - Browser chrome
 - Version 79.0.3945.130
- react-hook-form version: v4.8.2

**Additional context**
Im not sure why this is happening, as you can see im using the combination of  `mode: "onChange"` in the `useForm` hook to actually trigger the validation. Maybe I'm using it wrongly or even constructing the yup badly, but it seemed a bug in the yup validation functionality. Let me know if there is a way or if im actually setting it up wrongly, help would be much appreciated. Thank you.
Keep the nice work :)

bluebill1049:


bluebill1049:
https://codesandbox.io/s/react-hook-form-usefieldarray-c82kl

PR #1054, opened 2020-02-18
Update useFormContext.tsx

ccnokes:
Fail fast if `useFormContext` is used outside of a FormContext. This would have helped me earlier today when I was learning this library. Thanks!

### DIFF
--- a/src/useFormContext.tsx
+++ b/src/useFormContext.tsx
@@ -8,7 +8,7 @@ const FormGlobalContext = React.createContext<FormContextValues<
 
 export function useFormContext<T extends FieldValues>(): FormContextValues<T> {
   const context = React.useContext(FormGlobalContext) as FormContextValues<T>;
-  if (context) return context;
+  if (context !== undefined) return context;
   throw new Error('Missing FormContext');
 }
 
--- a/src/useFormContext.tsx
+++ b/src/useFormContext.tsx
@@ -11,7 +11,7 @@ export function useFormContext<T extends FieldValues>(): FormContextValues<T> {
   if (context === undefined) {
     throw new Error('useFormContext must be used within a FormContext');
   }
-  return context;
+  return context || throw new Error('Missing FormContext');
 }
 
 export function FormContext<T extends FieldValues>({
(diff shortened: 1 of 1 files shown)


src/useFormContext.tsx
```diff
@@ -7,7 +7,11 @@ const FormGlobalContext = React.createContext<FormContextValues<
 > | null>(null);
 
 export function useFormContext<T extends FieldValues>(): FormContextValues<T> {
-  return React.useContext(FormGlobalContext) as FormContextValues<T>;
+  const context = React.useContext(FormGlobalContext) as FormContextValues<T>;
+  if (context === undefined) {
+    throw new Error('useFormContext must be used within a FormContext');
+  }
+  return context;
```
bluebill1049:
```suggestion
    return context || throw new Error('Missing FormContext');
```
Thanks a lot of the PR ❤️ 
I am shorten the message to reduce the bundle size :)

ccnokes:
Good call @bluebill1049, thanks for the review!

bluebill1049:
oops sorry i stuffed up your PR.
```
if (context) {
  return context;
}

throw new Error('xxx');
```

bluebill1049:
sorry i did broken your code.
```
if (context) return context;

throw new Error('xxx')
```

ccnokes:
@bluebill1049 no worries! I'll work on getting it fixed today during lunch.

src/useFormContext.tsx
```diff
@@ -8,7 +8,7 @@ const FormGlobalContext = React.createContext<FormContextValues<
 
 export function useFormContext<T extends FieldValues>(): FormContextValues<T> {
   const context = React.useContext(FormGlobalContext) as FormContextValues<T>;
-  if (context) return context;
+  if (context !== undefined) return context;
```
bluebill1049: you can use isUndefinedOrNull method from the lib :)

ccnokes: Actually, because the default context value for `FormGlobalContext` is `null`,  the check here should be against `undefined` only. Only when it's `undefined` does it mean that there was no context registered in the tree.

ccnokes: Oh, but I see that you have an `isUndefined`. I'll use that instead 😀

bluebill1049: awesome thanks :)

bluebill1049:
Awesome thanks @ccnokes

PR #1061, opened 2020-02-20
support passing object to required for register valdation rules

### LINKED ISSUE #1060: Form validation with required taking an object does not work
**Describe the bug**
When we use object inside register to make field required, its value is not taken into consideration,

**To Reproduce**
1. create a form
2. try changing bold value from true to false - this does not work

      <label>Example</label>
      <input name="example" defaultValue="test" ref={register} />
      <label>ExampleRequired</label>
      <input
        name="exampleRequired"
        ref={register({
          required: { value: **true**, message: "test" },
          maxLength: 10
        })}
      />
      {errors.exampleRequired && <p>This field is required</p>}
      <input type="submit" />
    </form>

**Codesandbox link**
https://codesandbox.io/s/react-hook-form-get-started-0hiz9

**Expected behavior**
value should work

**Screenshots**

 - OS: windows 10
 - Browser chrom]

kotarella1110:
codesandbox: https://codesandbox.io/s/react-hook-form-get-started-lh2nf
close https://github.com/react-hook-form/react-hook-form/issues/1060

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -76,7 +76,7 @@ export type ValidateResult = string | boolean | undefined;
 export type Validate = (data: any) => ValidateResult | Promise<ValidateResult>;
 
 export type ValidationOptions = Partial<{
-  required: boolean | string | ValidationOptionObject<boolean>;
+  required: string | ValidationOptionObject<boolean>;
   min: ValidationOptionObject<number | string>;
   max: ValidationOptionObject<number | string>;
   maxLength: ValidationOptionObject<number | string>;
--- a/src/logic/validateField.test.ts
+++ b/src/logic/validateField.test.ts
@@ -48,6 +48,16 @@ describe('validateField', () => {
     });
 
     expect(
+      await validateField({} as any, false, {
+        ref: { type: 'text', value: '', name: 'test', setCustomValidity },
+        required: {
+          value: false,
+          message: 'required',
+        },
+      }),
+    ).toEqual({});
+
+    expect(
       await validateField(
         {
           current: {
--- a/src/logic/validateField.ts
+++ b/src/logic/validateField.ts
@@ -82,18 +82,22 @@ export default async <FormValues extends FieldValues>(
       (isCheckBox && !getCheckboxValue(options).isValid) ||
       (isRadio && !getRadioValue(options).isValid))
   ) {
-    const message = isString(required)
-      ? required
-      : getValueAndMessage(required).message;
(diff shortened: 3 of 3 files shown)


src/types.ts
```diff
@@ -76,7 +76,7 @@ export type ValidateResult = string | boolean | undefined;
 export type Validate = (data: any) => ValidateResult | Promise<ValidateResult>;
 
 export type ValidationOptions = Partial<{
-  required: boolean | string | ValidationOptionObject<boolean>;
+  required: string | ValidationOptionObject<boolean>;
```
bluebill1049: we should still allow boolean here?

`register({ required: true })`

bluebill1049:
what about this

line 79: 

`isBoolean(required) ? required : !!required`

src/types.ts
kotarella1110: boolean is unnecessary because it is allready defined in ValidationOptionObject.

https://github.com/react-hook-form/react-hook-form/blob/4d26f8cb3c86b3d35a1694a201332556be20b71f/src/types.ts#L70

bluebill1049: you are right. thanks :)

kotarella1110:
@bluebill1049 plz wait for merge due to a mistake😅

bluebill1049:
@kotarella1110 i think we only have to update the if statement?

`isBoolean(required) ? required : !!required`

kotarella1110:
> @kotarella1110 i think we only have to update the if statement?
> 
> `isBoolean(required) ? required : !!required`

@bluebill1049 What line is it?

kotarella1110:
> @kotarella1110 i think we only have to update the if statement?
> 
> `isBoolean(required) ? required : !!required`

@bluebill1049 I don't think so.

スクリーンショット 2020-02-21 9 19 50

If required is object ( `{ value: boolean; message: string; }`), the following if statement is true.

```
`isBoolean(required) ? required : !!required`
```

bluebill1049:
> > @kotarella1110 i think we only have to update the if statement?
> > `isBoolean(required) ? required : !!required`
> 
> @bluebill1049 I don't think so.
> 
> スクリーンショット 2020-02-21 9 19 50
> 
> If required is object ( `{ value: boolean; message: string; }`), the following if statement is true.
> Therefore, the following code doesn't pass unit tests.
> 
> `isBoolean(required) ? required : !!required`

Sounds good to me, merge :)

Issue #1039, opened 2020-02-16
Bug report: Cannot resolve dependency 'ts-polyfill/es2015-symbol' since v4.9.4


krnlde:
**Describe the bug**
After updating to `4.9.4` I immediatelly get this error on console:
```
×  [project_root]\node_modules\react-hook-form\dist\react-hook-form.ie11.js:7:8: Cannot resolve dependency 'ts-polyfill/es2015-symbol'
   5 | require('ts-polyfill');
   6 | require('ts-polyfill/es2015-iterable');
>  7 | require('ts-polyfill/es2015-symbol');
     |        ^
   8 | require('ts-polyfill/es2015-symbol-wellknown');
   9 | var React = require('react');
  10 |
```

**Additional context**
In my project I use the ie11 version of react-hook-form. This issue is probably related to the latest PR #1018

bluebill1049:
@vova-bezhenar are you having a similar issue?

kotarella1110:
I think the following lines need to be fixed.

https://github.com/react-hook-form/react-hook-form/blob/master/src/index.ie11.ts#L1-L4

```ts
import 'ts-polyfill';
import 'ts-polyfill/lib/es2015-iterable';
import 'ts-polyfill/lib/es2015-symbol';
import 'ts-polyfill/lib/es2015-symbol-wellknown';
```

kotarella1110:
@vova-bezhenar By the way, why are you importing `ts-polyfill/lib/es2015-iterable` and `ts-polyfill/lib/es2015-symbol` and `ts-polyfill/lib/es2015-symbol-wellknown`? 
Is it no good importing only `ts-polyfill`?

> https://github.com/ryanelian/ts-polyfill
> <img width="903" alt="スクリーンショット 2020-02-17 10 11 13" src="https://user-images.githubusercontent.com/12913947/74616925-db415b00-516d-11ea-8e1d-5ae07982b5bb.png">

bluebill1049:
worst case, we may need to revert that commit and patch this issue.

kotarella1110:
> Also, we need to install [`ts-polyfill`](https://github.com/ryanelian/ts-polyfill) to use ie11 version of `react-hook-form`.
> If we want to bundle with `ts-polyfill`, we should remove it from `external` in rollup config.
> 
> https://github.com/react-hook-form/react-hook-form/blob/master/rollup.ie11.config.js#L14

[...]

PR #1063, opened 2020-02-20
Support IE11

nowells:
Our published IE11 dist should work out of the box with IE11 without requiring the consumers to polyfill themselves. This attempts to solve https://github.com/react-hook-form/react-hook-form/issues/1039


bluebill1049:
hey @nowells can you please do us a favour to do a test locally with the build?

vova-bezhenar:
I've checked this on my end. Works like a charm.

nowells:
I work with @vova-bezhenar and he confirms that it works. So this looks good from our side. Thanks for your consideration of this fix!

PR #1072, opened 2020-02-21
Enable minified build for github action

stramel:
This should align us better with what bundlesize was showing


bluebill1049:
You want to remove bundle size?

stramel:
bundlesize integration will have to be disabled for this repo.

PR #1073, opened 2020-02-21
Enable minified build for github action

stramel:
NOTE: Created from the repo instead of my fork.

This should align us better with what bundlesize was showing


stramel:
This change is expected due to the fact it wasn't previously reporting on it.

Issue #1043, opened 2020-02-17
Provide example with Masked Input + Custom Input.


dskiba:
Hello, first wanna thank you all guys for the react-hook-form, really interesting thing :) 

I am trying RHF with react-input-mask and custom input
 https://codesandbox.io/s/new-star-rzmfb

And there is one problem with a cursor on Safari(13). I have already created a pull request in react-input-mask. 

So if this issue with safari will be fixed, would you be interested in me to create a pull request with this example?

dskiba:
fixed this issue 💃
good example: https://codesandbox.io/s/focused-kapitsa-n3u6c  
I should use react setState (or hook version) to chage value instead of setValues.

Now it works well and I can create a pull request with the contribution to examples sections if you are interested in it. 🤓

bluebill1049:
That's awesome. would you like to spend a PR and add this example to our examples folder?

bluebill1049:
can we do a version with Controller as well?

dskiba:
Yeah, I will do a PR with this example and with a Controller version.

bluebill1049:
Awesome thanks 👍❤️

bluebill1049:
https://spectrum.chat/react-hook-form/help/how-to-do-parse-and-format~2c0224a1-f3df-4e8a-821b-ea5e51328a32

bluebill1049:
I am going to close this issue, feel free to send PR @dskiba when you got the controller version ready :)

PR #1088, opened 2020-02-22
Add React Number Format mask input example

michael-land:
Related to #1043


bluebill1049:
thank you :)

cargallo:
The example throws Function components cannot be given refs. at console

PR #1091, opened 2020-02-24
fix #1090

### LINKED ISSUE #1090: Add a way to skip Contorller#commonTask getInputValue() call

When use Controller with custom input, If the value object is {id:1, value:"ABC"}, controller will set next value to "ABC" because getInputValue will try to get obj.value first. Should it be {id:1, value:"ABC"}? 

https://github.com/react-hook-form/react-hook-form/blob/master/src/controller.tsx#L59
and 
https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/getInputValue.ts#L13

Maybe we should only call `getInputValue` if `value` is an `SyntheticEvent` object? Can we rely on constructor check?
```tsx
if (e.constructor.name === "SyntheticEvent" && e.type === "change") {
  console.log("input change event");
}
```

Or can we add an option to skip it?

michael-land:
fixes #1090 

I tried to leave logic inside getInputValue but then I have to pass the whole event object function and It will break tests. So I decide to check the event object before pass it down

### DIFF
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -112,7 +112,7 @@ const Controller = <ControlProp extends Control = Control>({
         unregister(name);
       }
     };
-  }, [fieldArrayNamesRef, name, registerField, unregister]);
+  }, [name]);
 
   const shouldReValidateOnBlur = isOnBlur || isReValidateOnBlur;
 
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks a lot for the PR, I think what i got is probably simpler and less code: https://github.com/react-hook-form/react-hook-form/pull/1092

xiaoyu-tamu:
close. see #1092

PR #1092, opened 2020-02-24
close #1090 controller get value

### LINKED ISSUE #1090: Add a way to skip Contorller#commonTask getInputValue() call

bluebill1049:


bluebill1049:
beta:

yarn add https://pkg.csb.dev/react-hook-form/react-hook-form/commit/b4f974b7/react-hook-form

xiaoyu-tamu:
I'm just wondering what if user provide object `{ target: { id: 1, value: 'a' } }` as value? Looks like I still get 'a'. 

This happened when people use Mateiral-UI Autocomplete component or component that store whole object internal. e.g. 
```
<Autocomplete    
      options={[{ target: { id: 1, value: 'a' } }]}
      getOptionLabel={o => o?.target?.value} />

```

bluebill1049:
aha i see, will make the adjustment. My goal is to keep simple and try to write as less code as possible.

xiaoyu-tamu:
Definitely

bluebill1049:
let's check `e.type` as well, can't do `instanceof Event` due to react-native.

xiaoyu-tamu:
> let's check `e.type` as well, can't do `instanceof Event` due to react-native.

Sound good, we may need to update doc for controller to let developer know that when the value is following shape, they may get unexpected result. 
```
{
    target: {
         type:any,
         value: any,
   }
}
```

bluebill1049:
Yeah totally!

PR #1112, opened 2020-02-27
update ControllerProps type to improve DX

kotarella1110:
Fixed so that auto-completion works when rest props is entered like `ErrorMessage` Component!
This improve DX.

スクリーンショット 2020-02-27 21 44 37


bluebill1049:
wow this is really good.

bluebill1049 (approved):

PR #1127, opened 2020-03-01
added example for Material UI/Yup/Dynamic Reload, Reset

tsaockham:
Hello, took me ages to figure this all out, so I thought I'd share.  The main issue I was having was updating the default values when the prop changed, that and dealing with the controlled and uncontrolled component issues.  Setting the key value on the Control component handled the issue, however this then confused the validator on reset, so I had to create a custom reset method. Hopefully this will be useful to somebody, or if there is a better way I'd like to know.


tsaockham:
Found a bug that I fixed. have another updaet

PR #1128, opened 2020-03-01
Added example for Material UI/Yup/Dynamic Reload, Reset

tsaockham:
Second try... Hello, took me ages to figure this all out, so I thought I'd share. The main issue I was having was updating the default values when the prop changed, that and dealing with the controlled and uncontrolled component issues. Setting the key value on the Control component handled the issue, however this then confused the validator on reset, so I had to create a custom reset method. Hopefully this will be useful to somebody, or if there is a better way I'd like to know.

### DIFF
--- a/examples/materialUIYupDynamicReloadReset.tsx.js
+++ b/examples/materialUIYupDynamicReloadReset.tsx.js
@@ -1,0 +1,196 @@
+/*
+This example uses contolled componets, with yup valiation
+and can handle changes to the incoming prop. In this case
+the prop that is updated is called "accountvalues".  This 
+form can be used for add/edit via the "actionstate" prop.
+This works with Material-UI components and hooks.
+*/
+
+import React from "react";
+import ReactDOM from "react-dom";
+import clsx from "clsx";
+import PropTypes from "prop-types";
+import { makeStyles } from "@material-ui/styles";
+import {
+  Card,
+  CardHeader,
+  CardContent,
+  CardActions,
+  Divider,
+  Grid,
+  Button,
+  TextField
+} from "@material-ui/core";
+import { useForm, Controller } from "react-hook-form";
+import * as yup from "yup";
+
+const useStyles = makeStyles(theme => ({
+  root: {
+    display: "flex",
+    flexDirection: "column",
+    padding: 4
+  },
+  card: {
+    display: "flex",
+    flexDirection: "column",
+    padding: 4
+  }
+}));
+
+const App = props => {
+  const { className, staticContext, ...rest } = props;
+  let { accountvalues, cardtitle, actionstate } = props;
+  const classes = useStyles();
+
+  /* jsut resetting the prop here to get cosdesandbox to work
+  	 REMOVE THIS!! */
+  accountvalues = {
+    account_id: 1234,
+    first_name: "Test Name",
+    balance: 4567.0
+  };
+  // yup validation
+  const formSchema = yup.object().shape({
+    first_name: yup.string().required("First Name is required"),
+    balance: yup
+      .number()
+      .transform(value => (isNaN(value) ? undefined : value))
+      .required("Balance is a required number")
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks a lot @tsaockham for the PR

examples/materialUIYupDynamicReloadReset.tsx.js
```diff
@@ -0,0 +1,210 @@
...
+This example uses contolled componets, with yup valiation
+and can handle changes to the incoming prop. In this case
+the prop that is updated is called "accountvalues".  This
+form can be used for add/edit via the "actionstate" prop.
+This works with Material-UI components and hooks.
+*/
+
+import React, { useState, useEffect } from "react";
+import ReactDOM from "react-dom";
+import clsx from "clsx";
```
bluebill1049: please remove this

examples/materialUIYupDynamicReloadReset.tsx.js
```diff
@@ -0,0 +1,210 @@
...
+import PropTypes from "prop-types";
```
bluebill1049: remove this dep.

examples/materialUIYupDynamicReloadReset.tsx.js
```diff
@@ -0,0 +1,210 @@
...
+import { makeStyles } from "@material-ui/styles";
```
bluebill1049: remove this.

examples/materialUIYupDynamicReloadReset.tsx.js
```diff
@@ -0,0 +1,210 @@
...
+  CardActions,
+  Divider,
+  Grid,
+  Button,
+  TextField
+} from "@material-ui/core";
+import { useForm, Controller } from "react-hook-form";
+import * as yup from "yup";
+
+const useStyles = makeStyles(theme => ({
```
bluebill1049: remove extra style

bluebill1049:
hmm thanks for the PR, can we please isolate the example to a single issue and just for form, and remove all those other dependencies. take a look other examples.

examples/materialUIYupDynamicReloadReset.tsx.js
```diff
@@ -0,0 +1,210 @@
...
+  const resetForm = () => {
+    Object.keys(accountvalues).forEach(key => {
+      setValue(key, accountvalues[key]);
+    });
+    setSaveDisabled(true);
+  };
+
+  useEffect(() => {
+    setSaveDisabled(true);
+    resetForm();
```
bluebill1049: there is a `reset` api. is that not working for you?

vensauro:
the CSB is broken, need to export the component as default

vensauro:
sorry talk here, where is not my problem, but why is the problem that this example try to solve?  
 i don`t understant this.  
  
[i make some modification](https://codesandbox.io/s/withered-monad-jt2h9) to use the default reset api, and see the behavior to be same, i don`t see something?

tsaockham:
Okay, I've made a new sandbox to show the issue I was trying to solve.  https://codesandbox.io/s/vigilant-mclean-ovywx

To recreate the issue, open the console and play around with the broken and fixed versions.  Note that when you hit submit, the old values come, even though the fields in the form show the new values.  When you do a keypress on any of the fields, that field alone will update, but the other fields will remain as the old ones.  The reset API function visibly fixes the issues, but the data is still the old data.

Also, if you play around with the account options that have no name or no balance, you can see that the validation also does not reset in the broken form.

On the fixed version I loop over the values and run the API setValue against the incoming prop item, this fixes the issue.

Should I pull this PR and submit this as an issue?  Or is it something I'm doing wrong on my end?

Thanks.

bluebill1049:
thanks a lot for the example @tsaockham can we isolate to a simpler version (it is rather complex to understand and reproduce) and submit an issue with it? I am closing this PR and let's discuss in the issue panel.

PR #1139, opened 2020-03-03
Feature/reset omit state update

bluebill1049:
This feature allow you to reset the form value without reset the `formState`

`reset({}, { dirty, touched })`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-hook-form",
   "description": "Performant, flexible and extensible forms library for React Hooks",
-  "version": "4.10.1",
+  "version": "4.10.2-beta.1",
   "main": "dist/react-hook-form.js",
   "module": "dist/react-hook-form.es.js",
   "types": "dist/index.d.ts",
--- a/src/types.ts
+++ b/src/types.ts
@@ -332,12 +332,12 @@ export type ArrayField<
   KeyName extends string = 'id'
 > = FormArrayValues & Record<KeyName, string>;
 
-export type OmitResetState = {
-  errors?: boolean;
-  dirty?: boolean;
-  dirtyFields?: boolean;
-  isSubmitted?: boolean;
-  touched?: boolean;
-  isValid?: boolean;
-  submitCount?: boolean;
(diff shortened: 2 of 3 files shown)


kiansenggoh:
we will love this feature!

GohKianSeng (dismissed):
nice!!

cubase (approved):

Paquan:
This would be awesome!
If I understand right than we would have the option to reset maybe formState.dirty when the form gots submitted, right?

bluebill1049:
> This would be awesome!
> If I understand it right we would have the option to reset e.g. formState.dirty when the form gots submitted, right?

That's correct. you can omit state individually.

kiansenggoh:
When will the changed be implemented? Since there is an option for user to choose if they wan to clear the state, why still need up vote?

krnlde:
Ouuh, I could also make use of that feature. Right now I worked around it big time :D

kiansenggoh:
Hi KrnIde

Your workaround is using setValue?

bluebill1049:
> When will the changed be implemented? Since there is an option for user to choose if they wan to clear the state, why still need up vote?

Thanks for your feedback. It's always easy for me to introduce more code into the lib to fulfil users' feature request. However what you desire to have may not be the case for others, hence i tagged as `waiting for up vote`. React Hook Form is great in eyes because it's simplicity and size. That's also the reason i am hesitate to bring more features which potential can have a work around or simply not valuable for others. This lib is not build for **you** or **myself**, it's an open source for everyone, hence everything goes into it need careful thoughts and deserve a voice from everyone. I can see a lot of people want this feature, so i will merge this it most likely this weekend and publish.

krnlde:
> Hi KrnIde
> 
> Your workaround is using setValue?

Partly, but mostly by just avoiding the reset process :D

bluebill1049:
Thanks for everyone's vote. I will update the doc and publish this feature this weekend. Thanks for your patient. 🙏

Paquan:
Thanks. Feels like getting presents :D

PR #1157, opened 2020-03-05
Add UMD build target

hperl:
For compatibility with our development server I would like to add an UMD output to rollup.

PR #1159, opened 2020-03-06
fixed rollup output for umd

kotarella1110:
sorry.  It looks like we need to use [`output.globals`](https://rollupjs.org/guide/en/#outputglobals).

スクリーンショット 2020-03-06 10 02 18

related PR: https://github.com/react-hook-form/react-hook-form/pull/1157


bluebill1049:
Thanks a lot @kotarella1110 !

Issue #1119, opened 2020-02-28
formState.isValid is always true


lazyd3v:
**Describe the bug**
`formState.isValid` is not being updated on field change if mode='onChange', Context, Controller and custom onChange event handler are used. If I remove custom onChange from Controller and pass onChange directly in the input, everything works fine

**Codesandbox link**
 https://codesandbox.io/s/xenodochial-silence-wy7p3

**Expected behavior**
When input value is changed from `test` to `t`, it should display:
1. `formState.isValid` is changed from true to false
2. `Object.keys(errors).length` is changed from 0 to 1
3. `value` is `t`

**Actual behaviour**
1. `formState.isValid` is always false
2. `Object.keys(errors).length` is always 0
3. `value` is `t`

bluebill1049:
found issue:
```
onChange={([value]) => {
  return value[0];
}}
```
make sure you return string instead of array. 👍

lazyd3v:
Aha, okay, thanks! I thought arguments in Controller's onChange doesn't wrap arguments in additional array. For example, if MyInput called onChange like `onChange({ value: 'foo' }, { bar: 'baz'})`, Controller's onChange would receive `{ value: 'foo' }` as first argument and `{ bar: 'baz' }` as second, but now onChange receives these arguments wrapped in array, so array destructuring `([ value ]) =>` destructure passed args, not the argument itself

I think it would be nice to add documentation regarding this and probably make fix in types. What do you think? I would love to make a PR

lazyd3v:
From my point of view, it would be much nicer to pass arguments as arguments, without wrapping them into array, but it will break compatibility with older versions. So that's why I suggest to add documentation regarding this

bluebill1049:
> From my point of view, it would be much nicer to pass arguments as arguments, without wrapping them into array, but it will break compatibility with older versions. So that's why I suggest to add documentation regarding this

That's was a mistake that I did in v4, which should have passed `arguments`. I wouldn't change this for v4 consider it's a breaking change, let's leave it to v5.

bluebill1049:
also FYI for `arguments`, you will still need to use it as arguments[0]

PR #1169, opened 2020-03-07
improve types of Controller's onChange and onBlur

susisu:
Hi, thank you for creating this amazing library! It improved our DX incredibly.

We found that `Controller`'s `onChange` callback takes an array as its argument, but its type does not reflect it.
In this PR I changed `EventFunction` to prevent mistakes on its argument (like #1119, may not cover all cases though).
`EventFunction` is exported by the library and this change may affect users' code, but I guess it is OK in most cases...

### DIFF
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -67,7 +67,7 @@ const Controller = <
     return data;
   };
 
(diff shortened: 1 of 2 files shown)


bluebill1049:
awesome! and thank you too.

Issue #1184, opened 2020-03-10
Controller component doesn't apply changed validation rules


levente-gal:
Using react-hook-form version 5.0.3 and having a component like:
```javascript
function MyComp(props) {
return (
    <form onSubmit={handleSubmit(data => console.log(data))}>
      <Controller
        as={<Select options={options} />}
        control={control}
        rules={{ required: props.doValidate ? 'This field is Required': false }}
        onChange={([selected]) => {
          // React Select return object instead of value for selection
          return { value: selected };
        }}
        name="reactSelect"
        defaultValue={{ value: 'chocolate' }}
      />

      <button>submit</button>
    </form>
  );
}
```
It doesn't recognise the changed validation rules.

Note that the validation rule is based on a property
**required**: **props.doValidate** ? **'This field is Required'** : **false**

After a bit of investigation in the code, the problem seems to be that it doesn't re-register the input because the *useEffect* which calls `registerField` has only the `name` property as dependency but it should also have the *validation rules* too

bluebill1049:
i would recommend using `validate`. also provide a CSB.

levente-gal:
[CSB](https://codesandbox.io/s/react-hook-form-rhfinput-i6sxg) for reproducing the issue.

There is a switch on the top of the page which is used to control the validation of the React select combo.

bluebill1049:
> After a bit of investigation in the code, the problem seems to be that it doesn't re-register the input because the useEffect which calls registerField has only the name property as dependency but it should also have the validation rules too

want to submit a PR since you find the issue and solution.

levente-gal:
I'll do PR ASAP.

PR #1187, opened 2020-03-10
re-regeister controlled component on validation rule change

levente-gal:
added useEffect with the validation rules as dependency to re-register field and trigger re-validation if needed

fixes bug: https://github.com/react-hook-form/react-hook-form/issues/1184

### DIFF
--- a/cypress/fixtures/example.json
+++ b/cypress/fixtures/example.json
@@ -1,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
--- a/cypress/fixtures/example.json
+++ b/cypress/fixtures/example.json
@@ -1,5 +1,0 @@
-{
-  "name": "Using fixtures to represent data",
-  "email": "hello@cypress.io",
-  "body": "Fixtures are a great way to mock data for responses to routes"
-}
--- a/src/controller.test.tsx
+++ b/src/controller.test.tsx
@@ -338,3 +338,49 @@ describe('Controller', () => {
     expect(removeFieldEventListener).toBeCalled();
   });
 });
+
+it('should re-register on new validation rules`', () => {
+  const control = reconfigureControl();
+  const fieldsRef = {
+    current: {},
+  };
+  const register = jest.fn().mockImplementation((payload: any) => {
+    // @ts-ignore
+    fieldsRef.current[payload.name] = 'test';
+  });
+
+  const { rerender } = render(
+    <Controller
+      defaultValue=""
+      name="test"
+      as={'input' as 'input'}
+      rules={{ required: false }}
+      control={{
+        ...control,
+        fieldsRef,
+        register: register,
+      }}
+    />,
+  );
+
+  expect(register).toHaveBeenCalledWith({ name: 'test' }, { required: false });
+
+  rerender(
+    <Controller
+      defaultValue=""
+      name="test"
+      as={'input' as 'input'}
+      rules={{ required: true }}
+      control={{
+        ...control,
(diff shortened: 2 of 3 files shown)


cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
bluebill1049: i think this should be removed.

src/controller.test.tsx
```diff
@@ -338,3 +338,49 @@ describe('Controller', () => {
     expect(removeFieldEventListener).toBeCalled();
   });
 });
+
+it('should re-register on new validation rules`', () => {
```
bluebill1049: awesome!

src/controller.tsx
```diff
@@ -117,6 +118,13 @@ const Controller = <
     };
   }, [name]);
```
bluebill1049: i would just add `rules` here and remove the logic below. If user want to revalidate give them the option to that at `useEffect`.

bluebill1049 (changes requested):
Awesome stuff! thank you for your contribution.

src/controller.tsx
levente-gal: Can't add rules there because this effect deletes the value of the field because of the `unregister` call in the cleanup of the effect.
I think it cleaner to add another *useEffect* then adding some kind of condition(s) to avoid unregistering.

The validation I think should re-run automatically based on the validation mode set by the user. Think of having a validator based on the value of another field on the form. If the first field changes the validation should re-run automatically on the validated field.

bluebill1049: it's fine to be `unregister` and `re-register` again and the value won't be lost. It's cheap to run anyhow.

levente-gal: the value gets lost because of the `useEffect` above: it doesn't find the fieldRef by name (because of the unregister) and it re-registers and set's to the default value

bluebill1049: make sense.

src/controller.tsx
```diff
@@ -117,6 +118,13 @@ const Controller = <
...
 
+  React.useEffect(() => {
+    registerField();
+    if (touchedFieldsRef.current[name] && shouldValidate()) {
```
bluebill1049:
```suggestion
```

src/controller.tsx
```diff
@@ -117,6 +118,13 @@ const Controller = <
...
+      triggerValidation(name);
```
bluebill1049:
```suggestion
```

src/controller.tsx
```diff
@@ -117,6 +118,13 @@ const Controller = <
...
+    }
```
bluebill1049:
```suggestion
```

please this trigger validation, and let the user trigger themself.

bluebill1049:
```suggestion
```

please remove this trigger validation, and let user trigger themself at `useEffect`

levente-gal: The validation I think should re-run automatically based on the validation mode set by the user. E.g. If an input is required when a checkbox is checked and validation mode is `onChange`. This case I think it would be nice to re-validate the input field when the checkbox changes so the validation message can be shown when checkbox changes.
I don't see why to force the developers to trigger validation it can be triggered by the library. 

In my opinion if the library allows dynamic rules then it should re-validate when the validation rules change.

And without triggering the validation on changed rules the controlled components won't behave similarly to registered ones because the registered ones get re-validated on changed validation rules.

levente-gal: Please forget my previous message, I was wrong. I remove the trigger validation.

bluebill1049: awesome! thanks :) let's get it in.

bluebill1049:
Thanks very much for your contribution.

PR #1216, opened 2020-03-16
Fix bug in controlledInput example.

AndrewSouthpaw:
It seems to me like the `controlledInput.tsx` example has a bug in it. It looks like the intent of the code is to control the last name input, but the `onChange` handler is placed on the firstName field.

This PR moves full controls over to the lastName input.


bluebill1049:
thank you are right, i have updated the codesandbox as well.

PR #1233, opened 2020-03-17
fix Touched type for formState

kotarella1110:
The `value` of `formState.touched` should be `true`, not `FieldError`.
To fix this, I added an `Value` argument to the `NestedDataObject` to make it more flexible.


bluebill1049:
Awesome thanks @kotarella1110

Issue #1226, opened 2020-03-16
Document validationResolver result


kyeotic:
The current documentation for `validationResolver` gives very little information on what the return/result value should look like. A complete example that does not rely on knowledge of external libraries like **joi** or **yup** would be very helpful.

Specifically

* What shape is `return.values`?
* What shape is `return.errors`?

bluebill1049:
@kotarella1110 maybe you can help on improve the return type as well (if you got time) :)

kotarella1110:
@bluebill1049 I will look into it after!

bluebill1049:
thank you @kotarella1110 i can take care the doc afterwards 👍

kyeotic:
@bluebill1049 I think [there might be an issue with this](https://github.com/react-hook-form/react-hook-form/pull/1239#discussion_r395101911)

kotarella1110:
@kabriel What are the issue?

kyeotic:
A validation function will need to return an object when there is a validation problem; a type error will stop typescript from compiling. How is a validation function supposed to return an error object if its a type error?

PR #1239, opened 2020-03-18
improve type for validationResolver and validationSchema

kotarella1110:
I fixed return type of `validationResolver` and `validateWithSchema`.

スクリーンショット 2020-03-19 14 21 25
スクリーンショット 2020-03-19 14 21 41
スクリーンショット 2020-03-19 14 21 58

I added `EmptyObect` type to indicate that it is an empty object. For more information, see "[Typescript empty object and any difference - Stack Overflow](https://stackoverflow.com/questions/47339869/typescript-empty-object-and-any-difference)".

```ts
type EmptyObject = { [key in string | number]: never };

let a: EmptyObject;
a = {}; // ok
a = 'foo'; // error
a = 123; // error
a = { foo:'bar' }; // error
```

[TypeScript Playground](https://www.typescriptlang.org/play/?ssl=7&ssc=28&pln=1&pc=1#code/C4TwDgpgBAogtmUB5ARgKwgY2FAvFAbygG0BrCEKASwDsoBnYAJ1oHMoAfKGgVzhQhMAugC5uEAG6CoAXwDcAWABQygDYQcAQzHxEIVBmyKlmvIXlQA9JagB7UstP4A5ADNbt53Ks3BTW0yOZgCMAEwAzN7WUH4BQfhE7rYiziiaTM6yUb5M-oEqSuo4KGIE8sooZmXGlS5JXhUhETVVUEkpaRlZQA)

related: https://github.com/react-hook-form/react-hook-form/issues/1226

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -43,17 +43,19 @@ export type SchemaValidateOptions = Partial<{
   context: any;
 }>;
 
+type EmptyObject = { [key: string]: never };
+
 export type SchemaValidationSuccess<
   FormValues extends FieldValues = FieldValues
 > = {
   values: FormValues;
-  errors: {};
+  errors: EmptyObject;
 };
 
 export type SchemaValidationError<
   FormValues extends FieldValues = FieldValues
 > = {
-  values: {};
+  values: EmptyObject;
   errors: FieldErrors<FormValues>;
 };
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -43,7 +43,7 @@ export type SchemaValidateOptions = Partial<{
   context: any;
 }>;
 
-type EmptyObject = { [key: string]: never };
+export type EmptyObject = { [key: string]: never };
 
 export type SchemaValidationSuccess<
   FormValues extends FieldValues = FieldValues
--- a/src/types.ts
+++ b/src/types.ts
@@ -43,7 +43,7 @@ export type SchemaValidateOptions = Partial<{
   context: any;
 }>;
 
-export type EmptyObject = { [key: string]: never };
+export type EmptyObject = { [key in string | number]: never };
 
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -43,21 +43,39 @@ export type SchemaValidateOptions = Partial<{
...
-export type ValidationResolverResponse<FormValues> = {
-  values: FormValues | {};
-  errors: FieldErrors<FormValues> | {};
+export type EmptyObject = { [key in string | number]: never };
+
+export type SchemaValidationSuccess<
+  FormValues extends FieldValues = FieldValues
+> = {
+  values: FormValues;
+  errors: EmptyObject;
```
bluebill1049: oh cool!

bluebill1049: correct me if i am wrong here @kotarella1110 

Does that means if i return errors:
```
{
   firstName: 'inValid'
}
```
will result in type error?

kotarella1110: @bluebill1049 yes! that’s correct.

bluebill1049: so cool.

bluebill1049 (approved):

bluebill1049:
Thank you very much @kotarella1110 !

src/types.ts
kyeotic: Um, that doesn't sound like a good idea. A validation will need to return that object when there is a validation problem, but a type error will stop typescript from compiling. How is a validation function supposed to return an errror?

bluebill1049: it's or, you either provide empty object or object of errors.

kyeotic: OH! It results in `type=error` not a "type error". Sorry, I was confused. This is awesome!

PR #1251, opened 2020-03-20
remove unnecessary Promise.resolve in handleSubmit

kotarella1110:
The async function returns a promise without using Promise.resolve.
Therefore, Promise.resolve is unnecessary.

```ts
async function foo() {
  return Promise.resolve(1)
} // function foo(): Promise<number>

async function bar() {
  return 1
} // function bar(): Promise<number>
```

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -1030,7 +1030,7 @@ export function useForm<
               } = field;
 
               if (!fields[name]) {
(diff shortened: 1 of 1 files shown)


bluebill1049:
that's right! I think I need to revisit this and turn this into a for loop.

PR #1253, opened 2020-03-20
enable declaration map closes #1250

### LINKED ISSUE #1250: Enable declarationMap: true in tsconfig when publish build
Sometime I want to refer to library's implementation.

Now, react-hook-form has only `.d.ts` files. So I need to look for implementation in source code.

If enable `declarationMap: true`, TypeScript generates a sourcemap for each corresponding `.d.ts` file.
This makes happy me when source code reading

sisisin:
I checked this change following steps.

1. run `yarn run clean && yarn build && yarn pack`
2. run `yarn add react-hook-form/react-hook-form-v5.1.1.tgz` in my own repo
3. run go-to-definition

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,5 @@
 example/
 .circleci/
-src/
 test/
 .babelrc
 .flowconfig
--- a/.npmignore
+++ b/.npmignore
@@ -37,3 +37,4 @@ rollup.ie11.config.js
 rollup.min.config.js
 tsconfig.ie11.json
 .codesandbox/
+*.tgz
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -31,6 +31,7 @@ export function getConfig({
       typescript({
         tsconfig,
         clean: true,
+        useTsconfigDeclarationDir: true,
       }),
       ...plugins,
     ],
--- a/rollup.min.config.js
+++ b/rollup.min.config.js
@@ -3,7 +3,7 @@ import { terser } from 'rollup-plugin-terser';
 
 export default {
   input: 'src/index.ts',
-  plugins: [typescript(), terser()],
+  plugins: [typescript({ useTsconfigDeclarationDir: true }), terser()],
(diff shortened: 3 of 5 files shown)


.npmignore
```diff
@@ -1,6 +1,5 @@
 example/
 .circleci/
-src/
 test/
```
bluebill1049: wait do we need to publish he src folder as well?

sisisin: Yes, declarationMap needs source codes.
Because if it isn't exists `src/`, editor can't jump to original source code.

When run `go to definition` on some editor, TypeScript Language Service detect `.d.ts` file, and if it exists `.d.ts.map` , editor received original source file position from TypeScript Language Service.
And then, editor jump to the point of source code.

bluebill1049: hmmm @susisu I am not too sure about this then, I checked some of the popular packages (react-router, MUI and etc...) and none of them include `./src` folder, so far only you request to include the src folder. I will probably not going to go ahead with this, but tag this with waiting for upvote.

kotarella1110: @bluebill1049 To be fair, some of the popular packages like [rxjs](https://github.com/ReactiveX/rxjs), [immer](https://github.com/immerjs/immer) and [apollo-server](https://github.com/apollographql/apollo-server) set `declarationMap: true` and include the `src` folder. why don't you want to protect the `src` folder? Are you concerned about the size increase? (I'm neither for nor against this PR)

bluebill1049: hey @kotarella1110 I don't have a strong opinion about this either, but I did config the `npmingore` to be the exact files that a user needs to run the package and did try to exclude other files `/src`, my intention was so they can get the package quicker with `npm i/yarn`, but I can see @susisu 's need, that's why personally rather wait for more people upvote for such change before just merge it. let me know if you & @susisu  happy to let it sit for a little while and listen for others' opinion as well.

bluebill1049:
looks like the demand is pretty low. let's close it for now :)

PR #1273, opened 2020-03-25
improve validate function with for loop

bluebill1049:
- reduce code footprint with `for loop` instead of `reduce`.


kotarella1110 (approved):
awesome 💯

bluebill1049:
Ty for your review @kotarella1110

Issue #1245, opened 2020-03-19
Why require a separate `keyName` rather than use provided `id`?


AndrewSouthpaw:
This question relates to #964 ...

@bluebill1049 can you say more about the reasoning for setting a different `keyName` rather than using the existing `id` if provided? It seems reasonable to assume that the `id` would be unique, and if that's the case then why use a separate generated key at all?

I imagine there's some inner workings of the library that I'm not understanding, so I thought I'd check in. It seems like a leaky abstraction to require knowledge that `useFieldArray` will overwrite `id` in your data.

Thoughts?

(Also, love the library BTW, great work so far. I'm hoping to be able to adopt it within our team as we ramp up the use of React forms instead of Rails forms. Turns out react-hook-form perfectly fits the bill of reducing all the clutter of event handlers. Thank you! 🙏 )

bluebill1049:
aha, I forgot to updated the docs! my bad...

we have a new prop called `KeyName` which you can rename `id` to other stuff if `id` was used for your submit.

`id` is really a unique id for `key` attribute, bcause there are cases when user use `id` as actual field name, hence introduced `keyName` to enabled rename.

```
{
  control,
  name,
  keyName = 'id' as KeyName,
}
```

```
useFieldArray({
  control,
  name: 'test',
  keyName: 'bill',
})

{fields.map(field => <div key={field.bill} />)}
```

bluebill1049:
doc is updated.

bluebill1049:
going to close this issue, feel free to follow up with more questions.

AndrewSouthpaw:
Hi @bluebill1049 , thanks for the response.

I'm familiar with the option for `keyName`. My question is this: why overwrite the `id` attribute in the first place?

It seems reasonable to use a provided `id` if there is one. It could be stated as an invariant that the `id` attribute if provided is expected to be unique. That way, the code reads much more naturally.

As it stays, I have to use this `keyName` attribute in order to divert the generated key into a useless place, since the id on my data works perfectly well.

bluebill1049:
hey @AndrewSouthpaw for uncontrolled inputs, RHF will need unique id(key) for each input/div/row, hence the `id` was introduced and later on i realise that it will overwrite or ignore user's actual input field which named as `id` and hence introduced the `keyName`. hopefully this answers your question, it's always bit tricky to explain offline. let me know if this answers your question.

AndrewSouthpaw:
[...]

PR #1281, opened 2020-03-26
Don't clobber the keyName when already present in data for useFieldArray.

AndrewSouthpaw:
This PR addresses the suggestion raised in #1245. 

**Background**

The default `keyName` for `useFieldArray` is `id`. There was an issue raised where legitimate `id` values for objects were then being clobbered. A fix was introduced to customize the name of the key with `keyName`.

**The Problem**

It is highly likely that `id` values on an object would be unique. It is natural for a developer to think to use `id` as the `key` for their objects. The current behavior of `useFieldArray` requires knowledge that `id` would get clobbered without customization via `keyName`, when the existing `id` would actually suffice.

**Proposed solution**

Do not clobber the `keyName` value when it's already present. Augment the documentation to note that a unique `id` will be provided if absent, or will use an `id` if already present on the fields; and that the key must be unique. The docs will continue to explain how `keyName` can be used to customize the name of an auto-generated key.

### DIFF
--- a/src/logic/mapIds.test.ts
+++ b/src/logic/mapIds.test.ts
@@ -1,0 +1,14 @@
+import { appendId } from './mapIds';
+
+describe('appendId', () => {
+  it('should add an auto-generated ID', () => {
+    expect(appendId({ value: 42 }, 'id')).toEqual({
+      value: 42,
+      id: expect.any(String),
+    });
+  });
+
+  it('should not override the keyName if already present', () => {
+    expect(appendId({ value: 42, id: 1 }, 'id')).toEqual({ value: 42, id: 1 });
+  });
+});
--- a/src/logic/mapIds.ts
+++ b/src/logic/mapIds.ts
@@ -7,8 +7,8 @@ export const appendId = <FormArrayValues extends FieldValues = FieldValues>(
   value: FormArrayValues,
   keyName: string,
 ) => ({
-  ...(isObject(value) ? value : { value }),
   [keyName]: generateId(),
+  ...(isObject(value) ? value : { value }),
 });
 
 export const mapIds = (data: any, keyName: string) =>


AndrewSouthpaw:
(I'd be happy to update docs if I get the OK from you that you'd like this change merged.)

bluebill1049:
aha i see the problem now. the problem is when you submit the form, do you want the `ID` to be part of the submit? because if we doing this `ID` will get stripped, because useFieldArray will treat it as a Key ID.

bluebill1049:
Thanks for the PR ❤️ The following is what i refer to, `id` will get striped during submission, because it will be treated as a key instead form data. 

Screen Shot 2020-03-27 at 9 45 42 am

bluebill1049:
if my assumption above is correct, please let us know how we can improve on the doc, or feel free to send a PR for the docs too.

AndrewSouthpaw:
Hmm, I think you're missing the point of my suggestion, so let me try to explain a different way:

I have a collection of objects, that have unique `id`s, and I want to be able to `PUT` the `id` along with the data. Something like:

```
// data: [{ id: 1, firstName: 'Andrew' }, { id: 2, firstName: 'Billy' }]

<input name={`test[${index}].id`} value={item.id} style={{display: 'none'}} ref={register()} />
<input
  name={`test[${index}].firstName`}
  defaultValue={`${item.firstName}`}
  ref={register()}
/>
```

As a developer not paying close attention to the docs, I would expect `id` to be what I gave it, _not_ what was autogenerated by RHF.

I think that's a more positive DX that leads to less confusing code. Right now, I have to know that `id` gets overwritten, provide a different `keyName` in the `useFieldArray` setup to avoid it being overwritten, and it just leads to dev confusion overall.

Does that help clarify the purpose of my PR?

bluebill1049:
yes @AndrewSouthpaw I totally agree and on board with your suggestion, I haven't tested this out just yet, I am just worried `id` field with gets dropped off during submission. Let me test this out and let you know.

bluebill1049:
aha awesome, thank you! i just approved that you are right with the PR: https://codesandbox.io/s/react-hook-form-usefieldarray-with-keyname-4pix6 thanks a lot for your time, PR and explanation.

AndrewSouthpaw:
Great! If you want to point me to where the docs are stored I could update those as well.

bluebill1049:
https://react-hook-form.com/api#useFieldArray

Screen Shot 2020-03-28 at 11 56 19 am

i think we can leave a note there

https://github.com/react-hook-form/react-hook-form-website/tree/master/src/data

if you can duplicated the same note for different languages as well and that would be great.

thanks @AndrewSouthpaw

PR #1328, opened 2020-04-03
Feature: support getValues by field name

bluebill1049:
current behaviour: 

`getValues()['test']`

proposing to using 

`getValues('test')`

which is much cheaper to run. Rather gather entire form values, it will just get individual field value.

original post from spectrum: https://spectrum.chat/react-hook-form/help/possibility-for-getvalue-key-in-the-api~6f74e4f9-0fdb-43db-b69f-cf618c1df327


kotarella1110:
@bluebill1049 Does this feature support nested objects?

bluebill1049:
yes @kotarella1110 `getValue('test.kotaro')`

kotarella1110:
If we support this feature, we must consider the return type of `getValues`.

bluebill1049:
That's true @kotarella1110 feel free to commit on top of this PR. if you have any idea.

bluebill1049:
good point! cause it's not going to return formValues anymore!

kotarella1110:
@bluebill1049 This is difficult if the following issues are not resolved.
https://github.com/microsoft/TypeScript/issues/12754

bluebill1049:
@kotarella1110 we will have to do this for now:

`const getValues = (payload?: { nest: boolean } | string): FormValues | unknown => {`

kotarella1110:
@bluebill1049 It become `unknown`.

スクリーンショット 2020-04-03 15 40 37

[TypeScript Playground](https://www.typescriptlang.org/play/#code/C4TwDgpgBAKhDOwoF4oG8oEMBcVECcBLAOwHMoBfKAHygFdiBrYgewHdiBuIA)

We should use the unknown.

```ts
function getValues(payload?: { nest: boolean }): FormValues
function getValues(payload?: string): unknown
function getValues(payload?: { nest: boolean } | string): unknown
```

[TypeScript Playground](https://www.typescriptlang.org/play/?ssl=5&ssc=80&pln=5&pc=67#code/C4TwDgpgBAYg9gJwLYDUCGAbArhAzlAXigG8oBtADwC5dgEBLAOwHMBdKqNRkKAXwCh+AEwgBjDGgTQAZlkajg9OIyjMIwdNjwAKMGhAY4aIQH4OpRnmAcARnDgYIXPgEoO8ZJpy5hYiVKhZeUVlVXUvHT0DI1MOWgYWNyg5AGtGOAB3Rl9xSRk5BSUVNQ1Mb119Q2MzEihLWlt7R2deKAAfKHimZiTU9KzBEojcbQByYCtRl34hsp0LKw46HFcgA)

bluebill1049:
you are right! forgot function overloads.

bluebill1049:
want to get your thoughts on this one @kotarella1110 @JeromeDeLeon, i think this feature will screw up the return type. however this is a good perf enhancement for user to read particular field's data. 

1. We can release this and fix the type later on when TS support this.
2. create new API `getValue`, which grows the amount of APIs
3. Do nothing.

kotarella1110:
@bluebill1049 I personally probably won't use this feature due to type safety concerns. Developers who prefer JavaScript over TypeScript may not care much about this. I think we should wait up vote and decide whether to release this feature.

bluebill1049:
sounds good to me @kotarella1110

JeromeDeLeon:
I will later take a look on it tonight.

src/useForm.ts
```diff
@@ -1152,7 +1152,15 @@ export function useForm<
     reRender();
   };
 
-  const getValues = (payload?: { nest: boolean }): FormValues => {
+  function getValues(payload?: { nest: boolean }): FormValues;
+  function getValues(payload?: string): unknown;
```
JeromeDeLeon, line 1156: Mind if I ask why `unknown` instead of `any`?

JeromeDeLeon, line 1156:
```suggestion
  function getValues<Name extends FieldName<FormValues>>(
    payload?: Name
  ): FormValues[Name];
```

JeromeDeLeon: @kotarella1110

JeromeDeLeon:
> @bluebill1049 I personally probably won't use this feature due to type safety concerns. Developers who prefer JavaScript over TypeScript may not care much about this. I think we should wait up vote and decide whether to release this feature.

As someone not so great with TS, I'd like to know the concerns.

src/useForm.ts
bluebill1049: > unknown which is the type-safe counterpart of any. Anything is assignable to unknown, but unknown isn't assignable to anything but itself and any without a type assertion or a control flow based narrowing. Likewise, no operations are permitted on an unknown without first asserting or narrowing to a more specific type.

bluebill1049:
> > @bluebill1049 I personally probably won't use this feature due to type safety concerns. Developers who prefer JavaScript over TypeScript may not care much about this. I think we should wait up vote and decide whether to release this feature.
> 
> As someone that is not so great with TS, I'd like to know the concerns. 😄

In my honest opinion, functionality should be taken higher priority than `Type`. Type is a bonus to safe guard our application. However, i do respect concerns from @kotarella1110 and i would love to hear what everyone else think as well, hence tagging this PR with waiting for up vote.  We won't release this feature until the demand its out there and to a point overtake the Type concerns.

src/useForm.ts
bluebill1049: wow, actually will this working for us and remove unknown? @kotarella1110

bluebill1049: nvm, not really. because the nested object.

kotarella1110: Cause type error if the payload is a dot bracket.
We should use unknown.

スクリーンショット 2020-04-06 19 32 58

https://www.typescriptlang.org/play?ssl=7&ssc=21&pln=7&pc=25#code/C4TwDgpgBAYg9gJwLYDUCGAbArhAzlAXigG8AoKKAOz2AgBMAuE8iqW3YJjhAS0oHMWAX1IjSoSFAByaJNCIByah3oA6dsAWlx4aABUahWIlSYcuANoy5AXQDcQA

kotarella1110: This also has the same problem with `setValue` ...

https://github.com/react-hook-form/react-hook-form/blob/master/src/contextTypes.ts#L80-L93

JeromeDeLeon: [Codesandbox](https://codesandbox.io/s/react-hook-form-vts01)

Seems strange. :thinking:

JeromeDeLeon: I'm finding it hard to extract the path in TS to support this unless change how the API works like:
```js
getValues(obj => obj.firstName)

// or

getValues('hobbies', '0', 'name')
```

bluebill1049: only if we can return type base on the argument type

311chaos:
Does the typing become easier if `getValues` is left as is, and a new method is exposed for `getValue`?

bluebill1049:
> Does the typing become easier if `getValues` is left as is, and a new method is exposed for `getValue`?

That's was one of the options, but I am hesitated to introduce new API., after all, we moving from v1 to v5 and there no new API for users to remember.

bluebill1049:
looks like the demand is quite low, i am going to close this PR for now.

kotarella1110:
@bluebill1049 improved type 👍 

```ts
type FormValues = {
  key1: string;
  key2: number;
  key3: {
    key1: number;
    key2: boolean;
  };
};

const { getValues } = useForm<FormValues>;

getValues()
// function getValues(payload?: { nest: boolean; } | undefined): FormValues
getValues({ nest: true })
// function getValues(payload?: { nest: boolean; } | undefined): FormValues
getValues('key1')
// function getValues<"key1">(payload: "key1"): string
getValues('key2')
// function getValues<"key2">(payload: "key2"): number
getValues('key3.key1')
// function getValues(payload: string): unknown
getValues<boolean>('key3.key2')
// function getValues<boolean>(payload: string): boolean
```

[TypeScript Playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAYg9gJwLYDUCGAbArhAzlAXigG8AoKKAawhAEYAuKXYBASwDsBzAbnKpoBMjdliQAjCAl4VqIAMyMyFGTQZQR4ydOWyhUMXDgYIadtoC+vS6VIATCAGMMaBNABmWdg+Cs47KJwQwOjYeAAUYGggGHBotgD8iup4wIwGRib+5gCUjPDIITi4vPZOLu6e3r7+gcGYRQA8ACpQEAAewBDstviycG6wiKj1eAB8EVExcYxNuYMFI7gA2k0AuiWOzq5QHl4+fgFBhXjNrR1dPVCelOxwAO7s45HRsbaMzGxcc028pLXHuDC2VIAHoQTtKvsakdFhMXnFEiRksw0oZjKZuFBzFAAD5XboQNwcCC2Ob5YahXB-GGUsLEZGpKAsHBY4FgiF7aqHOq055TBJJdgpVEZDFY3H4+xEoWkvJDAHUnlFMIAclktBVbPBuyqB3+iwaACJ1YanpNXoxjapDXMPhxOIqAardJrQdrIVz9ZSjbpTXD+ZbfXMNBIEI7YWqaHIAHTq13snVQ7lOvkWpgse1za63B7h73pdGPZ1R2OCePuzl6mmNAuZM3wt7pz6cOa10xAA)

bluebill1049:
aha, that's cool!

`getValues<boolean>('key3.key2')` Let's do this!

bluebill1049:
amazing work @kotarella1110! This is awesome, I think we can get this in for the next release. Thank you for your amazing contribution again. ❤️

PR #1329, opened 2020-04-03
Lint test files and ignore `@ts-ignore` rule within them.

threehams:
During the last PR, I noticed that lint errors showed up in-editor, but this wasn't reflected in CI. It took me a bit of time to figure out if this was going to cause the CI checks to fail.

This just overrides the lint rules to allow `ts-ignore` within tests (and allows for other rules to be suppressed over time, as well). That should prevent the disconnect between code and tests for new contributors.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,15 @@ module.exports = {
     'react-hooks/exhaustive-deps': 'error',
     'no-console': 'error',
   },
+  overrides: [
+    {
+      files: ['*.test.ts', '*.test.tsx'],
+      rules: {
+        // Allow testing runtime errors when not using TS
+        '@typescript-eslint/ban-ts-ignore': 'off',
+      },
+    },
+  ],
   settings: {
     react: {
       pragma: 'React',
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -20,6 +20,15 @@ module.exports = {
     'react-hooks/exhaustive-deps': 'error',
     'no-console': 'error',
   },
+  overrides: [
+    {
+      files: ['*.test.ts', '*.test.tsx'],
+      rules: {
+        // Allow testing runtime errors when not using TS
```
JeromeDeLeon, line 27:
```suggestion
        // Allow testing runtime errors to suppress TS errors
```

JeromeDeLeon (changes requested):

.eslintrc.js
JeromeDeLeon: I'll allow this, but we should limit using `@ts-ignore` as much as possible.

bluebill1049: agree!

threehams:
hm. I looked at all the `ts-ignore` usages, and none of them are being used for testing invalid input (like I thought they might). I'll get this ready, but then get rid of all the `ts-ignore` in the next commit.

bluebill1049:
awesome thanks @threehams! @JeromeDeLeon you happy to approve this PR?

JeromeDeLeon (approved):

PR #1338, opened 2020-04-04
fix #1333 web issue around reset with radio and checkbox only

### LINKED ISSUE #1333: reset not working when form contains only checkbox and radio
**Describe the bug**
Reset not working when form contains only inputs checkbox and radio

**To Reproduce**
1. Go to 'https://codesandbox.io/s/react-hook-form-reset-form-61crk?from-embed'
2. Update radio and checkbox values
3. Scroll down and click on  'Custom Reset Field Values & Errors'
4. Values are not reset

https://codesandbox.io/s/react-hook-form-reset-form-61crk?from-embed

**Expected behavior**
Should reset values.

chrome-capture

bluebill1049:


src/utils/isRadioOrCheckbox.ts
```diff
@@ -0,0 +1,6 @@
+import isRadioInput from './isRadioInput';
+import isCheckBoxInput from './isCheckBoxInput';
+import { FieldElement } from '../types';
+
+export default (ref: FieldElement): boolean =>
```
kotarella1110, line 5: From the point of view of type guard, we think that we should:

```suggestion
export default (ref: FieldElement): ref is HTMLInputElement =>
```

bluebill1049:
thank you for reviewing @kotarella1110

PR #1346, opened 2020-04-05
TS `any` cleanup

JeromeDeLeon:
I was supposed to change below snippet:

```js
function getCheckboxValue(options) {
  // some code
  
  const { checked, value, attributes } = options[0].ref;

  return checked
    ? attributes && !isUndefined((attributes as any).value)
      ? isUndefined(value) || isEmptyString(value)
        ? validResult
        : { value: value, isValid: true }
      : validResult
  : defaultResult;
}
```

But to change it, I need to do below because `attributes` is of type [`NamedNodeMap`](https://developer.mozilla.org/en-US/docs/Web/API/NamedNodeMap) of which we need to use `getNamedItems`.

```js
attributes.getNamedItem("value")
```

However, a lot of breaking changes in the test are needed to update as well therefore decided not to because the simpler the better.

### DIFF
--- a/src/logic/validateField.ts
+++ b/src/logic/validateField.ts
@@ -85,7 +85,7 @@ export default async <FormValues extends FieldValues>(
       error[name] = {
         type: INPUT_VALIDATION_RULES.required,
         message: requiredMessage,
-        ref: isRadioOrCheckbox ? (fields[name] as any).options[0].ref : ref,
+        ref: isRadioOrCheckbox ? (fields[name] as Field).options?.[0].ref : ref,
         ...appendErrorsCurry(INPUT_VALIDATION_RULES.required, requiredMessage),
       };
       if (!validateAllFieldCriteria) {
--- a/src/utils/set.ts
+++ b/src/utils/set.ts
@@ -19,7 +19,7 @@ export default function set(object: FieldValues, path: string, value: any) {
       newValue =
         isObject(objValue) || isArray(objValue)
           ? objValue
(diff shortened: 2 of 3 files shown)


src/utils/set.ts
```diff
@@ -19,7 +19,7 @@ export default function set(object: FieldValues, path: string, value: any) {
       newValue =
         isObject(objValue) || isArray(objValue)
           ? objValue
-          : !isNaN(tempPath[index + 1] as any)
+          : !isNaN(+tempPath[index + 1])
```
JeromeDeLeon, line 22: This one uses unary operator to cast it to number but here's the catch, if the value of `tempPath[index + 1]` is `null`, casting it to number will result to `0` thus making the result of `!isNaN(0)` to `true` instead of `false`. We could also change it to `parseInt` if you want.

src/logic/validateField.ts
```diff
@@ -85,7 +85,7 @@ export default async <FormValues extends FieldValues>(
       error[name] = {
         type: INPUT_VALIDATION_RULES.required,
         message: requiredMessage,
-        ref: isRadioOrCheckbox ? (fields[name] as any).options[0].ref : ref,
+        ref: isRadioOrCheckbox ? (fields[name] as Field).options?.[0].ref : ref,
```
JeromeDeLeon, line 88: without `?.`, the lint yells for making object possibly of `undefined`

bluebill1049:
great!

PR #1377, opened 2020-04-09
fix deep property names in useFieldArray

Silverwolf90:
Since the `name` in `useFieldArray` can be a deep property (eg: `'field.subfield'`) there was a mismatch between the nested object returned by `getValues({ nest: true })` and shallow property access. 

This PR uses the `get` utility to support deep property access of the nested values object. 

As discussed in this spectrum chat:

https://spectrum.chat/react-hook-form/general/recursive-form-fields-problems-with-usefieldarray~08c2da1b-4dae-4966-ba45-40cd130b163f

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -105,7 +105,7 @@ export const useFieldArray = <
   };
 
   const mapCurrentFieldsValueWithState = () => {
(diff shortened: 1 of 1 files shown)


bluebill1049:
amazing! thank you very much.

PR #1407, opened 2020-04-14
Enable Proxy check React Native

### LINKED ISSUE #1379: Enable Proxy check with React Native
I have register screen with 7 TextInput, when call register function the validationResolver its triggered before mounted components and produce two initial renders, i use yup validation and fixed it, verifying its Screen was mounted;

The bug its fixed add isMounted mutable ref, but because happen this? what its the best solution?

```typescript

const useYupValidationResolver = (validationSchema: any) => {
  const isMounted = React.useRef<boolean | undefined>();
  React.useEffect(() => {
    isMounted.current = true;
  }, []);

  return React.useCallback(
    async (data, validationContext) => {
      if (!isMounted.current) //<-- this prevent re-render
        return {
          values: data,
          errors: {},
        };
      try {
        const values = await validationSchema.validate(data, {
          abortEarly: false,
        });
        return {
          values,
          errors: {},
        };
      } catch (errors) {
        return {
          values: {},
          errors: errors.inner.reduce(
            (allErrors: Array<any>, currentError: any) => ({
              ...allErrors,
              [currentError.path]: {
                type: currentError.type ?? 'validation',
                message: currentError.message,
              },
            }),
            {},
          ),
        };
      }
    },
    [validationSchema],
  );
};

export default useYupValidationResolver;

```
System:
    OS: Windows 10 10.0.18363
    CPU: (4) x64 Intel(R) Core(TM) i7-7500U CPU @ 2.70GHz
    Memory: 9.44 GB / 19.88 GB
  Binaries:
    Node: 10.16.3 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.18.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.9.0 - C:\Program Files\nodejs\npm.CMD
    Watchman: Not Found
  SDKs:
    Android SDK:
      API Levels: 23, 26, 27, 28, 29
      Build Tools: 23.0.1, 27.0.3, 28.0.3, 29.0.2
      System Images: android-28 | Intel x86 Atom_64, android-28 | Google APIs Intel x86 Atom, android-29 | Google APIs Intel x86 Atom
      Android NDK: 20.1.5948944
  IDEs:
    Android Studio: Version  3.5.0.0 AI-191.8026.42.35.6010548
  Languages:
    Java: 1.8.0_221 - C:\Program Files\Java\jdk1.8.0_221\bin\javac.EXE
    Python: 2.7.16 - C:\Python27\python.EXE
  npmPackages:
    @react-native-community/cli: Not Found
    react: Not Found
    react-native: 0.62.2 => 0.62.2

msvargas:
Add proxy-polyfill in react-native to increase performance, because avoid re-render unneccesary.
fix: #1379 

Example with [proxy-polyfill](https://www.npmjs.com/package/proxy-polyfill) package, call this code in your App.js or index.js:

```typescript
import 'proxy-polyfill';
```

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -118,7 +118,6 @@ export function useForm<
     !isWindowUndefined &&
     !isUndefined(window.HTMLElement);
   const isProxyEnabled =
-    (isWeb && 'Proxy' in window) ||
     isWeb ? 'Proxy' in window : typeof Proxy !== UNDEFINED;
   const readFormStateRef = React.useRef<ReadFormState>({
     dirty: !isProxyEnabled,
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -117,7 +117,9 @@ export function useForm<
     typeof document !== UNDEFINED &&
     !isWindowUndefined &&
     !isUndefined(window.HTMLElement);
-  const isProxyEnabled = isWeb && 'Proxy' in window;
+  const isProxyEnabled =
+    (isWeb && 'Proxy' in window) ||
+    (typeof navigator !== UNDEFINED && typeof Proxy !== UNDEFINED);
   const readFormStateRef = React.useRef<ReadFormState>({
     dirty: !isProxyEnabled,
     dirtyFields: !isProxyEnabled,
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -119,7 +119,7 @@ export function useForm<
     !isUndefined(window.HTMLElement);
   const isProxyEnabled =
     (isWeb && 'Proxy' in window) ||
-    (typeof navigator !== UNDEFINED && typeof Proxy !== UNDEFINED);
+    isWeb ? 'Proxy' in window : typeof Proxy !== UNDEFINED;
   const readFormStateRef = React.useRef<ReadFormState>({
     dirty: !isProxyEnabled,
     dirtyFields: !isProxyEnabled,


src/useForm.ts
```diff
@@ -117,7 +117,9 @@ export function useForm<
     typeof document !== UNDEFINED &&
     !isWindowUndefined &&
     !isUndefined(window.HTMLElement);
-  const isProxyEnabled = isWeb && 'Proxy' in window;
+  const isProxyEnabled =
+    (isWeb && 'Proxy' in window) ||
+    (typeof navigator !== UNDEFINED && typeof Proxy !== UNDEFINED);
```
bluebill1049, line 122:
```suggestion
    isWeb ? 'Proxy' in window : typeof Proxy !== UNDEFINED;
```

hope we can reduce some of the code above, what do you think?

bluebill1049: sorry been a pain in the a** for writing less code here, I care about the package size a lot, the less we have to write the better it is for me.

punisher97: For me its OK, dont worry

src/useForm.ts
```diff
@@ -117,7 +117,9 @@ export function useForm<
     typeof document !== UNDEFINED &&
     !isWindowUndefined &&
     !isUndefined(window.HTMLElement);
-  const isProxyEnabled = isWeb && 'Proxy' in window;
+  const isProxyEnabled =
+    (isWeb && 'Proxy' in window) ||
```
bluebill1049, line 121:
```suggestion
```

bluebill1049: sorry miss this line. we don't this anymore.

src/useForm.ts
punisher97: Or, its the same and less code
```
const isProxyEnabled = typeof Proxy != 'undefined';
```

bluebill1049:
thanks a lot for your PR.

PR #1426, opened 2020-04-15
Fix/example-with-FieldArray

DanielPopOut:
I removed k +1 because otherwise createArrayWithNumbers(size).map(number ... (line 22) starts with 1 instead of 0 leading to a non existing first field when data are submitted

### DIFF
--- a/examples/FieldArray.tsx
+++ b/examples/FieldArray.tsx
@@ -3,7 +3,7 @@ import ReactDOM from 'react-dom';
 import { useForm } from 'react-hook-form';
 
 function createArrayWithNumbers(length) {
-  return Array.from({ length }, (_, k) => k);
+  return Array.from({ length }, (_, k) => +k + 1);
 }
 
 export default function App() {
--- a/examples/FieldArray.tsx
+++ b/examples/FieldArray.tsx
@@ -3,7 +3,7 @@ import ReactDOM from 'react-dom';
 import { useForm } from 'react-hook-form';
 
 function createArrayWithNumbers(length) {
-  return Array.from({ length }, (_, k) => k + 1);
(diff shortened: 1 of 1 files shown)


examples/FieldArray.tsx
```diff
@@ -3,7 +3,7 @@ import ReactDOM from 'react-dom';
 import { useForm } from 'react-hook-form';
 
 function createArrayWithNumbers(length) {
-  return Array.from({ length }, (_, k) => k + 1);
+  return Array.from({ length }, (_, k) => k);
```
bluebill1049, line 6:
```suggestion
  return Array.from({ length }, (_, k) => +k + 1);
```

what about above?

DanielPopOut: The above does solve the problem : 
Case 1 :
Array.from({length: 3}, (_ , k) => k)  gives [0,1,2]
Thus, when I submit data, the result is : 
```
{
firstName: [ "dsadsa", "dazfe"]
lastName: [ "adszd", "azerg"]
email: [ "sadf", "aefzre"]
}
```

Case 2 (your updates) :
Array.from({length: 3}, (_ , k) => +k + 1)  gives [1,2,3] leading to a first empty value when data is submitted.
```
firstName:  [empty, "dsadsa", "dazfe"]
lastName:  [empty, "adszd", "azerg"]
email:  [empty, "sadf", "aefzre"]
```
 is it wanted ?

----------------------------------------------------
A simpler function for `createArrayWithNumbers(length)` could be to return `[...Array(length).keys()]` ?

Thks for this great library 🎉

bluebill1049: oh, that's not right. please send another PR lol

PR #1429, opened 2020-04-16
fix - examples/FieldArray

DanielPopOut:
I removed k +1 because otherwise createArrayWithNumbers(size).map(number ... (line 22) starts with 1 instead of 0 leading to a non existing first field when data are submitted


bluebill1049:
Thanks

PR #1435, opened 2020-04-16
Forward Controller ref to as component

gfpacheco:
I'm using ReactNative and I need TextInput's ref so I can set focus when I press enter on a previous input.

I tested locally and it worked, I don't know exactly how to write test for the ref prop.

I know I can write `<Controller as={<TextInput ref={ref} />} />` but I think this would be a more elegant way to do it.

### DIFF
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -15,20 +15,23 @@ const Controller = <
     | React.ComponentType<any>
     | keyof JSX.IntrinsicElements,
   ControlProp extends Control = Control
->({
-  name,
-  rules,
-  as: InnerComponent,
-  onBlur,
-  onChange,
-  onChangeName = VALIDATION_MODE.onChange,
-  onBlurName = VALIDATION_MODE.onBlur,
-  valueName,
-  defaultValue,
-  control,
-  onFocus,
-  ...rest
-}: ControllerProps<As, ControlProp>) => {
+>(
+  {
+    name,
+    rules,
+    as: InnerComponent,
+    onBlur,
+    onChange,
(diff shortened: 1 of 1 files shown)


bluebill1049:
Thanks for your PR.

Question: Why not just use register for TextInput? if you can get hold of the ref?

`<TextInput ref={register} />` you can share the ref usage as well 
https://react-hook-form.com/faqs#Howtosharerefusage

gfpacheco:
@bluebill1049 I swear I searched for any documentation on how to do that and didn't find.

I tried something similar, but I was unable to make it work. I don't know if it's something to do with it being a custom register.

bluebill1049:
can you provide an example? for Controller you probably won't need ref

gfpacheco:
Yes, I have it working with Controller, but I couldn't make it work using the custom register.

This is what a custom register usually looks like:

```
<TextInput ref={register({ name: 'firstName'}, { required: true })} />
```

Looking at the shared ref doc, this is what it should look like if I want to share the ref:

```
<TextInput
  ref={e => {
    register({ name: 'name' }, { required: true })(e);
    myRef.current = e;
  }}
/>
```

But it doesn't work, I get an error saying `register({ name: 'name' }, { required: true })` is not a function, is undefined.

bluebill1049:
for RN, you can just custom register at useEffect.
```
useEffect(() => {
  register('name')
}, [regsiter])
```

gfpacheco:
Of course, since I'm the one actually setting the value when it changes 🤦

bluebill1049:
no worries, we probably won't this PR for now. thanks for it tho.

PR #1448, opened 2020-04-19
RHF's enhancement

JeromeDeLeon:
My usage with this lib is quite stable and haven't come across any issues since way back last year. That's why I didn't know what to fix at the moment and just decided to enhance some parts of the lib be it code or performance improvements. I will be updating this PR as well as putting comments to some parts. Not that so familiar with TS; therefore, I will leave everything to master @kotarella1110 :smile: .

/cc @bluebill1049

### DIFF
--- a/cypress/integration/basic.ts
+++ b/cypress/integration/basic.ts
@@ -1,3 +1,5 @@
+/// <reference types="cypress" />
+
 context('basic form validation', () => {
   it('should validate the form and reset the form', () => {
     cy.visit('http://localhost:3000/basic/onSubmit');
@@ -33,9 +35,7 @@ context('basic form validation', () => {
     cy.get('input[name="max"]').type('21');
     cy.get('input[name="minDate"]').type('2019-07-30');
     cy.get('input[name="maxDate"]').type('2019-08-02');
-    cy.get('input[name="lastName"]')
-      .clear()
-      .type('luo');
+    cy.get('input[name="lastName"]').clear().type('luo');
     cy.get('input[name="minLength"]').type('b');
     cy.get('input[name="validate"]').type('test');
 
@@ -51,12 +51,8 @@ context('basic form validation', () => {
     cy.get('input[name="minRequiredLength"]').type('bi');
     cy.get('select[name="multiple"]').select(['optionA']);
     cy.get('input[name="radio"]').check('1');
-    cy.get('input[name="min"]')
-      .clear()
-      .type('11');
-    cy.get('input[name="max"]')
-      .clear()
-      .type('19');
+    cy.get('input[name="min"]').clear().type('11');
+    cy.get('input[name="max"]').clear().type('19');
     cy.get('input[name="minDate"]').type('2019-08-01');
     cy.get('input[name="maxDate"]').type('2019-08-01');
     cy.get('input[name="checkbox"]').check();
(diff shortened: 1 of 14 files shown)


cypress/integration/basic.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="cypress" />
```
JeromeDeLeon, line 1: I don't know if you guys have problem with this but I need this to type `cy` in my VSCode. I also tried adding it individually but turned out, that was unnecessary and only one file was needed and the rest will magically inherited the types of `cy`. I also tried using `tsconfig` but seemed quite complicated and bumped up with this [issue](https://github.com/cypress-io/cypress/issues/1152). Luckily, just using `ref` only once made it work.

src/logic/getRadioValue.ts
```diff
@@ -13,7 +13,7 @@ const defaultReturn: RadioFieldResult = {
 
 export default (options?: RadioOrCheckboxOption[]): RadioFieldResult =>
   isArray(options)
-    ? options.filter(Boolean).reduce(
```
JeromeDeLeon, line 16: @bluebill1049 , not sure why it's here and seemed unnecessary IMO and when you filter the `options` that resulted to `[]` it will not return default value as the test for it suggested below snippet:

```js
it('should return default value if not valid or empty options', () => {
  expect(getRadioValue(undefined)).toEqual({
    isValid: false,
    value: '',
  });
});
```

or I just read it wrong? :laughing:

cypress/integration/basic.ts
bluebill1049: lol, that's strange! that's fine we can keep it :)

src/logic/getRadioValue.ts
bluebill1049: I think you are right :)

src/utils/compareObject.ts
```diff
@@ -1,12 +1,8 @@
...
+  const objectBKeys = Object.keys(objectB);
   return (
-    objectAData.length === objectBData.length &&
-    objectAData.reduce(
-      (previous, [key, value]) =>
-        previous ? objectB[key] && objectB[key] === value : false,
-      true,
-    )
+    objectAKeys.length === objectBKeys.length &&
+    objectAKeys.every((key) => objectB[key] && objectB[key] === objectA[key])
```
bluebill1049, line 6: This is awesome! much less code. love it 👯

src/utils/isNullOrUndefined.ts
```diff
@@ -1,4 +1 @@
-import isUndefined from './isUndefined';
-
-export default (value: unknown): value is null | undefined =>
-  value === null || isUndefined(value);
+export default (value: unknown): value is null | undefined => value == null;
```
bluebill1049, line 1: is this enough to check `isUndefined`? maybe worth to give some unit tests.

bluebill1049:
awesome stuff! let's get this one to merge first! we can have more PRs :)

src/utils/isNullOrUndefined.ts
JeromeDeLeon: Yes, I've been using it for a long time and haven't come across any problems.

bluebill1049: awesome! I love less code.

JeromeDeLeon: FYI, upon checking, even [`lodash`](https://github.com/lodash/lodash/blob/master/isNil.js) is using it.

bluebill1049:
Screen Shot 2020-04-20 at 10 33 50 pm

i think this function can use every as well :)

JeromeDeLeon:
Yes, can be `reduced` more :rofl: 

```ts
export default (names: Set<string>, name: string) => [...names].every((current) => isMatchFieldArrayName(name, current));

bluebill1049:
another PR is coming from you 🧜🏻‍♂️`reduce` master

PR #1412, opened 2020-04-14
improve setValue type

kotarella1110:
improved type to support nest object 👍 

```ts
type FormValues = {
  string: string;
  number: number;
  object: {
    number: number;
    boolean: boolean;
  };
  array: {
    string: string;
    boolean: boolean;
  }[];
};

const { setValue } = useForm<FormValues>()

setValue('string', 'test')
// function setValue<"string", string>(name: "string", value: string, shouldValidate?: boolean | undefined): void
setValue('number', 1)
// function setValue<"number", number>(name: "number", value: number, shouldValidate?: boolean | undefined): void
setValue('number', 'error')
// ❌: type error
setValue([{ string: 'test' }])
// function setValue<"string">(namesWithValue: DeepPartial<Pick<FormValues, "string">>[], shouldValidate?: boolean | undefined): void
setValue([{ number: 'error' }])
// ❌: type error
setValue([{
  string: 'test',
  number: 1
}])
// function setValue<"string" | "number">(namesWithValue: DeepPartial<Pick<FormValues, "string" | "number">>[], shouldValidate?: boolean | undefined): void
setValue([
  { string: 'test' },
  { number: 1 },
])
// function setValue<"string" | "number">(namesWithValue: DeepPartial<Pick<FormValues, "string" | "number">>[], shouldValidate?: boolean | undefined): void
setValue('object', { number: 1 });
// function setValue<"object", { number: number }>(name: "object", value: DeepPartial<{ number: number; boolean: boolean }>, shouldValidate?: boolean | undefined): void
setValue('object.boolean', true);
// function setValue<"object.boolean", boolean>(name: "object.boolean", value: boolean, shouldValidate?: boolean | undefined): void
setValue('array', [{ string: 'test' }])
// function setValue<"array", { string: string; }[]>(name: "array", value?: (DeepPartial<{ string: string; boolean: boolean; }> | undefined)[] | undefined, shouldValidate?: boolean | undefined): void
setValue('array[1].boolean', true)
// function setValue<"array[1].boolean", boolean>(name: "array[1].boolean", value: boolean, shouldValidate?: boolean | undefined): void
setValue('array[1].boolean', 'noerror')
// function setValue<"array[1].boolean", string>(name: "array[1].boolean", value: string, shouldValidate?: boolean | undefined): void
setValue<string, boolean>('array[1].boolean', 'error')
// ❌: type error
setValue([{ array: [{ boolean: false }]}])
[...]

PR #1449, opened 2020-04-19
fix getValues type

### LINKED ISSUE #1446: [typescript]type error, when set `control` props to `Controller` component;
**Describe the bug**
in the latest version; use with typescript, if declare form data type use `interface`,  will get a type error, when set `control` props to `Controller` component;

**To Reproduce**
see the example in Codesandbox as line 25

https://codesandbox.io/s/react-hook-form-useform-template-nsmib

**Expected behavior**
no type error

kotarella1110:
close #1446
[CodeSandbox](https://codesandbox.io/s/react-hook-form-useform-template-dw0z5)

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -283,9 +283,9 @@ export type Control<FormValues extends FieldValues = FieldValues> = {
       : LiteralToPrimitive<U>,
     shouldValidate?: boolean,
   ): void;
-  getValues(): IsFlatObject<FormValues> extends true
-    ? FormValues
-    : Record<string, unknown>;
+  getValues(): IsFlatObject<FormValues> extends false
+    ? Record<string, unknown>
+    : FormValues;
   getValues<T extends boolean>(payload: {
     nest: T;
   }): T extends true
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -283,9 +283,9 @@ export type Control<FormValues extends FieldValues = FieldValues> = {
       : LiteralToPrimitive<U>,
     shouldValidate?: boolean,
   ): void;
-  getValues(): IsFlatObject<FormValues> extends true
-    ? FormValues
-    : Record<string, unknown>;
+  getValues(): IsFlatObject<FormValues> extends false
+    ? Record<string, unknown>
+    : FormValues;
```
kotarella1110, line 288: I just changed extends for Conditioal Types from `true` to `false`.

kotarella1110 (commented):
This is a strange bug...
why...

bluebill1049:
why is this impact `type` vs `interface`?

src/types.ts
bluebill1049: wow, that's so strange...

kotarella1110:
@bluebill1049 Sorry for my late reply.
I don't know why this bug caused.
It seems to be related to the type error I commented here https://github.com/react-hook-form/react-hook-form/pull/1412#issuecomment-614740110.
The type definition is correct, but conditional types with generics (`FormValues`) seem to have cases where inference doesn't work properly.

bluebill1049:
yea quite strange one would be good to know the reason (as learning), but not going to stress about it.

Issue #1453, opened 2020-04-20
[Controller] isNotFieldArray is not updated properly


deebov:
**Describe the bug**
I've been getting weird issues because of this bug. I opened an [issue](https://github.com/react-hook-form/react-hook-form/issues/1369) but it didn't get resolved so I decided to get my hands dirty 😅
Controller uses `useMemo` to calculate `isNotFieldArray`. But since the `useMemo` receives a ref as a dependency that does not guarantee that the functions inside `useMemo` will be triggered whenever the ref changes. https://github.com/facebook/react/issues/15164#issuecomment-474833861 I would be happy to provide a more detailed code but I cannot get into `node_modules` in CodeSandbox. But putting `console.log(isNotFieldArray, isNameInFieldArray(fieldArrayNamesRef.current, name), name)` after `isNotFieldArray` is defined would clearly show the problem ( if you are using `Controller` inside `fieldArray` ). What I came up with is to replace `fieldArrayNamesRef` with `fieldArrayNamesRef.current.size` in the deps list here https://github.com/react-hook-form/react-hook-form/blob/master/src/controller.tsx#L56

**To Reproduce**

1. Go to https://codesandbox.io/s/react-hook-form-controller-jsgm6
2. Select `Strawberry` as a value for the first input
3. Append another input
4. Remove the input you've just added
5. The value of the first input get resetted

**Codesandbox link (Required)**
https://codesandbox.io/s/react-hook-form-controller-jsgm6

bluebill1049:
oh, cool @deebov thanks for the investigation. would like you like to send a PR if you have a solution?

deebov:
> oh, cool @deebov thanks for the investigation. would like you like to send a PR if you have a solution?

sure!

bluebill1049:
thanks a lot @deebov if you are having issues, I will be your fallback so no stress about it.

deebov:
@bluebill1049 would you give a little explanation about this part of the code https://github.com/react-hook-form/react-hook-form/blob/4c865d35465b2d8f4e855c5285c4bc25ae2fe45c/src/controller.tsx#L118

![image](https://user-images.githubusercontent.com/28673719/79720826-bfd40700-82fa-11ea-9d95-205d596aa8e0.png)

[...]

PR #1454, opened 2020-04-20
fix isNotFieldArray update

deebov:


src/controller.tsx
```diff
@@ -53,10 +53,12 @@ const Controller = <
   const isCheckboxInput = isBoolean(value);
   const shouldReValidateOnBlur = isOnBlur || isReValidateOnBlur;
   const rulesRef = React.useRef(rules);
-  const isNotFieldArray = React.useMemo<boolean>(
-    () => !isNameInFieldArray(fieldArrayNamesRef.current, name),
-    [fieldArrayNamesRef, name],
+  const isFieldArray = React.useMemo<boolean>(
+    () => isNameInFieldArray(fieldArrayNamesRef.current, name),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [fieldArrayNamesRef, fieldArrayNamesRef.current.size, name],
```
bluebill1049, line 59: event `fieldArrayNamesRef.current` doens't pick up the diff?

src/controller.tsx
```diff
@@ -53,10 +53,12 @@ const Controller = <
...
   );
+  const isNotFieldArray = !isFieldArray;
```
bluebill1049, line 61: maybe combine the line above to save some code?

src/controller.tsx
deebov: first lint complaint was "fieldArrayNamesRef is a necessary dep." the second complaint was "fieldArrayNamesRef.current.size is not necessary"

src/controller.tsx
```diff
@@ -119,9 +118,7 @@ const Controller = <
     registerField();
```
bluebill1049, line 118:
```suggestion
```

src/controller.tsx
deebov: sorry, I couldn't get what you mean

bluebill1049:
awesome stuff! thanks a lot for your effort @deebov

deebov:
> awesome stuff! thanks a lot for your effort @deebov

thank you ) I think it's not finished yet https://github.com/react-hook-form/react-hook-form/issues/1453#issuecomment-616474048

deebov:
> > awesome stuff! thanks a lot for your effort @deebov
> 
> thank you ) I think it's not finished yet [#1453 (comment)](https://github.com/react-hook-form/react-hook-form/issues/1453#issuecomment-616474048)

I can post the screenshots if you want

src/controller.tsx
```diff
@@ -115,15 +112,7 @@ const Controller = <
...
-  React.useEffect(() => {
-    registerField();
-
-    return () => {
-      if (isNotFieldArray) {
-        unregister(name);
-      }
-    };
-  }, [name, unregister, fieldArrayNamesRef, registerField, isNotFieldArray]);
+  React.useEffect(() => unregister(name), [unregister, name]);
```
bluebill1049, line 115:
```suggestion
  React.useEffect(() => () => unregister(name), [unregister, name]);
```

src/controller.tsx
```diff
@@ -53,12 +53,10 @@ const Controller = <
   const isCheckboxInput = isBoolean(value);
   const shouldReValidateOnBlur = isOnBlur || isReValidateOnBlur;
   const rulesRef = React.useRef(rules);
-  const isFieldArray = React.useMemo<boolean>(
+  const isNotFieldArray = !React.useMemo<boolean>(
     () => isNameInFieldArray(fieldArrayNamesRef.current, name),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    [fieldArrayNamesRef, fieldArrayNamesRef.current.size, name],
+    [fieldArrayNamesRef, name],
```
deebov, line 58: i think we shouldn't remove fieldArrayNamesRef.current.size

deebov: because fieldArrayNamesRef is a ref and its current value is Set which is going to be same everytime. Because react uses Object.is to compare previous and current values

deebov: yes, `fieldArrayNamesRef.current.size` is an ideal solution but in this case it's the only source-of-truth

bluebill1049: > because fieldArrayNamesRef is a ref and its current value is Set which is going to be same everytime. Because react uses Object.is to compare previous and current values

hmmm, you have a valid point there.

bluebill1049: > yes, `fieldArrayNamesRef.current.size` is an ideal solution but in this case it's the only source-of-truth

it's not a guarantee as well then, if 1 get removed and another gets added, so size still the same.

deebov: https://github.com/facebook/react/issues/15164#issuecomment-474833861

bluebill1049: I don't think we can safely memo this value here.

bluebill1049: because the reference is the same `Set` and size does not guarantee the diff.

deebov: > > yes, `fieldArrayNamesRef.current.size` is an ideal solution but in this case it's the only source-of-truth
> 
> it's not a guarantee as well then, if 1 get removed and another gets added, so size still the same.

sorry I meant `fieldArrayNamesRef.current.size` is **not** an ideal solution

deebov: > I don't think we can safely memo this value here.

we could remove `memo` here

bluebill1049: yeah i think memo need to be removed.

PR #1471, opened 2020-04-22
V6

bluebill1049:
### 📖 Documentation

- [x] Update V6 documenation: https://github.com/react-hook-form/react-hook-form-website/pull/252
- [x] New Page TS 
  - [x] scope/limit exposed Types
- [x] Include selection under API page to change version
- [x] Migration Guide V5 -> V6

### 🚨 Breaking changes

- [x] remove `validationSchema` and embrace validation resolver
https://github.com/react-hook-form/react-hook-form-resolvers

**Reason:** we would love to support all kinds of schema validation instead of just Yup, also it's not fair to carry that extra logic for users who don't use schema at all. also perhaps there are lighter options than only `Yup` as well.

| Name | Size | Url |
|--|--|--|
| Superstruct | 4.5KB | [link](https://github.com/ianstormtaylor/superstruct) |
| Yup | 20KB | [link](https://github.com/jquense/yup) |
| Joi | 29KB | [link](https://github.com/hapijs/joi) |

```diff
- useForm({ validationSchema: schema })
+ useForm({ resolver: yupResolver(schema) })
```

```ts
import React from 'react';
import { yupResolver } from 'react-hook-form-resolvers';

const SignupSchema = yup.object().shape({
  name: yup.string().required(),
  age: yup.number().required(),
});

const App = () => {
  const { register, handleSubmit } = useForm({
    resolver: yupResolver(SignupSchema), // yup, joi and even your own.
  });

  return (
    <>
        <label>Test</label>
        <input name="test" ref={register} />
        <input type="submit" />
      </form>
    </>
  );
};
```

---------------------

- [x] reanme `validationResolver` to resolver
- [x] reanme `validationContext` to context
- [x] reanme `validateCriteriaMode` to `criteriaMode`

**Reason:** less for the user to type, consistent with the rest of the APIs. This is a generic validation library, it makes sense to keep naming generic and validation should already be the context for this whole lib.  

```diff
- validationResolver,
- validationContext,
+ resolver,
+ context,
```

---------------------

- [x] `Controller` props `onChange` change from array arguments to object arguments

**Reason:** This should be the expected behavior for function's `arguments`. 

```diff
- onChange = {[e] => console.log(e)};
+ onChange = {e => console.log(e)};
```

---------------------

- [x] `triggerValidation` rename to `trigger`

**Reason:** Less for a user to type and consistent with the rest of the APIs. However, if this having issue or conflict with your existing code base, you have the option to rename it in the deconstructor.

```diff
- triggerValidation();
+ trigger();
```

---------------------

- [x] `FormContext` rename to `FormProvider`

**Reason:** Align name with all those big libraries such as Redux, styled-components and etc.

```diff
- <FormContext />
+ <FormProvider />
```

- [x]  export `FormContext`

**Reason:** Some users may want to use `FormContext.Consumer`.

```jsx
import { FormContext } from 'react-hook-form';

// in your connected component
render() {
  return (
    <FormContext.Consumer>
      {({ register, handleSubmit }) => {
        // do something with the useForm methods here
      }}
    </FormContext.Consumer>
  )
};
```

---------------------

- [x] remove `nest` option for `watch`& `getValues`, so data return from both methods will be in `FormValues` shape.

**Reason:** Consistency for form values and better type support.

```diff
- getValues({ nest: true }); // { test: { data: 'test' }}
- watch({ nest: true });  // { test: { data: 'test' }}
+ getValues(); // { test: { data: 'test' }}
+ watch();  // { test: { data: 'test' }}
````

---------------------

- [x] `dirtyFields` change type from `set` to `object`

**Reason**: consistency form state API and simplify the codebase. Reduced codebase size with object data shape.

---------------------

- [x] `dirty` change name to `isDirty`

**Reason**: consistency form state name for boolean data.

###  🥃 New Features

- [x] TS: `NestedValue` 

**Reason**: The field value can be array/object such as Material-UI Multiple Select and custom registered field (Custom Register).

Currently, errors object expects an array value to always be multiple fields, but it can also be a single field with an array value. To solve this, it supports NestedValue to determine if it is the field value (array/object).

```ts
import { useForm, NestedValue } from 'react-hook-form';

type FormValues = {
  key1: string;
  key2: number;
  key3: NestedValue<{
    key1: string;
    key2: number;
  }>;
  key4: NestedValue<string[]>
};

const { errors } = useForm<FormValues>();

errors?.key1?.message // no type error
errors?.key2?.message // no type error
errors?.key3?.message // no type error
errors?.key4?.message // no type error
```

---------------------

- [x] `useWatch` (new) subscribe to registered inputs.

**Reason:** This could isolate the re-render within a particular component without trigger re-render at the form level. This is different from `watch` API, which will re-render at the root of `useForm`

```ts
<input name="test" ref={register} />

function IsolateReRender() {
  const { state } = useWatch({
    name: 'test',
    control,
    defaultValue: 'default'
  });

  return <div>{state}</div>
}
```

---------------------

- [x] `getValues()` support array of field names

 **Reason:** API Consistency with `watch`

```ts
+ getValues(['test', 'test1']); // { test: 'test', test1: 'test1' }
````

---------------------

- [x] `useForm({ mode: 'all' })` support all validation

 **Reason:** so each input can validate with blur and change. https://github.com/react-hook-form/react-hook-form/issues/1222

```ts
+ useForm({ mode: 'all' })
````

### CODE REVIEW

**Comment by JeromeDeLeon** [comment]
Oh! I just pushed my changes directly to v6 branch.  :disappointed:

**Comment by bluebill1049** [comment]
> Oh! I just pushed my changes directly to v6 branch. 😞

That's all good! wow you fixed it...

File: src/useForm.ts [hunk 1]
```diff
@@ -1154,6 +1153,13 @@ export function useForm<
     reRender();
   };
 
+  const getValue = <T extends string, U extends unknown>(
```
bluebill1049, line 1156: @kotarella1110 let me know if the type is correct.

kotarella1110: I will fix this later 👍

bluebill1049: Thanks @kotarella1110

bluebill1049:
@kotarella1110 you want to include more examples/update on the `NestedValue` section?

azizur:
@bluebill1049 Will there be a backward compatibility for `validationSchema` maintained for some period? I just migrated to using `react-hook-form` and I want keep up with upgrades, but same time I also want to launch my startup without having to invest more time on migration just when I made significant progress on the app.

bluebill1049:
> @bluebill1049 Will there be a backward compatibility for `validationSchema` maintained for some period? I just migrated to using `react-hook-form` and I want keep up with upgrades, but same time I also want to launch my startup without having to invest more time on migration just when I made significant progress on the app.

Hey, @azizur It will just be search-replace, which I will include that in the migration guide.

```diff
- validationSchema: schema,
+ import { yupResolver } from 'react-hook-form-resolvers';
+ validationResolver: yupResolver(schema);
```

We are moving those schema specific logic out from the main lib, and open more opportunities other than just `yup`.

joejordan:
When it comes to renaming triggerValidation, validationResolver, validationContext, personally as a new RHF user I like the descriptive terms better than the shorter more generic terms. It gives me a clearer indication of what to expect.

bluebill1049:
> When it comes to renaming triggerValidation, validationResolver, validationContext, personally as a new RHF user I like the descriptive terms better than the shorter more generic terms. It gives me a clearer indication of what to expect.

Thanks for your feedback. After all, this is a validation library. I think it makes sense to be generic for a lib instead of a hook that we build inside our project. Also, you can still rename them as you want as well. It also keeps things consistent with the rest of the API.

bluebill1049:
6.0.0-rc.1 is out

jtomaszewski:
Why merge by squasing? :( That ruins commit history a lot, makes code/changes harder to understand. Just a suggestion for the future to use squash only for squashing small related-to-each-other (often WIP-like) commits.

PR #1487, opened 2020-04-24
improve register type

kotarella1110:
Fixed to auto completion works on flat objects.

flat object (typesafe):

スクリーンショット 2020-04-24 14 17 08

nested object (no-typesafe):

スクリーンショット 2020-04-24 14 17 26


bluebill1049:
awesome stuff @kotarella1110 i am sure one day TS will support some type of regex key name.

PR #1491, opened 2020-04-25
V6 useForm: global nest config for errors & values

bluebill1049:
I think the issue is between `values` and `errors` are not always in sync with each other, purpose to have two separate `config`. Let me know what do you think @kotarella1110 

By default `errors` and `values` will be all nested object, and yet still allow users to config to be flat. We can then probably have a separate type for Values and Errors (if errors are different from values).

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -104,6 +104,8 @@ export type UseFormOptions<
   validationContext: ValidationContext;
   submitFocusError: boolean;
   validateCriteriaMode: 'firstError' | 'all';
+  nestedError: boolean;
+  nestedValue: boolean;
 }>;
 
 export type MutationWatcher = {
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -75,6 +75,8 @@ export function useForm<
   defaultValues = {},
   submitFocusError = true,
   validateCriteriaMode,
+  nestedError = true,
+  nestedValue = true,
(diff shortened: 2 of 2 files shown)


kotarella1110:
@bluebill1049 this feature has type safety problems.
if only one is `true`, can’t be infer type as flollows:

```ts
const { getValues, errors } = useForm<{
  key: string;
  'nested.key': string
}>({
  nestedError: true,
  nestedValue: false,
});

getValues()
// type:
// {
//   key: string;
//   'nested.key': string;
// }

errors
// type: unknown
```

bluebill1049:
> @bluebill1049 this feature has type safety problems.
> if only one is `true`, can’t be infer type as flollows:
> 
> ```ts
> const { getValues, errors } = useForm<{
>   key: string;
>   'nested.key': string
> }>({
>   nestedError: true,
>   nestedValue: false,
> });
> 
> getValues()
> // type:
> // {
> //   key: string;
> //   'nested.key': string;
> // }
> 
> errors
> // type: unknown
> ```

I will think of having two separate generics to counter this issue:

<Values, Errors, Context>

by default 

```
useForm<{ firstName: string, nested: { age: string } }>()

errors.nested.age
```

```
useForm<{ firstName: string, nested: { age: string } }, { ['nested.age']: string }>({
  nestedErrors: false
})

// Assumption: Values type can merge with Errors

errors['nested.age']
```

PR #1497, opened 2020-04-26
V6: remove nest option

bluebill1049:
- [x] remove `nest` option for `watch`& `getValues`, so data return from both methods will be in `FormValues` shape.

```diff
- getValues({ nest: true }); // { test: { data: 'test' }}
- watch({ nest: true });  // { test: { data: 'test' }}
+ getValues(); // { test: { data: 'test' }}
+ watch();  // { test: { data: 'test' }}
````


bluebill1049:
let me know if you guys have feedback or concerns.

kotarella1110:
@bluebill1049 awesome! I've always thought this option was unnecessary!
I'll fix it because the type is incorrect.

Issue #987, opened 2020-02-06
Typescript: The property 'message' doesn't exists in 'FieldError | FieldErrors<any>[]'


cargallo:
**Describe the bug**

The property 'message' doesn't exists in 'FieldError | FieldErrors<any>[]'.
The property 'message' doesn't exists in type 'FieldErrors<any>[]'.ts(2339)

This happens in typescript when trying to access errors messages from Autocomplete component from yup validation schema

**Codesandbox link**

Look at line 48!!!

https://codesandbox.io/s/material-demo-o15mc

**Screenshots**

![image](https://user-images.githubusercontent.com/13320290/73969895-f94fc580-48fa-11ea-8f3f-2ae6054faf47.png)

fabiodcorreia:
I also have the same problem, but if I do console.log(errors.name) the message property is there. I think it's just something with the typescript validation.

bluebill1049:
when you have sometimes, can you please take a look @kotarella1110 🙏

cargallo:
In the meantime a temporary workarround could be:

helperText={errors.multiple && (errors.multiple as any).message}

fabiodcorreia:
Works for me, here is a smaller version.
```typescript
{(errors.nameInput as any)?.message}
```

kotarella1110:
@fabiodcorreia How is this?

```ts
interface IServicio {
  multiple: unknown;
}
```

fabiodcorreia:
I think that's for @cargallo, I my case I don't have that interface or Material UI :)

> @fabiodcorreia How is this?
> 
> ```diff
> interface IServicio {
> -  multiple: any[];
> +  multiple: unknown;
> }
> ```

cargallo:
> @fabiodcorreia How is this?
> 
> ```diff
> interface IServicio {
> -  multiple: any[];
> +  multiple: unknown;
> }
> ```

@kotarella1110 That works, but I still think that there is a problem in the errors object handling in typescript, I had troubles with yup nested objects too. I used any[] to show that neither in that case the code was working, but it should be a custom object array.

kotarella1110:
@bluebill1049 How should this be fixed? Do you have any suggestions?

bluebill1049:
no worries @kotarella1110 i will take a closer look at this over the weekend.

bluebill1049:
@cargallo did you updated the CSB? I don't see the error line now.

cargallo:
[...]

PR #1503, opened 2020-04-26
support NestedValue type

### LINKED ISSUE #1421: Typings for `errors` and `validationResolver` are very restrictive
Both `errors` and `validationResolver` make strong assumptions about the format of the errors (when using TypeScript, of course).
The project page says it supports a variety of validation libraries (none of which I use) or custom validation. I partially rely on `ajv` and partially write my own validation. In both cases I use completely flat errors (e.g. `{ "cases[0].caseData.profileInformation.age": "should be a number" }`), which is incompatible with the typings enforced by the library.
So I feel like (at least with TypeScript) it does not actually support using custom validation. Or is there a technical reason why errors need to be shaped a certain way? AFAIC errors are created and consumed by user-code only and the library only passes it around?

I think it would be good to allow a third, optional type being passed to `useForm` which indicates the error object shape. It could default to the current type (which seems to work fine for a variety of use cases).

I've resorted to using a bunch of `as any` casts and sometimes tricking the TypeScript compiler by combining untyped objects with not using the dot notation (which is bad style in the end), e.g.
```
  const errors = {}
  if (!values.caseSetId) {
    // eslint-disable-next-line dot-notation
    errors['caseSetId'] = 'Select a case set'
  }
```
But this doesn't work when I want to put an error on an array field. It will always want an array of errors, which IMHO doesn't make sense if I simply want to say "Please add at least one X"

### LINKED ISSUE #1536: Separate field errors and form value types
1) Why errors can not be accessed by passing controller name?
2) Why type of `errors` is same as form values, where form values can be similar but have different type. As you can see from the example below, value of the `mcc` field is `IMCC`. But the field name is `mcc`. So if I want to access errors of `mcc` field, it will return an error just because `errors` and `formValues` are sharing same type passed from `useForm`.

````typescript
interface IMCC {
  code: string,
  value: string,
}

type FormData = {
  mcc: IMCC;
  input: string;
};
````

### PR DESCRIPTION
**Pull request by kotarella1110**
The field value can be array/object such as [Material-UI  Multiple Select](https://material-ui.com/components/selects/#multiple-select) and custom registered field ([Custom Register](https://react-hook-form.com/api/#register)).
Currently, `errors` object expects an array value to always be multiple fields, but it can also be a single field with array value.
To solve this, it supports `NestedValue` to determine if it is the field value (array/object).

Usage:

```ts
import { useForm, NestedValue } from 'react-hook-form';

type FormValues = {
  key1: string;
  key2: number;
  key3: NestedValue<{
    key1: string;
    key2: number;
  }>;
  key4: NestedValue<string[]>
};

const { errors } = useForm<FormValues>();

errors?.key1?.message // no type error
errors?.key2?.message // no type error
errors?.key3?.message // no type error
errors?.key4?.message // no type error
```

Related Issues:

*  https://github.com/react-hook-form/react-hook-form/issues/1421
*  https://github.com/react-hook-form/react-hook-form/issues/987


src/types.ts
```diff
@@ -31,6 +31,24 @@ export type FieldValue<FormValues extends FieldValues> = FormValues[FieldName<
...
 
+declare const $NestedValue: unique symbol;
+
+export type NestedValue<
+  TValue extends unknown[] | Record<string, unknown> =
+    | unknown[]
+    | Record<string, unknown>
+> = {
+  [$NestedValue]: never;
+} & TValue;
```
kotarella1110, line 42: @bluebill1049 @JeromeDeLeon Is this type of naming "NestedValue" OK?

src/types.ts
```diff
@@ -31,6 +31,24 @@ export type FieldValue<FormValues extends FieldValues> = FormValues[FieldName<
...
+
+export type UnpackedFieldValues<TFieldValues extends FieldValues> = {
+  [Key in keyof TFieldValues]: TFieldValues[Key] extends NestedValue<infer U>
+    ? U
+    : TFieldValues[Key] extends object
+    ? UnpackedFieldValues<TFieldValues[Key]>
+    : TFieldValues[Key];
+};
```
kotarella1110, line 50: @bluebill1049 @JeromeDeLeon I can't come up with a good naming. Is this type any good naming?

src/types.ts
bluebill1049: sounds good to me :)

src/types.ts
bluebill1049: TFieldValues sounds good, maybe send a second Pr to change all our generic to follow that pattern.

kotarella1110: @bluebill1049 yeah! I will do it.
What do you think about naming `UnpackedFieldValues`?

bluebill1049: prefer `TFieldValues` sounds simpler

kotarella1110: Is the following prefer?

```diff
- export type UnpackedFieldValues<TFieldValues extends FieldValues> = {
+ export type TFieldValues<TFieldValues extends FieldValues> = {
```

I think the `T` should only be prefixed to generics parameter as follows:

```ts
type Parameter = string;
type Example<TParameter extends Parameter> = TParameter;
```

bluebill1049: sounds good!

src/types.ts
```diff
@@ -31,16 +31,38 @@ export type FieldValue<FormValues extends FieldValues> = FormValues[FieldName<
...
+export type UnpackedDeepPartial<T> = {
+  [K in keyof T]?: T[K] extends NestedValue<infer U>
+    ? U
+    : T[K] extends Array<infer U>
+    ? Array<UnpackedDeepPartial<U>>
+    : T[K] extends ReadonlyArray<infer U>
+    ? ReadonlyArray<UnpackedDeepPartial<U>>
+    : T[K] extends { [key: string]: unknown }
+    ? UnpackedDeepPartial<T[K]>
+    : T[K];
```
kotarella1110, line 65: @bluebill1049 I changed `DeepPartial` to `UnpackedDeepPartial`.
Do you have any objections to this?

kotarella1110 (commented):
~~@bluebill1049 completed! I added some tests, but I will add type tests later.~~

kotarella1110:
@bluebill1049 completed 👯

bluebill1049:
AWESOME!!!

src/types.ts
```diff
@@ -31,16 +31,38 @@ export type FieldValue<FormValues extends FieldValues> = FormValues[FieldName<
...
+  [K in keyof T]: NonUndefined<T[K]> extends NestedValue<infer U>
+    ? U
+    : NonUndefined<T[K]> extends Array<infer U>
+    ? Array<Unpacked<U>>
+    : NonUndefined<T[K]> extends ReadonlyArray<infer U>
+    ? ReadonlyArray<Unpacked<U>>
+    : NonUndefined<T[K]> extends object
+    ? Unpacked<T[K]>
+    : T[K];
+};
```
bluebill1049, line 54: How are this `UnPack` works?

`UnPack<{ tests: 'test', test2: [1, 2, 3] }>`

src/useFieldArray.ts
```diff
@@ -210,7 +210,7 @@ export const useFieldArray = <
 
   const mapCurrentFieldsValueWithState = () => {
     const currentFieldsValue: Partial<FormArrayValues>[] = get(
-      getValues({ nest: true }),
+      getValues(),
```
bluebill1049, line 213: nice pickup.

src/types.ts
kotarella1110: @bluebill1049 oh... Not what I was expecting...

result:

```ts
Unpacked<{ tests: 'test', test2: [1, 2, 3] }>;
// {
//     tests: "test";
//     test2: (3 | 1 | 2)[];
// }
```

expected:

```ts
Unpacked<{ tests: 'test', test2: [1, 2, 3] }>;
// {
//     tests: "test";
//     test2: [1, 2, 3];
// }
```

[TypeScript Playground](https://www.typescriptlang.org/play/?ssl=154&ssc=5&pln=151&pc=24#code/CYUwxgNghgTiAEYD2A7AzgF3gEgHIkxGADUoIBXEALnnJQEsBHS+NATwFsAjJCAbgBQAkAA8ADkhhYMbMQnyESZSgB4B8eABUFGIqQoJRulMDTwoKNgG0AuvAA+8JFwBW4LAF5zl2w6ev3AQA+eC8Ab3V4KzwCXSUDGxoUEAA3EBhBAF94ADItHT1lEEFhcUlpWXlUAFUTEAAzemTgFU0Qr014IxATMzpQRub4AH54ZLSYeBpNEtEJKXgZOXhasSgwAGsiVvb4CI0rAGl4JvgttiR6rUT4XBq6we3NI5sQ7t7b2MKDFSb69JWQUiGlG1WBU1u9wGTSeLzeImMpngAEEYDAoGxfih-pNqkCNCCUWiMSpVusti08fiCTQ7ihatDmq04V0ET0kQAlEBQYCoCBsVHozF-AF48GjLk8vkC4mYsmbbZU8G0qENGEtZ6HV6sxFmZxuMAYcUrFBrBUauHKrQvLIlAD0dvg9WgWH17jQAiWCAAkqbyBhkaE9pFzgBGGiYGBNADmgg05wATElyNx0rbPZV4L6xP6AEJB-ZnEBsUPDCMYKMoWMh4tJz6KfSqQvx4vh1gVmNxgmJ5OpjKRTJBdNerN+jAAYQLNZL5cr1ZbbATZfrcUbIBUkZjtiHAkyghH2f9ABEpwu25uq13E8uCvEm+Cw7POw-a72uGmBzuFwBmWlfO-rhe0bbumDpjP+-gGhgHoHmOACip5FjOwYEkhpZPpeL6Lm+H4aHuu77pmh4YAAYohj7tnOtjDkRY4AOLka2GHzkhdbNmhzFdguS5-g2RQbh2VYgQONHLMRAASjHIUBV61je-5rio7EUTJWF1igKbvv2eFfkhv4od2TErt8qjgho7GoSpgksahmS2OCunccuPC8NyKBcXpNAWRxlHPrZIkERmYljt6iG6JgaA0AA5OFGBRV2sV1lYoYADTwAmaXfjY6ZBQg8oUsRgZePl2yFTuYEjiVwCFVOYGWUZql1YZ2FjJpH5gZkuUmmaBVjvmxWmuSpV9eVjqVYN5rEf1wZNWey5AX4-Rqs0gizaxXkCGt9XSdZq2Oqh23qW1-ZrfhHVdVVxGTgNPXDTmE6jYsmaXWO10zftzXnrtm0fU5zG+I4S2PMAe3wJ1423dVY4njdQ0tMRR6PRDcMI7Vv0+Y16PXhtW2fZxP0HYdOEnej2SAw86qg9t+lAdRBPg89E29fdCGw+aKjEXBSOM5DnNowdFERLjc3MYtFMrQThPccTVMEmdjoM8sL33WRbMUhzY4kdzStM0QxGq+9AsNdZdPncjk30UGVUa-ddHa3lutQ7b-PbV9c6y9LhtSxj33C6x80m3Y5OMkQHtg6DisO7zY6SWrd3+uJ9vdSjMcu3jvmYWt2Ne1LVnu5Luevq1fZh2TtDi6HBc-jj6OuzX3tGzt+d+3Z2VV43PHwC5EBuWLIcg+31P1w36eYw38sHRPkfJxb92hXH8MhUnyv+vPOcaLFkXwAARLF29h4lNAABTfn4oZ+AmACUpsK0AA)

kotarella1110: @bluebill1049 I fixed this!

bluebill1049: awesome stuff! thanks @kotarella1110 for explains.

bluebill1049 (approved):

bluebill1049:
Going into V6! 👯

PR #1508, opened 2020-04-27
feature: useWatch custom hook

bluebill1049:
- [ ] `useWatch` (new) subscribe to registered inputs.

**Reason:** This could isolate the re-render within a particular component without trigger re-render at the form level. This is different from `watch` API, which will re-render at the root of `useForm`

```ts
<input name="test" ref={register} />

function IsolateReRender() {
  const data = useWatch({
    name: 'test',
    control,
    defaultValue: 'default'
  });

  return <div>{data}</div>
}
```


bluebill1049:
I am giving up on the `type` with useWatch :( I will pass the ball to you @kotarella1110 🙏

bluebill1049:
please review this feature whenever you have sometimes too 🙏

src/useWatch.ts
```diff
@@ -0,0 +1,57 @@
...
+import isUndefined from './utils/isUndefined';
+import isString from './utils/isString';
+import generateId from './logic/generateId';
+import { Control, UseWatchProps } from './types';
+
+export const useWatch = <TWatchValues, TControlProp extends Control = Control>({
+  control,
+  name,
+  defaultValue,
+}: UseWatchProps<TControlProp>): TWatchValues => {
```
kotarella1110, line 12: I think you can infer `TWatchValues` ​​from `ControlProp`.
I will fix this.
May I commit directly to this branch?
Or should I fix this after merging this PR?

bluebill1049: yea please commit into the branch 👍

bluebill1049:
i will merge this PR first @kotarella1110 to avoid more conflicts on the way 🙏

svekl:
@bluebill1049 trying out V6 branch and want to use `useWatch` to watch for array of objects. I use `useFieldArray` with inputs with names like
```
foo[0].bar
foo[0].baz
foo[1].bar
foo[1].baz
foo[2].bar
foo[2].baz
```

it gives me form object like
```
{
  foo: [
    {
      bar: 'something',
      baz: 'something',
    },
    {
      bar: 'something',
      baz: 'something',
    },
    {
      bar: 'something',
      baz: 'something',
    },
  ]
}
```
When I try to use `useWatch` to watch on `foo` object it doesn't work. Is this scenario supported?

Tried to look inside `useWatch` and, if I understand right, it is setting up watches, but does not ever call `updateWatchValue` inside. Is it because I don't have `foo` input?

bluebill1049:
Do you have a codesandbox? @svekl i will take a look

svekl:
@bluebill1049 thanks, will try to create one, can you, please, give me advise on how to create it based on V6 branch?

bluebill1049:
clone the repo and checkout V6 branch, create a codesandbox, and coy the entire `src` folder into it. let me know the steps to reproduce the concern as well

svekl:
@bluebill1049 just reproduced same use case in codesandbox and found few scenarios
https://codesandbox.io/s/charming-night-vylfp

**Scenario 1(`register` is used):**
Type something into any of `Watch working:` fields

In this case watch works absolutely fine

**Scenario 2(`Controller` is used):**
Type something into any of `Watch not working:` fields

Expected:
Watch rerenders component as it does in Case 1

**Scenario 3(`fieldArray.append` is used):**
Press `Add foo` button

Expected:
Watch rerenders component as it does in Case 1

**Scenario 4(`fieldArray.append` is used):**
Press `Add foo` button
Type something into any of `Watch working:` or `Watch not working:` fields of new `Foo` item

Expected:
Watch rerenders component as it does in Case 1

bluebill1049:
Thanks @svekl i will be looking into it 👍

bluebill1049:
fixed: https://codesandbox.io/s/strange-greider-o2gmd?file=/src/App.tsx will patch tonight with test coverage.

svekl:
@bluebill1049 Thank you for all you work! Can you, please, also take a look at Scenario 3 and Scenario 4? Watch also doesn't fire on `fieldArray.append`

bluebill1049:
thanks @svekl take a look the updated: https://codesandbox.io/s/strange-greider-o2gmd?file=undefined

svekl:
@bluebill1049 wow, great work! Thanks for addressing it so fast!

ziron0:
@bluebill1049 I tried out the new useWatch feature in our project. Experienced some weird behaviour, so I made a sandbox about it:

https://codesandbox.io/s/infallible-lamarr-uuedg

Seems like when you use watch and useWatch the same time (see the "foo" form) controlled input updates only the "watch", uncontrolled the "useWatch". 

**To Reproduce**

1. Type into FOO form - foo1 input
2. Error: only watch result updated
3. Type into FOO form - foo2 input
4. Error: only useWatch result updated

bluebill1049:
thanks @ziron0 I will investigate the issue.

bluebill1049:
found the issue @ziron0 and patched, however, I probably not going to recommend such usage.  watch() is already trigger global re-render at the root, and it's not a good idea to trigger unneed re-render at child level.

ziron0:
Thanks for the fast patch! @bluebill1049 
We use the watch multiple places in our app (probably few invalid reasons though). 
I was switching them 1 by 1 that's how I found out that they don't "like" each other.

bluebill1049:
@ziron0 good find 👍 after V6 useWatch is probably the one you want to smooth your app's perf 👍

PR #1509, opened 2020-04-27
fix getValues type for FormContext

### LINKED ISSUE #1507: [TS] Types of property 'setError' are incompatible.
**Describe the bug**
This is a ts typings related bug. If the default values contain an array, types of setError are incompatible. This happens when tsconfig has strict set to false, in strict mode the TS error doesn't occur.

```
Type '{ children: Element; register<Element extends FieldElement<MyFormValues> = FieldElement<MyFormValues>>(): (ref: Element) => void; register<Element extends FieldElement<...> = FieldElement<...>>(validationOptions: Partial<...>): (ref: Element) => void; register(name: string, validationOptions?: Partial<...>): void; r...' is not assignable to type 'FormContextValues<MyFormValues | Record<string, unknown>>'.
  Types of property 'setError' are incompatible.
...
```

https://codesandbox.io/s/strange-flower-bs6u3?file=/src/MyForm.tsx

kotarella1110:
close #1507 

CodeSandbox: https://codesandbox.io/s/elegant-brahmagupta-28e3m?file=/src/MyForm.tsx

### DIFF
--- a/src/contextTypes.ts
+++ b/src/contextTypes.ts
@@ -115,7 +115,7 @@ export type FormContextValues<FormValues extends FieldValues = FieldValues> = {
     omitResetState?: OmitResetState,
   ) => void;
   getValues(): IsFlatObject<FormValues> extends false
-    ? Record<string, unknown>
+    ? Record<string, any>
     : FormValues;
   getValues<T extends boolean>(payload: {
     nest: T;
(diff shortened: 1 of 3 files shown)


bluebill1049:
Thanks @kotarella1110

PR #1531, opened 2020-04-29
🐞 fix react-hook-form#1520 return proper proxy object

Photonios:


bluebill1049:
https://codesandbox.io/s/exciting-merkle-gir6v?file=/src/app.test.js with your code change and test from your issue report.

Issue #1535, opened 2020-04-29
[v5.6.1-beta.2] DefaultValues empty string gives undefined


CyanoFresh:
**Describe the bug**
When using `defaultValues` and default value is `""`, `Controller` sets `value` to `undefined` which converts input to uncontrolled.

Only for dot path names. I think problem is somewhere in `utils/get()`.

**To Reproduce**
1. Go to https://codesandbox.io/s/react-hook-form-defaultvalues-ncfnj?file=/src/index.js
2. Open console
3. See that field value with dot path is undefined. The same field accessed by `getValues()[name]` works as expected

**Codesandbox link (Required)**
https://codesandbox.io/s/react-hook-form-defaultvalues-ncfnj?file=/src/index.js

**Additional context**
v5.6.1-beta.2

Sorry if I wasn't meant to report issues for beta

PR #1537, opened 2020-04-29
fix get() for empty string values

CyanoFresh:
Related to https://github.com/react-hook-form/react-hook-form/issues/1535

### DIFF
--- a/src/utils/get.ts
+++ b/src/utils/get.ts
@@ -10,13 +10,9 @@ export default (obj: any, path: string, defaultValue?: any) => {
       obj,
     );
 
-  if (isUndefined(result) || result === obj) {
-    if (!obj[path] && obj[path] !== '') {
-      return defaultValue;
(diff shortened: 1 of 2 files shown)


src/utils/get.ts
```diff
@@ -9,7 +9,14 @@ export default (obj: any, path: string, defaultValue?: any) => {
...
+
+  if (isUndefined(result) || result === obj) {
+    if (!obj[path] && obj[path] !== '') {
+      return defaultValue;
+    }
+
+    return obj[path];
+  }
+
+  return result;
```
bluebill1049, line 21:
```suggestion

  return isUndefined(result) || result === obj
    ? isUndefined(obj[path])
      ? defaultValue
      : obj[path]
    : result;
```

bluebill1049:
Thanks a lot for the PR.

PR #1538, opened 2020-04-29
change naming from FormValues to TFieldValues

kotarella1110:

PR #1542, opened 2020-04-30
improve generics type naming

kotarella1110:
# Generics Type Parameter Naming Conventions

## simple (abstract) type

if generics type parameter does not `extends` keyword to denote our constraint, etc:

* `T`: Type
* `P`: Generics Parameter
* `K`: Key
* `V`: Value
* `A`: Function Argument
* `R`: Return
* `T`, `U` ,`V`  etc: 1nd, 2rd, 3th types
* `T`, `T1`, `T2` etc: 1nd, 2rd, 3th types

example:

```ts
export type IsAny<T> = boolean extends (T extends never ? true : false)
  ? true
  : false;
export type Assign<T extends object, U extends object> = T & Omit<U, keyof T>;
```

## meaningful (concrete) type

if generics type parameter `extends` keyword to denote our constraint, etc:

add `T` prefix.

example:

```ts
export type OnSubmit<TFieldValues extends FieldValues> = (
  data: Unpacked<TFieldValues>,
  event?: React.BaseSyntheticEvent,
) => void | Promise<void>;
export type ValidationOption<
  TValidationValue extends ValidationValue = ValidationValue
> = TValidationValue | ValidationValueMessage<TValidationValue>;
```

Related PR: https://github.com/react-hook-form/react-hook-form/pull/1538

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,6 +19,10 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/no-empty-function': 'off',
     '@typescript-eslint/no-object-literal-type-assertion': 'off',
+    '@typescript-eslint/generic-type-naming': [
+      'error',
+      '(^[A-Z]\\d?$|^T[A-Z][a-zA-Z]+$)',
+    ],
     'react-hooks/rules-of-hooks': 'error',
     'react-hooks/exhaustive-deps': 'error',
     'react/display-name': 'warn',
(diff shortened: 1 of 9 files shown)


.eslintrc.js
```diff
@@ -19,6 +19,10 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/no-empty-function': 'off',
     '@typescript-eslint/no-object-literal-type-assertion': 'off',
+    '@typescript-eslint/generic-type-naming': [
+      'error',
+      '(^[A-Z]\\d?$|^T[A-Z][a-zA-Z]+$)',
+    ],
```
kotarella1110, line 25: I found a awesome eslint rule 👯

bluebill1049: wohooo!

bluebill1049 (approved):

PR #1544, opened 2020-04-30
remove ErrorMessage component

kotarella1110:
Remove `ErrorMessage` component and move back to its own repo: https://github.com/react-hook-form/react-hook-form-error

**Reason:** TS support is getting better with [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and this component's usage is quite low through the year.

 ```diff
- import { ErrorMessage } from 'react-hook-form'
+ import { ErrorMessage } from 'react-hook-form-error'
```


bluebill1049:
Thank you @kotarella1110

PR #1549, opened 2020-05-01
rename FormContext to FormProvider

kotarella1110:
1. `FormContext` rename to `FormProvider`

**Reason:** Align name with all those big libraries such as Redux, styled-components and etc.

```diff
- <FormContext />
+ <FormProvider />
```

2. export `FormContext`

**Reason:** Some users may want to use `FormContext.Consumer`.

```jsx
import { FormContext } from 'react-hook-form';

// in your connected component
render() {
  return (
    <FormContext.Consumer>
      {({ register, handleSubmit }) => {
        // do something with the useForm methods here
      }}
    </FormContext.Consumer>
  )
};
```

### DIFF
--- a/src/useFormContext.tsx
+++ b/src/useFormContext.tsx
@@ -2,9 +2,9 @@ import * as React from 'react';
 import { FieldValues } from './types';
 import { FormContextValues, FormProviderProps } from './contextTypes';
 
-const FormContext = React.createContext<FormContextValues<FieldValues> | null>(
-  null,
-);
+export const FormContext = React.createContext<FormContextValues<
+  FieldValues
+> | null>(null);
 
 export const useFormContext = <
   TFieldValues extends FieldValues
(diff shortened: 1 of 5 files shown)


src/useFormContext.tsx
```diff
@@ -1,28 +1,21 @@
 import * as React from 'react';
 import { FieldValues } from './types';
-import { FormContextValues, FormProps } from './contextTypes';
+import { FormContextValues, FormProviderProps } from './contextTypes';
 
-const FormGlobalContext = React.createContext<FormContextValues<
+export const FormContext = React.createContext<FormContextValues<
```
kotarella1110, line 5: @bluebill1049 Exported `FormContext` for users who want to use `FormContext.Consumer`.
What do you think?

```ts
import { FormContext } from 'react-hook-form';

// in your connected component
render() {
  return (
    <FormContext.Consumer>
      {({ register, handleSubmit }) => {
        // do something with the useForm methods here
      }}
    </FormContext.Consumer>
  )
};
```

bluebill1049 (approved):
awesome!!

PR #1569, opened 2020-05-05
improve FieldName type

kotarella1110:


src/contextTypes.ts
```diff
@@ -35,24 +35,14 @@ export type FormContextValues<
     validationOptions: ValidationOptions,
   ): (ref: TFieldElement | null) => void;
   register(
-    name: IsFlatObject<TFieldValues> extends true
-      ? Extract<keyof TFieldValues, string>
-      : string,
+    name: FieldName<TFieldValues>,
```
bluebill1049, line 38: oh wow, so much simpler!

src/types.ts
```diff
@@ -23,13 +23,19 @@ export type LiteralToPrimitive<T extends any> = T extends string
 
 export type FieldValues = Record<string, any>;
 
-export type FieldName<TFieldValues extends FieldValues> =
+export type InternalFieldName<TFieldValues extends FieldValues> =
   | (keyof TFieldValues & string)
   | string;
```
bluebill1049, line 28: isn't this just `string`?

bluebill1049 (approved):
Amazing! thank you @kotarella1110

src/types.ts
kotarella1110: @bluebill1049 Yes! This is just `string`.
I thought about replacing it with `string`, 
but changed the `FieldName` used internally to `InternalFieldName` in case the type support was improved.
Do you think you should replace it with `string`?

bluebill1049: yea I agree with you. Keep our fingers crossed for the TS update in the near future.

PR #1572, opened 2020-05-05
rename and improve Unpacked type

kotarella1110:


bluebill1049:
so much Ts improvement! loving them! thanks @kotarella1110

PR #1592, opened 2020-05-07
improve ErrorMessage type

kotarella1110:


bluebill1049:
thanks @kotarella1110 🙏 much cleaner

kotarella1110:
@bluebill1049 Should util functions be impose? Or do you want to leave it as is?

bluebill1049:
you mean export util functions? I think we should export less, only export unnecessary things.

kotarella1110:
I used to export only `get` function for ErrorMessage component, but I will remove this.

bluebill1049:
thanks @kotarella1110

PR #1599, opened 2020-05-08
Add German README

pmaier983:
Added a German README and connected it to the other README translation.


bluebill1049:
awesome thank you @pmaier983

PR #1606, opened 2020-05-10
improve Control type

kotarella1110:
Currently, when fixing useForm methods type such as `register` and `getValues`​, we need to fix both the `Control` type and `UseFormMethods` type.
This can cause omissions and mistakes.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { UseFormMethods } from './contextTypes';
 
 export type IsAny<T> = boolean extends (T extends never ? true : false)
   ? true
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -244,50 +245,12 @@ export type FieldValuesFromControl<
   TControl extends Control
 > = TControl extends Control<infer TFieldValues> ? TFieldValues : never;
 
-export type Control<TFieldValues extends FieldValues = FieldValues> = {
+export type Control<TFieldValues extends FieldValues = FieldValues> = Pick<
```
bluebill1049, line 248: OMG!!! genius!!

bluebill1049:
best thing ever!

PR #1610, opened 2020-05-11
fix IsFlatObject type for Date and FileList

kotarella1110:


bluebill1049:
do we need to release this patch?

kotarella1110:
@bluebill1049 Yes. By releasing this, we can correctly judge flat object.

release before:

```ts
type FormValues1 = {
  date: Date;
}

type FormValues2 = {
  file: FileList;
}

type A = IsFlatObject<FormValues1>;
// reuslt: false, expected: true
type B = IsFlatObject<FormValues2>;
// result: false, expected: true
```

release after:

```ts
type FormValues1 = {
  date: Date;
}

type FormValues2 = {
  file: FileList;
}

type A = IsFlatObject<FormValues1>;
// reuslt: true
type B = IsFlatObject<FormValues2>;
// result: true
```

PR #1624, opened 2020-05-13
improve Build

bluebill1049:


src/useForm.ts
```diff
@@ -961,7 +961,8 @@ export function useForm<
           isRadioOrCheckbox && field.options
             ? field.options[field.options.length - 1]
             : field,
-        isRadioOrCheckbox: isRadioOrCheckbox || isSelectInput(ref),
+        isRadioOrCheckbox:
+          isRadioOrCheckbox || isSelectInput(ref as FieldElement),
```
bluebill1049, line 965: @kotarella1110 i had to cast those type here for now, FYI

PR #1626, opened 2020-05-13
be able to use regex in triggerValidation

konstantinkreft:
First of all, awesome project!

I encountered a problem with `triggerValidation` when I wanted to validate a subset of fields. 
For example, afaik there is no way to trigger validation for all `person` and all `children` fields  for the following form example:

```ts
interface Form {
  person: {
    firstName: string;
    lastName: string;
  };
  children: {
    firstName: string;
    lastName: string;
  }[];
  foo: string;
  bar: string;
}
```

With the proposed change that would be possible by calling 
```ts
triggerValidation(/(person.*|children.*)/)
```

That would be awesome especially for use cases in conjunction with `useFieldArray` where the user can add and delete fields e.g. as in the. example above.

### DIFF
--- a/app/src/triggerValidation.tsx
+++ b/app/src/triggerValidation.tsx
@@ -8,6 +8,8 @@ function TriggerValidation() {
     test: string;
     test1: string;
     test2: string;
+    test3: string;
+    test4: string;
   }>();
 
   renderCounter++;
@@ -23,6 +25,12 @@ function TriggerValidation() {
       <input name="test2" ref={register({ required: true })} />
       <div id="test2Error">{errors.test2 && 'required'}</div>
 
+      <input name="test3" ref={register({ required: true })} />
+      <div id="test3Error">{errors.test3 && 'required'}</div>
+
+      <input name="test4" ref={register({ required: true })} />
+      <div id="test4Error">{errors.test4 && 'required'}</div>
+
       <button
         id="single"
         type="button"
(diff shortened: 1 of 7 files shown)


JeromeDeLeon:
Thanks for the feedback. As for the PR, I would like to know your use case of wanting to do that? Maybe we could around it and think of other solutions available instead of doing that, I'm not against it. Its just that, we want to minimize the size as little as possible. We could flag this as "wanted feature" or some sort and wait for other people to see if they want this to be implemented.

bluebill1049:
Thanks a lot for the PR!  I think we need to invest time in a wholist solution for `regex`. I don't think this is a bad idea, I think we probably need to plan better to support other API as well, so it's consistent through this lib's APIs.

kreftk:
I thought about it and agree that there should be a wholist solution if implemented. Also, it is a very valid point that there might not be a demand for it right now.

I found a solution with the control object to get it working without bloating the bundle size of the package.

```ts
const valid = await triggerValidation(
  Object.keys(control.fieldsRef.current).filter(key =>
    regex.test(key),
  ),
);
```

I think for now it would be enough to add a section to the advanced documentation.

bluebill1049:
hey, @kreftk honestly I wouldn't use anything inside the `control` object as they are not part of the API and not versioned, so breaking change can occur.

kreftk:
that's okay for my current use case but yeah it should not be documented then.

PR #1648, opened 2020-05-18
limit types to be exported

kotarella1110:
limit the types export and so we can document all the types been exported.

* NestedValue
* Resolver
* OnSubmit
* Control


bluebill1049:
are there any more types we can export to our users? if that's it, it's good as well. We will document down those 4 types on the website.

kotarella1110:
@bluebill1049 I think there are other types to export. After v6 release, we should add it at the user's request.

bluebill1049:
sounds great to me @kotarella1110

PR #1660, opened 2020-05-20
improve setValue and watch type

kotarella1110:


src/types/contextTypes.ts
```diff
@@ -68,15 +66,14 @@ export type UseFormMethods<TFieldValues extends FieldValues = FieldValues> = {
...
+  >(
     name: TFieldName,
-    value: TFieldName extends keyof TFieldValues
-      ? IsAny<TFieldValues[TFieldName]> extends true
-        ? any
-        : TFieldValues[TFieldName] extends NestedValue<infer U>
-        ? U
-        : UnpackNestedValue<DeepPartial<TFieldValues[TFieldName]>>
-      : LiteralToPrimitive<TFieldValue>,
+    value: TFieldValue extends NestedValue<infer U>
```
bluebill1049, line 74: if the field value, matches nestedValue's value return it's type?

kotarella1110: @bluebill1049 what do you mean?

```ts
const { setValue } = useForm<{ test: NestedValue<{ key: string }> }>();

setValue('test', { key: 'value' });
```

bluebill1049: oh sorry, I read my own sentence again and it's confusing... 

so the type was meant to infer the type inside NestedValue right?

kotarella1110: @bluebill1049 yes. you're completely right!

PR #1670, opened 2020-05-22
change structure types

kotarella1110:
```
src/types/
├── form.ts
├── index.ts
├── props.ts
└── utils.ts
```


kotarella1110:
@bluebill1049 Changed the structure types. What do you think about this structure?

bluebill1049:
> @bluebill1049 Changed the structure types. What do you think about this structure?

love it! looks good and much more organized.

bummey:
Some types (like `FieldValues` and `UnpackNestedValue`) need to be imported from 'react-hook-form/dist/types/form' after this change, is that intended?

bluebill1049:
> @bummey @Photonios We can export any types on user request.
> The reason for limiting exports is that renaming or fixing the exported type can impact users.
> 
> @bluebill1049 This is no problem in exporting it, but I think that we need to add the following sentence somewhere in the document.
> "Types not documented may change at any time by the RHF team."

Thanks @kotarella1110 let's discuss what we want to officially export.

https://github.com/react-hook-form/react-hook-form/discussions/1835

PR #1689, opened 2020-05-24
auto focus with useFieldArray

bluebill1049:
**UX:** improve user experience by autofocus on the newly append item.

Original discussion: https://github.com/react-hook-form/react-hook-form/discussions/1686

**Idea:** 
```
append({ test: 'test' }); // will auto focus into the new field
append({ test: 'test' }, false); // disable auto focus
```

**Affected API:**
- `append`
- `prepend`
- `insert`

**Todo:**
- [x] append automation for action without focus

CSB: https://codesandbox.io/s/react-hook-form-usefieldarray-focus-last-added-2px1i?file=/src/src/useFieldArray.ts

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -173,7 +173,7 @@ export const useFieldArray = <
       shouldRender = true;
     }
 
-    focusIndexRef.current = shouldFocus ? fields.length : -1;
+    focusIndexRef.current = shouldFocus ? allFields.current.length : -1;
 
     shouldRenderFieldArray(shouldRender);
   };
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -36,6 +36,7 @@ export const useFieldArray = <
   keyName = 'id' as TKeyName,
 }: UseFieldArrayOptions<TKeyName, TControl>) => {
   const methods = useFormContext();
+  const focusIndexRef = React.useRef(-1);
   const {
     isWatchAllRef,
     resetFieldArrayFunctionRef,
@@ -58,6 +59,7 @@ export const useFieldArray = <
     renderWatchedInputs,
   } = control || methods.control;
   let shouldRender;
+
   const getDefaultValues = () => [
     ...get(
       fieldArrayDefaultValues.current[getFieldArrayParentName(name)]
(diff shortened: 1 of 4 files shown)


src/useFieldArray.ts
```diff
@@ -415,13 +425,28 @@ export const useFieldArray = <
...
     }
+
+    if (focusIndexRef.current > -1) {
+      for (const key in fieldsRef.current) {
+        const field = fieldsRef.current[key];
+        if (
+          key.startsWith(`${name}[${focusIndexRef.current}]`) &&
+          field!.ref.focus
+        ) {
+          field!.ref.focus();
```
JeromeDeLeon, line 436: Can it be below snippet?

```js
if (key.startsWith(`${name}[${focusIndexRef.current}]`)) {
  field!.ref.focus?.();
}
```

bluebill1049: thanks @JeromeDeLeon didn't want to use optional chaining, it will increase the bundle size with polyfills.

src/useFieldArray.ts
```diff
@@ -208,6 +214,7 @@ export const useFieldArray = <
     }
 
     shouldRenderFieldArray(shouldRender);
+    focusIndexRef.current = shouldFocus ? 0 : -1;
```
JeromeDeLeon, line 217: We can move `-1` and `0` out and replace it with `NOT_FOUND_INDEX` and `START_INDEX` respectively or whatever you see fit for it, for readability. This could be done later for final touch of course. :smile:

src/useFieldArray.ts
JeromeDeLeon: I see. Make sense :smile:

JeromeDeLeon:
I tried the csb and added another input element. Is the behavior correct that the last field will get the focus? I could see that the code is looping through all rows until it finds the focus-able field index and try to focus until the last field.

JeromeDeLeon:
> I tried the csb and added another input element. Is the behavior correct that the last field will get the focus? I could see that the code is looping through all rows until it finds the focus-able field index and try to focus until the last field.

LOL, I just read the discussion. Disregard this one.

Issue #397, opened 2019-10-21
MUI RadioGroup - setValue() doesn't update the value


MinaAmrouche:
Hello,

I use Material UI RadioGroup component and I would like to change the value of this radio input by using `setValue()` when another field in my form changes, but it doesn't seem to have any control on the value of RadioGroup.

Here is a simple example of what I'm trying to do:
https://codesandbox.io/s/withered-cloud-1xeq2

When the value of the first block changes, `setValue` is called to update the value of the second block but nothing happens.

Am I missing something ?

bluebill1049:
hey @michaelbirchler,

I think you probably want to post this issue at Material-UI. I don't think material-ui is updating the UI when `checked` set to `true`

chungweileong94:
Hi @MinaAmrouche,

MUI works differently, the `checked` prop of the `<input/>` in the `RadioGroup` component can only trigger by the user interaction.

From my experience, in order to set the value of the radio group, you have to create a custom component with state to wrap around the RadioGroup component

bluebill1049:
thanks @chungweileong94 for helping, I think MUI is not supporting uncontrolled components. that's the main reason, while react hook form is based on uncontrolled component to have better performance and trigger less re-rendering.

MinaAmrouche:
Thank you for your answer @chungweileong94.
I've decided to create my own Radio component based on MUI style and it works just fine.
You're right @bluebill1049, MUI doesn't support uncontrolled components entirely.

bluebill1049:
Good to hear @MinaAmrouche ❤️🙏🏻

karimja:
Mina

Would you mind sharing the code for your custom component. I am keen to use this library over formik but want to ensure that it will work with MUI.It would be good to see how you have developed the custom component

MinaAmrouche:
@karimja here is my previous example with the custom Radio component :
https://codesandbox.io/s/quiet-currying-2gy6l

I pretty much copied-paste the MUI Radio component so that it looks almost exactly the same and now I can do whatever I want with it.
I hope it helps.

I had problems with 2 MUI components so far:
* Select
* Radio
After spending a few hours trying to make them behave like I needed, creating custom ones seemed to be a good solution for me. But it depends on your use cases of course.

bluebill1049:
going to close this issue @karimja feel free to continue the discussion in here tho. thanks for your help @MinaAmrouche <3

keemor:
[...]

PR #1694, opened 2020-05-25
Add the excellent MUI comprehensive CSB

dandv:
Per https://github.com/react-hook-form/react-hook-form/issues/397#issuecomment-633350553


bluebill1049:
It's actually more than just MUI, may be renamed to Controller example? 

`Controller - (MUI, React-Select and etc)`

PR #1695, opened 2020-05-25
Typo fix: "Miss name" -> "Missing name"

dandv:
Though I saw "**Missing** name" [here](https://github.com/react-hook-form/react-hook-form/discussions/1693)??


bluebill1049:
was try to save few bytes here, you got me hahah

PR #1697, opened 2020-05-25
add TypeScript example

kotarella1110:
- [x] NestedValue
- [x] Native Multiple Input
- [x] OnSubmit
- [x] Resolver
- [x] UseFormMethods
- [x] Control

PR #1698, opened 2020-05-25
fix handleSubmit type for native multiple input

kotarella1110:
In the case of native multiple input, submit value type is wrong.

```ts
const { register, errors, handleSubmit } = useForm<{
   email: NestedValue<string[]>;
}>();

const onSubmit = handleSubmit(data => {
  // data.email type is incorrect. The string is correct, not the string[]
  // (parameter) data: {
  //  email: string[];
  // }
  console.log(data);
});

<input
  multiple
  type="email"
  name="email"
  list="email"
  ref={register({ required: 'This is required.' })}
/>
```

The solution is to add a generics parameter to handleSubmit and allow the user to define the type of submit value.

```ts
handleSubmit<{ email: string }>(data => {
  console.log(data);
});
// const handleSubmit: <{
//     email: string;
// }>(callback: OnSubmit<{
//     email: string;
// }>) => (e?: React.BaseSyntheticEvent<object, any, any> | undefined) => Promise<void>
```

Resolved CodeSandbox: https://codesandbox.io/s/react-hook-form-native-multiple-input-tyhue

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -371,8 +371,8 @@ export type UseFormMethods<TFieldValues extends FieldValues = FieldValues> = {
   getValues<TFieldName extends keyof TFieldValues>(
     names: TFieldName[],
   ): UnpackNestedValue<Pick<TFieldValues, TFieldName>>;
-  handleSubmit: (
-    callback: OnSubmit<TFieldValues>,
+  handleSubmit: <TSubmitFieldValues extends FieldValues = TFieldValues>(
+    callback: OnSubmit<TSubmitFieldValues>,
   ) => (e?: React.BaseSyntheticEvent) => Promise<void>;
   control: Control<TFieldValues>;
 };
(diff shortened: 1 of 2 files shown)


src/types/form.ts
```diff
@@ -371,8 +371,8 @@ export type UseFormMethods<TFieldValues extends FieldValues = FieldValues> = {
   getValues<TFieldName extends keyof TFieldValues>(
     names: TFieldName[],
   ): UnpackNestedValue<Pick<TFieldValues, TFieldName>>;
-  handleSubmit: (
-    callback: OnSubmit<TFieldValues>,
+  handleSubmit: <TSubmitFieldValues extends FieldValues = TFieldValues>(
```
bluebill1049, line 374: cool!

bluebill1049 (approved):
Awesome to give users the option to overwrite the values type

kotarella1110:
@bluebill1049 Would you like to include this PR's codesandbox example in this PR (https://github.com/react-hook-form/react-hook-form/pull/1697) as well?

bluebill1049:
Yes plz. thanks

PR #1705, opened 2020-05-26
fix watch and setValue type

kotarella1110:


bluebill1049:
thanks @kotarella1110

PR #1706, opened 2020-05-26
export UseFormMethods type

kotarella1110:
There are a certain number of users who want to use `UseFormMethods` type.

related: https://github.com/react-hook-form/react-hook-form/pull/1697


bluebill1049:
thanks @kotarella1110

PR #1711, opened 2020-05-27
#1580 Controller with render props API

### LINKED ISSUE #1580: Controller requires it's props for custom components
**Describe the bug**
When using `Controller` with a custom input component, causes the `Controller` to require the `value` and `onChange` props to be required.

**To Reproduce**
1. Open codesandbox
2. Observe the type errors

https://codesandbox.io/s/react-hook-form-controller-template-qrtco

**Expected behavior**
I would expect the typings to be smart enough to know if I have already passed the values or to know that the properties are injected.

**Additional context**
Also appears to be an issue with `<input name="test" />`

bluebill1049:
### 🔍Origin

This PR is led by a TS bug #1580 where required props will require to fulfill again at the `props`.

Example CSB: https://codesandbox.io/s/react-hook-form-controller-template-qrtco

```tsx
type MyInputProps = {
  value: any;
  onChange: () => void;
};

function MyInput(props: MyInputProps) {
  return <input name="FirstName" {...props} />;
}

<Controller
  as={<input name="FirstName" />} // result in TS error for required onChange and value 
  defaultValue="FirstName"
  control={control}
/>
```

-------

### ✏️Solutions

Render prop was purposed by @kotarella1110 which also lead to a discussion around shaping better API. Here are the options 

-------

**OPTION A** only support Render Props for `Controller`

**cons:** 

1. verbose for simple usage 
2. impact API consistency between `ErrorMessage` and `Controller`

**pros:**

1. consistent usage with Controller
2. less code inside lib
3. predictable result 

-------

**OPTION B** support both `as` and `render` 

**cons:** 

1. less intuitive consider providing two options: `as` vs `render`. this could lead to confusion.
2. lack of consistent usage

**pros:**

1. consistent API across `ErrorMessage` & `Controller`
2. more code inside lib
3. simple usage option, plug-in & play
4. the community seems to be keener with both options 

-------

❌ **OPTION C** `render` support render props and component

This approach having issues with good/easy `TS` support and having issue `useFieldArray` (which lead input to remount)

Screen Recording 2020-05-30 at 10 46 28 am

-------

### 📧Poll result

The above options lead us into discussion and poll on twitter. Below is a good result.

image

-------

### 📦What's next?

The plan is to release an **RC.2 version** with **option B** and waiting for the community's feedback for **2 weeks** before releasing the new V6. During the period, if users raise concerns over the new API and we will re-look at **option A** again. 

-------

### 📘Changes

- [x] Controller

Less of a breaking change for existing users, simple usage remain while still give the user the full control with render prop. Remove the following props 

- `onChange` 
- `onChangeName` 
- `onBlur`
- `onBlurName`
- `valueName`

Simple usage as below, will not be impacted:

```tsx
<Controller as={TextField} control={control} name="test" />
```

Usage involed configration and those remmoved props:

```diff
-<Controller 
-  as={CustomInput} 
-  valueName="textValue"
-  onChangeName="onTextChange"
-  control={control} 
-  name="test"  
-/>
+<Controller 
+  render={({ onChange, onBlur, value }) => (
+     <CustomInput onTextChange={onChange} onBlur={onBlur} textValue={value} />
+  )}
+  control={control} 
+  name="test"  
+/>
```

-------

- [x] ErrorMessage

change render props from `children` to `render`, usage change from the ability to use both to either. you can either use `as` or `render`. for multiple error messages, you will have to use `render`

```tsx
<ErrrorMessage errors={errors} as="p" />
<ErrrorMessage errors={errors} render={({message}) => {message}} />
```

-------

**Note:** Both `ErrorMessage` and `Controller` contains `as` and `render` props. Users should  either use `as` or `render`.

```tsx
❌<Controller as={TextField}  render={(props) => <TextField {...props} />}/>
❌<ErrorMessage as={TextField}  render={({ message }) => {message} }/>

✅<Controller as={TextField}  />
✅<Controller render={(props) => <TextField {...props} />}/>

✅<ErrorMessage as={TextField} />
✅<ErrorMessage render={({ message }) => {message} }/>
```

### DIFF
--- a/cypress/integration/controller.ts
+++ b/cypress/integration/controller.ts
@@ -11,7 +11,7 @@ context('controller basic form validation', () => {
     cy.get('#switch').contains('switch Error');
 
     cy.get('#input-checkbox input').click();
-    cy.get('#input-radio-group input').click({ multiple: true });
+    cy.get('input[name="gender1"]').first().click();
     cy.get('#input-textField input').type('test');
     cy.get('#input-select > div > div').click();
     cy.get('.MuiPopover-root ul > li:first-child').click();
@@ -19,7 +19,7 @@ context('controller basic form validation', () => {
     cy.get('#input-ReactSelect > div').click();
     cy.get('#input-ReactSelect > div > div').eq(1).click();
 
-    cy.get('p').should('have.length', 2);
+    cy.get('p').should('have.length', 0);
     cy.get('#renderCount').contains('8');
   });
 
--- a/cypress/integration/controller.ts
+++ b/cypress/integration/controller.ts
@@ -19,7 +19,7 @@ context('controller basic form validation', () => {
     cy.get('#input-ReactSelect > div').click();
     cy.get('#input-ReactSelect > div > div').eq(1).click();
 
-    cy.get('p').should('have.length', 0);
+    cy.get('p').should('have.length', 2);
     cy.get('#renderCount').contains('8');
   });
 
--- a/cypress/integration/controller.ts
+++ b/cypress/integration/controller.ts
@@ -19,8 +19,8 @@ context('controller basic form validation', () => {
     cy.get('#input-ReactSelect > div').click();
     cy.get('#input-ReactSelect > div > div').eq(1).click();
 
-    cy.get('p').should('have.length', 2);
-    cy.get('#renderCount').contains('8');
+    cy.get('.container > p').should('have.length', 6);
+    cy.get('#renderCount').contains('2');
   });
 
   it('should validate the form with onBlur mode and reset the form', () => {
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import isBoolean from './utils/isBoolean';
 import isUndefined from './utils/isUndefined';
-import isReactComponentClass from './utils/isReactComponentClass';
 import get from './utils/get';
 import set from './utils/set';
 import getInputValue from './logic/getInputValue';
@@ -15,11 +14,10 @@ import { ControllerProps } from './types/props';
 const Controller = <TControl extends Control = Control>({
   name,
   rules,
-  render: InnerComponent,
+  render,
   defaultValue,
   control,
   onFocus,
-  ...rest
 }: ControllerProps<TControl>) => {
   const methods = useFormContext();
   const {
@@ -154,22 +152,11 @@ const Controller = <TControl extends Control = Control>({
       );
   }
 
-  return isReactComponentClass(InnerComponent) ? (
-    <InnerComponent
-      {...{
-        onChange,
-        onBlur,
-        ...{ [isCheckboxInput ? 'checked' : VALUE]: value },
-        ...rest,
-      }}
-    />
-  ) : (
-    InnerComponent({
-      onChange,
-      onBlur,
-      value,
-    })
-  );
+  return render({
+    onChange,
+    onBlur,
+    value,
+  });
 };
 
 export { Controller };
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -142,16 +142,12 @@ const Controller = <TControl extends Control = Control>({
   function onChange(
     ...callbackOrEvent: any[]
   ): ((...event: any[]) => void) | void {
-    if (!isFunction(callbackOrEvent[0])) {
-      setValue(name, commonTask(...callbackOrEvent), shouldValidate());
-    }
+    const [firstArg] = callbackOrEvent;
 
-    return (event: any) =>
-      setValue(
-        name,
-        commonTask(callbackOrEvent[0](...event)),
-        shouldValidate(),
-      );
+    return isFunction(firstArg)
+      ? (event: any) =>
+          setValue(name, commonTask(firstArg(event)), shouldValidate())
+      : setValue(name, commonTask(...callbackOrEvent), shouldValidate());
   }
 
   return render({
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import isBoolean from './utils/isBoolean';
 import isUndefined from './utils/isUndefined';
+import isReactComponentClass from './utils/isReactComponentClass';
 import get from './utils/get';
 import set from './utils/set';
 import getInputValue from './logic/getInputValue';
@@ -11,22 +12,15 @@ import { VALUE } from './constants';
 import { Control, Field } from './types/form';
 import { ControllerProps } from './types/props';
 
-const Controller = <
-  TAs extends
-    | React.ReactElement
-    | React.ComponentType<any>
-    | keyof JSX.IntrinsicElements,
-  TControl extends Control = Control
->({
+const Controller = <TControl extends Control = Control>({
   name,
   rules,
-  render,
-  as,
+  render: InnerComponent,
   defaultValue,
   control,
   onFocus,
   ...rest
-}: ControllerProps<TAs, TControl>) => {
+}: ControllerProps<TControl>) => {
   const methods = useFormContext();
   const {
     defaultValuesRef,
@@ -152,15 +146,15 @@ const Controller = <
     onBlur: handleBlur,
   };
 
-  return as
-    ? React.isValidElement(as)
-      ? React.cloneElement(as, props)
-      : React.createElement(as as string, props)
-    : render({
-        value,
-        handleChange,
-        handleBlur,
-      });
+  return isReactComponentClass(InnerComponent) ? (
+    <InnerComponent {...props} />
+  ) : (
+    InnerComponent({
+      value,
+      handleChange,
+      handleBlur,
+    })
+  );
 };
 
 export { Controller };
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -136,8 +136,23 @@ const Controller = <TControl extends Control = Control>({
     }
   };
 
-  const onChange = (event: any): any =>
-    setValue(name, commonTask(event), shouldValidate());
+  function onChange(...event: any[]): void;
+  function onChange(
+    callback: (...args: any[]) => any,
+  ): (...event: any[]) => void;
+  function onChange(
+    ...callbackOrEvent: any[]
+  ): ((...event: any[]) => void) | void {
+    if (typeof callbackOrEvent[0] !== 'function') {
+      setValue(name, commonTask(callbackOrEvent), shouldValidate());
+    }
+    return (event: any) =>
+      setValue(
+        name,
+        commonTask(callbackOrEvent[0](...event)),
+        shouldValidate(),
+      );
+  }
 
   return isReactComponentClass(InnerComponent) ? (
     <InnerComponent
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -145,6 +145,9 @@ const Controller = <
 
   function onChange(...event: any[]): void;
   function onChange(
+    callback: (...args: any[]) => any,
+  ): (...event: any[]) => void;
+  function onChange(
     ...callbackOrEvent: any[]
   ): ((...event: any[]) => void) | void {
     const [firstArg] = callbackOrEvent;
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -63,7 +63,7 @@ const Controller = <TControl extends Control = Control>({
     });
 
   const commonTask = (event: any[]) => {
-    const data = getInputValue(event, isCheckboxInput);
+    const data = getInputValue(event[0], isCheckboxInput);
     setInputStateValue(data);
     valueRef.current = data;
     return data;
@@ -147,7 +147,7 @@ const Controller = <TControl extends Control = Control>({
     return isFunction(firstArg)
       ? (event: any) =>
           setValue(name, commonTask(firstArg(event)), shouldValidate())
-      : setValue(name, commonTask(...callbackOrEvent), shouldValidate());
+      : setValue(name, commonTask(callbackOrEvent), shouldValidate());
   }
 
   return render({
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -9,7 +9,7 @@ import skipValidation from './logic/skipValidation';
 import isNameInFieldArray from './logic/isNameInFieldArray';
 import { useFormContext } from './useFormContext';
 import { VALUE } from './constants';
-import { Control, Field } from './types/form';
+import { Control, Field, EventFunction } from './types/form';
 import { ControllerProps } from './types/props';
 
 const Controller = <TControl extends Control = Control>({
@@ -19,7 +19,6 @@ const Controller = <TControl extends Control = Control>({
   defaultValue,
   control,
   onFocus,
-  ...rest
 }: ControllerProps<TControl>) => {
   const methods = useFormContext();
   const {
@@ -140,20 +139,15 @@ const Controller = <TControl extends Control = Control>({
     setValue(name, commonTask(event), shouldValidate());
 
   const props = {
-    name,
-    ...rest,
-    onChange: onChange,
-    onBlur: onBlur,
+    onChange,
+    onBlur,
+    value,
   };
 
   return isReactComponentClass(InnerComponent) ? (
     <InnerComponent {...props} />
   ) : (
-    InnerComponent({
-      value,
-      onChange,
-      onBlur,
-    })
+    InnerComponent(props)
   );
 };
 
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -62,8 +62,8 @@ const Controller = <TControl extends Control = Control>({
       isSubmitted,
     });
 
-  const commonTask = (...event: any[]) => {
-    const data = getInputValue(event[0], isCheckboxInput);
+  const commonTask = (event: any[]) => {
+    const data = getInputValue(event, isCheckboxInput);
     setInputStateValue(data);
     valueRef.current = data;
     return data;
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -9,7 +9,7 @@ import skipValidation from './logic/skipValidation';
 import isNameInFieldArray from './logic/isNameInFieldArray';
 import { useFormContext } from './useFormContext';
 import { VALUE } from './constants';
-import { Control, Field, EventFunction } from './types/form';
+import { Control, Field } from './types/form';
 import { ControllerProps } from './types/props';
 
 const Controller = <TControl extends Control = Control>({
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -7,7 +7,7 @@ import getInputValue from './logic/getInputValue';
 import skipValidation from './logic/skipValidation';
 import isNameInFieldArray from './logic/isNameInFieldArray';
 import { useFormContext } from './useFormContext';
-import { VALUE } from './constants';
+import { VALIDATION_MODE, VALUE } from './constants';
 import { Control, Field } from './types/form';
 import { ControllerProps } from './types/props';
 
@@ -148,8 +148,8 @@ const Controller = <
   const props = {
     name,
     ...rest,
-    onChange: handleChange,
-    onBlur: handleBlur,
+    [VALIDATION_MODE.onChange]: handleChange,
+    [VALIDATION_MODE.onBlur]: handleBlur,
   };
 
   return as
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -7,8 +7,8 @@ import getInputValue from './logic/getInputValue';
 import skipValidation from './logic/skipValidation';
 import isNameInFieldArray from './logic/isNameInFieldArray';
 import { useFormContext } from './useFormContext';
-import { VALIDATION_MODE, VALUE } from './constants';
-import { Control, EventFunction, Field } from './types/form';
+import { VALUE } from './constants';
+import { Control, Field } from './types/form';
 import { ControllerProps } from './types/props';
 
 const Controller = <
@@ -20,12 +20,8 @@ const Controller = <
 >({
   name,
   rules,
-  as: InnerComponent,
-  onBlur,
-  onChange,
-  onChangeName = VALIDATION_MODE.onChange,
-  onBlurName = VALIDATION_MODE.onBlur,
-  valueName,
+  render,
+  as,
   defaultValue,
   control,
   onFocus,
@@ -80,9 +76,6 @@ const Controller = <
     return data;
   };
 
-  const eventWrapper = (event: EventFunction) => (...arg: any): any =>
-    setValue(name, commonTask(event(...arg)), shouldValidate());
-
   const registerField = React.useCallback(() => {
     if (!isNotFieldArray) {
       removeFieldEventListener(fieldsRef.current[name] as Field, true);
@@ -135,38 +128,39 @@ const Controller = <
     }
   });
 
+  const handleBlur = () => {
+    if (
+      readFormStateRef.current.touched &&
+      !get(touchedFieldsRef.current, name)
+    ) {
+      set(touchedFieldsRef.current, name, true);
+      reRender();
+    }
+
+    if (shouldReValidateOnBlur) {
+      trigger(name);
+    }
+  };
+
+  const handleChange = (event: any): any =>
+    setValue(name, commonTask(event), shouldValidate());
+
   const props = {
     name,
     ...rest,
-    ...(onChange
-      ? { [onChangeName]: eventWrapper(onChange) }
-      : {
-          [onChangeName]: (event: any): any =>
-            setValue(name, commonTask(event), shouldValidate()),
-        }),
-    [onBlurName]: (...args: any) => {
-      if (onBlur) {
-        onBlur(...args);
-      }
-
-      if (
-        readFormStateRef.current.touched &&
-        !get(touchedFieldsRef.current, name)
-      ) {
-        set(touchedFieldsRef.current, name, true);
-        reRender();
-      }
-
-      if (shouldReValidateOnBlur) {
-        trigger(name);
-      }
-    },
-    ...{ [valueName || (isCheckboxInput ? 'checked' : VALUE)]: value },
+    onChange: handleChange,
+    onBlur: handleBlur,
   };
 
-  return React.isValidElement(InnerComponent)
-    ? React.cloneElement(InnerComponent, props)
-    : React.createElement(InnerComponent as string, props);
+  return as
+    ? React.isValidElement(as)
+      ? React.cloneElement(as, props)
+      : React.createElement(as as string, props)
+    : render({
+        value,
+        handleChange,
+        handleBlur,
+      });
 };
 
 export { Controller };
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -150,7 +150,7 @@ const Controller = <TControl extends Control = Control>({
       : setValue(name, commonTask(callbackOrEvent), shouldValidate());
   }
 
-  return React.createElement(render, {
+  return render({
     onChange,
     onBlur,
     value,
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -3,7 +3,6 @@ import isBoolean from './utils/isBoolean';
 import isUndefined from './utils/isUndefined';
 import get from './utils/get';
 import set from './utils/set';
-import isFunction from './utils/isFunction';
 import getInputValue from './logic/getInputValue';
 import skipValidation from './logic/skipValidation';
 import isNameInFieldArray from './logic/isNameInFieldArray';
@@ -143,19 +142,8 @@ const Controller = <
     }
   };
 
-  function onChange(...event: any[]): void;
-  function onChange(
-    callback: (...args: any[]) => any,
-  ): (...event: any[]) => void;
-  function onChange(
-    ...callbackOrEvent: any[]
-  ): ((...event: any[]) => void) | void {
-    const [firstArg] = callbackOrEvent;
-
-    return isFunction(firstArg)
-      ? (event: any) =>
-          setValue(name, commonTask(firstArg(event)), shouldValidate())
-      : setValue(name, commonTask(callbackOrEvent), shouldValidate());
+  function onChange(...event: any[]): void {
+    setValue(name, commonTask(event), shouldValidate());
   }
 
   const handlerProps = {
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -150,7 +150,7 @@ const Controller = <TControl extends Control = Control>({
       : setValue(name, commonTask(callbackOrEvent), shouldValidate());
   }
 
-  return render({
+  return React.createElement(render, {
     onChange,
     onBlur,
     value,
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -158,10 +158,14 @@ const Controller = <
       : setValue(name, commonTask(callbackOrEvent), shouldValidate());
   }
 
+  const handlerProps = {
+    onChange: React.useCallback(onChange, []),
+    onBlur: React.useCallback(onBlur, []),
+  };
+
   const props = {
     ...rest,
-    onChange,
-    onBlur,
+    ...handlerProps,
     ...{ [isCheckboxInput ? 'checked' : VALUE]: value },
   };
 
@@ -171,8 +175,7 @@ const Controller = <
       : React.createElement(as as string, props)
     : render
     ? render({
-        onChange,
-        onBlur,
+        ...handlerProps,
         value,
       })
     : null;
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -159,9 +159,9 @@ const Controller = <
   }
 
   const props = {
+    ...rest,
     onChange,
     onBlur,
-    ...rest,
     ...{ [isCheckboxInput ? 'checked' : VALUE]: value },
   };
 
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -142,9 +142,9 @@ const Controller = <
     }
   };
 
-  function onChange(...event: any[]): void {
+  const onChange = (...event: any[]) => {
     setValue(name, commonTask(event), shouldValidate());
-  }
+  };
 
   const handlerProps = {
     onChange: React.useCallback(onChange, []),
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -63,8 +63,8 @@ const Controller = <TControl extends Control = Control>({
       isSubmitted,
     });
 
-  const commonTask = (event: any) => {
-    const data = getInputValue(event, isCheckboxInput);
+  const commonTask = (...event: any[]) => {
+    const data = getInputValue(event[0], isCheckboxInput);
     setInputStateValue(data);
     valueRef.current = data;
     return data;
@@ -144,7 +144,7 @@ const Controller = <TControl extends Control = Control>({
     ...callbackOrEvent: any[]
   ): ((...event: any[]) => void) | void {
     if (typeof callbackOrEvent[0] !== 'function') {
-      setValue(name, commonTask(callbackOrEvent), shouldValidate());
+      setValue(name, commonTask(...callbackOrEvent), shouldValidate());
     }
     return (event: any) =>
       setValue(
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -12,14 +12,22 @@ import { VALUE } from './constants';
 import { Control, Field } from './types/form';
 import { ControllerProps } from './types/props';
 
-const Controller = <TControl extends Control = Control>({
+const Controller = <
+  TAs extends
+    | React.ReactElement
+    | React.ComponentType<any>
+    | keyof JSX.IntrinsicElements,
+  TControl extends Control = Control
+>({
   name,
   rules,
+  as,
   render,
   defaultValue,
   control,
   onFocus,
-}: ControllerProps<TControl>) => {
+  ...rest
+}: ControllerProps<TAs, TControl>) => {
   const methods = useFormContext();
   const {
     defaultValuesRef,
@@ -150,11 +158,24 @@ const Controller = <TControl extends Control = Control>({
       : setValue(name, commonTask(callbackOrEvent), shouldValidate());
   }
 
-  return render({
+  const props = {
     onChange,
     onBlur,
-    value,
-  });
+    ...rest,
+    ...{ [isCheckboxInput ? 'checked' : VALUE]: value },
+  };
+
+  return as
+    ? React.isValidElement(as)
+      ? React.cloneElement(as, props)
+      : React.createElement(as as string, props)
+    : render
+    ? render({
+        onChange,
+        onBlur,
+        value,
+      })
+    : null;
 };
 
 export { Controller };
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -3,6 +3,7 @@ import isBoolean from './utils/isBoolean';
 import isUndefined from './utils/isUndefined';
 import get from './utils/get';
 import set from './utils/set';
+import isFunction from './utils/isFunction';
 import getInputValue from './logic/getInputValue';
 import skipValidation from './logic/skipValidation';
 import isNameInFieldArray from './logic/isNameInFieldArray';
@@ -141,9 +142,10 @@ const Controller = <TControl extends Control = Control>({
   function onChange(
     ...callbackOrEvent: any[]
   ): ((...event: any[]) => void) | void {
-    if (typeof callbackOrEvent[0] !== 'function') {
+    if (!isFunction(callbackOrEvent[0])) {
       setValue(name, commonTask(...callbackOrEvent), shouldValidate());
     }
+
     return (event: any) =>
       setValue(
         name,
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -151,8 +151,8 @@ const Controller = <TControl extends Control = Control>({
   }
 
   return render({
-    onChange,
-    onBlur,
+    onChange: React.useCallback(onChange, []),
+    onBlur: React.useCallback(onBlur, []),
     value,
   });
 };
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -142,9 +142,8 @@ const Controller = <
     }
   };
 
-  const onChange = (...event: any[]) => {
+  const onChange = (...event: any[]) =>
     setValue(name, commonTask(event), shouldValidate());
-  };
 
   const handlerProps = {
     onChange: React.useCallback(onChange, []),
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -122,7 +122,7 @@ const Controller = <TControl extends Control = Control>({
     }
   });
 
-  const handleBlur = () => {
+  const onBlur = () => {
     if (
       readFormStateRef.current.touched &&
       !get(touchedFieldsRef.current, name)
@@ -136,14 +136,14 @@ const Controller = <TControl extends Control = Control>({
     }
   };
 
-  const handleChange = (event: any): any =>
+  const onChange = (event: any): any =>
     setValue(name, commonTask(event), shouldValidate());
 
   const props = {
     name,
     ...rest,
-    onChange: handleChange,
-    onBlur: handleBlur,
+    onChange: onChange,
+    onBlur: onBlur,
   };
 
   return isReactComponentClass(InnerComponent) ? (
@@ -151,8 +151,8 @@ const Controller = <TControl extends Control = Control>({
   ) : (
     InnerComponent({
       value,
-      handleChange,
-      handleBlur,
+      onChange,
+      onBlur,
     })
   );
 };
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -7,7 +7,7 @@ import getInputValue from './logic/getInputValue';
 import skipValidation from './logic/skipValidation';
 import isNameInFieldArray from './logic/isNameInFieldArray';
 import { useFormContext } from './useFormContext';
-import { VALIDATION_MODE, VALUE } from './constants';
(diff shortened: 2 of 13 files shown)


bluebill1049:
Let me know what your guys think please, thanks for @kotarella1110's suggestion.

src/controller.tsx
```diff
@@ -135,38 +128,39 @@ const Controller = <
...
+      !get(touchedFieldsRef.current, name)
+    ) {
+      set(touchedFieldsRef.current, name, true);
+      reRender();
+    }
+
+    if (shouldReValidateOnBlur) {
+      trigger(name);
+    }
+  };
```
kotarella1110, line 143: I think you need to do the following to make the onBlur event handleable by the user:

https://codesandbox.io/s/react-hook-form-issue-1580-59soi?file=/src/src/controller.tsx:2544-3180

```ts
const commonBlurTask = () => {
  if (
    readFormStateRef.current.touched &&
    !get(touchedFieldsRef.current, name)
  ) {
    set(touchedFieldsRef.current, name, true);
    reRender();
  }

  if (shouldReValidateOnBlur) {
    trigger(name);
  }
};

function handleBlur(...args: any): void;
function handleBlur(handler: EventFunction): (...args: any) => void;
function handleBlur(handlerOrArgs: any): ((...args: any) => void) | void {
  if (typeof handlerOrArgs !== 'function') {
    commonBlurTask();
  }
  return (...args: any) => {
    handlerOrArgs(args);
    commonBlurTask();
  };
}
```

```ts
<Controller 
  name="test" 
  control={control} 
  render={({ handleChange, handleBlur, value}) => <CustomInput
      value={value}
      onChange={handleChange}
      onBlur={handleBlur((...args) => { console.log(args) })}
    />
/>
```

kotarella1110 (changes requested):
I think it's better to remove `as` prop, leaving only render prop.
The reason is that if you use as prop, we need to leave `onBlur`, `onChange`, `onChangeName`, `onBlurName`, `onBlur`, `valueName` props.

```ts
export type ControllerProps<TControl extends Control = Control> = {
  name: FieldName<FieldValuesFromControl<TControl>>;
  as?: React.ReactElement;
  rules?: ValidationOptions;
  onFocus?: () => void;
  onBlur?: EventFunction;
  onChangeName?: string;
  onBlurName?: string;
  valueName?: string;
  render?: (data: {
    handleChange: EventFunction;
    handleBlur: EventFunction;
    value: any;
  }) => React.ReactElement;
  defaultValue?: unknown;
  control?: TControl;
};
```

What if we want to customize the return value? We should use `onChange` prop.

```ts
<Controller
  name="test" 
  control={control}
  as={<CustomInput />}
  onChange={(e, value) => value}
/>
```

What if the `onChange` event name is `onTextChange`? We should use `onChangeName` prop.

```ts
<Controller
  name="test" 
  control={control}
  as={<CustomInput />}
  onChange={(e, value) => value}
  onChangeName="onTextChange"
/>
```

If we have only render prop, `Controller` don't need these props.

```ts
export type ControllerProps<TControl extends Control = Control> = {
  name: FieldName<FieldValuesFromControl<TControl>>;
  rules?: ValidationOptions;
  onFocus?: () => void;
  render: (data: {
    handleChange: EventFunction;
    handleBlur: EventFunction;
    value: any;
  }) => React.ReactElement;
  defaultValue?: unknown;
  control?: TControl;
};
```

```ts
<Controller 
  name="test" 
  control={control} 
  render={({ handleChange, handleBlur, value}) => <CustomInput
      value={value}
      onTextChange={(e, value) => value}
      onBlur={handleBlur}
    />
/>
```

With render props only, it's simpler with fewer props.
RHF users can also easily predict what they are doing in the controller by passing event handlers or values ​​to the Controlled Component (`<ControlledComponent value={value} onChange={handleChange} onBlur={handleBlur} />`).

Note: I understand that removing `as` prop is a breaking change and will impact existing users.

bluebill1049:
Thanks, @kotarella1110 I think what I refer is to use either, maybe `as` should rename to `element` like react-router. so the user still gets easy to use (if nothing onChange, onBlur, doesn't need to be renamed).

bluebill1049:
I did a similar thing with `react-simple-animate`. users can use the following syntax
```
<Animate>Test</Animate>

<Aniamte render={} /> // if you want to customise the dom node
```

src/controller.tsx
```diff
@@ -135,38 +128,39 @@ const Controller = <
...
+
+  const handleChange = (event: any): any =>
+    setValue(name, commonTask(event), shouldValidate());
```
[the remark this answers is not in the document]
bluebill1049: Yes, you are right, please commit that change. 🙏  (to my branch)

src/controller.tsx
bluebill1049: we don't really worry about what's return by `onBlur`.

kotarella1110: OK. If RHF users wants to do something in the blur event, they need to do the following:

```ts
<Controller 
  name="test" 
  control={control} 
  render={({ handleChange, handleBlur, value}) => <CustomInput
      value={value}
      onChange={handleChange}
      onBlur={(e) => {
        // do something here
        // console.log(e)
        handleBlur(e)
      }}
    />
/>
```

src/controller.tsx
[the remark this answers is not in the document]
kotarella1110: @bluebill1049 I will fix 👍

bluebill1049:
after @stramel 's feedback I think the following is best:

```
// Supported
<Controller name="test" control={control} render={<TextFields myUniqueProp={state} />} />
<Controller name="test" control={control} render={({onChange, onBlur, ...props}) => <TextFields handleChange={onChange} handleBlur={onBlur} {..props} />} />
```

src/controller.tsx
bluebill1049: yes, we should document that.

kotarella1110:
@bluebill1049 if rename handleChange to onChange and handleBlur to onBlur, user can simplify the code.

```diff
export type ControllerProps<TControl extends Control = Control> = {
  name: FieldName<FieldValuesFromControl<TControl>>;
  rules?: ValidationOptions;
  onFocus?: () => void;
  render: (data: {
-    handleChange: EventFunction;
+   onChange: EventFunction;
-    handleBlur: EventFunction;
+   onBlur: EventFunction;
    value: any;
  }) => React.ReactElement;
  defaultValue?: unknown;
  control?: TControl;
};
```

```ts
<Controller 
  name="test" 
  control={control} 
  render={(props) => <CustomInput {...props} />
/>
```

bluebill1049:
love it @kotarella1110!

bluebill1049:
> after @stramel 's feedback I think the following is best:
> 
> ```
> // Supported
> <Controller name="test" control={control} render={<TextFields myUniqueProp={state} />} />
> <Controller name="test" control={control} render={({onChange, onBlur, ...props}) => <TextFields handleChange={onChange} handleBlur={onBlur} {..props} />} />
> ```

what do you think about this proposal @kotarella1110 ?

JeromeDeLeon:
I assumed this fix the TS that's why we want to provide render props, am I right? @bluebill1049  @kotarella1110

bluebill1049:
> I assumed this fixes the TS that's why we want to provide render props, am I right? @bluebill1049 @kotarella1110

That's the origin, but at the same time, I think this will improve the API.

stramel:
Do we have preference on the prop name? 

`as`, `render`, `element`, `input` or other suggestions?

JeromeDeLeon:
I agreed with all the decision of @kotarella1110 , like we should really remove `as` as this will not be necessary anymore unless there's a fix to it for TS. As for @stramel's idea, 

```jsx
<Controller name="test" control={control} render={<TextFields myUniqueProp={state} />} />

// is the same as below?

<Controller name="test" control={control} as={<TextFields myUniqueProp={state} />} />
```

Is there any reason for supporting it in `render` prop? Just asking :wink:

I've come to like this idea of `render` thing. :smile:

bluebill1049:
> Do we have preference on the prop name?
> 
> `as`, `render`, `element`, `input` or other suggestions?

I think `render` is pretty standard for render props. probably have to change `ErrorMessage` to follow this as well.

bluebill1049:
> I agreed with all the decision of @kotarella1110 , like we should really remove `as` as this will not be necessary anymore unless there's a fix to it for TS. As for @stramel's idea,
> 
> ```js-jsx
> <Controller name="test" control={control} render={<TextFields myUniqueProp={state} />} />
> 
> // is the same as below?
> 
> <Controller name="test" control={control} as={<TextFields myUniqueProp={state} />} />
> ```
> 
> Is there any reason for supporting it in `render` prop? Just asking 😉
> 
> I've come to like this idea of `render` thing. 😄

Yes, I agree.

what about the following option:
```
<Controller name="test" control={control} render={<TextFields myUniqueProp={state} />} />
<Controller name="test" control={control} render={({onChange, onBlur, ...props}) => <TextFields handleChange={onChange} handleBlur={onBlur} {..props} />} />
```

so if users don't require any prop wiring, they can just use slot it in without the need to render props.

JeromeDeLeon:
I haven't seen any cases where below applies, I have only seen `render function` so far.

```jsx
<Controller name="test" control={control} render={<TextFields myUniqueProp={state} />} />
```

JeromeDeLeon:
Though the reason is valid :) We could move it out to `element` prop like you said.

bluebill1049:
Happy to listen everyone's thoughts, I just thought to render prop is convenient when you need to wire those onChange and value, for standard stuff `render={<TextField />}` is short and sweet. I think most people try to follow the pattern expose `value`, `onBlur` and `onChange`.

stramel:
The main thing that I'm worried about is the typing on the element without the render props. It may have to be render props or component. That would allow it to be short and sweet.

JeromeDeLeon:
I'm good with using `render` like what @stramel suggested. :smile:

bluebill1049:
> The main thing that I'm worried about is the typing on the element without the render props. It may have to be render props or component. That would allow it to be short and sweet.

Yes, i agree, 

Props `render` support `renderProps` or `component` is the best, short and sweet.

stramel:
I also like the fact this cuts a few bytes off the size.

bluebill1049:
let me know your thoughts as well @kotarella1110 if we can all align, I will proceed with the change.

kotarella1110:
`render={<CustomInput />}` is very simple.
But if `onChange` is required in the `CustomInput`'s props, it will cause type error.

```ts
type CustomInput = {
  onChange: (e: any) => any; // required
};
<Controller name="test" control={control} render={<CustomInput />} /> // type error
```

bluebill1049:
> `render={<CustomInput />}` is very simple.
> But if `onChange` is required in the `CustomInput`'s props, it will cause type error.
> 
> ```ts
> type CustomInput = {
>   onChange: (e: any) => any; // required
> };
> <Controller name="test" control={control} render={<CustomInput />} /> // type error
> ```

I think if it's required, then user has the opportunity to use `renderProps`. I guess it's optional.

bluebill1049:
I think this will tick both boxes, in terms of simplicity and type support.

stramel:
Yeah, I was trying to come up with a solution around it but I think the only way would be to do component.

So I think everyone is in agreement?

```
// Support

render={MyInput}
render={({ onChange, onBlur, value, ...props }) => <MyInput handleChange={onChange} handleBlur={onBlur} myValue={value} {...props} customProp={state.isCustom} />}
```

kotarella1110:
> Yeah, I was trying to come up with a solution around it but I think the only way would be to do component.
> 
> So I think everyone is in agreement?
> 
> ```
> // Support
> 
> render={MyInput}
> render={({ onChange, onBlur, value, ...props }) => <MyInput handleChange={onChange} handleBlur={onBlur} myValue={value} {...props} customProp={state.isCustom} />}
> ```

awesome! all OK!

bluebill1049:
I will start working on it tonight, but feel free to commit to my branch. if you find the thing is missing and incorrect.

bluebill1049:
maybe we should fix ErrorMessage as well, so the API is consistent. 

```
<ErrorMessage render={} />
```

and remove `children` and `as` props.

kotarella1110:
> @kotarella1110 maybe we should fix ErrorMessage as well, so the API is consistent.
> 
> ```
> <ErrorMessage render={} />
> ```
> 
> and remove `children` and `as` props.

I agree with you.
This is OK?

```ts
// not support
<ErrorMessage render={CustomMessage} />
// support
<ErrorMessagee rennder={({ message }) => <CustomMessage>{message}</CustomMessage>} />
```

bluebill1049:
yea sounds good @kotarella1110 consistent API is great.

kotarella1110:
I will fix ErrorMessage component 👍

src/controller.tsx
```diff
@@ -152,15 +146,15 @@ const Controller = <
...
-    ? React.isValidElement(as)
-      ? React.cloneElement(as, props)
-      : React.createElement(as as string, props)
-    : render({
-        value,
-        handleChange,
-        handleBlur,
-      });
+  return isReactComponentClass(InnerComponent) ? (
+    <InnerComponent {...props} />
```
bluebill1049, line 150: we can't use `clone`?

src/controller.tsx
```diff
@@ -152,15 +146,15 @@ const Controller = <
...
-  return as
-    ? React.isValidElement(as)
-      ? React.cloneElement(as, props)
-      : React.createElement(as as string, props)
-    : render({
-        value,
-        handleChange,
-        handleBlur,
-      });
+  return isReactComponentClass(InnerComponent) ? (
```
bluebill1049, line 149: do we need this check? or `isValidElement` is good enough?

src/utils/isReactComponentClass.ts
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+export default (component: any): component is React.ComponentClass<any> =>
```
bluebill1049, line 3: will `isValidElement` enough for the check instead of this utility function?

src/controller.tsx
bluebill1049: hmmm i see <Test /> will return false. is there any better way then the function below?

bluebill1049: https://codesandbox.io/s/unruffled-mahavira-x5hnw?file=/src/App.js it doesn't really work.

kotarella1110: `isReactComponentClass` is is a function that validates a class component.

```ts
class ClassComponent extends React.Component {
  render() {
    return <div />;
  }
}

const FunctionComponent = () => {
  return <div />;
};

isReactComponentClass(ClassComponent) // true
isReactComponentClass(FunctionComponent) // true
```

https://codesandbox.io/s/silly-chaplygin-tb74k

bluebill1049: sorry, why are we testing for Class component?

src/controller.tsx
```diff
@@ -135,38 +122,54 @@ const Controller = <
...
+      );
+  }
+
+  return isReactComponentClass(InnerComponent) ? (
+    <InnerComponent
+      {...{
+        onChange,
+        onBlur,
+        ...{ [isCheckboxInput ? 'checked' : VALUE]: value },
+        ...rest,
```
kotarella1110, line 163: I don't think rest props are unnecessary. I think users should use render prop if they want to pass custom props to `ControlledComponent`.

```ts
type CustomInputProps = {
  custom: string;
  onChange?: (...args: any[]) => any;
  onBlur?: (...args: any[]) => any;
  value?: string;
}

<Controller 
  name="test" 
  control={control} 
  render={(props) => <CustomInput {...props} custom="custom" />
/>
```

bluebill1049: what if user just wants to add a className?

`<Controller className="test" name="test" control={control} />`

src/controller.tsx
kotarella1110: `React.ComponentType` is union of `React.ComponentClass` and `React.FunctionComponent`.


```ts
type React.ComponentType<P = {}> = React.ComponentClass<P, any> | React.FunctionComponent<P>
```

The Controlled Input may also be a Class Component.

bluebill1049: but the check right now said if it's a class component do <Xxxx /> otherwise render props?

src/controller.tsx
JeromeDeLeon: Like what @stramel said
> ```
> // Support
> 
> render={MyInput}
> render={({ onChange, onBlur, value, ...props }) => <MyInput handleChange={onChange} handleBlur={onBlur} myValue={value} {...props} customProp={state.isCustom} />}
> ```

If we would support passing `MyInput` directly then we need `rest` but for me since we now have render props, I will leave everything to the `InnerComponent` instead of passing it to `Controller`. Just my opinion though.

src/controller.tsx
kotarella1110: @bluebill1049 Yes. Just as you said.

src/controller.tsx
kotarella1110: I think user should do the following:

```ts
<Controller 
  name="test" 
  control={control} 
  render={(props) => <CustomInput {...props} className="test" />
/>
```

src/controller.tsx
JeromeDeLeon: Couldn't we just check if it's a component, because either way they are both component that can be rendered on screen right?

src/controller.tsx
bluebill1049: hmmm, I think mixed with `Component ` with `render props` sounds good on paper, but during implementation is not great, especially with the `component` type check that @kotarella1110 introduced. I think maybe we should make a decision on the following

1. use two props `as` or `render`
con: two props
pro: we still have good DX with `as` when a component is simple

2. use render props only
con: lost the DX with a simple component
prop:  less headache and easy to implement.

src/controller.tsx
bluebill1049: yes, @JeromeDeLeon that's what I am after here, we shouldn't differentiate Class vs Functional component. but after reading through StackOverflow, 

it's not a simple check. https://stackoverflow.com/questions/33199959/how-to-detect-a-react-component-vs-a-react-element

src/controller.tsx
kotarella1110: I agree with the second proposal.

bluebill1049: How about this plan? Let's go all the way with `render props` first. Then worry about component after. at least we nail one goal first before the other.

bluebill1049: if people miss `as` we can introduce it again and it won't be **type** safe.

bluebill1049: make sure you do a code review on this PR when it's ready @JeromeDeLeon. this is your territory that we are messing around 👍

bluebill1049: sorry for taking the wrong direction and wasting your effort @kotarella1110 😭

kotarella1110: @bluebill1049 do not worry 👍

src/utils/isReactComponentClass.ts
JeromeDeLeon: Shouldn't be just return a boolean? I just tried it with codesandbox and it returns an object.

bluebill1049: this is removed :)

src/controller.tsx
JeromeDeLeon: Will do. Later tonight.

src/controller.tsx
JeromeDeLeon: Yes, I also read that one and think that it's not as simple as it is.

src/controller.tsx
```diff
@@ -135,38 +121,40 @@ const Controller = <
...
+  ): (...event: any[]) => void;
+  function onChange(
+    ...callbackOrEvent: any[]
+  ): ((...event: any[]) => void) | void {
+    const [firstArg] = callbackOrEvent;
+
+    return isFunction(firstArg)
+      ? (event: any) =>
+          setValue(name, commonTask(firstArg(event)), shouldValidate())
+      : setValue(name, commonTask(...callbackOrEvent), shouldValidate());
```
bluebill1049, line 150: @kotarella1110 can you review what I did?

kotarella1110: @bluebill1049 wow! greate! Thanks.
Does `getInputValue` inside `commonTask` work correctly if this is multiple event args?

kotarella1110: I missed this commit (https://github.com/react-hook-form/react-hook-form/pull/1711/commits/a93e713e9f1f870ca5d6c3821d7b31c89b961772). I think it works correctly 💯

bluebill1049:
create another PR to support both 
`as` and `render` consider we are already doing with `ErrorMessage`. so we can keep API consistent with support `as` and `render` obviously `as` will have an issue with required type, so use `render props` to avoid issues like that, however, we still can provide flexibility to users. This also leads to a minimum change to ErrorMessage as well, the only thing to change is `children` to `render`.

bluebill1049:
Lost a few hrs of sleep last night to think over this API, here are my thoughts and feel free to raise your objection.

This PR is trigger by `Type` related issue which cause by `as` which makes sense to provider solution with this with render prop, but it doesn't/shouldn't mean we should kill the flexible with `as`. 

--------

- `as` will be remained for simple usage assume most of the external component does follow standard component, which is exposed with `onChange` and `value` as optional not required type.
- `onChange, onChangeName, onBlur, onBlurName, valueName` will be remmoved
- render props to support custom methods to support those methods have been removed.

JeromeDeLeon:
Could you show a sample snippet of using `as` and `render` props?

bluebill1049:
> Could you show a sample snippet of using `as` and `render` props?

They don't work together, just like ErrorMessage. you use either `as` or `render`

```tsx
<Controller as={TextField} control={control} name="test" />
```

or

```tsx
<Controller render={props => <TextField {...props} />} control={control} name="test" />
```

JeromeDeLeon:
I understand your decision and seems valid for me. The only difference I could see is customization. `as` prop works out of the box while keeping the simplicity to use and `render` prop is for customization but I could only see myself using the `as` prop as I don't have any use cases for customization right now aside when `onChangeName` or `valueName` is different and even the introduction of below snippet, I still prefer to write it like what you did above.

```jsx
<Controller as={<CustomInput />} />

// I still prefer it this way,
<Controller as={CustomInput} />
```

bluebill1049:
> I understand your decision and seems valid for me. The only difference I could see is customization. `as` prop works out of the box while keeping the simplicity to use and `render` prop is for customization but I could only see myself using the `as` prop as I don't have any use cases for customization right now aside when `onChangeName` or `valueName` is different and even the introduction of below snippet, I still prefer to write it like what you did above.
> 
> ```js-jsx
> <Controller as={<CustomInput />} />
> 
> // I still prefer it this way,
> <Controller as={CustomInput} />
> ```

For this part we will keep API as what we have today:

```
// all valid
<Controller as={<CustomInput />} />
<Controller as={"input"} />
<Controller as={CustomInput} />
```

stramel:
Wait since this is for v6, why are we not removing the `as`?

bluebill1049:
> Wait since this is for v6, why are we not removing the `as`?

because of its simplicity as I have demonstrated above. ErrorMessage shares the same usage with `as` and `render`.

stramel:
Why do we need both props? That will just result in a confusing API. The simplicity can still remain. The single prop can cover all cases.

Personally, I very much dislike the string use-case. I don't really see a benefit to it.

bluebill1049:
> Why do we need both props? That will just result in a confusing API. The simplicity can still remain since the single prop can cover all cases.
> 
> Personally, I very much dislike the string use-case. I don't really see a benefit to it.

I understand your concern with `as` with string, and so does is an optional and personal preference thing.

JeromeDeLeon:
> Why do we need both props? That will just result in a confusing API. The simplicity can still remain since the single prop can cover all cases.
> 
> Personally, I very much dislike the string use-case. I don't really see a benefit to it.

I see, I do think string-case is invalid because `Controller` is for Custom Component that doesn't support ref and string-case like `input`, `select`, `textarea` doesn't need `Controller` at all at makes thing more complicated. We could omit the support for string-case here or DOMElement.

bluebill1049:
I am also looking at the API at the wholist point of view.

remove `as` means we need to remove `as` at `ErrorMessage`, and I still believe as is the simplest way to render an error message at `ErrorMessage`, some users may prefer `as` with its simplicity and some will have to force to use `renderProps` for the sake of custom event name. I think give people the option is not the bad thing, it shouldn't create too much confusion for the API. 

it's simple as using `as` if you don't need to configure anything, otherwise use render props to wire it up yourself.

kotarella1110:
> `render={<CustomInput />}` is very simple.
> But if `onChange` is required in the `CustomInput`'s props, it will cause type error.
> 
> ```ts
> type CustomInput = {
>   onChange: (e: any) => any; // required
> };
> <Controller name="test" control={control} render={<CustomInput />} /> // type error
> ```

@bluebill1049 If the above issue occurs, is this solution users need to use `React.ComponetType` or render prop as below?

```ts
<Controller name="test" control={control} as={CustomInput} />
```

or

```ts
<Controller name="test" control={control} render={props => <CustomInput {...props} />} />
```

bluebill1049:
> > `render={<CustomInput />}` is very simple.
> > But if `onChange` is required in the `CustomInput`'s props, it will cause type error.
> > ```ts
> > type CustomInput = {
> >   onChange: (e: any) => any; // required
> > };
> > <Controller name="test" control={control} render={<CustomInput />} /> // type error
> > ```
> 
> @bluebill1049 If the above issue occurs, is this solution users need to use `React.ComponetType` or render prop as below?
> 
> ```ts
> <Controller name="test" control={control} as={CustomInput} />
> ```
> 
> or
> 
> ```ts
> <Controller name="test" control={control} render={props => <CustomInput {...props} />} />
> ```

Simple usage stays with `as`:

```
<ErrorMessage as={<StyledParagraph />} errors={errors} />
<Controller as={<TextField} />} name="test" control={control} />
```

Type error or custom node structure use render props

```
<ErrorMessage errors={errors} render={({ message }) => {
  <div><span className="error">{message}</span></div>
}} />
<Controller 
  name="test" 
  control={control} 
  render={props => <div><CustomInput {...props} /></div>}
/>
````

**Comment by kotarella1110** [comment]
@bluebill1049 Do you not support `React.ComponetType` and `keyof JSX.IntrinsicElements`?

```ts
// not support?
<Controller as={TextField} name="test" control={control} />
```

bluebill1049:
> @bluebill1049 Do you not support `React.ComponetType` and `keyof JSX.IntrinsicElements`?
> 
> ```ts
> // not support?
> <Controller as={TextField} name="test" control={control} />
> ```
> 
> Also, do you support rest props?
> 
> ```ts
> <Controller as={<TextField />} name="test" control={control} custom="custom" />
> ```

What you and @JeromeDeLeon have done with the Controller will stay as it is for today (a little less breaking change for the existing users), except we will only remove `onChangeName`, `onBlurName`, `valueName` and replaced with render props approach.

stramel:
I still really don't like the idea of offering both `as` and `render`. The API is not intuitive. What happens when I pass both props?

bluebill1049:
> I still really don't like the idea of offering both `as` and `render`. The API is not intuitive. What happens when I pass both props?

That's what we have inside ErrorMessage:

```
  return InnerComponent ? (
    React.isValidElement(InnerComponent) ? (
      React.cloneElement(InnerComponent, props)
    ) : (
      React.createElement(InnerComponent as string, props)
    )
  ) : (
    <React.Fragment {...props} />
  );
```

bluebill1049:
In fact, `as` is used by `styled-component` and `chakra UI` as well.

Take an example here:
 
```
<Box as="button" rounded="md" bg="tomato" color="white" px={4} h={8} />
```

<Box as="button" rounded="md" bg="tomato" color="white" px={4} h={8}>
  {YourContent}
</Box>
```

**Comment by bluebill1049** [comment]
This Component is been there for a while: I didn't hear people complain about the usage just yet: https://react-hook-form.com/api#ErrorMessage

**Comment by bluebill1049** [comment]
Look away from Controller for a sec, what about ErrorMessage

```
<ErrorMessage as={StyledError} name="Test" />

<ErrorMessage render={({ message }) => <StyledError>{message}</StyledError>} name="Test" /> //how inconvient it is for something simple
```

**Comment by bluebill1049** [comment]
shared a vote on twitter: https://twitter.com/HookForm/status/1266240189195669505 decision will be made by tmr after the vote.

**Comment by bluebill1049** [comment]
If we can easily diff between `callback` and `component` with `render`, then it's good to move ahead, but reality the check is not great and pretty.

**Comment by stramel** [comment]
Is there a reason that we need to check? Can't we just call the function using `createElement`? In reality, it's just an inline component...

**Comment by bluebill1049** [comment]
> Is there a reason that we need to check? Can't we just call the function using `createElement`? In reality, it's just an inline component...

can you elaborate on this? 
```
render={() => {}}
render={Test}
```
how can we know which one to use React.createElement()?

**Comment by stramel** [comment]
Isn't all of this essentially equal?

```
const Test = () => <input />
render={Test}
```

vs

```
render={() => <input />}
```

It's essentially an inline component.

**Comment by bluebill1049** [comment]
> Isn't all of this essentially equal?
> 
> ```
> const Test = () => <input />
> render={Test}
> ```
> 
> vs
> 
> ```
> render={() => <input />}
> ```
> 
> It's essentially an inline component.

where are all the `props`?

**Comment by bluebill1049** [comment]
> Isn't all of this essentially equal?
> 
> ```
> const Test = () => <input />
> render={Test}
> ```
> 
> vs
> 
> ```
> render={() => <input />}
> ```
> 
> It's essentially an inline component.

I think you are right, we can always just do `React.creatElement('xxx', props)`, it doesn't need a check.

**Comment by JeromeDeLeon** [comment]
[Codesandbox](https://codesandbox.io/s/suspicious-wing-d9n1f?file=/src/App.js) I tried it, @stramel was right. I didn't know that.

**Comment by stramel** [comment]
It's late here but I will post some more detailed thoughts in the morning

**Comment by JeromeDeLeon** [comment]
Thanks @stramel I learned new stuff today :confetti_ball:

**Comment by bluebill1049** [comment]
> It's late here but I will post some more detailed thoughts in the morning

I didn't proceeds with render for all because all the `React.prototype` check. This is great! my god wasted almost 1 day discuss this...

**Comment by JeromeDeLeon** [comment]
I must say this discussion is worthwhile. I enjoyed being part of this.

**Comment by bluebill1049** [comment]
i guess this is what we after?

```
return React.createElement(render, {
    onChange,
    onBlur,
    value,
});
```

**Comment by JeromeDeLeon** [comment]
We could even passed the `name` and `controller`

**Comment by bluebill1049** [comment]
> We could even passed the `name` and `controller`

what do you mean?

**Comment by JeromeDeLeon** [comment]
I mean below snippet. I saw that you passed the `control` and `name` instead of putting it inside `props` or that doesn't make sense?

> ```tsx
> <Controller render={props => <TextField {...props} />} control={control} name="test" />
> ```

**Comment by bluebill1049** [comment]
> I mean below snippet. I saw that you passed the `control` and `name` instead of putting it inside `props` or that doesn't make sense?
> 
> > ```tsx
> > <Controller render={props => <TextField {...props} />} control={control} name="test" />
> > ```

Below looks right to me.
`<Controller render={props => <TextField {...props} />} control={control} name="test" />`

**Comment by bluebill1049** [comment]
ok running into the same problem again...

Screen Shot 2020-05-29 at 5 28 17 pm
Screen Shot 2020-05-29 at 5 28 09 pm

**Comment by JeromeDeLeon** [comment]
Tell me if I'm wrong, but to my understanding, `control` is only being used inside `Controller` and not the `InnerComponent`, that's why we never passed it to `props` right?

For `name`, not all Components support `name` prop like `react-select`.

For `Controller`'s use case:

```jsx
<Controller
  control={control}
  name="reactSelect"
  render={({ onChange, ...restProps }) => (
    <Select
      {...restProps}
      onChange={onChange(selected => selected)}
      options={options}
    />
  )}
/>
```

bluebill1049:
> Tell me if I'm wrong, but to my understanding, `control` is only being used inside `Controller` and not the `InnerComponent`, that's why we never passed it to `props` right?
> 
> For `name`, not all Components support `name` prop like `react-select`.
> 
> For `Controller`'s use case:
> 
> ```js-jsx
> <Controller
>   control={control}
>   name="reactSelect"
>   render={({ onChange, ...restProps }) => (
>     <Select
>       {...restProps}
>       onChange={onChange(selected => selected)}
>       options={options}
>     />
>   )}
> />
> ```

Yes correct, control, and name is for RHF.

bluebill1049:
I think either:

1. `renderProps` all the way or (unless someone can solve my problem above)
2. `support` both separately. 

let's see the votes tmr.

JeromeDeLeon:
> ok running into the same problem again...
> 
> Screen Shot 2020-05-29 at 5 28 17 pm
> 
> Screen Shot 2020-05-29 at 5 28 09 pm
> 
> we still need to differentiate callback (as a component with payloads) vs react component in order to let the type work correctly, back to square one.

What file did you encounter this one? 
Screenshot from 2020-05-29 15-49-04

bluebill1049:
> > ok running into the same problem again...
> > Screen Shot 2020-05-29 at 5 28 17 pm
> > Screen Shot 2020-05-29 at 5 28 09 pm
> > we still need to differentiate callback (as a component with payloads) vs react component in order to let the type work correctly, back to square one.
> 
> What file did you encounter that one?
> Screenshot from 2020-05-29 15-49-04

update render type to

```
render:
    | ((data: {
        onChange: {
          (...event: any[]): void;
          (callback: (...args: any[]) => any): (...event: any[]) => void;
        };
        onBlur: () => void;
        value: any;
      }) => React.ReactElement)
    | React.ComponentType<any>;
```
then you will see the error in test file with controller.

JeromeDeLeon:
That's problematic. Couldn't we just support `render={Input}` and `render={props => <Input {...props} />}`?

bluebill1049:
> That's problematic. Couldn't we just support `render={Input}` and `render={props => <Input {...props} />}`?

That's what I try to achieve here. :(

JeromeDeLeon:
Any reason why you need `React.ComponentType<any>`?

bluebill1049:
> Any reason why you need `React.ComponentType<any>`?

for this `render={Checkbox}`.

bluebill1049:
upon some investigation `render={Test}` || `render={() => {}}`is not a good pattern to follow and TS is not going to sing happily as well.

1. `render` prop all the way
2. `render` && `as` (similar as  react-router)

will decide that tmr and move on with this API.

bluebill1049:
also worth mentioning if we going `render` prop only, for API consistency then make sense to remove `ErrorMessage` from this lib, because `as` props should be gone as well.

1. lose some package weight
2. the component is becoming too verbose anyhow

`<ErrorMessage render={({ message }) => <p>{messge}</p>} name="test" errors={errors} />`

vs

`errors.test?.messaga`

it's almost pointless to have such a component.

kotarella1110:
> ok running into the same problem again...
> 
> Screen Shot 2020-05-29 at 5 28 17 pm
> 
> Screen Shot 2020-05-29 at 5 28 09 pm
> 
> we still need to differentiate callback (as a component with payloads) vs react component in order to let the type work correctly, back to square one.

I will look into whether I can.

bluebill1049:
> > ok running into the same problem again...
> > Screen Shot 2020-05-29 at 5 28 17 pm
> > Screen Shot 2020-05-29 at 5 28 09 pm
> > we still need to differentiate callback (as a component with payloads) vs react component in order to let the type work correctly, back to square one.
> 
> I will look into whether I can.

Thanks @kotarella1110 enjoy your Friday, let's not spending more effort on this until tmr.

JeromeDeLeon:
I don't know Controller's solution. I looked at how `react-router` implemented their own types and seemed like they separated it with `component` and `render`. I could only passed the function component but when passing Class component, I couldn't. I'll let  @kotarella1110 or @stramel handle this. :laughing:

bluebill1049:
> I don't know Controller's solution. I looked at how `react-router` implemented their own types and seemed like they separated it with `component` and `render`. I could only passed the function component but when passing Class component, I couldn't. I'll let @kotarella1110 or @stramel handle this. 😆

That's basically `as` and `render`.

stramel:
@bluebill1049 You should only need `React.ComponentType<{ value: any; onBlur: () => void; onChange: (...args: any[]) => void; }>`. That should support both `TextFields` and `(props) => <TextFields {...props} />`

bluebill1049:
> @bluebill1049 You should only need
> 
> ```tsx
> React.ComponentType<{ value: any; onBlur: () => void; onChange: (...args: any[]) => void; }>
> ```
> 
> That should support both `TextFields` and `(props) => <TextFields {...props} />`

Thanks! make sense, will give it a go tmr morning.

JeromeDeLeon:
Is that so? I tried it out earlier with ComponentType but having a problem with 

```tsx
class Component extends Component<ControllerRenderProps, any> { }
```

stramel:
@JeromeDeLeon I modified your CodeSandbox from earlier, everything seems to be working for me. https://codesandbox.io/s/funny-nightingale-xpvuy?file=/src/App.tsx

bluebill1049:
Screen Shot 2020-05-30 at 9 14 29 am

let's not worry about `ComponentType` with render props, keep it as pure render props. let's see the vote result and decided what I suggest above for next.

stramel:
The problem there is you're not converting the props. Your Checkbox component doesn't take whatever you specified.

bluebill1049:
> The problem there is you're not converting the props. Your Checkbox component doesn't take whatever you specified.

```
<Controller
  render={TextField}
  name="TextField"
  control={control}
  rules={{ required: true }}
/>
```

MUI TextField accepts `onChange, onBlur, value`, but still cry with TS. I think this proof is better to keep render props as just render props.

Screen Shot 2020-05-30 at 10 06 36 am

bluebill1049:
I am getting renderProp only pass the automation, and then prepare for the next step.

stramel:
It's just a typing issue with conflicting onBlur props

bluebill1049:
> It's just a typing issue with conflicting onBlur props

Yes, it's not compatible with `onBlur` but we probably going to have more issues like this with other libs? that's why I think keeping it as render props is actually easier right?

bluebill1049:
> It's just a typing issue with conflicting onBlur props

FYI: I am updating all the automation app to render props for now.

bluebill1049:
currently investigating an issue with Controller render props with useFieldArrya, somehow input lost focus... feel like mount and remount.

bluebill1049:
Screen Recording 2020-05-30 at 10 46 28 am

bluebill1049:
@kotarella1110 can you please update your ErrorMessage component from 'children' to 'render' as well?

bluebill1049:
Screen Shot 2020-05-30 at 5 46 53 pm

kotarella1110:
> @kotarella1110 can you please update your ErrorMessage component from 'children' to 'render' as well?

I will do this tomorrow.

src/controller.tsx
```diff
@@ -135,38 +121,40 @@ const Controller = <
...
+
+    return isFunction(firstArg)
+      ? (event: any) =>
+          setValue(name, commonTask(firstArg(event)), shouldValidate())
+      : setValue(name, commonTask(callbackOrEvent), shouldValidate());
+  }
+
+  return render({
+    onChange,
+    onBlur,
```
kotarella1110, line 155: Do you think that we can avoid unnecessary rendering with useCallback memo?

bluebill1049:
> > @kotarella1110 can you please update your ErrorMessage component from 'children' to 'render' as well?
> 
> I will do this tomorrow.

nvm, I already did :) hehe just a small comment.

src/controller.tsx
bluebill1049: 🤔 hmmm good point.

kotarella1110:
> > > @kotarella1110 can you please update your ErrorMessage component from 'children' to 'render' as well?
> > 
> > 
> > I will do this tomorrow.
> 
> nvm, I already did :) hehe just a small comment.

Sorry...
Reply is slow on weekends because I need to take care of my child 👶

bluebill1049:
> > > > @kotarella1110 can you please update your ErrorMessage component from 'children' to 'render' as well?
> > > 
> > > 
> > > I will do this tomorrow.
> > 
> > 
> > nvm, I already did :) hehe just a small comment.
> 
> Sorry...
> Reply is slow on weekends because I need to take care of my child 👶

aha no worries @kotarella1110 didn't know you are a daddy! hahaha, take care little Kotaro. ❤️

kotarella1110:
> > > > > @kotarella1110 can you please update your ErrorMessage component from 'children' to 'render' as well?
> > > > 
> > > > 
> > > > I will do this tomorrow.
> > > 
> > > 
> > > nvm, I already did :) hehe just a small comment.
> > 
> > 
> > Sorry...
> > Reply is slow on weekends because I need to take care of my child 👶
> 
> aha no worries @kotarella1110 didn't know you are a daddy! hahaha, take care little Kotaro. ❤️

hahaha I'm a daddy. My child was born when I joined the React Hook Form 💚

JeromeDeLeon:
Wow! That was great! Congratulations!! :confetti_ball:

kotarella1110:
Thank you very much @JeromeDeLeon 💚

bluebill1049:
> > > > > > @kotarella1110 can you please update your ErrorMessage component from 'children' to 'render' as well?
> > > > > 
> > > > > 
> > > > > I will do this tomorrow.
> > > > 
> > > > 
> > > > nvm, I already did :) hehe just a small comment.
> > > 
> > > 
> > > Sorry...
> > > Reply is slow on weekends because I need to take care of my child 👶
> > 
> > 
> > aha no worries @kotarella1110 didn't know you are a daddy! hahaha, take care little Kotaro. ❤️
> 
> hahaha I'm a daddy. My baby was born when I joined React Hook Form team 💚

Congrats! happy daddy! 👨‍👩‍👦

kotarella1110:
Thanks @bluebill1049 💚 
I raise babies (My Baby and RHF) with care 👨‍👧

bluebill1049:
> Thanks @bluebill1049 💚
> I raise babies (My Baby and RHF) with care 👨‍👧

hahaha sounds funny 4 dads... (1 dad left, he is a good dad as well, his name is Barry) RHF is like a little baby, it's growing each day with our cares!

src/controller.tsx
bluebill1049: I did try `useCallback`, but it caught issue during automation test... and I didn't investigate further, feel free to take a look if you have time.

cypress/integration/controller.ts
```diff
@@ -11,16 +11,16 @@ context('controller basic form validation', () => {
...
     cy.get('#input-textField input').type('test');
     cy.get('#input-select > div > div').click();
     cy.get('.MuiPopover-root ul > li:first-child').click();
     cy.get('#input-switch input').click();
     cy.get('#input-ReactSelect > div').click();
     cy.get('#input-ReactSelect > div > div').eq(1).click();
 
-    cy.get('p').should('have.length', 2);
-    cy.get('#renderCount').contains('8');
+    cy.get('.container > p').should('have.length', 6);
```
JeromeDeLeon, line 22: The reason why it failed before was because we're getting all P tags. To solve that, get only the P tags under `.container`.

src/controller.tsx
```diff
@@ -135,38 +129,56 @@ const Controller = <
...
+  ): ((...event: any[]) => void) | void {
+    const [firstArg] = callbackOrEvent;
+
+    return isFunction(firstArg)
+      ? (event: any) =>
+          setValue(name, commonTask(firstArg(event)), shouldValidate())
+      : setValue(name, commonTask(callbackOrEvent), shouldValidate());
+  }
+
+  const handlerProps = {
```
JeromeDeLeon, line 161: I hoisted it above to avoid `conditional memoizing` rule of hooks.

cypress/integration/controller.ts
```diff
@@ -11,16 +11,16 @@ context('controller basic form validation', () => {
...
+    cy.get('#renderCount').contains('2');
   });
 
   it('should validate the form with onBlur mode and reset the form', () => {
```
JeromeDeLeon, line 26: Here's a sample of test name where `reset` button didn't get called even though it was mentioned.

cypress/integration/controller.ts
bluebill1049: awesome stuff @JeromeDeLeon thank you

cypress/integration/controller.ts
bluebill1049: thanks @JeromeDeLeon

JeromeDeLeon: I will patch it together with providing types for `cy` for code review. :smile:

src/controller.tsx
```diff
@@ -135,38 +129,56 @@ const Controller = <
...
 
-      if (shouldReValidateOnBlur) {
-        trigger(name);
-      }
-    },
-    ...{ [valueName || (isCheckboxInput ? 'checked' : VALUE)]: value },
+  function onChange(...event: any[]): void;
+  function onChange(
+    callback: (...args: any[]) => any,
+  ): (...event: any[]) => void;
```
kotarella1110, line 149: I have improved it possible to customize the return value to match the existing API, but this may not be necessary for render props.
I think we can remove it to keep the API simple, what do you think?

```ts
<Controller 
  render={({ onChange, ...props }) => (
    <CustomInput
      {...props}
      onChange={(e) => {
        const customValue = custom(e.target.value);
        onChange(customValue);
      }}
    />
  )}
  control={control} 
  name="test"  
/>
```

bluebill1049: sounds great.

bluebill1049: done.

bluebill1049: Do you want to do a commit on this one? I think i stuff it up by removing that overload function.

kotarella1110: @bluebill1049 yes. I will fix soon!

bluebill1049: ty

PR #1727, opened 2020-05-29
Fix undefined ref in options

### LINKED ISSUE #1718: Error in dynamic radio group options + calling setValue
**Describe the bug**
I have a radio group. The number of options are dynamic based on conditions. In my code, after alternating between options twice I get an error when trying to use `setValue`. In the code sandbox this will happen after 7 alternations.
I noticed the error does not occur if I do not use `setValue` and it appears the `field.options` keep growing in size with each change.

**To Reproduce**
1. Go to https://codesandbox.io/s/empty-tree-qff1x?file=/src/App.js
2. Alternate clicking between `Button 1` and `Button 2`
3. After 7 times you will see an error

https://codesandbox.io/s/empty-tree-qff1x?file=/src/App.js

 - OS: Mac OSx
 - Browser Chrome, Firefox, Safari

JeromeDeLeon:
Fix #1718

@bluebill1049 , At first I put the `filter` to `setFieldValue` function but resulted in below image
Screenshot from 2020-05-29 12-43-49

then I tried to move it to `registerFieldsRef` function instead and resulted in below image:
Screenshot from 2020-05-29 12-52-10

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -218,7 +218,7 @@ export function useForm<
           : rawValue;
 
       if (isRadioInput(ref) && options) {
-        options.filter(Boolean).forEach(
+        options.forEach(
           ({ ref: radioRef }: { ref: HTMLInputElement }) =>
             (radioRef.checked = radioRef.value === value),
         );
@@ -897,7 +897,7 @@ export function useForm<
       field = isRadioOrCheckbox
         ? {
             options: [
-              ...((field && field.options) || []),
+              ...((field && field.options.filter(Boolean)) || []),
               {
                 ref,
                 mutationWatcher,
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -218,7 +218,7 @@ export function useForm<
           : rawValue;
 
       if (isRadioInput(ref) && options) {
-        options.forEach(
+        options.filter(Boolean).forEach(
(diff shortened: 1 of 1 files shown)


bluebill1049:
wow, thanks so much for picking this up @JeromeDeLeon

src/useForm.ts
```diff
@@ -897,7 +897,7 @@ export function useForm<
       field = isRadioOrCheckbox
         ? {
             options: [
-              ...((field && field.options) || []),
+              ...((field && field.options.filter(Boolean)) || []),
```
bluebill1049, line 900: by the way, i belive we have `unique`

JeromeDeLeon: I will try to look into that. I'm currently testing it to see if I could remove the `[empty x 3, {...}, {...}]` to make it `[{...}, {...}]` because right now I'm still seeing it even though I used `filter`

JeromeDeLeon:
@bluebill1049 , Could you take a look at why E2E test failed? seems to be related to error in connection.

bluebill1049:
> @bluebill1049 , Could you take a look at why E2E test failed? seems to be related to error in connection.

Thanks @JeromeDeLeon i will take this over and sort out over the weekend. 🙏

PR #1740, opened 2020-05-31
Add displayName for easier dev-tools debugging

pmaier983:
This change will allow for a slightly easier debugging experience by attaching a displayName to the context for use in the React DevTools, and is recommended by the [React Docs](https://reactjs.org/docs/context.html#contextdisplayname). I've also attached an example of how this change will effect the debugging experience.

From:
Pre-displayName

To:
Post-displayName

I'm unsure of what would be the best display name to use, so I gave it my best shot (not be against a change tho).


src/useFormContext.tsx
```diff
@@ -4,6 +4,8 @@ import { FormProviderProps } from './types/props';
 
 export const FormContext = React.createContext<UseFormMethods | null>(null);
 
+FormGlobalContext.displayName = 'ReactHookFormGlobalContext';
```
bluebill1049, line 7:
```suggestion
FormGlobalContext.displayName = 'ReactHookFormGlobalContext';
FormContext.displayName = 'RHFContext';
```

bluebill1049: Thanks for this awesome PR by the way.

PR #1741, opened 2020-05-31
Added DefaultValuesAtRender type

keiya01:
I added `DefaultValuesAtRender` type. 
Because `defaultValuesAtRenderRef` variable is used in several place, and `defaultValuesAtRenderRef`'s type is complicated.


bluebill1049:
nice! much cleaner.

bluebill1049:
Thank you @keiya01

PR #1742, opened 2020-05-31
Fixed useForm types

keiya01:
I fixed ref's types because `as` expression is used on `registerFieldsRef()` of `useForm()`.


bluebill1049:
wow nice. thank you again @keiya01

PR #1743, opened 2020-05-31
Fixed getIsFieldsDifferent logic

keiya01:
I refactored `getIsFieldsDifferent()`.
I think `isMatch` variables is not needed on `getIsFieldsDifferent()`.


bluebill1049:
wow awesome! less code is great!

PR #1748, opened 2020-06-01
Fixed error type

keiya01:
I fixed `error` variable type in `validateField()`.

### DIFF
--- a/src/logic/validateField.ts
+++ b/src/logic/validateField.ts
@@ -19,10 +19,12 @@ import appendErrors from './appendErrors';
 import { INPUT_VALIDATION_RULES } from '../constants';
 import {
   Field,
-  FieldErrors,
   FieldValues,
   FieldRefs,
   Message,
+  FieldError,
+  InternalFieldName,
+  FieldErrorMap,
 } from '../types/form';
 
 export default async <TFieldValues extends FieldValues>(
@@ -30,7 +32,7 @@ export default async <TFieldValues extends FieldValues>(
   validateAllFieldCriteria: boolean,
   {
     ref,
-    ref: { type, value, name },
+    ref: { type, value },
     options,
     required,
     maxLength,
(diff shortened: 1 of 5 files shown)


src/logic/validateField.ts
```diff
@@ -40,9 +42,10 @@ export default async <TFieldValues extends FieldValues>(
     pattern,
     validate,
   }: Field,
-): Promise<FieldErrors<TFieldValues>> => {
+): Promise<FieldErrorMap<TFieldValues>> => {
   const fields = fieldsRef.current;
-  const error: any = {};
+  const name = ref.name as InternalFieldName<TFieldValues>;
+  const error = {} as FieldErrorMap<TFieldValues>;
```
kotarella1110, line 48: When fixing to Partial, you should fix the following instead of `as`:

```ts
const error: FieldErrorMap<TFieldValues> = {};
```

kotarella1110 (changes requested):
@keiya01 Thank you for many PRs 👍 
@bluebill1049 The internal code of RHF contains a mixture of deep and flat objects.

```tsx
<input name="nested.test" />

// deep
type FieldErrors = {
  nested: {
    test: FieldError
  };
};
// flat
type FieldErrorMap = {
  'nested.test': FieldError
};
```

It's good to be able to see the differences between the two by looking at the types. I think a good name is important here.
ex: `FieldErrorMap` => `FlatFieldErrors`

Is this any good naming?

bluebill1049:
> @keiya01 Thank you for many PRs 👍
> @bluebill1049 The internal code of RHF contains a mixture of deep and flat objects.
> 
> ```tsx
> <input name="nested.test" />
> 
> // deep
> type FieldErrors = {
>   nested: {
>     test: FieldError
>   };
> };
> // flat
> type FieldErrorMap = {
>   'nested.test': FieldError
> };
> ```
> 
> It's good to be able to see the differences between the two by looking at the types. I think a good name is important here.
> ex: `FieldErrorMap` => `FlatFieldErrors`
> 
> Is this any good naming?

That sounds really good!

kotarella1110:
@bluebill1049 I think the types like fieldRef (`FieldRefs`) are also flat. I will fix these later.

src/logic/validateField.ts
kotarella1110: @keiya01 Can you rename from `FieldErrorMap` to `FlatFieldErrors` ?

keiya01: OK!

keiya01:
Thank you for the review.
I fixed code :)

src/logic/validateField.ts
```diff
@@ -69,9 +72,6 @@ export default async <TFieldValues extends FieldValues>(
         ? appendErrorsCurry(maxType, message)
         : appendErrorsCurry(minType, message)),
     };
-    if (!validateAllFieldCriteria) {
-      return error;
-    }
```
kotarella1110, line 74: good point 👍

kotarella1110 (approved):

PR #1754, opened 2020-06-02
Reduce re-rendering on handleSubmit

JeromeDeLeon:


JeromeDeLeon:
I'll push my changes later and summarize of what I have found.

cypress/integration/useFieldArray.ts
```diff
@@ -80,7 +80,7 @@ context('useFieldArray', () => {
     cy.get('#submit').click();
     cy.get('#result').contains('{}');
 
-    cy.get('#renderCount').contains('35');
+    cy.get('#renderCount').contains('27');
```
JeromeDeLeon, line 83: This is one of the gap I'm talking about.

src/controller.tsx
```diff
@@ -38,7 +38,7 @@ const Controller = <
     trigger,
     mode: { isOnSubmit, isOnBlur, isOnChange },
     reValidateMode: { isReValidateOnBlur, isReValidateOnSubmit },
-    formState: { isSubmitted },
+    isSubmittedRef,
```
JeromeDeLeon, line 41: I found out about this one though, if keep using the `formState`, for every keystroke, the whole component will re-render.

src/controller.tsx
```diff
@@ -57,6 +57,7 @@ const Controller = <
   const onFocusRef = React.useRef(onFocus);
   const isNotFieldArray = !isNameInFieldArray(fieldArrayNamesRef.current, name);
   rulesRef.current = rules;
+  const isSubmitted = isSubmittedRef.current;
```
JeromeDeLeon, line 60: The reason I did this because, I found out that we're using `useCallback` and at first, the value of it is `false` but when you hit submit, the value becomes `true` but not in the case of when accessing it inside `shouldValidate` and reading it because we are using `useCallback` once, it memoize the function once and keep the value to `false` even though it changed.

src/useForm.ts
```diff
@@ -130,7 +130,7 @@ export function useForm<
   const readFormStateRef = React.useRef<ReadFormState>({
     isDirty: !isProxyEnabled,
     dirtyFields: !isProxyEnabled,
-    isSubmitted: isOnSubmit,
+    isSubmitted: !isProxyEnabled,
```
JeromeDeLeon, line 133: @bluebill1049 Not sure why you did this, Could you explain? thanks :smile:

src/useForm.ts
```diff
@@ -1067,7 +1068,14 @@ export function useForm<
         isSubmittedRef.current = true;
         isSubmittingRef.current = false;
         submitCountRef.current = submitCountRef.current + 1;
-        reRender();
+
+        if (
+          readFormStateRef.current.isSubmitted ||
```
JeromeDeLeon, line 1073: As you can see, I only put the `reRender` method inside because `isSubmittedRef` is being used on the other part of the app hence need to be updated and so for the rest of the `ref`.

JeromeDeLeon (commented):
I divided the commits to two, the former which I examined and have questions on it and the latter to which I do not understand whats going with `useFieldArray` at all. You could switch it to it and examine it. For the rest, it has minimal impact with only 1 render being saved and I don't know if it's worth pushing, but with `useFieldArray`, a large gap can be seen.

My feelings for it is kinda not wanna push this through because there's a lot for something so simple like skipping the check.

@bluebill1049 probably skip it because it needs to re-render in case there's an error anyway. See [this](https://github.com/react-hook-form/react-hook-form/blob/43af1e0ebac5583024b9f97fb9da2944be0efaad/src/useForm.ts#L1064), there's no `reRender` getting called.

JeromeDeLeon:
It happened again! :laughing: connection error on e2e. @bluebill1049

JeromeDeLeon:
At first glance, it was so simple with just checking if it was necessary to render or not, but as I proceed with test, I found it complex.

JeromeDeLeon:
If this wouldn't be push through, it's fine by me :smile: I'm more than happy learning with this lib. :+1:

cypress/integration/useFieldArray.ts
bluebill1049: strange...

src/controller.tsx
bluebill1049: this one is good!

src/controller.tsx
bluebill1049: 👍

src/useForm.ts
bluebill1049: `submitted` doesn't really care about proxy because that finally statement with the re-render.

bluebill1049:
> I divided the commits to two, the former which I examined and have questions on it and the former to which I do not understand whats going with `useFieldArray` at all. You switch it to it and examined it. For the rest, it has minimal impact with only 1 render being saved and I don't know if it's worth pushing, but with `useFieldArray`, a large gap can be seen.
> 
> My feelings for it is kinda not wanna push this through because there's a lot for something so simple like skipping the check.
> 
> @bluebill1049 probably skip it because it needs to re-render in case there's an error anyway. See [this](https://github.com/react-hook-form/react-hook-form/blob/43af1e0ebac5583024b9f97fb9da2944be0efaad/src/useForm.ts#L1064), there's no `reRender` getting called.

it's intentional because of that final statement. `try` bit

bluebill1049:
> If this wouldn't be push through, it's fine by me 😄 I'm more than happy learning with this lib. 👍

I will take closer look today, i think there are some good stuff in here. good one.

PR #1772, opened 2020-06-04
Reduced re-render in useFieldArray()

keiya01:
I removed `reRender()` in `useFieldArray()`. 
Because `reRender()` had been executed if `swap()` is executed when `formState.touched` or `formState.dirtyFields` was changed.


bluebill1049:
great pick up!

PR #1779, opened 2020-06-04
fix ControllerProps type

kotarella1110:
remove `onChangeName`, `onBlurName`, `valueName` from ControllerProps type.


bluebill1049:
oops thanks for picking this up.

Issue #1766, opened 2020-06-03
Error focus passed to wrong input when using register & Controller onFocus prop


mush159:
**Describe the bug**
I love being able to autoFocus on inputs wrapped by Controller but I've noticed when its used on a form that contains components that use a mix of `register` and `Controller` it prioritises the autofocus to the inputs that use `register`.

I figure this is due to inputs using `register` being registered with the form before `Controller` components and when iterating over the fields with an error in [focusOnErrorField.ts](https://github.com/react-hook-form/react-hook-form/blob/8a08a04f2741667a900feec5b2b873377abaed2b/src/logic/focusOnErrorField.ts#L8) it iterates in the order that the inputs were registered.

This isnt ideal but a simple workaround I'm using now is to use `Controller` for all inputs, even if they could just use `register`.

Is there any way to respect the order in which the inputs are layed out when deciding which input to focus on?

**To Reproduce**
1. Go to codesandbox link below
2. Click on "submit query" button
3. See that "LastName" input has been focused upon.

**Codesandbox link (Required)**
https://codesandbox.io/s/react-hook-form-controller-template-g1hyy

**Expected behavior**
In the example above I would expect the "FirstName" input to recieve focus.

**Desktop (please complete the following information):**
 - OS: OSX Catalina
 - Browser: Firefox
 - Version: 77

**Additional context**

bluebill1049:
Yes, you are absolutely correct on how focus takes the order, and your solution sounds about right as well. right now, I don't think there is another way as we follow the order of input get registered, because Controller wrapper the input and register after useEffect, so it will be after native ref register. I don't think we will introduce another API to get the order right for now, as it will complicate the API as we all know react hook form is simple and I love to keep it that way for as long as i can (if there is walk around as you mentioned above). however, if a lot of people report this issue, we will consider introducing additional API to solve this.

SAGARACH65:
@bluebill1049 Was this behavior removed in the later versions? Because i cant find this code in the latest version v6.15.5

bluebill1049:
> @bluebill1049 Was this behavior removed in the later versions? Because i cant find this code in the latest version v6.15.5

yes, no longer useLayoutEffect. what's your issue?

SAGARACH65:
@bluebill1049 I had the same problem. I have form inputs with rendered controller in between native input elements.

So the error focus goes to the last elements before the controlled elements in the middle of the form. I just want the error to be shown sequentially from the top.

[...]

PR #1789, opened 2020-06-05
🐞 Register controller field using useLayoutEffect

mush159:
This change fixes the issue described here: https://github.com/react-hook-form/react-hook-form/issues/1766


bluebill1049:
Thanks, looks good to me.

Issue #1749, opened 2020-06-01
Rules of Controller does not react to changes


joaoeudes7:
First, the lib is very cool!! Thanks!

My problem is in set the rule in Controller:
```js
{{ required: needRegister, min: 3 }}
```
The property rules of Controller does not not react to `needRegister`

**Codesandbox link (Required)**
- [Controller](https://codesandbox.io/s/react-hook-form-controller-template-08og8?file=/src/index.js) Template

**Expected behavior**
React to changes in rules

**Desktop (please complete the following information):**
 - OS: Arch Linux Amd64
 - Browser Firefox
 - Version 77

bluebill1049:
we cache the rules object inside the controller, hence the validation rules is not changing.

- use `validate` function combined `getValues()`
- `unregister` that input, so input gets re-registered with updated `rules`.

acfasj:
> we cache the rules object inside the controller, hence the validation rules is not changing. (so the user doesn't have to memo the rules object themself)
> 
> **Solutions**
> 
> * use `validate` function combined `getValues()` (assume the toggle input is register with hook form as well)
> * `unregister` that input, so input gets re-registered with updated `rules`.

what if the `validate` function in the Controller rules need to use props?

kotarella1110:
@bluebill1049 Controlled Input and Uncntrolled Input are inconsistent. Shouldn't this be fixed?

https://codesandbox.io/s/react-hook-form-controller-template-oolpt

bluebill1049:
> > we cache the rules object inside the controller, hence the validation rules is not changing. (so the user doesn't have to memo the rules object themself)
> > **Solutions**
> > 
> > * use `validate` function combined `getValues()` (assume the toggle input is register with hook form as well)
> > * `unregister` that input, so input gets re-registered with updated `rules`.
> 
> seems like my validate function was cached, too. and re-registered looks bad

You mean `unregister` at `useEffect`?

bluebill1049:
> @bluebill1049 Controlled Input and Uncntrolled Input are inconsistent. Shouldn't this be fixed?
> 
> https://codesandbox.io/s/react-hook-form-controller-template-oolpt

happy to see a simple solution for this. 🙏

kotarella1110:
@bluebill1049 I think should be fix the line below. we should custom compare deps (rulesRef).

https://github.com/react-hook-form/react-hook-form/blob/v5.7.2/src/controller.tsx#L110

[...]

PR #1791, opened 2020-06-06
👨🏻‍💻 enable controller register rules on each render

### LINKED ISSUE #1770: Controller rules reset
_| create a new issue, so we don't notify and annoy others._

I changed the previous code as much as i could to simulate my real problem. 
Can you take a look? https://codesandbox.io/s/controller-rules-npe18?file=/src/App.tsx

As you can see, required erros appear even though the fields are not required in that moment.
Thats why i thought on re-registering both hour fields whenever the switched are toggled.
Thanks.

_Originally posted by @pedroguia in https://github.com/react-hook-form/react-hook-form/issues/1749#issuecomment-638512677_

bluebill1049:
we had quite few issues/questions around change rules with controller due to cache inside the controller, i have been thinking over the nights about this issue and decided to take register each render as a tradeoff for exchange better DX, I did have documentation updated for V6 and the more I have to write about it make me think how unintuitive for a developer.

fix #1770 #1749


src/controller.tsx
```diff
@@ -78,26 +76,30 @@ const Controller = <
   };
 
   const registerField = React.useCallback(() => {
-    if (!isNotFieldArray) {
-      removeFieldEventListener(fieldsRef.current[name] as Field, true);
-    }
+    if (fieldsRef.current[name]) {
+      register(name, rules);
```
bluebill1049, line 80: if input already been registered, we just keep updating the rules.

JeromeDeLeon:
Why close this one? What kind of rules are they talking about? is it about dynamic rules and since it cached inside rules, it didn't work as they expected? Am I right?

bluebill1049:
Yes I tried, didn’t work well. Yes dynamic rules.

bluebill1049:
i am going to take a look at it again.

PR #1793, opened 2020-06-06
Reduced code in setInternalValue that is used in useForm

keiya01:
I removed conditional expression in `setInternalValue`. Because `setDirty` method return `boolean`, but it was used with `isBoolean` method.


bluebill1049:
wow, thank you! so good! i swear i look the code more than 100 times and didn't pick this one up.

PR #1807, opened 2020-06-07
Fix readme and contributor guidleines

pmaier983:
Add some clarify text to the CONTRIBUTING.md.

The linking between docs was not working 50% of the time because the routes were incorrect. I fixed the routes.

Remove the flags to mirror the English Readme.


bluebill1049:
awesome and thank you!

Issue #1709, opened 2020-05-27
How to properly manage conditional fields?


rpmonteiro:
Hi Bill, fantastic job on this library and maintaining and closing issues like a champion. I absolutely love using it, and recommend it to pretty much everyone!

I'm not sure if this is a bug, or if I'm the bug (most likely lol), but I can't seem to manage conditional fields in a way that's predictable and doesn't make me tear my hair out.

I'm very sorry if I'm the one being a complete doofus. Your feedback would be much appreciated.
I tried to reproduce a minimal version on the codesandbox to illustrate the problem. Let me know if you want me to add more details.

**Describe the bug**
Values get reset without me doing so excplicitly.
Values get reset between mount/unmount of the form (I think this is by design, as it follows the normal form behaviour)
I've tried to manage it with useEffect() and register/unregister, but it's so difficult and the code so hacky... it's a nightmare :(

**To Reproduce**
Inside the codebox

**Codesandbox link (Required)**
https://codesandbox.io/s/react-hook-form-useform-template-gi2f7

**Expected behavior**
I want to have a bunch of conditional/hidden fields that I can manage, with the values staying where they are, with validation still being applied to them, without having to hack away like crazy and using setValue, watch(), handle onChange, onBlur, onFocus, etc etc etc... without making me want to cry lol

**Desktop (please complete the following information):**
 Mac, latest version with latest chromium browser

bluebill1049:
oh my god, I just shared a message on twitter today. Is this the problem that you having?

https://twitter.com/bluebill1049/status/1265505918852173824

rpmonteiro:
Yes, very close to it. I've tried that approach, managing the register myself, but with more than 2 or 3 fields, it becomes crazy verbose, specially if you want to have the same behaviour as others, and manage blur/focus states, and validation seems to not be predictable...

bluebill1049:
> but with more than 2 or 3 fields, it becomes crazy verbose, specially if you want to have the same behaviour as others, and manage blur/focus states, and validation seems to not be predictable...

do you have a `codesandbox` that I can have a play with? 

1. custom register (not working for you)
2. toggle visibility
3. wrapper Controller around your toggled inputs

rpmonteiro:
I'll prepare one asap

rpmonteiro:
Ok, I gave it another shot:
https://codesandbox.io/s/react-hook-form-useform-template-fd2zy?file=/src/index.tsx

[...]

Issue #186, opened 2019-08-05
Unmounting and re-mounting existing fields gives weird issues


Nickman87:
**Describe the bug**
I'm making a wizard type input flow and thus I'm mounting and unmounting input fields whenever I need them.
In this flow I'm using a mix of manual inputs and actual registered components. De manual input is stable and works as expected. The registered one however does not.

1. When re-mounting it, it forgets it's previously set value
2. In my actual implementation, my watch often doesn't get registered properly or something and I do not get component updates when I type in the input (it seems to be working stable in my sample...)

**To Reproduce**
I've made a small copy of my setup here: https://codesandbox.io/s/react-hook-form-basic-tqfe9?fontsize=14

The first button input is completely manual and if you go back to that step, you will see that your choice is correctly remembered.
The second input is a number input, if you have entered something I want to show a next button (therefore the watch) and I also want to move to the next panel if you press enter (therefore the strange register code).
If you enter an age and come back to that panel, you will see that it is empty.

**Expected behavior**
As the field is registered, and the value is entered into the form, I would expect it to remain there and when I re-mount the input field with the same name, that I see my previous value again.

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: 76.0.3809.87 (64-bits)

bluebill1049:
This is actually the expected behavior:

1. for registered input with ref, when input unmount. The input will get automatically `unregistered` hence value is lost
2. for custom registered input without ref, when input unmount. you will have to manually `unregister` yourself to during `useEffect` ( i will update that in the documentation)

- Solution:
1. To achieve your Wizzard form. I would recommend using `validationFields`, so restrict fields to be validated per step. So you basically show and hide inputs.
2. Build the form to remember input state with state management lib: I have built one maybe can be useful to you https://github.com/bluebill1049/little-state-machine

Nickman87:
That seems a bit extreme for a simple usecase not? I mean, a simple boolean inside react-hook-form that switched between remembering and forgetting would be enough to make this work I think?

bluebill1049:
your example unmounts a component, it shouldn't remember the state.

bluebill1049:
as long as you do not unmount the input, then the state will remain. Just like a normal input component.

bluebill1049:
https://codesandbox.io/s/react-hook-form-unmounting-fields-0tvoe

Nickman87:
[...]

PR #1809, opened 2020-06-08
config for auto un-register

### LINKED ISSUE #1713: V6: Message type requires a cast to string when used on a string only field.
**Describe the bug**
This is more of an annoyance with the typing. This requires all props that you feed `errors.fieldName?.message` into to support both `string` and `ReactElement` or always require a cast to the correct typing.

https://codesandbox.io/s/react-hook-form-useform-template-s2687?file=/src/index.tsx

**Expected behavior**
Not to have to cast every usage of an error message

**Screenshots**
image

 - OS: Windows 10
 - Browser Chrome
 - Version 81

bluebill1049:
**V6 feature:**

This feature will solve an issue when users want to persis their inputs without `unregister` invoked by `mutationObserver`. 

```tsx
useform({
  shouldUnregister: false
})
```

above config will disable `unregister` when inputs get unmounted and give user the option to keep value inside the hook store.

https://codesandbox.io/s/autounregister-4e91k?file=/src/src/logic/findRemovedFieldAndRemoveListener.ts

fix #1713 for the error message to be strictly typed for `string`.

### DIFF
--- a/app/src/app.tsx
+++ b/app/src/app.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { BrowserRouter as Router, Route } from 'react-router-dom';
+import AutoUnregister from './autoUnregister';
 import Basic from './basic';
 import Watch from './watch';
 import BasicSchemaValidation from './basicSchemaValidation';
@@ -87,6 +88,7 @@ const App: React.FC = () => {
         exact
         component={CustomSchemaValidation}
       />
+      <Route path="/autoUnregister" exact component={AutoUnregister} />
       <Route path="/useWatch" exact component={useWatch} />
     </Router>
   );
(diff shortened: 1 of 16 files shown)


JeromeDeLeon (dismissed):
LGTM! Seems pretty useful feature for conditional fields.

JeromeDeLeon:
[#186](https://github.com/react-hook-form/react-hook-form/issues/186) [#1709](https://github.com/react-hook-form/react-hook-form/issues/1709) These issues could be resolved by this feature, I suppose?

JeromeDeLeon:
This looks great really! I think it could improved our case for conditional rendering especially when it has the ability to set value even when field is not render on screen. Greate Job!

bluebill1049:
Thanks @JeromeDeLeon yes, I think a lot of users still live in the Controlled input land, so this will let them breathe, while we still embrace native form validation behavior.

PR #1827, opened 2020-06-10
Reduced code in executeSchemaOrResolverValidation method

keiya01:
I removed conditional expression in `executeSchemaOrResolverValidation()`, because this method is used with only `trigger()`, and duplicated conditional expression.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -350,41 +350,36 @@ export function useForm<
         | InternalFieldName<TFieldValues>
         | InternalFieldName<TFieldValues>[],
     ) => {
-      if (resolverRef.current) {
-        const { errors } = await resolverRef.current(
-          getFieldArrayValueByName(fieldsRef.current),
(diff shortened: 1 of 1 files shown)


src/useForm.ts
```diff
@@ -350,41 +350,36 @@ export function useForm<
...
     ) => {
-      if (resolverRef.current) {
-        const { errors } = await resolverRef.current(
-          getFieldArrayValueByName(fieldsRef.current),
-          contextRef.current,
-          validateAllFieldCriteria,
-        );
-        const previousFormIsValid = isValidRef.current;
-        isValidRef.current = isEmptyObject(errors);
+      // Use non-null assertion due to resolverRef is checked with trigger.
```
bluebill1049, line 353:
```suggestion
```

bluebill1049:
totally make sense! thank you! love code reduce.

PR #1844, opened 2020-06-11
fix #1843 dirty with useForm and useFieldArray

### LINKED ISSUE #1843: Bug with dirty combined with useForm and useFieldArray
Hi again. I have been trying to get this working with useFieldArray and have had some success. However I wanted to check if this is the correct way to do it?

https://codesandbox.io/s/dazzling-lalande-yx5mt?file=/index.js

The form data is modifying and submitting correctly, but the problem is that the dirty state isn't correct.

    If you edit the text box dirty is true.
    If you check the checkbox number 3 dirty is true.
    If you uncheck checkbox number 3 dirty is false.

This shouldn't happen as the textbox is still dirty.

Am I doing something wrong? Thanks in advance.

_Originally posted by @mike-jewell in https://github.com/react-hook-form/react-hook-form/issues/1829#issuecomment-642498323_

bluebill1049:


bluebill1049:
Fixed CSB: https://codesandbox.io/s/intelligent-architecture-1prbs?file=/src/src/useForm.ts

PR #1846, opened 2020-06-11
reduced code for removeFieldEventListener in useForm

keiya01:
- I reduced code in `removeFieldEventListener()`, because this function's conditions are unused
- I think `handleChangeRef` will not be undefined, because this function is defined before `removeFieldEventListener()` is called
- And `field` will be not undefined too, because `field` is checked before this function is called


bluebill1049:
awesome reduction!

PR #1857, opened 2020-06-12
Reduced code with useForm

keiya01:
- I fixed conditional expression
- `watchFieldsHookRenderRef` is used in only useWatch and this value is defined if `watchFieldsHookRef` is defined
- `validateResolver()` is used when resolver is defined only

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -261,15 +261,16 @@ export function useForm<
         unset(dirtyFieldsRef.current, name);
       }
 
-      isDirtyRef.current = isFieldArray
-        ? getIsFieldsDifferent(
+      isDirtyRef.current =
+        (isFieldArray &&
+          getIsFieldsDifferent(
             getFieldArrayValueByName(
               fieldsRef.current,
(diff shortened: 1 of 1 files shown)


src/useForm.ts
```diff
@@ -261,15 +261,16 @@ export function useForm<
         unset(dirtyFieldsRef.current, name);
       }
 
-      isDirtyRef.current = isFieldArray
-        ? getIsFieldsDifferent(
+      isDirtyRef.current =
```
bluebill1049, line 264: smart!

src/useForm.ts
```diff
@@ -410,9 +411,7 @@ export function useForm<
   );
 
   const isFieldWatched = (name: string) =>
-    isWatchAllRef.current ||
-    watchFieldsRef.current.has(name) ||
-    watchFieldsRef.current.has((name.match(/\w+/) || [])[0]);
```
bluebill1049, line 415: we do actually need this bit.

keiya01: oops... I understood.

PR #1861, opened 2020-06-13
Increase test for useForm

keiya01:
- I renamed arg name for `watchInternal` in types file
- I removed unnecessary conditional expression in `unregister()`
- I increased test for `useForm()`


bluebill1049:
Those unit test coverage is awesome!

PR #1874, opened 2020-06-15
Fix/yarn start app

pmaier983:
Upgrade yarn.lock and bump react-scripts version.

- why bump react-scripts version?
	- From a fresh install I was unable to ```yarn start:app```. Received the following error:
	RHF-Broken-app
	- Bumping react-scripts fixes this error.

- why upgrade yarn.lock?
	- It's important to keep dep's up to date in a test app for the off chance that some dependency change causes RHF to no longer work perfectly. Example: Joi is somehow updated, and RHF can no longer resolver JOI's schema. (Very unlikely, and honestly impossible, but its still best to stay safe)

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -18,7 +18,7 @@
     "react-hook-form": "^3.27.0",
     "react-hook-form-resolvers": "0.0.1-beta.1",
     "react-router-dom": "^5.1.2",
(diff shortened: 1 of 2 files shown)


bluebill1049:
Thanks @pmaier983 :)

PR #1891, opened 2020-06-17
remove getValues from control

bluebill1049:
remove getValues() from control object, less memory footprint as well.


keiya01:
LGTM!

PR #1900, opened 2020-06-18
Feature/set value should not set `dirty` to true and give users options

bluebill1049:
`setValue` API refinement

**Reason**

The current API is confusing with support both name as string and array of values. In addition, programmatically set a field should give users the option to set `dirty`, we shouldn't set dirty to `true` when user setValue programmatically by default. 

`setValue(fieldName, values, config)`

---

`setValue('test', 'value',  { shouldValidate: false, shouldDirty: false })`

### DIFF
--- a/src/useForm.test.tsx
+++ b/src/useForm.test.tsx
@@ -1031,6 +1031,26 @@ describe('useForm', () => {
         });
 
         expect(transformToNestObject).not.toHaveBeenCalled();
+        expect(result.current.formState.dirtyFields.test).toBeFalsy();
+      });
+
+      it('should set name to dirtyFieldRef if field value is different with default value and isDirty is true', () => {
+        const { result } = renderHook(() =>
+          useForm<{ test: string }>({
+            defaultValues: { test: 'default' },
+          }),
+        );
+        result.current.control.readFormStateRef.current.isDirty = true;
+
+        act(() => {
+          result.current.register('test');
+        });
+
+        act(() => {
+          result.current.setValue('test', '1', { shouldDirty: true });
+        });
+
+        expect(transformToNestObject).not.toHaveBeenCalled();
         expect(result.current.formState.dirtyFields.test).toBeTruthy();
       });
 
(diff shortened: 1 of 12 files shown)


src/useForm.test.tsx
```diff
@@ -1155,24 +1108,29 @@ describe('useForm', () => {
...
-              'test[0].name': 'default_update',
-              'test[1].name': 'default1',
-              'test[2].name': 'default2',
-            },
-          ]);
-          expect(result.current.formState.dirtyFields.test!).toEqual([
-            { name: true },
-          ]);
+          expect(transformToNestObject).toBeCalled();
+          // expect((transformToNestObject as any).mock.calls[2]).toEqual([
```
JeromeDeLeon, line 1124: Have you forgot to delete this or you have something still in your mind? :wink:

JeromeDeLeon (dismissed):
Do you have something in mind in case users wanted to set multiple values?

src/useForm.test.tsx
bluebill1049: lol! good pick up.

bluebill1049:
> Do you have something in mind in case users wanted to set multiple values?

they have to invoke multiple times or use `reset()` with `getValues()`

src/useForm.ts
```diff
@@ -411,6 +371,56 @@ export function useForm<
...
+        if (field) {
+          setFieldValue(field, value[key]);
+
+          if (config.shouldDirty) {
+            setDirty(fieldName);
+          }
+
+          if (config.shouldValidate) {
+            trigger(fieldName as FieldName<TFieldValues>);
+          }
```
keiya01, line 400: I think this condition is unnecessary. Because `trigger` method support array field.

keiya01 (commented):
I have only one question. Other than that, it looks good to me!

src/useForm.ts
bluebill1049: Thanks @keiya01 which condition? `if` ?

keiya01: Yes, `if`.

JeromeDeLeon:
I think its important to let users know of this although I agree with you their. I have never use `setValue` with multiple values because the word says it all, `setValue` not `setValues`, :smile:

bluebill1049:
> I think its important to let users know of this although I agree with you their. I have never use `setValue` with multiple values because the word says it all, `setValue` not `setValues`, 😄

agree,  i have already updated the doc.

src/useForm.ts
bluebill1049: without it will validate input each time users set a value, it's actually optional

bluebill1049: by the way, your unit tests r so good!!

bluebill1049: check this out: https://coveralls.io/github/bluebill1049/react-hook-form?branch=master

keiya01: Oh... I understood. Sorry.

keiya01 (approved):
LGTM :)

jviall:
@bluebill1049 Hi there, I'm confused because I can't seem to get this functionality to work for me. Particularly the setting a field to `dirty === false` when using `setValue`. Here's a sandbox forked from one of the doc's examples where I try to use `setValue` to "reset" an individual field back to its default value:

https://codesandbox.io/s/react-hook-form-formstate-dirty-touched-submitted-forked-opflm?file=/src/index.js

bluebill1049:
@jviall can you submit a GitHub discussion question instead?

PR #1902, opened 2020-06-18
reduced code with useWatch

keiya01:
I think `value` will not be `undefined`. Do you think about it ?


bluebill1049:
> I think `value` will not be `undefined`. Do you think about it ?

initial render will be `undefined`, if `defaultValues` or `defaultValue` is not supplied :)

keiya01:
I understood. This PR close.
Thank you :)

bluebill1049:
Thank you @keiya01 for trying to reduce code! love them!

PR #1921, opened 2020-06-19
Fixed isHTMLElement method

keiya01:
I think the way to check if value is HTMLElement is sufficient only `instanceof`. What does it have problem ?


bluebill1049:
i think you are right :) https://stackoverflow.com/questions/384286/how-do-you-check-if-a-javascript-object-is-a-dom-object

less code!

bluebill1049:
Thank you @keiya01 <3

PR #1947, opened 2020-06-24
Refactor useForm test

keiya01:
- I removed unnecessary `act()` that wrapped method don't call re-render
- I removed unnecessary `waitFor()` that don't need to wait for finishing wrapped method
- I fixed `container.querySelector()` to `screen.getByRole()` because `container.querySelector` is not recommended
- I refactored `setError` test because there are duplicated tests
- I fixed some method to using from `screen`


bluebill1049:
Awesome! This is huge clean up for the tests and improvements!

Issue #1888, opened 2020-06-16
v6 useWatch nested fields feature suggestions / queries / request


alexbakerdev:
**Is your feature request related to a problem? Please describe.**
I've been using `useWatch`in a relatively complex form for the last week involving previewing combinations of values, and subsequent dropdown options based on previous inputs. Some of the values in the form are nested objects that are dynamically added and removed.

Here are a few situations that have caught me out:

1. useWatch doesn't get the initial value from the form, it just returns the user defined defaultValue, even if there is a defined value in the form
2. changing the name value passed to useWatch doesn't change what the watcher watches. ie

```
// if colorName changes the colorValue won't match the newly named field.
const colorValue = useWatch({ name: `color.${colorName}.value` ])
```
    
3. registering and unregistering of nested paths doesn't trigger updates on watches of their ancestors.
4. setting values to undefined doesn't tigger a watcher

**Describe the solution you'd like**
I've had a go at fixing each of these things and was wondeirng if they are something react-hook-form wants.

1. Ideally the watch returns the computed value of the field if it exists, without waiting for a change.
2. useWatch tears down its watchers and sets up new ones when its parameter changes, and immediately returns the correct value for the new watched path.

I've got a [small patch](https://github.com/alexbakerdev/react-hook-form/commit/b19a7f1324dd210c50a91401d121f7c424e9b985) that implements a non optimal version of this.

It causes an extra rerender by setting state in a useEffect. I think it could be refactored to do the work in the use effect directly in the hook using some refs. Which will avoid that extra render allowing the hook to return the initial getValues.

3. registering and unregistering fields should update watchers. Implementing this involved invoking renderWatchedInputs on register and unregister, and searching a bunch of field paths in renderWatchedInputs [like this](https://github.com/alexbakerdev/react-hook-form/commit/122e517d46727262bfb1d69a229d3f745e0b9e7a)

    - Implementing this made me wonder if it's sufficient for `watchFieldsHookRef[id]` to just keep track of the provided field names. This would mean [not adding it to that set when children are discovered](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/assignWatchFields.ts#L37) since my change means those fields have to be searched twice (once for an exact match, once for a starts with).

[...]

PR #1968, opened 2020-06-25
fix useWatch defaultValue

bluebill1049:
FYI @keiya01 i think this will resolve issue 1 of that ticket.

here is the bug: https://codesandbox.io/s/react-hook-form-watch-default-values-ki6s8?file=/src/Watch.js

https://github.com/react-hook-form/react-hook-form/issues/1888

### DIFF
--- a/src/useWatch.ts
+++ b/src/useWatch.ts
@@ -48,7 +48,13 @@ export function useWatch<TWatchFieldValues>({
       ? isString(name)
         ? get(defaultValuesRef.current, name)
         : isArray(name)
-        ? name.map((inputName) => get(defaultValuesRef.current, inputName))
+        ? name.reduce(
+            (previous, inputName) => ({
+              ...previous,
+              [inputName]: get(defaultValuesRef.current, inputName),
+            }),
+            {},
+          )
         : {}
       : defaultValue,
   );
--- a/src/useWatch.ts
+++ b/src/useWatch.ts
@@ -3,6 +3,8 @@ import { useFormContext } from './useFormContext';
 import isUndefined from './utils/isUndefined';
 import isString from './utils/isString';
 import generateId from './logic/generateId';
+import get from './utils/get';
+import isArray from './utils/isArray';
 import {
   UseWatchOptions,
   FieldValues,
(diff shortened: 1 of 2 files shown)


src/useWatch.ts
```diff
@@ -35,12 +37,18 @@ export function useWatch<TWatchFieldValues>({
...
+    watchInternal,
+    defaultValuesRef,
+  } = control || methods.control;
   const [value, setValue] = React.useState<unknown>(
     isUndefined(defaultValue)
       ? isString(name)
-        ? defaultValue
+        ? get(defaultValuesRef.current, name)
+        : isArray(name)
+        ? name.map((inputName) => get(defaultValuesRef.current, inputName))
```
keiya01, line 51: I think `name.map` is not working correctly.

My suggestion is the following.

```ts
name.reduce((obj, inputName) => ({
  ...obj, 
  [inputName]: get(defaultValuesRef.current, inputName)
}), {})
```

keiya01 (commented):
Thank you for fixing issue. But why is this PR Draft ?
This PR resolve issue 1.

src/useWatch.ts
bluebill1049: aha, you are right! still, in the draft, because I still need to update those unit tests.

bluebill1049: i think we can save some bytes, later on, there are some common logic here with `watch`

keiya01: I understood.

keiya01: Yes, I think `useWatch` can  reduce code as well. There are some logic that is used in `watch`.

bluebill1049: 100% agree :)

bluebill1049:
Screen Shot 2020-06-26 at 1 36 21 pm
@keiya01 i run test locally and get above, any idea.
`node v12.16.1`

keiya01:
Are these bugs resolved on test ? I could not confirm these bugs...

bluebill1049:
> Are these bugs resolved on test ? I could not confirm these bugs in local environment...

really? if you supply `defaultValues` at `useForm`, is `useWatch` return default value correctly?

bluebill1049:
here is the bug: https://codesandbox.io/s/react-hook-form-watch-default-values-ki6s8?file=/src/Watch.js

keiya01:
Yes... In my local env, all test passing.

bluebill1049:
> Yes... In my local env, all test passing.

oh you refer to the `tests`, yeah strange... don't worry about it.

keiya01:
Maybe node version is cause. My node version is v13.9.0 ...

bluebill1049:
13 works 🥳 @keiya01 let me know if you are happy with this PR, we can merge it and save the code afterwards.

keiya01:
@bluebill1049 LGTM🎉

moulinraphael:
Thanks a lot @bluebill1049  !

PR #1994, opened 2020-06-29
close #1990 issue with setValue, getValues, watch and reset with shouldUnregister: false

### LINKED ISSUE #1990: Form's values get lost after hide / show form in render
**Describe the bug**
If I have Form with values and I want to show custom input to specify custom value and hide From, after Form shows in render it 's values get lost and you need to touch any control to see those values again. 

**To Reproduce**
1. load https://codesandbox.io/s/react-hook-form-set-inputselect-value-z87ce
2. in `Age group` select `+ add`. This hides Form element and show single input field where you can add custom age number. 
3. enter number `4` and press Enter
4. open console and see `formValues` is completely empty `Object {}`
5. scroll up and enter in input `First Name` anything, now in console `formValues` has form values
6. or press `submit` button and form has submit data even if `formValues` was empty after step 3

https://codesandbox.io/s/react-hook-form-set-inputselect-value-z87ce

**Expected behavior**
`formValues` should not be empty

macosx (latest), chrome (latest)

bluebill1049:
This PR aims to fix the following API when input is unmounted and `shouldUnregister` set to `false`. Users will still allow using the following API for input that's unmounted.

- [x] watch
- [x] getValues
- [x] setValue
- [x] reset

close #1990


JeromeDeLeon:
What about `reset`?

bluebill1049:
> What about `reset`?

good point, i will need to update that in reset as well. thanks @JeromeDeLeon

src/useForm.ts
```diff
@@ -1050,6 +1050,7 @@ export function useForm<
 
     if (values) {
       defaultValuesRef.current = values;
+      unmountFieldsStateRef.current = values;
```
bluebill1049, line 1053: @JeromeDeLeon can you help add a test for this if you can? going to start "real" work soon. otherwise, i will do it tonight.

JeromeDeLeon:
I really thought it was the right behavior that I will not be able to set values to field even with `shouldUnregister` be set to `false` until I read the related issue LOL.

src/useForm.ts
JeromeDeLeon: I'm really sorry. I was on my break when I got this notif. I was about to take a look at it tonight also. :disappointed:

bluebill1049:
> I really thought it was the right behavior that I will not be able to set values to field even with `shouldUnregister` be set to `false` until I read the related issue LOL.

I think it makes sense to support that behavior.

src/useForm.ts
bluebill1049: don't be sorry at all. I will sort out after work 👍

src/useForm.test.tsx
```diff
@@ -1199,6 +1205,16 @@ describe('useForm', () => {
         expect(result.current.formState.dirtyFields?.test).toBeUndefined();
       });
     });
+
+    it('should set unmounted input value when field is not found', () => {
```
JeromeDeLeon, line 1209: Am I right? It would still persist data even when the field doesn't really exist like small number of people wanted to have this feature for a long time instead of adding hidden field they just use `setValue`?

bluebill1049: yes, that's correct, basically, you can still update the store inside the hook. I think that's consistent behavior otherwise it would be confusing to users. personally I prefer auto `unregsiter`.

bluebill1049: your thoughts?

JeromeDeLeon: I'm good. I just suddenly thought about some users wanting that kind of feature.

JeromeDeLeon (dismissed):

src/useForm.test.tsx
bluebill1049: i did think twice about this as well, if an input didn't get unregistered (with `shouldUnregistered: false`), then the user should be able to setValue, watch, getValues and reset them for consistency.

bluebill1049: let me get @stramel and @kotarella1110's involvement as well, not rush to release this as part of v6.

JeromeDeLeon: Yeah. This also just occur to me the moment I saw this thing.

keiya01 (dismissed):
LGTM🙏 
I'm sorry for the delay in reviewing.

bluebill1049:
> LGTM🙏
> I'm sorry for the delay in reviewing.

Thanks @keiya01 ❤️

bluebill1049:
@kotarella1110 do you want to take a quick look before i merge this PR in?

PR #1275, opened 2020-03-25
allow React.ReactElement in register option's message

kotarella1110:
Todo:

- [x] add unit test
- [x] improve type

Usage:

```tsx
register({
  required: <CustomErrorMessage message="required" />,
  max: {
    value: 1000,
    message: <CustomErrorMessage message="max" />,
  },
  min: {
    value: 10,
    message: <CustomErrorMessage message="min" />,
  },
  validate: {
    test: value => value === '100' || <CustomErrorMessage  message="validate" />,
  },
})
```

```tsx
<ErrorMessage
  name="test"
  errors={errors}
  message={<CustomErrorMessage message="ErrorMessage component" />}
/>
```

```tsx
// Single Error
setError(
  "test",
  "singleError",
  <CustomErrorMessage message="single error" />
)
// Single Errors
setError("test", {
  single1: <CustomErrorMessage message="single error1" />,
  single2: <CustomErrorMessage message="single error2" />
})
// Multiple Errors
setError([
  {
    type: "multiple",
    name: "test1",
    message: <CustomErrorMessage message="multiple error" />
  },
  {
    type: "multiple",
    name: "test2",
    message: <CustomErrorMessage message="multiple error" />
  }
])
```

Test it out: https://codesandbox.io/s/react-hook-form-issue-1266-khtse

close #1266

PR #1999, opened 2020-06-30
Allow a function to be passed to `message` for validations

et:
When react-hook-form is used with [`react-i18next`](https://github.com/i18next/react-i18next), we get a type error when setting up messages for validations, e.g.

```tsx
<input
  name="title"
  ref={register({
    required: t("form.validations.required"),
    maxLength: {
      value: 50,
      message: t("form.validations.maxLength"),
    },
  })}
/>
```

Screen Shot 2020-06-30 at 4 41 07 PM

In #1275, `isMessage` was adjusted to accept a `ReactElement`. This PR proposes to allow a function such that we can seamlessly use i18n.

### DIFF
--- a/src/utils/isMessage.ts
+++ b/src/utils/isMessage.ts
@@ -2,6 +2,12 @@ import { isValidElement } from 'react';
 import isString from '../utils/isString';
 import isObject from '../utils/isObject';
 import { Message } from '../types/form';
+import isFunction from './isFunction';
 
-export default (value: unknown): value is Message =>
-  isString(value) || (isObject(value) && isValidElement(value));
+export default (value: unknown): value is Message => {
+  return (
+    isString(value) ||
+    (isObject(value) && isValidElement(value)) ||
+    (isFunction(value) && isString(value()))
+  );
+};
--- a/src/utils/isMessage.test.ts
+++ b/src/utils/isMessage.test.ts
@@ -5,6 +5,7 @@ describe('isBoolean', () => {
   it('should return true when value is a Message', () => {
     expect(isMessage('test')).toBeTruthy();
     expect(isMessage(React.createElement('p'))).toBeTruthy();
+    expect(isMessage(() => 'p')).toBeTruthy();
   });
 
   it('should return false when value is not a Message', () => {


src/utils/isMessage.ts
```diff
@@ -2,6 +2,12 @@ import { isValidElement } from 'react';
...
 import { Message } from '../types/form';
+import isFunction from './isFunction';
 
-export default (value: unknown): value is Message =>
-  isString(value) || (isObject(value) && isValidElement(value));
+export default (value: unknown): value is Message => {
+  return (
+    isString(value) ||
+    (isObject(value) && isValidElement(value)) ||
+    (isFunction(value) && isString(value()))
```
et, line 11: The downside of this is that we have to evaluate the the function (`isString(value())`) to get the return value of the function. I'm not sure if there's a smarter way to do this in typescript or if it's even necessary to check the return value -- I simply opted to check the return value because of this [test](https://github.com/et/react-hook-form/blob/a4dd2a843ffe83e1197fe5b4f2a3531cae3ac720/src/utils/isMessage.test.ts#L19). If it's not important, it would probably be more wise to omit the test.

et:
cc @bluebill1049 and @kotarella1110 curious to hear your thoughts 😄

bluebill1049:
hey @et, Thanks for the PR,  sorry to get back to you late, too much with V6 release. I think to make things simple, maybe consider this: https://github.com/react-hook-form/react-hook-form/issues/2025#issuecomment-653379883 we had support JSX and string for a message before, it creates type related issue and hopefully, we don't get back to that situation again.

et:
Hey @bluebill1049 - no worries on the delay! I figured out a workaround by changing:

```tsx
<input
  name="title"
  ref={register({
    required: t("form.validations.required")
  })}
/>
```

to

```tsx
<input
  name="title"
  ref={register({
    required: {
      value: true
      message: t("form.validations.required")
    }
  })}
/>
```


I could have also just simply casted:


```tsx
<input
  name="title"
  ref={register({
    required: t("form.validations.required") as string
  })}
/>
```

but I try to avoid `as`.


With that being said, I feel like https://github.com/react-hook-form/react-hook-form/issues/2025#issuecomment-653379883 isn't the most ideal workaround. That makes sense if the validation rules were backend driven. However, if that's not the case, then it is way more elegant to simply define the validation rules as close as possible to the JSX code (especially since this is how react-hook-form advocates for).

I'm not familiar of previous type related issues but do you see this still being an issue with this PR? 

Appreciate the feedback!

src/utils/isMessage.ts
pke: Not sure the evaluation is required here. What if the function is async? Is that supported as well?

et: Did you catch what I was saying? It's because of this [test](https://github.com/et/react-hook-form/commit/35fcde32c258351be2bb9f6587e5e4a12aa65a0c#diff-47be23272ed62f7a2542f06a421a2b94R18). 
I agree that it seems unnecessary but @bluebill1049 or @kotarella1110 should make the call that it's safe to omit the test.

pke: I read what you wrote. I was raising the question if the function could also be async.

et: Given that this would be called in the render function of a component, using an async function would be a major anti-pattern in react. I can’t think of an easy way to prevent this for react-hook-form given the way the `isMessage` function is using `unknown`. Open to suggestions.

et:
@bluebill1049 - any chance this could get in? I'm open to making any changes you recommend.

bluebill1049:
hey @et sorry I have been busy with issues and other stuff, I think for now we won't introduce this feature until more users request for such. again thanks for putting time into this PR.

et:
No worries! Thanks for getting back to me. Users who come across this can use the typecasting I posted above.

PR #2000, opened 2020-07-01
Replaced from mutationobserver-shim to jest-dom

keiya01:


bluebill1049 (approved):
Awesome! thanks @keiya01

PR #2040, opened 2020-07-03
🐞 fix unset nested array object

euphocat:
Hello ! I've noticed a strange behavior while working with form errors. I spotted a problem in the `unset` method. I think my fix might be a bit dirty... But I'm not confortable with that `unset` method (with that while loop and stuff).

### DIFF
--- a/src/utils/unset.test.ts
+++ b/src/utils/unset.test.ts
@@ -198,4 +198,30 @@ describe('unset', () => {
 
     expect(unset(data, 'name[0]')).toEqual({});
   });
+
+  it('should not remove nested empty array item', () => {
+    const data = {
+      scenario: {
+        steps: [
+          {
+            content: {
+              question: 'isRequired',
+            },
+          },
+        ],
+      },
+    };
+
+    expect(unset(data, 'scenario.steps[1].messages[0]')).toEqual({
+      scenario: {
+        steps: [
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks for the PR, can you demonstrate us with the issue in a codesandbox?

euphocat:
hum maybe I can, but I think it's a bit tricky... isn't the test enough? :)

bluebill1049:
Thanks @euphocat 👍 @kotarella1110 can you please review the change, I think the last refactor break this.

euphocat:
Ok because it's fun, I was able to reproduce it with a bit of my setting: https://codesandbox.io/s/react-hook-form-usefieldarray-nested-arrays-8dbws
It uses io-ts, and call `trigger`

bluebill1049:
thanks @euphocat 👍

keiya01:
Thank you @euphocat for sending PR.
Did this problem occur by executing `trigger()`? If so, I think you might want to write test about this problem with `trigger()`.

keiya01 (approved):
LGTM!

euphocat:
Thank you for your help and reactivity !

bluebill1049:
Thank you for your reporting and PR @euphocat 🙏

PR #2047, opened 2020-07-03
remove unnecessary `deps` in `useForm` hook

mh-alahdadian:
this PR will remove refs in dependency list of `useCallback` and also will move some static variables out of hook body to not need to add them to dependency list 

** NOTE: isWindowUndefined and isWeb and isProxyEnabled are static(constant) variables and can not be changed after first run of js file on same engine **

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -66,6 +66,15 @@ import {
 } from './types/form';
 import { LiteralToPrimitive, DeepPartial, NonUndefined } from './types/utils';
 
+export const isWindowUndefined = typeof window === UNDEFINED;
+export const isWeb =
+  typeof document !== UNDEFINED &&
+  !isWindowUndefined &&
+  !isUndefined(window.HTMLElement);
+export const isProxyEnabled = isWeb
+  ? 'Proxy' in window
+  : typeof Proxy !== UNDEFINED;
+
 export function useForm<
   TFieldValues extends FieldValues = FieldValues,
   TContext extends object = object
@@ -122,12 +131,6 @@ export function useForm<
     modeChecker(mode),
   ).current;
   const validateAllFieldCriteria = criteriaMode === VALIDATION_MODE.all;
-  const isWindowUndefined = typeof window === UNDEFINED;
-  const isWeb =
-    typeof document !== UNDEFINED &&
-    !isWindowUndefined &&
-    !isUndefined(window.HTMLElement);
-  const isProxyEnabled = isWeb ? 'Proxy' in window : typeof Proxy !== UNDEFINED;
   const readFormStateRef = React.useRef<ReadFormState>({
     isDirty: !isProxyEnabled,
     dirtyFields: !isProxyEnabled,
@@ -189,7 +192,7 @@ export function useForm<
         return true;
       }
     },
-    [reRender, resolverRef],
+    [reRender],
   );
 
   const setFieldValue = React.useCallback(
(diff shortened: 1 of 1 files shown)


src/useForm.ts
```diff
@@ -66,6 +66,13 @@ import {
 } from './types/form';
 import { LiteralToPrimitive, DeepPartial, NonUndefined } from './types/utils';
 
+const isWindowUndefined = typeof window === UNDEFINED;
+const isWeb =
+  typeof document !== UNDEFINED &&
+  !isWindowUndefined &&
+  !isUndefined(window.HTMLElement);
+const isProxyEnabled = isWeb ? 'Proxy' in window : typeof Proxy !== UNDEFINED;
```
bluebill1049, line 74: Thanks for this PR, have we consider in SSR?

MHA15: also in SSR when this file run for the first time it will calculate these variables and will never changed 
in client side again this file will be run and will calculate these variables for it self

src/useForm.ts
```diff
@@ -141,8 +139,16 @@ export function useForm<
...
-  resolverRef.current = resolver;
+
+  const options = {
+    context,
+    resolver,
+    shouldUnregister,
+    shouldFocusError,
+    validateAllFieldCriteria: criteriaMode === VALIDATION_MODE.all,
+  };
+  const optionsRef = React.useRef<typeof options>({} as typeof options);
```
bluebill1049, line 150: This actually increase the package size, let's keep them flat, please.

MHA15: increase package size? are you sure? I think this will decrease it (because of using less useRef )
let's not talk about package size in this one , It's difference might be 1KB or less in development mode, and also as I said before using less hooks are more important

I will change these in next version (createControl function) and will remove all refs (don't worry about it)

keiya01: I actually saw increasing file size.
I think you should remove only unnecessary `deps` because bundling `options` increase bundle size.

MHA15: thanks for your attention
I had run `yarn bundlesize` and I got 8.42KB in my branch and 8.41KB on master? am I true?
are we talking about 10B difference? :)

bluebill1049: yes, correct. sorry, every B counts here.

bluebill1049: > I will change these in next version (createControl function) and will remove all refs.

Thanks, I am looking forward to that.

src/useForm.ts
JeromeDeLeon: Couldn't we move this to `utils` folder? If every function is hoisted out of the lib, please move it to respective folder. That way we could keep the main file clean as much as possible. Thank you. Nice work! :smile:

JeromeDeLeon (changes requested):
I do appreciate your work about using `optionsRef` instead of spreading them but we do value bundle size so much like what @bluebill1049 and @keiya01 told you. Thanks again! we value your contribution to the enhancement of this library.

src/useForm.ts
MHA15: third commit has been removed from PR

src/useForm.ts
MHA15: yes I do want it too, but I didn't find any related file in utils and I didn't know where should I put them, can you help me and tell which file I should put them?

bluebill1049:
Thanks @MHA15

src/useForm.ts
bluebill1049: no worries @MHA15 i will take it over and sort it out 👍 get your contribution in first 🥳

MHA15:
> Thanks @MHA15
that is my pleasure to participate on this awesome library

PR #2057, opened 2020-07-04
Added test for SSR

keiya01:
I added test for SSR. This test case detect error that occur in initial loading(ex: `useLayoutEffect`, `window` object, etc...).


bluebill1049:
Awesome! 🥳 thanks @keiya01

PR #2058, opened 2020-07-04
create own file for env detect

bluebill1049:


bluebill1049:
@JeromeDeLeon going to keep those variable inline 😱

JeromeDeLeon:
Why? Though its fine for me. Just curious.

bluebill1049:
+326 B 😭

JeromeDeLeon:
Oh god why? Why is it that when you move it to a seprate file makes a big difference than just spreading it inside?
 :disappointed:

JeromeDeLeon:
Maybe they were creating import statement and such that's why its adding few lines of code hence the result of such size. Okay, we couldn't do about it. Yeah, great better not move it if its the case.

bluebill1049:
> Maybe they were creating import statement and such that's why its adding few lines of code hence the result of such size. Okay, we couldn't do about it. Yeah, great better not move it if its the case.

yea, worth testing it out anyhow :( it's ugly because save us some bytes.

PR #2067, opened 2020-07-05
Added test for Controller with React Native

keiya01:
- [x] added Controller test with React Native
- [x] fixed jest config for React Native


bluebill1049 (approved):
Bravo!

PR #2072, opened 2020-07-05
improve watch API with useFieldArray delete

### LINKED ISSUE #2031: [v6] - watch returns defaultValue after moving an item in a fieldArray
**Describe the bug**
Not sure if this is an error or is it by design. Having a `fieldArray`, when you move one of the items,  the watch for the FA first returns the default value of the array and then the updated value. I'm not sure why the watch runs twice, but if it does, shouldn't it return the latest value instead of the default?

**To Reproduce**
1. add 4 items to the list
2. move 1 item
3. see console log (it shows first the watch result with the initial value of the FA and then the updated value)

https://codesandbox.io/s/react-hook-form-usefieldarray-mnemu?file=/src/index.js

### LINKED ISSUE #2052: Registering new field when form is invalid can make the form valid
**Describe the bug**
when the form is invalid and a new field is registered, the validation resolver is passed initial state, which may result in the form become incorrectly valid
i might be missing something since this library is very well thought out and i see the code that intentionally passes initial data when the form is invalid so please excuse me if i made a mistake

**To Reproduce**
1. Go to https://codesandbox.io/s/react-hook-form-validationresolver-g7ett?file=/src/App.js
2. Click "Include username" checkbox to cause form to be in invalid state
3. Click "Toggle username field" to show username
4. See error - form is is valid state/validation is passed initial form data, causing the form to be incorrectly valid

https://codesandbox.io/s/react-hook-form-validationresolver-g7ett?file=/src/App.js

**Expected behavior**
when registering new inputs when form is invalid, the form will be passed the current state(merged with initial state?)

 - OS: macOS
 - Browser: chrome
 - Version 83

bluebill1049:
React Hook Form's form states live inside each input, which leads to difficultly to track `useFieldArray` inputs value during deletion, so this PR is aimed to improve this behavior. After this PR, user `fields` as `watch`'s defaultValue to assure `watch` return correct inputs value.

**usage:** 

```
const { fields } = useFieldArray()
watch('test', fields);
```

- remove additional function
- using `getValues()` in `useFieldArray` and `useForm` with `resolver`
- update unit tests

close #2031
close #2052


src/useForm.ts
```diff
@@ -309,7 +305,7 @@ export function useForm<
         | InternalFieldName<TFieldValues>[],
     ) => {
       const { errors } = await resolverRef.current!(
-        getFieldArrayValueByName(fieldsRef.current),
+        getValues() as any,
```
bluebill1049, line 308: @kotarella1110 can you please assist with this type issue?

JeromeDeLeon (dismissed):
LGTM! I believe this doesn't affect how useWatch works for useFieldArray, right?

bluebill1049:
> LGTM! I believe this doesn't affect how useWatch works for useFieldArray, right?

no, it doesn't. thanks for the review ❤️

bluebill1049:
@keiya01 I love your 👀 icon LOL.

keiya01:
👀  is very useful lol
I reviewed the PR! I think we should write test about this behavior.
Could I write test about this behavior?

bluebill1049:
> 👀 is very useful lol
> I reviewed the PR! I think we should write test about this behavior.
> Could I write test about this behavior?

Bring your 🔥 🚀

src/useFieldArray.test.tsx
```diff
@@ -11,6 +11,10 @@ jest.mock('./logic/generateId', () => ({
 }));
 
 describe('useFieldArray', () => {
+  beforeEach(() => {
+    jest.resetAllMocks();
+  });
+
```
keiya01, line 17: sorry, I have one question.
Why did you add this code ?

bluebill1049: One of the test pass when i run individual and failed when i run them all, i guess it was to do with mock function, so i just reset it.

keiya01: I see. Thank you.

keiya01:
@bluebill1049 
I'm sorry. Testing this behavior is too difficult. Therefore I will try to write test if I come up with any solution.🙏

keiya01 (approved):

bluebill1049:
> @bluebill1049
> I'm sorry. Testing this behavior is too difficult. Therefore I will try to write test if I come up with any solution.🙏

no worries at all, thanks for trying out! we have plenty of time anyhow.

PR #2121, opened 2020-07-08
Fix docs contributing links with valid path

zenby:
At the moment at `docs/README**-**.md` files we can see the links with invalid path at `Contributors` title. I think it's better to have the link to `root/CONTRIBUTING.md` file if we don't have translations of this files to each language directly.


bluebill1049:
Thanks, agree.

PR #2127, opened 2020-07-09
Added test when removed watched fields with useFieldArray

keiya01:
I added test for `useFieldArray` but this test is not working correctly. Because `getCurrentFieldsValues()` don't return correct value. 
Therefore `getCurrentFieldsValues()` return the following value when I `remove` field after I `append` field and `input` value.

```js

// after removed

[
  // first re-render
  [
    {
       "0": "1",
       "1": "1",
       "2": "1",
       "id": "1",
       "test": "",
    },
    {
      "0": "2",
      "1": "2",
      "2": "2",
      "id": "1",
      "test": "",
    },
 ],
// second re-render
 [
   "111",
   "222",
 ],
]

```

Is this result bugs? or is my implementation incorrect ?

#2072


keiya01:
@bluebill1049 
I represented this problem but I think this problem is bugs. 
Please look at the following example.

https://codesandbox.io/s/react-hook-form-usefieldarray-eq5y9?file=/src/index.js

bluebill1049:
> @bluebill1049
> I represented this problem but I think this problem is bugs.
> Please look at the following example.
> 
> https://codesandbox.io/s/react-hook-form-usefieldarray-eq5y9?file=/src/index.js

good find!

bluebill1049:
Do you want to investigate on this issue?

keiya01:
1. click `append` button
2. input value
3. click `delete` button
4. you can check strange output when you see logs

bluebill1049:
oh I see the problem because for useFieldArray need to be an object, which I should document this down on the website

so the name should be 'test[0].xxx'

keiya01:
I have an idea. I think this bugs are caused by `getCurrentFieldsValues` in `useFieldArray`.

```ts

const getCurrentFieldsValues = () =>
    watchFieldsRef.current.has(name)
      ? get(getValues(), name).map(
          (item: Partial<TFieldArrayValues>, index: number) => ({
            ...allFields.current[index],
            ...item, // this line is wrong
          }),
        )
      : allFields.current;

```

keiya01:
Oh I see ...

keiya01:
I'm sorry. And thank you for checking this issue.

bluebill1049:
the reason why we need `object` is that we inject `id` into the field item.

bluebill1049:
> I'm sorry. And thank you for checking this issue.

Thank you for writing those really useful tests! i am so confident on making changes now, it's crazy! You made a valid point as well, i will update the documentation so users aware of this too.

bluebill1049:
FYI: here is the ticket for myself: https://github.com/react-hook-form/react-hook-form/issues/2141

keiya01:
> Thank you for writing those really useful tests! i am so confident on making changes now, it's crazy! You made a valid point as well, i will update the documentation so users aware of this too.

Thank you very much🙏 💪

> FYI: here is the ticket for myself: #2141

I check this. Thank you.

src/useFieldArray.test.tsx
```diff
@@ -1344,6 +1278,67 @@ describe('useFieldArray', () => {
...
+      await waitFor(() =>
+        expect(renderedItems).toEqual([
+          [
+            { id: '1', value: '111' },
+            { id: '1', value: 'test' },
+            { id: '1', value: '222' },
+          ],
+          [{ value: '111' }, { value: '222' }, { value: '' }],
+        ]),
+      );
```
keiya01, line 1340: first render is when value was inserted.
second render is when value was watched.

@bluebill1049 
I have a question. Are this result values correct when I inserted value ?

bluebill1049: doesn't look right, looks like that's a bug.

keiya01: I try to investigate this.

bluebill1049: Thanks @keiya01! Take your time, no rush. 🙏

keiya01: @bluebill1049 
I'm sorry ...
This bug was caused by my test code ... 🙇‍♂️

keiya01:
@bluebill1049 
This PR have been completed! Could you review this?🙏

bluebill1049 (approved):
amazing! our testing wizard. 🧙

PR #2140, opened 2020-07-09
working in progress with Controller

### LINKED ISSUE #2139: Controller improvement
Hi there

I updated to RHF v6, good job guys (although it was quite hard to work through all those breaking changes)

One of the issues that came up is when implementing the `Controller` component where we used rhf input before.

# 1. Issue with boolean values

An issue that took me some time to resolve, is with `boolean-value component that are not checkboxes`

Example:
```js
<Controller
  control={control}
  defaultValue={false}
  name="Go to the beach or not?"
  as={
    <MuiSelect>
       <MuiMenuItem value={false} label="Stay home" />
       <MuiMenuItem value label="Go to the beach" />
    </MuiSelect>
   }/>
```

The above code does not work because of the following line in the Controller component:
https://github.com/react-hook-form/react-hook-form/blob/master/src/controller.tsx#L60

Because it is appending a `checked` prop in my Select component, instead of normal `value`:
https://github.com/react-hook-form/react-hook-form/blob/master/src/controller.tsx#L149

(it needs to be rewritten with manually passing onChange and value props)

Suggestion:
- Do not treat boolean value as exclusively restricted to checkbox components?
- Instead, maybe test the element to see if it correspond to a native checkbox element?
- Or, add both `checked` and `value` props (not sure if it wouldn't raise some warning though)?

# 2. Improvement suggestion

In our tests, we usually grab fields by their `name` prop, which is quite standard (for eg. in Capybara tests)./ The controller does not forward the name prop to the DOM element, which requires us to manually add it which is redundant (both on the Controller component, and the `as` or `render` content).

IS there a reason why you are not adding it? It would be a nice addition and would allow us to simplify our code that uses Controller component.

_Originally posted by @pandaiolo in https://github.com/react-hook-form/react-hook-form/discussions/2130_

bluebill1049:
`checkbox` will not be determined inside the `Controller`, instead, use `render` prop. This is due to detecting `checkbox` before render leads to bug, because `select` component can have value as `true` or `false`, and it shouldn't be the case to determine if it's a `checkbox`.

- [x] fix tests
- [x] upddate doc

close #2139

### DIFF
--- a/app/src/controller.tsx
+++ b/app/src/controller.tsx
@@ -50,11 +50,7 @@ export default function Field(props: any) {
   renderCount++;
 
   return (
-    <form
-      onSubmit={handleSubmit((d) => {
-        console.log(d);
-      })}
-    >
+    <form onSubmit={handleSubmit(() => {})}>
       <div className="container">
         <section id="input-checkbox">
           <label>MUI Checkbox</label>
--- a/app/src/controller.tsx
+++ b/app/src/controller.tsx
@@ -50,15 +50,24 @@ export default function Field(props: any) {
   renderCount++;
 
   return (
-    <form onSubmit={handleSubmit(() => {})}>
+    <form
+      onSubmit={handleSubmit((d) => {
+        console.log(d);
+      })}
+    >
       <div className="container">
         <section id="input-checkbox">
           <label>MUI Checkbox</label>
           <Controller
-            as={<Checkbox />}
             name="Checkbox"
             control={control}
             rules={{ required: true }}
+            render={(props) => (
+              <Checkbox
+                {...props}
+                onChange={(e) => props.onChange(e.target.checked)}
+              />
+            )}
           />
         </section>
 
@@ -122,9 +131,14 @@ export default function Field(props: any) {
         <section id="input-switch">
           <label>MUI Switch</label>
           <Controller
-            as={<Switch value="checkedA" />}
             name="switch"
             rules={{ required: true }}
+            render={(props) => (
+              <Switch
+                {...props}
+                onChange={(e) => props.onChange(e.target.checked)}
+              />
+            )}
             control={control}
           />
         </section>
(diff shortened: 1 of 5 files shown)


kotarella1110 (approved):
Awesome 🎊 
Not only does this solve the problem, it also reduces the amount of code 👍

bluebill1049:
> Awesome 🎊
> Not only does this solve the problem, it also reduces the amount of code 👍

Thanks @kotarella1110 🙏

keiya01 (approved):
LGTM!

bluebill1049:
> LGTM!

Thank you @keiya01 for the review.

taschetto:
Hey, @bluebill1049, just opened https://github.com/react-hook-form/react-hook-form/issues/2151 before I was this PR.

Would you care providing an example?

I've tried changing this:

`<Controller as={Checkbox} name="Checkbox" control={control} />`

To this:

`<Controller render={props => <Checkbox {...props} />} name="Checkbox" control={control} />`

Without any success...

Thank you! :-)

bluebill1049:
@taschetto here you go:

```
<Controller
  name="Checkbox"
  control={control}
  render={props => (
    <Checkbox
      onChange={e => props.onChange(e.target.checked)}
      checked={props.value}
     />
  )}
/>
```

remove the checkbox logic inside the Controller, it creates bug whenever the user using value as `boolean`. `render` prop is the way to go.

Screen Shot 2020-07-11 at 9 56 19 am

as will pass down three props `value, onChange, onBlur`

helloanil:
Wasn't this a breaking change? Until 6.0.2, following code returns `true` or `false` in the data:

```js
<Controller
  as={Checkbox}
  name="Checkbox"
  type="checkbox"
  control={control}
/>
```

However now, it returns `undefined` and `""`, or `false` and `"false"` depending on your default value. If I had built my app on top of this logic, this would have for sure broken my form.

debrice:
I might be wrong but it seems that this update also broke radio input type when using a boolean value.

PR #2162, opened 2020-07-11
fixed watch test

keiya01:


src/useForm.test.tsx
```diff
@@ -320,85 +323,6 @@ describe('useForm', () => {
     });
   });
 
-  describe('watchInternal', () => {
```
keiya01, line 323: I moved these test because I think these test should be performed with `useWatch`.
And I removed unnecessary test cases.

bluebill1049 (approved):
LGTM!

src/useForm.test.tsx
bluebill1049: sounds good to me.

PR #2166, opened 2020-07-11
Fix getValues when shouldUnregister set to false

JeromeDeLeon:
This is the reproducible bug when using `getValues`.
https://codesandbox.io/s/clever-spence-tcdn1

I didn't notice it when merging the PR from before but this problem showed for me when I created a multi-step form and only the last values got displayed on the screen even though I set `shouldUnregister` to `false`.

### DIFF
--- a/src/logic/getFieldsValues.ts
+++ b/src/logic/getFieldsValues.ts
@@ -14,8 +14,12 @@ export default <TFieldValues extends FieldValues>(
     | { nest: boolean },
 ) => {
   const output = {} as TFieldValues;
+  const fields = {
+    ...fieldsRef.current,
+    ...(unmountFieldsStateRef && unmountFieldsStateRef.current),
+  };
 
-  for (const name in fieldsRef.current) {
+  for (const name in fields) {
     if (
       isUndefined(search) ||
       (isString(search)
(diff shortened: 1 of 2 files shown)


src/logic/getFieldsValues.ts
```diff
@@ -14,8 +14,12 @@ export default <TFieldValues extends FieldValues>(
     | { nest: boolean },
 ) => {
   const output = {} as TFieldValues;
+  const fields = {
+    ...fieldsRef.current,
+    ...(unmountFieldsStateRef && unmountFieldsStateRef.current),
```
bluebill1049, line 19:
```suggestion
    ...unmountFieldsStateRef.current,
```

bluebill1049: It defaults to `{}` you don't need conditional there.

JeromeDeLeon: The TS is complaining about this one because it may possibly be `undefined`

bluebill1049: oh really. let's get it in then! thanks, @JeromeDeLeon.

bluebill1049 (approved):
👍

src/logic/getFieldsValues.ts
JeromeDeLeon: Welcome :heart:

PR #2167, opened 2020-07-11
remove extra object compare logic

bluebill1049:


JeromeDeLeon (dismissed):
LGTM! This is awesome! :fire:
Great! We don't need to separate it to a file unless this is being used or going to be used elsewhere.
Pretty neat!! :confetti_ball:

pmaier983 (dismissed):
LGTM!

JeromeDeLeon:
Oh, for checking of length, I was thinking this yesterday if its worth checking and decided not to because the objects might be pretty small and `error.types` still got iterated from start to finish.

Also, the placement is incorrect because the objects were already  compared before checking if its worth comparing the two based on there length.

Check the length first, then compare. 😬😬

bluebill1049:
Send a PR😘

bluebill1049:
Maybe worthy to beef up the tests too. So we are more confident.

PR #2171, opened 2020-07-12
lose weight by move single used function inline

bluebill1049:


src/logic/mapIds.test.ts
```diff
@@ -1,22 +0,0 @@
-import { appendId } from './mapIds';
-
-jest.mock('../logic/generateId', () => ({
```
pmaier983, line 3: I don't love removing tests if we are still using the functions. Would it be too much a size burden to export appendId?

bluebill1049: Agree. Will try to cover the tests in usefieldarry

bluebill1049: I confirm those tests are reduduant, as @keiya01 has already converted those in his tests. :) 
Screen Shot 2020-07-12 at 3 01 00 pm

bluebill1049:
gain is too little, closing...

keiya01 (approved):
LGTM :)

JeromeDeLeon (approved):
LGTM! :cocktail: :cocktail:

bluebill1049:
Thanks @JeromeDeLeon & @keiya01's review

PR #2175, opened 2020-07-12
fix #2153 issue with reValidate blur mode

### LINKED ISSUE #2153: reValidateMode: 'onBlur' is not working as expected
this is my setup 

useForm({
    mode: 'onSubmit',
    reValidateMode: 'onBlur',
    defaultValues: {
      email: '',
    },

**Describe the bug**
the revalidation is triggered after every change

**To Reproduce**
1. I submit the form with no errors (all validation rules get passed)
2. I come back to previous screen (in my case) and I start to edit the email field again, at this moment the reValidation will happen after each change and not only after onBlur event

**Expected behavior**
the reValidation should happen after onBlur or onSubmit event as per my config

bluebill1049:
- fix skip validation logic
- reduce re-render amount
- reduce code weight


JeromeDeLeon:
Am I seeing it correct? You removed a variable but it adds a byte to it. :rofl:

bluebill1049:
> Am I seeing it correct? You removed a variable but it adds a byte to it. 🤣

I am very confused here too, but i think i am still working on the proper fix here anyhow.

JeromeDeLeon:
Yeah, fine! I just noticed it. The bot assigned to that task of checking file size is probably laughing now lol

keiya01:
I think we should test about this issue with useForm. Could you write test?

bluebill1049:
> I think we should test about this issue with useForm. Could I write test?

sure, 🧙‍♂️please

keiya01:
I added test :)

keiya01 (approved):
LGTM 🧙‍♂️

bluebill1049:
nice! thanks @keiya01 🙏 🧙

PR #2183, opened 2020-07-13
improved bundle size with production code

keiya01:
I think error message (like `console.warn`) is needed only in development mode. Therefore we need to remove this line with tree shaking.


bluebill1049:
would be nice if can avoid increate the bundle size 👀

keiya01:
@bluebill1049 
hmmm, do you think about increasing bundle size in development environment.

bluebill1049:
oh, wait you are right! sorry 🙏 let's go ahead with it

keiya01:
Is there code else that is used only in development environment?

bluebill1049:
> Is there code else that is used only in development environment?

don't think so,

bluebill1049 (approved):
👍

PR #2230, opened 2020-07-15
fix #2220 with getValues()

### LINKED ISSUE #2229: reset issue on 6.0.5 with useWatch
Bill,

I upgraded my project to 6.0.5 and a form with a structure similar to the one from the codesandbox below stopped functioning.

https://codesandbox.io/s/usewatch-bug-605-pm6uw?file=/src/App.tsx

Steps to reproduce:
1) useForm is called with a initial defaultValues property
2) after a little while, some data is fetched from the server and the component asks the form to reset to the data fetched from the server (this is simulated with a setTimeout on the codesandbox)
3) whenever a value is updated in the form, the result from useWatch will always return the defaultValues that were set in step 2, instead of the updated values (inspect the console log to see the values being logged)

The codesandbox above has a working example of the bug. If I change the version back to 6.0.4, it works fine. The problem is with the reset call in the object returned from useForm. After the reset call, useWatch stops working correctly. In my project I refactored the code to not call the reset and it now works, but I think other people might go through the same problem.

_Originally posted by @eduleite in https://github.com/react-hook-form/react-hook-form/discussions/2228_

bluebill1049:
@JeromeDeLeon `getFieldsValues` are flat object once merge with unmounted input data which will lead to error, so I just do a merge at getValues function now.

- [ ] follow up a unit test at getValues() at useForm

close #2229

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -557,9 +557,10 @@ export function useForm<
       );
     }
 
-    return transformToNestObject(
-      getFieldsValues(fieldsRef, unmountFieldsStateRef),
-    );
+    return transformToNestObject({
+      ...unmountFieldsStateRef.current,
+      ...getFieldsValues(fieldsRef, unmountFieldsStateRef),
+    });
   }
 
   const validateResolver = React.useCallback(
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -557,10 +557,7 @@ export function useForm<
       );
     }
 
-    return transformToNestObject({
-      ...unmountFieldsStateRef.current,
-      ...getFieldsValues(fieldsRef, unmountFieldsStateRef),
-    });
+    return getFieldsValues(fieldsRef, unmountFieldsStateRef);
   }
 
   const validateResolver = React.useCallback(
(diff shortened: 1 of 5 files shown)


src/useForm.ts
```diff
@@ -557,9 +557,10 @@ export function useForm<
       );
     }
 
-    return transformToNestObject(
-      getFieldsValues(fieldsRef, unmountFieldsStateRef),
-    );
+    return transformToNestObject({
+      ...unmountFieldsStateRef.current,
+      ...getFieldsValues(fieldsRef, unmountFieldsStateRef),
```
JeromeDeLeon, line 562: I will try to look into it later. It seems like `unmountFieldsStateRef` will be useless beacause `fieldsRef` only contains the mounted fields that's why I made this changes so `getFieldValue` will only get the mounted fields' value.

bluebill1049: you have a valid point there.

keiya01:
LGTM!!

keiya01 (approved):

JeromeDeLeon:
It really did break with fix last time sorry about that.

bluebill1049:
> It really did break with fix last time sorry about that.

No man, not at all. It's all for the improvement. If you are happy with this, let's patch it.

JeromeDeLeon (approved):

bluebill1049:
Thanks @JeromeDeLeon and @keiya01's review.

PR #2238, opened 2020-07-15
Loosen Resolver type constraints (a bit)

blakepettersson:
The reason we need to loosen the type constraints of `Resolver` is due
to the fact that io-ts can refine a type to something else other than
`TFieldValues`. In order to get around that, we define another type
parameter which is by default instantiated to `TFieldValues` in order
to preserve backwards compatibility, but which can be overridden by
a specific implementation of `Resolver`.

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -56,28 +56,30 @@ export type SubmitHandler<TFieldValues extends FieldValues> = (
   event?: React.BaseSyntheticEvent,
 ) => void | Promise<void>;
 
-export type ResolverSuccess<TFieldValues extends FieldValues = FieldValues> = {
-  values: UnpackNestedValue<TFieldValues>;
+export type ResolverSuccess<T> = {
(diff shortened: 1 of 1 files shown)


bluebill1049:
Do you try to return values which is not what you defined in your `FormValues`? I am not sure we need to loosen up the type in here.

kotarella1110 (commented):
umm...
resolvers allow custom validation, but the return type is not arbitrary.
value should be `FieldValues` type.

blakepettersson:
So I spent some time digging a bit further into this... the original issue is that the default validation type that io-ts gives is not parsable as a `FieldValues` type (which was why I tried to make the resolver not being depend on `FieldValues`). In the end, it turns out that io-ts provides an escape hatchet for such situations, which means this PR is not necessary. Sorry about that!

bluebill1049:
all good @blakepettersson please send us more PR if you have good ideas in the long run. 🙏

PR #2250, opened 2020-07-16
fix #2240 issue with unmounted data gets updated when shouldUnregister

### LINKED ISSUE #2240: remove doesnt work on useFieldArray when reset has been done on form
**Describe the bug**
Hello! I hope you are doing well!

The form does not remove the data but the UI elements are removed when using `useFieldArray` after a `reset` has been done on the form. If I provide the data as `defaultValues` instead of the reset, when I remove an item, it works as intended.

**To Reproduce**
1. After the form has loaded, hit submit.
2. You will see that there are 3 items in the array in the console log of the form data.
3. Delete an item and hit submit again and you will see 3 items still in the array in the console.log.

https://codesandbox.io/s/react-hook-form-usefieldarray-nested-arrays-cku66?file=/src/index.js

**Expected behavior**
I expect when the UI changes, I hit remove on an element, after I hit submit it reflects what is actually on the form.

**Additional context**
Originally I thought the `reset` was just like providing `defaultValues` to the form but it seems to act completely different. `reset` actually seems to register data differently with the form than using `defaultValues`

If I need to be more clear in my context let me know. Also, for some reason append and prepend do work, just not remove. Those only work after I call register after the reset.

bluebill1049:
fix #2240

### DIFF
--- a/src/useForm.test.tsx
+++ b/src/useForm.test.tsx
@@ -711,8 +711,10 @@ describe('useForm', () => {
         }),
       );
 
-      result.current.setValue('test', '1');
-      result.current.setValue('checkbox', ['1', '2']);
+      act(() => {
+        result.current.setValue('test', '1');
+        result.current.setValue('checkbox', ['1', '2']);
+      });
 
       expect(result.current.control.unmountFieldsStateRef.current).toEqual({
         checkbox: ['1', '2'],
@@ -839,7 +841,9 @@ describe('useForm', () => {
     it('should not work if field is not registered', () => {
       const { result } = renderHook(() => useForm());
 
-      result.current.setValue('test', '1');
+      act(() => {
+        result.current.setValue('test', '1');
+      });
 
       expect(result.current.control.fieldsRef.current['test']).toBeUndefined();
     });
--- a/src/useForm.test.tsx
+++ b/src/useForm.test.tsx
@@ -704,7 +704,7 @@ describe('useForm', () => {
       });
     });
 
-    it('should not set unmountFieldsState value when shouldUnregister is set to false', async () => {
+    it('should set unmountFieldsState value when shouldUnregister is set to false', async () => {
       const { result } = renderHook(() =>
         useForm<{ test: string; checkbox: string[] }>({
           shouldUnregister: false,
@@ -714,7 +714,10 @@ describe('useForm', () => {
       result.current.setValue('test', '1');
       result.current.setValue('checkbox', ['1', '2']);
 
-      expect(result.current.control.unmountFieldsStateRef.current).toEqual({});
+      expect(result.current.control.unmountFieldsStateRef.current).toEqual({
+        checkbox: ['1', '2'],
+        test: '1',
+      });
     });
 
     it('should set nested value correctly ', () => {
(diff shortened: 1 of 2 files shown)


src/useForm.test.tsx
```diff
@@ -694,6 +704,25 @@ describe('useForm', () => {
       });
     });
 
+    it('should not set unmountFieldsState value when shouldUnregister is set to false', async () => {
```
keiya01, line 707: this test has not been able to test correctly too.
You should check `unmountFieldsStateRef` after executed `unmount()` or `unregister()`.

src/useForm.test.tsx
```diff
@@ -492,6 +492,16 @@ describe('useForm', () => {
       });
     });
 
+    it('should not reset unmountFieldsState value when shouldUnregister set to false ', () => {
```
keiya01, line 495: this test has not been able to test correctly.
You should check unmountFieldsStateRef after executed unmount() or unregister().

keiya01 (changes requested):
I reviewed code!

src/useForm.test.tsx
bluebill1049: thanks @keiya01 the usage is actually during setValue.

src/useForm.test.tsx
bluebill1049: this one is getting updated during `reset` api

keiya01: oops... this test is correct... I'm sorry.

src/useForm.test.tsx
keiya01: But I think [this line](https://github.com/react-hook-form/react-hook-form/blob/fix-2240/src/useForm.ts#L416) is not called because `fieldsRef` is exist.(https://github.com/react-hook-form/react-hook-form/blob/fix-2240/src/useForm.ts#L407)

bluebill1049: good one! I will get that fixed. thank you 🕵️

bluebill1049:
Thanks @keiya01 i have updated the tests, please re-review 👍

src/useForm.test.tsx
```diff
@@ -492,6 +492,16 @@ describe('useForm', () => {
...
+      const { result } = renderHook(() => useForm());
```
keiya01, line 496: Does this line need to specify `shouldUnRegister: false` ?

bluebill1049: wow thanks, almost tested for nothing...

src/useForm.test.tsx
```diff
@@ -492,6 +492,20 @@ describe('useForm', () => {
...
+        useForm({
+          shouldUnregister: false,
+        }),
+      );
+
+      result.current.register('test');
+
+      act(() => result.current.reset({ test: 'test' }));
+
+      expect(result.current.control.unmountFieldsStateRef.current).toEqual({});
```
keiya01, line 506: Does this test failure ?

bluebill1049: yea you are right it should fail... why unit test didn't fail.

bluebill1049: thanks @keiya01 🙏

keiya01: hmmm ... I don't know...
It's scary😅

keiya01 (approved):
LGTM :)
Thank you very much🙏

bluebill1049:
> LGTM :)
> Thank you very much🙏

Thank you very much for your patient review! sorry for those mistakes...

keiya01:
Don't worry :)

PR #2254, opened 2020-07-16
expose ArrayField type

bluebill1049:
expose: `ArrayField`  type

https://github.com/react-hook-form/react-hook-form/discussions/2174#discussioncomment-38042

- [ ]  update doc

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -9,5 +9,6 @@ export {
   ValidationRules,
   FieldError,
   Field,
+  ArrayField,
   Mode,
 } from './form';


src/types/index.ts
```diff
@@ -9,5 +9,6 @@ export {
   ValidationRules,
   FieldError,
   Field,
+  ArrayField,
```
kotarella1110, line 12: I think better rename to `ArrayFieldState`.
What do you think ?

bluebill1049: how about `ArrayFields`?

kotarella1110: `ArrayField` type is not array type.
`ArrayField` or `ArrayFieldState` is better than `ArrayFields` 👍

bluebill1049: you are right. shall we just keep `ArrayField` then? it's a represent the shape of FieldArray item.

bluebill1049:
```
const toniosCat: ArrayField<{ test: string}> = {
  test: 'test',
  id: 'xxx'
};
```

looks good as above :)

kotarella1110: let's keeep `ArrayField` 👍

kotarella1110 (approved):

src/types/index.ts
bluebill1049: thanks for the review :)

PR #2267, opened 2020-07-17
do not passing refs to logic functions

mh-alahdadian:
logic functions are getting data and calculate depend on it so they don't have callback function which needs to access  latest value in ref in future and it just want to access current value of ref which we can pass it directly 
in another way logic functions should not import React and should not depend on it so we shouldn't use refs in them 

in this PR tests has been changed and remove current creating in them but I do not know what should I do with snapshot , please someone fix snapshots 

thank you

### DIFF
--- a/src/logic/findRemovedFieldAndRemoveListener.test.ts
+++ b/src/logic/findRemovedFieldAndRemoveListener.test.ts
@@ -19,9 +19,7 @@ describe('findMissDomAndClean', () => {
     it('should return default fields value if nothing matches', () => {
       document.body.contains = jest.fn(() => true);
       const fields = {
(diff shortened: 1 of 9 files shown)


MHA15:
@bluebill1049

bluebill1049:
Thanks for the PR @MHA15 i will take a look this issue probably a bit later.

MHA15:
> Thanks for the PR @MHA15 i will take a look this issue probably a bit later.

thank you and please fix snapshots of tests I did not work with them yet

bluebill1049:
hey @MHA15 I think I will update progressively, function by function rather in one PR.

PR #2274, opened 2020-07-18
Added useWatch test about partial re-render

keiya01:


bluebill1049 (approved):
Thanks @keiya01 LGTM.

PR #2275, opened 2020-07-18
reduce getPath logic

bluebill1049:
- reduce `getPath` logic

it's always surprising and great to keep reading the code and find anything can be removed to reduce the lib package size. 💃


bluebill1049:
I am pretty confident with those changes because @keiya01's tests.

JeromeDeLeon (approved):
LGTM!❤️

PR #2286, opened 2020-07-20
Increased useWatch test

keiya01:


src/useWatch.test.tsx
```diff
@@ -215,6 +215,46 @@ describe('useWatch', () => {
       expect(childCount).toBe(1);
     });
 
+    it("should not external re-render when field name don't match", async () => {
```
bluebill1049, line 218:
```suggestion
    it("should not re-render external component when field name don't match", async () => {
```

keiya01: Thank you @bluebill1049 :)

bluebill1049 (approved):
LGTM! thanks @keiya01

PR #2305, opened 2020-07-21
Allow resolver to be sync or async

Satyam:
Resolver may now return an object with `values` and `errors` objects
or a Promise returning the same object.

Commit contains a copy of an existing resolver test with the resolver function changed to synchronous. It fails with v6.0.8 but now passes.  
I also changed the type definition for `Resolver` so that it now returns a `ResolverResult` besides a `Promise<ResolverResult>`

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -85,7 +85,7 @@ export type Resolver<
   values: TFieldValues,
   context?: TContext,
   validateAllFieldCriteria?: boolean,
-) => Promise<ResolverResult<TFieldValues>>;
+) => Promise<ResolverResult<TFieldValues>> | ResolverResult<TFieldValues>;
 
 export type UseFormOptions<
   TFieldValues extends FieldValues = FieldValues,
--- a/src/useForm.test.tsx
+++ b/src/useForm.test.tsx
@@ -2427,6 +2427,46 @@ describe('useForm', () => {
         expect(renderCount).toBe(2);
       });
 
+      it('with sync resolver it should contain error if value is invalid with resolver', async () => {
+        const mockResolver = jest.fn();
+        const resolver = (data: any) => {
+          if (data.test) {
+            return { values: data, errors: {} };
+          }
+          mockResolver();
+          return {
+            values: data,
+            errors: {
+              test: {
+                message: 'resolver error',
+              },
+            },
+          };
+        };
+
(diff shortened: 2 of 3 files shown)


bluebill1049:
Thanks for the PR, will take a look at it soon.

bluebill1049:
This looks fine by me 👍 @kotarella1110 would you like to take a look too?

keiya01 (approved):
LGTM! Thank you very much.

kotarella1110:
@bluebill1049 If I remember correctly, return value is only support async resolver (Promise) since v6.
https://github.com/react-hook-form/react-hook-form/pull/1471/commits/4895d9fea1142b24605700da9c03885e1a2bbb18#diff-41df78d86ed4de80aae9b0b9560feb2bR93-R94
In v5 supported both sync and async, I think that there is no problem in restoring it.

bluebill1049:
> @bluebill1049 If I remember correctly, return value is only support async resolver (Promise) since v6.
> [4895d9f#diff-41df78d86ed4de80aae9b0b9560feb2bR93-R94](https://github.com/react-hook-form/react-hook-form/commit/4895d9fea1142b24605700da9c03885e1a2bbb18#diff-41df78d86ed4de80aae9b0b9560feb2bR93-R94)
> It is supported both sync and async in v5, I think that there is no problem in revert it.

Yes, you are right. It was due to consistency, but it turns out probably support both its easier for developers. If you are happy to revert it back to async/sync,  you happy to approve?

kotarella1110:
> > @bluebill1049 If I remember correctly, return value is only support async resolver (Promise) since v6.
> > [4895d9f#diff-41df78d86ed4de80aae9b0b9560feb2bR93-R94](https://github.com/react-hook-form/react-hook-form/commit/4895d9fea1142b24605700da9c03885e1a2bbb18#diff-41df78d86ed4de80aae9b0b9560feb2bR93-R94)
> > It is supported both sync and async in v5, I think that there is no problem in revert it.
> 
> Yes, you are right. It was due to consistency, but it turns out probably support both its easier for developers. If you are happy to revert it back to async/sync, would you like to approve?

sure!

kotarella1110 (approved):

bluebill1049:
Thanks @keiya01 & @kotarella1110's review.

PR #2313, opened 2020-07-22
fix #2312 issue useFieldArray with react beautiful DND

### LINKED ISSUE #2312: After upgrading to v6 field array value changes no longer persist after a drag and drop using React Beautiful DnD.
After upgrading to v6 field array value changes no longer persist after a drag and drop using React Beautiful DnD.

https://codesandbox.io/s/react-hook-form-usefieldarray-kit8r?file=/src/index.js

**Steps to Reproduce:**
1. Enter a value in the two inputs shown
2. Using the 'Drag Me' element drop one input into the place of the other

**Result:**
Data entered into both fields are reset to the default
**Expected Result:**
Data entered moves with the input

Switch React Hook Form version to 5.7.2 in the code sandbox and repeat the steps. The changed values move as expected.

Any idea what has changed?

_Originally posted by @nlrowe in https://github.com/react-hook-form/react-hook-form/discussions/2309_

bluebill1049:
fix #2312

### DIFF
--- a/src/useFieldArray.test.tsx
+++ b/src/useFieldArray.test.tsx
@@ -1300,7 +1300,6 @@ describe('useFieldArray', () => {
             } as any,
             dirtyFieldsRef,
             touchedFieldsRef,
-            getValues: () => ({ test: [] }),
           }),
           name: 'test',
         }),
@@ -1492,7 +1491,6 @@ describe('useFieldArray', () => {
                 dirtyFields: true,
               },
             } as any,
-            getValues: () => ({ test: [] }),
             dirtyFieldsRef,
           }),
           name: 'test',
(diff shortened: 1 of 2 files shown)


src/useFieldArray.test.tsx
```diff
@@ -1863,7 +1857,6 @@ describe('useFieldArray', () => {
                 'test[1]': { ref: { name: 'test[1]', value: { test: '2' } } },
               },
             },
-            getValues: () => ({ test: [] }),
```
bluebill1049, line 1866: I think those tests will all have to rewrite the way @keiya01 is written, more of integration tests then just test `useFieldArray`.

keiya01: I think so too. We should not use `reconfigureControl()`.

bluebill1049: totally agree. you happy for this to move in first?

keiya01: It's fine :)

PR #2300, opened 2020-07-21
fix #2298 issue around isValid with reset API

bluebill1049:
close #2298

Issue #2298, opened 2020-07-20
formState isValid issue with `reset`


bluebill1049:
I want the submit button to be disabled if the form is in an invalid state. However, the button remains disabled until I clear a field & type again. Below is a code snippet for producing this issue. I am quite new to react, any idea what could be wrong here?

```
function TestForm() {
    const {register, handleSubmit, formState, reset} = useForm({mode: "onChange"});
    const {isValid} = formState;

    useEffect(() => {
        const data = {name: "John", count: 10};
        reset(data);
    }, [reset]);

    return (
        <div>
            <h1>Form state valid: {isValid.toString()}</h1>
            <form onSubmit={handleSubmit(d => console.log(d))}>
                <input type="text" name="name"
                       ref={register({required: true})}/>

                <input type="text" name="count"
                       ref={register({required: true})}/>

                <input type="submit" disabled={!isValid}/>
            </form>
        </div>
    );
}
```
![image](https://user-images.githubusercontent.com/15045608/87881229-56ef4e80-ca11-11ea-9830-d8cd83ae4b8f.png)

_Originally posted by @salmanbabri in https://github.com/react-hook-form/react-hook-form/discussions/2281_

PR #2314, opened 2020-07-22
reverted #2300

keiya01:
I reverted #2300 because this change did not resolve issue. This problem is caused by `validateField()` is asynchronous. That is, this problem is caused by `reset()` is executed before previous `validateField()` is completed executing. 
I tried to fix this issue but I could not resolve. However we can avoid this problem  by using `setTimeout()`.

codesandbox: https://codesandbox.io/s/winter-cache-ipq9b?file=/src/App.js

PR: #2300
Issue: #2298

### DIFF
--- a/src/useForm.test.tsx
+++ b/src/useForm.test.tsx
@@ -201,6 +201,39 @@ describe('useForm', () => {
 
       expect(result.current.getValues()).toEqual({ test: 'test' });
     });
+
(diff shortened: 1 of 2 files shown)


bluebill1049:
wow, you are right @keiya01 because the `validate` is async. thanks for fixing this.

PR #2347, opened 2020-07-24
fix Chinese Traditional README wrong html tag

chouandy:


bluebill1049:
oops i miss that during the last review. thanks.

PR #2358, opened 2020-07-25
fix #2310 useWatch with useFieldArray delete

### LINKED ISSUE #2310: useWatch value become empty after delete field array item
**Describe the bug**

**To Reproduce**
1. Go to https://codesandbox.io/s/react-hook-form-v6-usewatch-mjnow?file=/src/index.js
2. Click on append
3. Click on append
4. Click on delete
5. See JSON output

**Expected behavior**

it should preserve value same as `watch`

**Screenshots**
image
image

bluebill1049:
- [x] unit test

close #2310

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -443,14 +443,16 @@ export const useFieldArray = <
     if (isWatchAllRef.current) {
       reRender();
     } else if (watchFieldsRef) {
+      let shouldRenderUseWatch = true;
       for (const watchField of watchFieldsRef.current) {
         if (watchField.startsWith(name)) {
           reRender();
+          shouldRenderUseWatch = false;
           break;
         }
       }
 
-      shouldRenderFieldArray();
+      shouldRenderUseWatch && shouldRenderFieldArray();
     }
 
     if (focusIndexRef.current > -1) {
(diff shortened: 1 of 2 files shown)


src/useFieldArray.ts
```diff
@@ -443,12 +443,16 @@ export const useFieldArray = <
     if (isWatchAllRef.current) {
       reRender();
     } else if (watchFieldsRef) {
+      let shouldRenderUseWatch = true;
       for (const watchField of watchFieldsRef.current) {
         if (watchField.startsWith(name)) {
           reRender();
+          shouldRenderUseWatch = false;
```
keiya01, line 450: Isn't this extra `re-render` ?

bluebill1049: reRender will trigger the re-render at the root (useForm) level, so we probably can skip the extra re-render for `useWatch`.

keiya01: Oh I see. I misunderstood ...

keiya01 (approved):
LGTM :)

src/useFieldArray.ts
bluebill1049: thanks for the detailed review @keiya01 🙏

Issue #2308, opened 2020-07-21
Setting default value


mrgazanfarli:
**Describe the bug**
When taking the form methods by using `useForm<Type>({ defaultValues: values })` , I cannot get values by using `getValues()`

**To Reproduce**
1. Get form methods from `useForm<Type>({ defaultValues: values })`
2. Use `<FormProvider {...formMethods}>{children}</FormProvider>` 
3. Get values from `useFormContext` inside a nested component
4. And the values will be `{}`

**Codesandbox link (Required)**
https://codesandbox.io/s/brave-wescoff-s8i8k

**Expected behavior**
`getValues()` method had to return default values as I had provided them

**Desktop (please complete the following information):**
 - OS: macOS Catalina
 - Browser: Chrome v84.0.4147.89

nigarjafarr:
+
I am facing same issue while using last version react-hook-form

bluebill1049:
hey guys, this is not a bug, but when you are using getValue during the render, its meant to retrieve inputs values?

gilbarbara:
Hey,

I was using getValues with my Input component but apparently it's not working with 6.x.
Did it change? I couldn't find anything about this in the CHANGELOG.

bluebill1049:
hey @gilbarbara can you sure a codesandbox?

mrgazanfarli:
> hey guys, this is probably not a bug, but when you are using getValue during the render, its meant to retrieve inputs values?

yes, I just want to take values (at least default ones), but the method does not return them. It was OK with the 5.x versions. After upgrading the library, I observe this problem very often

bluebill1049:
can you share your usage @mrgazanfarli?

mrgazanfarli:
@bluebill1049 my usage scenario is the same as I wrote on codesandbox

mrgazanfarli:
I use `useForm` and provide `defaultValues`, but I cannot get them. This is the simplest way of describing the problem :)

bluebill1049:
looks like everyone is expected getValues to return defaultValues, we will sort it out then.

mrgazanfarli:
> looks like everyone is expected getValues to return defaultValues, we will sort it out then.

If I use useForm and have no changes, then getValues should return the default values

bluebill1049:
would you be able to take care of this one @keiya01?

keiya01:
@bluebill1049 I can do it. I will check this issue.

bluebill1049:
[...]

PR #2359, opened 2020-07-25
Enhance #2308

keiya01:
- I fixed `getFieldValue` and `getFieldsValues` then related methods are changed too.
- By these changes, `handleSubmit` has been to pass `defaultValues` when fields is not registered.

issue: #2308

### DIFF
--- a/src/logic/findRemovedFieldAndRemoveListener.ts
+++ b/src/logic/findRemovedFieldAndRemoveListener.ts
@@ -8,7 +8,15 @@ import isArray from '../utils/isArray';
 import unset from '../utils/unset';
 import unique from '../utils/unique';
 import isUndefined from '../utils/isUndefined';
-import { Field, FieldRefs, FieldValues, Ref } from '../types/form';
+import {
+  Field,
+  FieldRefs,
+  FieldValues,
+  Ref,
+  FieldValue,
+  UnpackNestedValue,
+} from '../types/form';
+import { DeepPartial } from '../types/utils';
 
 const isSameRef = (fieldValue: Field, ref: Ref) =>
   fieldValue && fieldValue.ref === ref;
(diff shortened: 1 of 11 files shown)


src/logic/findRemovedFieldAndRemoveListener.ts
```diff
@@ -19,7 +27,13 @@ export default function findRemovedFieldAndRemoveListener<
   fieldsRef: React.MutableRefObject<FieldRefs<TFieldValues>>,
   handleChange: ({ type, target }: Event) => Promise<void | boolean>,
   field: Field,
-  unmountFieldsStateRef: React.MutableRefObject<Record<string, any>>,
+  optionalState: {
```
bluebill1049, line 30: Minor: 🙏 sorry to be the pain in the ass, I think this will increase the bundle size, any chance we can try just flat it? If too much trouble it's ok to leave as it is too.

src/logic/getFieldValue.ts
```diff
@@ -40,7 +53,9 @@ export default function getFieldValue<TFieldValues extends FieldValues>(
     return value;
   }
 
-  if (unmountFieldsStateRef) {
-    return unmountFieldsStateRef.current[name];
-  }
+  return (
+    optinalState &&
+    (optinalState.unmountFieldsStateRef.current[name] ||
+      optinalState.defaultValuesRef.current[name])
```
bluebill1049, line 59: What about this?
```
{
  ...defaultValuesRef.current,
  ...unmountFieldsStateRef.current,
}
```

otherwise, we will need to check empty object.

bluebill1049:
Solid PR! just some minor comments.

keiya01:
Thank you for reviewing. I will fix your review!

src/logic/getFieldValue.ts
keiya01: I think `getFieldValue` should return `undefined` if `name` is undefined.

bluebill1049: you are right!

bluebill1049:
Thank you very much for the fix the args... sorry for that as well, it must be tedious to change. 🙏

bluebill1049 (approved):
🎉 amazing `getValues()` update!

PR #2360, opened 2020-07-25
fixed #2339

### LINKED ISSUE #2339: reset issue on 6.0.8 with useWatch
**Describe the bug**
Looks like `reset` do not notify `useWatch` subscription. However, `setValue` notify both `useWatch` and `watch`

I'm not sure it's bug or not. If this is not a bug, should we update the doc to make it more clear?

From my understand, useWatch === watch except useWatch isolate re-rendering at the component level.  The doc says `useWatch` Behaves similarly to the `watch` API. 

it's not clear what's the differences between them

**To Reproduce**
1. Go to https://codesandbox.io/s/react-hook-form-v6-usewatch-gci0c?file=/src/index.js
2. Click on use reset
3. See output

2. Click on use set value
3. See output

 
**Expected behavior**

`useWatch` and `watch` should both get notify when use `reset` api

keiya01:


bluebill1049 (approved):
oh wow! simple and effective fix. Thank you 🎯

PR #2377, opened 2020-07-27
new feature onTouched mode

bluebill1049:
This feature is related to user experience behavior.

`onBlur` will only trigger validation with a blur event.

`onTouched` mode will only trigger validation after input gets touched.

- [x] automation
- [x] unit tests


bluebill1049:
TS question: @kotarella1110 
if i change this line (https://github.com/react-hook-form/react-hook-form/pull/2377/files#diff-a922078008692a8cd2e441a97dd8fa24R96) to `Omit<xxx, 'all', 'onTouch'>` then i get an issue with `keyof mode` in `SkipValidation` function, do you know how to solve this problem?

kotarella1110:
> TS question: @kotarella1110
> if i change this line (https://github.com/react-hook-form/react-hook-form/pull/2377/files#diff-a922078008692a8cd2e441a97dd8fa24R96) to `Omit<xxx, 'all', 'onTouch'>` then i get an issue with `keyof mode` in `SkipValidation` function, do you know how to solve this problem?

what line is that written on?

bluebill1049:
oops that link didn't work...

`form.ts`
Screen Shot 2020-07-29 at 7 05 04 pm

kotarella1110:
@bluebill1049 You should use `Exclude` instead of `Omit` 👍

bluebill1049:
oh thanks @kotarella1110

src/useForm.test.tsx
```diff
@@ -2832,4 +2832,55 @@ describe('useForm', () => {
...
+        );
+      };
+
+      render(<Component />);
+
+      screen.getByRole('textbox').focus();
+
+      await actComponent(async () => {
+        await fireEvent.blur(screen.getByRole('textbox'));
+      });
```
JeromeDeLeon, line 2861: Kinda confuse, will it trigger also even without doing blur event? because in my understanding, `onTouch` will validate the moment I focus/touch the field, right? and then again, you triggered the `onChange` function then validation run, but should it run? I'm kinda confuse with this behavior. I'm good with this feature just confuse sorry :sweat_smile:

bluebill1049: good question :) touched only gets updated after user leave input focused. so the first focus with change will not trigger validation. only after they have focused on the inputs.

bluebill1049: perhaps we should name it `onTouched`

bluebill1049: make sense. updating...

JeromeDeLeon: Sorry I forgot to follow up. Thanks for your explanation. It somehow makes sense.

Shouldn't we have revalidate mode for this or blur is enough?

bluebill1049: after input been touched, the validation will be switched to `onChange`. That's the difference.

src/useForm.ts
```diff
@@ -649,7 +649,7 @@ export function useForm<
 
     set(errorsRef.current, name, {
       ...error,
-      ref: fieldsRef.current[name]?.ref,
+      ref: (fieldsRef.current[name] || {})!.ref,
```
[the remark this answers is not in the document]
JeromeDeLeon: Maybe because of it's compiled version is longer hence causing more bytes? Am I right? @bluebill1049

src/useForm.test.tsx
JeromeDeLeon: Thank you 😁

src/useForm.ts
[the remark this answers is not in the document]
bluebill1049: yes :( it will add a polyfill.

src/useForm.test.tsx
bluebill1049: Some users/UX prefer this type of validation. Personally I prefer `onSubmt` but it's good to give them the option too, otherwise, they will have to dot the following:

```
useForm({ mode: 'onChange' })

{touched.test && errors.test.message}
```

src/useForm.ts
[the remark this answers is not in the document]
adhrinae: Understood 👍

src/useForm.test.tsx
JeromeDeLeon: Same haha I also prefer to handle validation onSubmit 🤣

src/useForm.test.tsx
```diff
@@ -2832,4 +2832,55 @@ describe('useForm', () => {
...
+
+      expect(screen.getByText('This is required.')).toBeInTheDocument();
```
keiya01, line 2863:
```suggestion
      expect(screen.queryByText('This is required.')).toBeInTheDocument();
```

I think you should `screen.queryByText()` because `getBy*` method throw error when specified element could not found.

src/useForm.test.tsx
```diff
@@ -2832,4 +2832,55 @@ describe('useForm', () => {
...
+
+      await actComponent(async () => {
+        fireEvent.input(screen.getByRole('textbox'), {
+          target: {
+            value: '',
+          },
+        });
+      });
+
+      expect(screen.getByText('This is required.')).toBeInTheDocument();
```
keiya01, line 2883:
```suggestion
      expect(screen.queryByText('This is required.')).toBeInTheDocument();
```

keiya01 (changes requested):
This is awesome feature 🎉

src/useForm.test.tsx
bluebill1049: awesome! thanks @keiya01 🙏

pmaier983 (approved):
LGTM! Awesome work :D

bluebill1049:
Thanks for the review @keiya01 and @pmaier983. Will merge this in once all the remaining issues are resolved. we can release this with @adhrinae's feature too.

PR #2390, opened 2020-07-27
fix #2382 issue with defaultValues inside submit values

### LINKED ISSUE #2382: [useFieldArray] The value will be the default value when removed all fields
**Describe the bug**
The value will be the default value when removed all fields. If the arguments of useForm has defaultValues.

**To Reproduce**
1. Go to the codesandbox provided below
2. Remove all array fields
3. Click submit
4. See console output

https://codesandbox.io/s/react-hook-form-usefieldarray-nested-arrays-cqvfs

**Expected behavior**
The data of the submission should be an empty array.

**Desktop:**
 - OS: macOS
 - Browser: chrome
 - Version: 84.0.4147.89

bluebill1049:
@keiya01 this is what i meant. (feel free to take over whether you think this is right or wrong solution)

### DIFF
--- a/cypress/integration/useFieldArray.ts
+++ b/cypress/integration/useFieldArray.ts
@@ -170,9 +170,7 @@ describe('useFieldArray', () => {
     cy.get('ul > li').should('have.length', 0);
 
     cy.get('#submit').click();
-    cy.get('#result').contains(
-      '{"data":[{"name":"test"},{"name":"test1"},{"name":"test2"}]}',
-    );
+    cy.get('#result').contains('{}');
 
     cy.get('#append').click();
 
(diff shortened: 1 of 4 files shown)


keiya01:
Thank you. It looks good to me.
I was worried data of `getValues()` and `handleSubmit()` have no consistent. But I noticed this change has no problem because we can do opt-in `defaultValues` by using `getValues()`.

cypress/integration/useFieldArray.ts
```diff
@@ -170,9 +170,7 @@ describe('useFieldArray', () => {
     cy.get('ul > li').should('have.length', 0);
 
     cy.get('#submit').click();
-    cy.get('#result').contains(
-      '{"data":[{"name":"test"},{"name":"test1"},{"name":"test2"}]}',
-    );
+    cy.get('#result').contains('{}');
```
bluebill1049, line 173: can't believe the mistake stays there for long.

kotarella1110 (approved):
Awesome!
I think in the docs, clarifying the difference between handleSubmit and getValues ​​about defaultValues ​​will avoid user confusion.
What do you think?

bluebill1049:
will do @kotarella1110

bluebill1049:
Screen Shot 2020-07-28 at 5 30 00 pm

PR #2407, opened 2020-07-29
Fix 2406

### LINKED ISSUE #2406: `setValue()` does not blur and therefore `formState.isValid` is not usable
**Describe the bug**
`setValue()` does not trigger _blur_ so `formState.isValid` isn't true even when `mode: blur`

But tabbing into (past) the field does make `formState.isValid` true.

https://codesandbox.io/s/rhf-react-select-setvalue-isvalid-g9dk2?file=/src/index.js was as close as I got but couldn't get it to actually hide the Submit Button based on `formState.isValid` even though it's working for my own project.

 - OS: Mac
 - Browser: Edge
 - Version: 84.0.522.44

**Additional context**
I also tried:

* `<Controller ... rules={{ required: true }} ... as React Select ... />`
* `reValidateMode: 'all'`

FYI: docs site mentions `validateCriteriaMode` but that seems like it's maybe an outdated reference???

bluebill1049:
fix #2406

### DIFF
--- a/src/__mocks__/reconfigureControl.ts
+++ b/src/__mocks__/reconfigureControl.ts
@@ -44,7 +44,12 @@ export const reconfigureControl = (
   removeFieldEventListener: jest.fn(),
   errorsRef: { current: {} },
   touchedFieldsRef: { current: {} },
-  mode: { isOnSubmit: false, isOnBlur: false, isOnChange: false },
+  mode: {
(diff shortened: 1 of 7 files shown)


keiya01 (approved):
LGTM!

bluebill1049:
thank you for the review @keiya01

PR #2418, opened 2020-07-30
fix #2403 issue with re-render valid formState in useFieldArray

### LINKED ISSUE #2403: useFieldArray remove(index) is not refreshing formState.isValid
The formState.isValid continues in "false" when remove items from a filedArray.

1. Go to Sandbox
2. Click on "Append" and the Submit button will be disabled
3. "Delete" second item.
4. Submit button continues disabled !!

[Codesandbox](https://codesandbox.io/s/formstate-is-not-still-valid-after-removing-items-eevbe)

**Expected behavior**
The submit button should be enabled (formState.isValid == true)

**Additional context**
I think it's related to this old ticket https://github.com/react-hook-form/react-hook-form/issues/1097

bluebill1049:


src/useFieldArray.test.tsx
```diff
@@ -974,6 +974,65 @@ describe('useFieldArray', () => {
       expect(formState.isDirty).toBeFalsy();
     });
 
+    it('should update isValid formState when item removed', async () => {
```
bluebill1049, line 977: @keiya01 could you please take a look at this test? why it's failing? I have test it the code functional and ti's working 🤔 

https://codesandbox.io/s/formstate-is-not-still-valid-after-removing-items-by3e7?file=/src/App.js:1629-1636

keiya01: maybe, you should wrap append and remove with `await actComponent(async () => ...)`.

bluebill1049: Thanks @keiya01 i will do some debug in the code, see what's going on 👍

src/useFieldArray.test.tsx
```diff
@@ -974,6 +974,70 @@ describe('useFieldArray', () => {
...
+      let formState: any;
+      const Component = () => {
+        const { register, control, formState: tempFormState } = useForm({
+          mode: 'onChange',
```
keiya01, line 981: I see😅

keiya01 (approved):
LGTM :)

src/useFieldArray.test.tsx
bluebill1049: LOL my bad.

keiya01: @bluebill1049 
I think we should output error message for developer. what do you think?
bundle size is no problem in production if we use `if (process.env.NODE_ENV !== 'production') { ... }’ statement.

bluebill1049: yes, but this one is hard to leave a console for user, unless you have a good suggestion?

keiya01: we can use ‘Proxy’ API that is used by ‘formState’ variables.
I think we can check if ‘formState.isValid’ is invoked with ‘onChange’ mode or not.

bluebill1049: that's pretty smart! we can give it a try. 👍

keiya01: I will try to implement it :)

PR #2432, opened 2020-07-31
add utility functions test

kotarella1110:


src/utils/set.test.ts
```diff
@@ -0,0 +1,60 @@
+import set from './set';
+
+describe('set', () => {
+  it('should set the correct values', () => {
+    const test1 = { a: [{ b: { c: 3 } }] };
+    expect(set(test1, 'a[0].b.c', 4)).toEqual(4);
+    expect(test1.a[0].b.c).toEqual(4);
+
+    const test2 = { foo: { bar: 'baz' } };
+    expect(set(test2, 'foo.arr[0]', 3)).toEqual(3);
```
kotarella1110, line 10: @bluebill1049 Is this the expected return value?
I was expecting the following return value.

```ts
expect(set(test2, 'foo.arr[0]', 3)).toEqual({
  foo: {
    bar: 'baz',
    arr: [3],
  },
});
```

bluebill1049:
```
set(test2, 'foo.arr[0]', 3)
expect(test2).toEqual(3);
```
This works?

kotarella1110: this works

```ts
const test2 = { foo: { bar: 'baz' } };
const returnValue = set(test2, 'foo.arr[0]', 3);
expect(test2).toEqual(3); // <= expected?
expect(test2).toEqual({
  foo: {
    bar: 'baz',
    arr: [3],
  },
}); // correct!
```

bluebill1049: `expect(returnValue).toEqual(3);` expected because this method mutates the original object.

bluebill1049 (approved):
awesome! thanks @kotarella1110  for refining this lib with more tests. 🙏

PR #2447, opened 2020-08-01
fixed logic to output error message

keiya01:
I fixed logic to output error message when registered field name is flat array. Because this logic is not supported nested array.


bluebill1049 (approved):
LGTM! Thank you @keiya01

PR #2456, opened 2020-08-02
Update triggerFieldValidation.tsx example to get rid of `triggerValidation`

kalmanh:
Trigger validation is not a "thing" anymore. It disappeared from the API and all this example is currently doing is confusing :confused: the daylights out of people.


bluebill1049:
Thank you @kalmanh 🙏

PR #2458, opened 2020-08-02
Provide isSubmitting for Resolver

tedliang:
Provide isSubmitting=true for Resolver when it's triggered by handleSubmit.

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -85,6 +85,7 @@ export type Resolver<
   values: TFieldValues,
   context?: TContext,
   validateAllFieldCriteria?: boolean,
+  isSubmitting?: boolean,
 ) => Promise<ResolverResult<TFieldValues>> | ResolverResult<TFieldValues>;
 
 export type UseFormOptions<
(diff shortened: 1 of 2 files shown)


bluebill1049:
can you try to inject into the `context` object?

tedliang:
Thanks for your instant response! 
Yes, we can. Would be good the library can provide it. You can see the implement is very straight forward.

tedliang:
According to the [handleSubmit documentation](https://react-hook-form.com/api#handleSubmit)
> Note: You can pass an async function for asynchronous validation. eg:
> handleSubmit(async (data) => await fetchAPI(data))

But the errors returned from fetchAPI are not being handled yet. 

When the form is submitted to the backend, errors will be returned if the data is invalid. 

We did a workaround with the resolver to handle the errors for this scenarios. That's why isSubmitting is required in resolver.

It will be ideal if the handleSubmit api can handle errors (similar to resolver).

tedliang:
Please review the proper solution: https://github.com/react-hook-form/react-hook-form/pull/2515

PR #2461, opened 2020-08-03
Feature/better error message

bluebill1049:
DX! thanks @kotarella1110 to fix our build which stripe all the none production code.

### DIFF
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -95,7 +95,7 @@ const Controller = <
   const registerField = React.useCallback(() => {
     if (process.env.NODE_ENV !== 'production' && !name) {
       // eslint-disable-next-line no-console
-      return console.warn('📋 Field is missing name attribute.');
+      return console.warn('📋 Field is missing name prop.');
     }
 
     if (fieldsRef.current[name]) {
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -74,6 +74,11 @@ const Controller = <
   };
 
   const registerField = React.useCallback(() => {
+    if (process.env.NODE_ENV !== 'production' && !name) {
+      // eslint-disable-next-line no-console
+      return console.warn('Field is missing name attribute.');
+    }
+
     if (fieldsRef.current[name]) {
       fieldsRef.current[name] = {
         ref: fieldsRef.current[name]!.ref,
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -59,20 +59,17 @@ const Controller = <
 
   if (process.env.NODE_ENV !== 'production') {
     if (isUndefined(value)) {
-      // eslint-disable-next-line no-console
       console.warn(
-        '📋 Controller defaultValue or useForm defaultValues is missing.',
+        '📋 Controller `defaultValue` or useForm `defaultValues` is missing.',
       );
     }
 
     if (as && render) {
-      // eslint-disable-next-line no-console
-      console.warn('📋 Should use either as or render prop.');
+      console.warn('📋 Should use either `as` or `render` prop.');
     }
 
     if (!control) {
-      // eslint-disable-next-line no-console
-      console.warn('📋 Controller is missing control prop.');
+      console.warn('📋 Controller is missing `control` prop.');
     }
   }
 
@@ -94,8 +91,7 @@ const Controller = <
 
   const registerField = React.useCallback(() => {
     if (process.env.NODE_ENV !== 'production' && !name) {
-      // eslint-disable-next-line no-console
-      return console.warn('📋 Field is missing name prop.');
+      return console.warn('📋 Field is missing `name` prop.');
     }
 
     if (fieldsRef.current[name]) {
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -57,6 +57,25 @@ const Controller = <
   const onFocusRef = React.useRef(onFocus);
   const isSubmitted = isSubmittedRef.current;
 
+  if (process.env.NODE_ENV !== 'production') {
+    if (isUndefined(value)) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        '📋 Controller defaultValue or useForm defaultValues is missing.',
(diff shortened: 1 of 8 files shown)


bluebill1049:
@keiya01 i think we have a test case which fails sometimes :(
Screen Shot 2020-08-03 at 5 28 57 pm

keiya01:
Maybe, this error is occurred by cache in jest. You can resolve this error by executing `yarn test --clearCache`.

src/controller.tsx
```diff
@@ -57,6 +57,22 @@ const Controller = <
...
+      console.warn(
+        '📋 Controller `defaultValue` or useForm `defaultValues` is missing.',
+      );
+    }
+
+    if (as && render) {
+      console.warn('📋 Should use either `as` or `render` prop.');
+    }
+
+    if (!control) {
```
keiya01, line 71:
```suggestion
    if (!control && !methods.control) {
```

src/useFieldArray.ts
```diff
@@ -102,6 +102,16 @@ export const useFieldArray = <
 
   allFields.current = fields;
 
+  if (process.env.NODE_ENV !== 'production') {
+    if (!control) {
```
keiya01, line 106:
```suggestion
    if (!control && methods.control) {
```

keiya01 (changes requested):
I reviewed this PR. And I sent commit. :)

bluebill1049:
awesome! thanks @keiya01 good pick up, miss the context 🙏

keiya01:
I fixed some code.

keiya01 (dismissed):
LGTM 🧙‍♂️

bluebill1049:
thanks @keiya01

keiya01 (approved):

gregfenton:
I have code that was happily running in dev with 6.1.2.  Upgrading to 6.7.2 I now get these warnings about my Controller defaultValues being 'undefined'.  I set my defaultValues to the previously saved value (if any) for a given form.  If the form hasn't been saved previously, then the defaultValue is actually `undefined`.

Is there a doc update (or just hint me the answer here  ;) ) as to how to handle this situation?  What should defaultValue be set to when there isn't actually a default value?

Thanks in advance!

bluebill1049:
> I have code that was happily running in dev with 6.1.2. Upgrading to 6.7.2 I now get these warnings about my Controller defaultValues being 'undefined'. I set my defaultValues to the previously saved value (if any) for a given form. If the form hasn't been saved previously, then the defaultValue is actually `undefined`.
> 
> Is there a doc update (or just hint me the answer here ;) ) as to how to handle this situation? What should defaultValue be set to when there isn't actually a default value?
> 
> Thanks in advance!

an empty string will do, undefined may lead to issue with Controller.

gregfenton:
I am using `<Controller>` throughout a complex, nested, and dynamically generated form.

Will the above approach work for fields that will otherwise contain non-string values?  In particular, the field that 6.7.2 is complaining about should either be empty (currently `undefined`) or will contain an object `{comment: theTextValue, photos: [...theUrlArray]}`.

So will setting that field to empty string (`''`) be okay for "no value" and set the object when there is a comment and/or values in the URL array?  Is there any other thing I should be thinking about when it comes to setting, unsetting and resetting fields within the form?

Thanks for the help (and your overall contributions!!)

bluebill1049:
> I am using `<Controller>` throughout a complex, nested, and dynamically generated form.
> 
> Will the above approach work for fields that will otherwise contain non-string values? In particular, the field that 6.7.2 is complaining about should either be empty (currently `undefined`) or will contain an object `{comment: theTextValue, photos: [...theUrlArray]}`.
> 
> So will setting that field to empty string (`''`) be okay for "no value" and set the object when there is a comment and/or values in the URL array? Is there any other thing I should be thinking about when it comes to setting, unsetting and resetting fields within the form?
> 
> Thanks for the help (and your overall contributions!!)

as long as you can supply a `defaultValue` at Controller or useForm `defaultValues`, it should be fine. I mean it's not the end of world if you don't supply `defaultValue`, but it's definitely better practice and less issue later on.

gregfenton:
I read the descriptions of `defaultValue` (Controller and useForm), then the links to React's controlled & uncontrolled components...maybe it is just me but none of it jumps out saying "here is how `defaultValue` impacts the behaviour of the component/form/submission/etc.".  IMO it is up to the reader to wade through the various pages and APIs and build out the picture in their own minds.  Or is there one key "here is how it works" doc I am failing to absorb?

Questions I still have:

- when do defaultValue(s) get used - initial load, submission (success/fail), form reset?
- does the `value` prop of a control/form actually get the `defaultValue`, or is defaultValue directly referenced when `value` is empty/null/undefined?

bluebill1049:
> I read the descriptions of `defaultValue` (Controller and useForm), then the links to React's controlled & uncontrolled components...maybe it is just me but none of it jumps out saying "here is how `defaultValue` impacts the behaviour of the component/form/submission/etc.". IMO it is up to the reader to wade through the various pages and APIs and build out the picture in their own minds. Or is there one key "here is how it works" doc I am failing to absorb?
> 
> Questions I still have:
> 
> * when do defaultValue(s) get used - initial load, submission (success/fail), form reset?
> * does the `value` prop of a control/form actually get the `defaultValue`, or is defaultValue directly referenced when `value` is empty/null/undefined?

1. like I suggested above follow the rule to provide `defaultValue` (none undefined) which why the warning is in the first place.
2. have a look at the source code, which may give you better idea: https://github.com/react-hook-form/react-hook-form/blob/master/src/controller.tsx#L57
3. `defaultValue` is for the initial input value, just like `<input defaultValue="test" />`
4. > maybe it is just me but none of it jumps out saying "here is how defaultValue impacts the behaviour of the component/form/submission/etc."
it's open-source, you can contribute and help too.

gregfenton:
Happy to do so and will.  Just wanted to be sure I wasn't missing something obvious.

PR #2480, opened 2020-08-05
improved controller test

keiya01:


bluebill1049 (approved):
This looks good! we can eventually remove that mock file! Thank you @keiya01 (sorry for the late review)

Issue #1614, opened 2020-05-11
handleSubmit possibility to catch when validation fails


gcourtemanche:
**Is your feature request related to a problem? Please describe.**
I'm using snackbars to show errors to the user and I need to create them one time when the validation fails. I can't rely on the `errors` object like when you show the error as text next to your field. I'm using `triggerValidation` at the moment, but it doesn't handle the focus on html elements.

**Describe the solution you'd like**
If it's possible that `handleSubmit` returns `Promise<boolean>` like with `triggerValidation`, it will solve my problem.

**Describe alternatives you've considered**
The best solution I have so far, but the focus is not 100% working and it's hacky.

`handleSubmit` override:
```
  const handleSubmit: FormContextValues<FormValues>['handleSubmit'] = (
    callback
  ) => async (event) => {
    event?.preventDefault();

    const isValid = await form.triggerValidation();

    if (isValid) {
	  const data = form.getValues();
      try {
        await callback(data);
      } catch (err) {
        // Log unexpected errors from your callback function
      }
    } else {
      Object.values(form.errors).forEach((error) => {
        const { message } = error as FieldError;
        errorSnackbar(message);
      });
    }
  };
```

In my form fields component:
```
  const { name } = props;
  const { errors } = useFormContext();

  const firstError = Object.keys(errors)[0];
  const errorsString = Object.keys(errors).toString();

    // Doesn't work when submitting 2 times with the same errors because errorsString doesn't change
  useEffect(() => {
    if (firstError === name) {
      elementRef.current.focus();
    }
  }, [firstError, errorsString]); 
```

**Additional context**

bluebill1049:
you can upvote on this issue: https://github.com/react-hook-form/react-hook-form/issues/1574

jpodpro:
has this been resolved? i cannot believe there would be no non-hack way to respond to an invalid form!

bluebill1049:
@jpodpro take back your shitty comment, nobody is making open source to listen bullshit like that.

jpodpro:
@bluebill1049 apologies man, didn't mean to come off as such. just very surprised with all the features that this isn't supported. i take this as a no.

bluebill1049:
> @bluebill1049 apologies man, didn't mean to come off as such. just very surprised with all the features that this isn't supported. i take this as a no.

[...]

Issue #1574, opened 2020-05-05
Callback when form is invalid


feclist:
**Is your feature request related to a problem? Please describe.**
We are running into this problem where we need to have some side effect triggered when a form is invalidated. Now we've tried getting to that point by two methods:
1. Writing a useEffect that listens for the `errors` object to change. This doesn't consistently work as only the first invalidation triggers this effect not any subsequent validations where the errors don't change.
2. using setError on the custom `validate` function in register. It seems that once all these functions have run (and some/all of them return false) all the errors set in validation are overridden by the internal error type of `validation`
EDIT: 3. I got it to work using the following hook below. By listening to the `errors` and `formState.submitCount` I know when a new submit happened. 
```tsx
const { register, handleSubmit, errors, setValue, formState } = useForm()

useEffect(() => {
  // Trigger callback when there's errors
  if (Object.keys(errors).length !== 0) {
    onInvalid()
  }
}, [JSON.stringify(errors), formState.submitCount])
```

**Describe the solution you'd like**
It would be great to have `handleSubmit` to take two functions one callback for a successful submit and a second argument for a validation error callback. Something that looks like: `handleSubmit(successCallback: function, invalidCallback: function)`. This would give more flexibility to the user when the user wants to rely on the internal validation, but does need to have a side effect triggered once validation fails. 

**Describe alternatives you've considered**
The three alternatives listed above is what we've come up with. The third one did get me the desired result, but even though this works it would still be better to have a direct callback instead of an inferred one.

**Additional context**
If anything remains unclear I can elaborate :)

bluebill1049:
Thanks a lot for you suggestion :) I will leave for my backlog and attach waiting up vote tag.

zelucena:
This should be very useful. I use a state to temporarily disable my submit buttons so the user can't trigger more than one request at a time. 
I got the function `useYupValidationResolver` from the docs at V5 but I also manually added the following to the catch segment:
`
if (typeof handleErrors === 'function') {
    handleErrors(errors);
}
`

bluebill1049:
@zelucena this is already shipped.

Asrover:
```
useEffect(() => {
     if (formState.isSubmitted && !formState.isValid) {
         // Error callback
     }
 }, [formState.submitCount])
```

PR #2515, opened 2020-08-07
Handle errors when handleSubmit callback return errors

tedliang:
According to the [handleSubmit documentation](https://react-hook-form.com/api#handleSubmit)

```
Note: You can pass an async function for asynchronous validation. eg:
handleSubmit(async (data) => await fetchAPI(data))
```

But the errors returned from fetchAPI are not being handled yet. 
This merge request will make it work.

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -59,10 +59,20 @@ export type ValidationMode = {
 
 export type Mode = keyof ValidationMode;
 
-export type SubmitHandler<TFieldValues extends FieldValues> = (
-  data: UnpackNestedValue<TFieldValues>,
+export type SubmitHandler<
+  TSubmitFieldValues extends FieldValues,
+  TFieldValues extends FieldValues
+> = (
+  data: UnpackNestedValue<TSubmitFieldValues>,
   event?: React.BaseSyntheticEvent,
-) => void | Promise<void>;
+) =>
+  | void
+  | SubmitError<TFieldValues>
+  | Promise<void | SubmitError<TFieldValues>>;
+
+export type SubmitError<TFieldValues extends FieldValues = FieldValues> = {
+  errors: FieldErrors<TFieldValues>;
+};
 
 export type ResolverSuccess<TFieldValues extends FieldValues = FieldValues> = {
   values: UnpackNestedValue<TFieldValues>;
@@ -365,7 +375,7 @@ export type UseFormMethods<TFieldValues extends FieldValues = FieldValues> = {
     names: TFieldName[],
   ): UnpackNestedValue<Pick<TFieldValues, TFieldName>>;
   handleSubmit: <TSubmitFieldValues extends FieldValues = TFieldValues>(
-    callback: SubmitHandler<TSubmitFieldValues>,
+    callback: SubmitHandler<TSubmitFieldValues, TFieldValues>,
   ) => (e?: React.BaseSyntheticEvent) => Promise<void>;
   control: Control<TFieldValues>;
 };
--- a/src/useForm.test.tsx
+++ b/src/useForm.test.tsx
@@ -1461,6 +1461,26 @@ describe('useForm', () => {
       expect(callback).toBeCalled();
     });
 
+    it('should set errors when callback return errors', async () => {
+      const { result } = renderHook(() => useForm<{ test: string }>());
+      const callback = jest.fn().mockReturnValue({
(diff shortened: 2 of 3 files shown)


bluebill1049:
Thanks @tedliang for your PR, let's have a quick discussion next time :) This issue is already working in progress: https://github.com/react-hook-form/react-hook-form/issues/1574 maybe you want to collaborate with @adhrinae ?

adhrinae:
@tedliang 's work is much better than I was doing.

@bluebill1049 Do you think #1614 is covered by this PR? I'm not sure about that.

bluebill1049:
> @tedliang 's work is much better than I was doing. So we can merge this PR :)
> 
> @bluebill1049 Do you think #1614 is covered by this PR? I'm not sure about that.

I think we should have the second argument is the callback for error.

adhrinae:
> > @tedliang 's work is much better than I was doing. So we can merge this PR :)
> > @bluebill1049 Do you think #1614 is covered by this PR? I'm not sure about that.
> 
> I think we should have the second argument is the callback for error.
> 
> ```
> handleSubmit(onSubmit, onError)
> ```

I agree. and somehow when I take a second look at this PR, I doubt that there will be many cases handling errors in the callback of `handleSubmit`.

Because `handleSubmit` already takes care of errors at first, then pass `data` to the callback arg.

bluebill1049:
@adhrinae we can always invoke the second callback when there is an error, and it's easier to documented the API as well, rather than hidden the logic inside the `handleSubmit`.

adhrinae:
Right. I'm just reviewing [the current implementation status](https://github.com/react-hook-form/react-hook-form/blob/31bf519f3bacc9baffe0fed90147413e706fc250/src/useForm.ts#L964-L985).

Then I need to work on invoking the second callback like `onError`.
I was going too slow solving this issue 🐢.

tedliang:
I'm confused. Do the user need to provide or call onError?

adhrinae:
@tedliang I think this PR has different concern from what @bluebill1049 and I think. I'm not saying you're wrong :)
#1574 is focused on 'when the validation failed at the time of form submission'.
However, you have a concern that after the validation success, such as there might be a chance to get an error from API server. Is it correct?

I'd like to ask you more about the usage of your implementation. I usually thought that if an asynchronous action returns error, it should be handled in `useEffect`. or, you can use `setError` in the callback.

and also to @bluebill1049, I don't have a context about why `handleSubmit` only returns `void` before. Is it because the users can use `setError` in the callback?

bluebill1049:
> @tedliang I think this PR has different concern from what @bluebill1049 and I think. I'm not saying you're wrong :)
> #1574 is focused on 'when the validation failed at the time of form submission'.
> However, you have a concern that after the validation success, such as there might be a chance to get an error from API server. Is it correct?
> 
> I'd like to ask you more about the usage of your implementation. I usually thought that if an asynchronous action returns error, it should be handled in `useEffect`. or, you can use `setError` in the callback.
> 
> and also to @bluebill1049, I don't have a context about why `handleSubmit` only returns `void` before. Is it because the users can use `setError` in the callback?

you are right @adhrinae :) 

@tedliang I do appreciate your PR as I stated above, but we should have this discussion ahead of time to make your contribution worth your effort, there are many things to be considered before adding a feature on top of it.

`handleSubmit` doesn't take any return value and I am not saying it's a bad idea, but it certainly increases the complexity and things for the user to learn. 

with your use case above, it would not hit the `onError` callback, `onError` callback is when form validation failed. what you want is just `setError` inside onSubmit.

tedliang:
Thanks. Make sense to me now. 

Yep. I'm totally fine with the setError approach for solving my problem. 

Would be good that it can handle multiple errors in one call.😇

bluebill1049:
> Thanks. Make sense to me now.
> 
> Yep. I'm totally fine with the setError approach for solving my problem.
> 
> Would be good that it can handle multiple errors in one call.😇

Let's keep this PR @tedliang I think there are some good values in here we can contribute on top of it once we get to this feature: https://github.com/react-hook-form/react-hook-form/issues/1614

again thanks very much for your contribution and understanding. ❤️

bluebill1049:
> Would be good that it can handle multiple errors in one call.

yes, we didn't do `setErrors` and `setValues`, because you can probably easily map it instead another API to keep, keep hook form API small :)

PR #2528, opened 2020-08-08
Cast checkbox value to string explicitly

Toleckk:
Fix rare error 'value.includes is not a function', that appears on controlled form with checkboxes array resetting, if fields order was changed.

**Steps to reproduce the behavior:**
1. Go to https://codesandbox.io/s/clever-pond-zjequ
2. Add few fields
3. Click on any checkbox but last

**Expected behavior**
Value of checkbox should change after click

image

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -227,7 +227,7 @@ export function useForm<
               ({ ref: checkboxRef }) =>
                 (checkboxRef.checked = (value as string | boolean)
                   .toString()
-                  .includes(checkboxRef.value))
+                  .includes(checkboxRef.value)),
             )
           : (options[0].ref.checked = !!value);
       } else {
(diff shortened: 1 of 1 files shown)


src/useForm.ts
```diff
@@ -225,9 +225,9 @@ export function useForm<
         options.length > 1
           ? options.forEach(
               ({ ref: checkboxRef }) =>
-                (checkboxRef.checked = (value as string).includes(
-                  checkboxRef.value,
-                )),
+                (checkboxRef.checked = (value as string | boolean)
```
bluebill1049, line 228: `String(value)` is perhaps shorter? (see if can save few bytes)

bluebill1049:
Thanks @Toleckk for the PR 👍

PR #2532, opened 2020-08-09
Invoke error handler when validation failed during form submission

### LINKED ISSUE #1574: Callback when form is invalid

adhrinae:
This PR resolves #1574, #1614.

Just made `handleSubmit` takes `onError` callback as a second argument.
Personally I wanted to refactor the parameter signature of `handleSubmit` something like `({ onValid, onInvalid })`, However, that kind of change breaks backward compatibility.
I tried to overload the `handleSubmit` function to take different arguments, but It didn't work.
image

I'm not sure whether I wrote tests correctly, Please feel free to give me feedback.


adhrinae:
🤔 Why the other unit test get failed...? Will fix it later.

app/src/app.tsx
```diff
@@ -90,6 +91,11 @@ const App: React.FC = () => {
       />
       <Route path="/autoUnregister" exact component={AutoUnregister} />
       <Route path="/useWatch" exact component={useWatch} />
+      <Route
+        path="/form-submit-with-on-error"
+        exact
+        component={formSubmitWithOnError}
```
pmaier983, line 97: #nitpicking (feel free to ignore)

formSubmitWithHandleOnError may be a better name for the component and path. A bit more descriptive.

src/useForm.ts
```diff
@@ -930,6 +931,7 @@ export function useForm<
   const handleSubmit = React.useCallback(
     <TSubmitFieldValues extends FieldValues = TFieldValues>(
       callback: SubmitHandler<TSubmitFieldValues>,
+      onError?: SubmitErrorHandler<TFieldValues>,
```
kotarella1110, line 934: You should use `TSubmitFieldValues` instead of `TFieldValues`.

```ts
SubmitErrorHandler<TSubmitFieldValues>
```

app/src/formSubmitWithOnError.tsx
```diff
@@ -0,0 +1,55 @@
...
+  lastName: yup.string().max(5).required(),
+});
+
+const formSubmitWithOnError: React.FC = () => {
+  const [show, setShow] = React.useState(true);
+  const [invalidMessage, setInvalidMessage] = React.useState('');
+  const { register, handleSubmit, formState } = useForm<{
+    firstName: string;
+    lastName: string;
+    select: string;
```
pmaier983, line 21: The FormTypes (select, radio, checkbox) don't seem to be used? This may be part of the error you're trying to create, but if not may want to simplify this (have the error just be the lack of a lastName.

src/useForm.ts
```diff
@@ -998,6 +1000,9 @@ export function useForm<
             ...errorsRef.current,
             ...fieldErrors,
           };
+          if (onError) {
+            await onError(fieldErrors);
```
kotarella1110, line 1004: I think you need to pass the second argument event here.

```
await onError(fieldErrors, e);
```

src/types/form.ts
```diff
@@ -64,6 +64,11 @@ export type SubmitHandler<TFieldValues extends FieldValues> = (
   event?: React.BaseSyntheticEvent,
 ) => void | Promise<void>;
 
+export type SubmitErrorHandler<TFieldValues extends FieldValues> = (
```
kotarella1110, line 67: I think `SubmitErrorHandler` will be the type to expose to users. Which is better naming?
`SubmitErrorHandler` vs `ErrorHandler`

cypress/integration/basic.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference types="cypress" />
```
pmaier983, line 1: Is this needed? (just curious, cuz I don't see it in any other cypress files)

app/src/formSubmitWithOnError.tsx
```diff
@@ -0,0 +1,55 @@
...
+    >
+      <input name="location" ref={register} placeholder="location" />
+      {show && (
+        <input name="firstName" ref={register} placeholder="firstName" />
+      )}
+      <input name="lastName" ref={register} placeholder="lastName" />
+      <input name="age" ref={register} placeholder="age" />
+      <div id="isValid">{JSON.stringify(formState.isValid)}</div>
+      <div id="renderCount">{renderCounter}</div>
+      <div id="invalid-message">{invalidMessage}</div>
```
pmaier983, line 49: I think it might be best to leave the excess divs out of the PR. Similar to the excess Types, it might be best to keep this test case as simple as possible.

pmaier983 (dismissed):
Awesome Addition! Most of my comments are just #nitpicking. Take a look and tell me what you think 😄

kotarella1110:
> 🤔 Why the other unit test gets failed...? Will fix it later.

The unit test sometimes fails .
You may pass by rerunning the unit test.
 I think @keiya01  is investigating the cause of this 👍

cypress/integration/basic.ts
JeromeDeLeon: I was investigating it before. As you can see, the reason why it is here is because `cy` is not define hence doesn't have proper `types` and to resolve this, either put that line per file or look for a solution that could allow us to have cypress types just by putting it on cy config or so and I haven't gotten that far yet with my investigation before.

That's my only opinion when I used it before but removed it because of its repetition.

pmaier983: I looked into this: seems we can add ``` "types": ["cypress"]``` to the tsconfig in app to solve the problem (https://docs.cypress.io/guides/tooling/intelligent-code-completion.html#Reference-type-declarations-via-tsconfig)

src/useForm.ts
adhrinae: Right, I forgot to pass the event object. 😅

src/types/form.ts
adhrinae: As I thought this is not a generic error handler, so I picked `SubmitErrorHandler`.

src/useForm.ts
adhrinae: But `TSubmitFieldValues` throws type mismatch. because it doesn’t match with `errors`’s type.

cypress/integration/basic.ts
adhrinae: I was going to create another PR for handling cypress type errors. but will try to fix it in this PR.

app/src/formSubmitWithOnError.tsx
adhrinae: I’ll make it simpler. I just brought codes from `isValid.tsx`.

app/src/app.tsx
adhrinae: You're right. At first my intention was using path like `/basic/onError`. However this matches `basic/:mode` path so It renders an unnecessary page.

Perhaps I need to think a proper name for this page. Any suggestions? 🙂

app/src/formSubmitWithOnError.tsx
adhrinae: Sorry, but I don’t understand what ”leave the excess divs out” means. Does it mean that it is better to remove `#invalid-message` div?

JeromeDeLeon (commented):
LOL I forgot to submit this pending review last night.

src/useForm.ts
```diff
@@ -930,6 +931,7 @@ export function useForm<
   const handleSubmit = React.useCallback(
     <TSubmitFieldValues extends FieldValues = TFieldValues>(
       callback: SubmitHandler<TSubmitFieldValues>,
```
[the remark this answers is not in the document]
adhrinae: Good point. I prefer `onValid`, `onInvalid` though.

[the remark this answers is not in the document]
JeromeDeLeon: Its fine by me as long as we keep the callback consistence. 😁

bluebill1049:
Thanks very much for everyone's review! 🎉❤️and @adhrinae's work. 🥃

src/useForm.test.tsx
```diff
@@ -1750,6 +1750,49 @@ describe('useForm', () => {
     });
   });
 
+  describe('handleSubmit with onError callback', () => {
```
keiya01, line 1753: That's great 🤩

src/useForm.ts
adhrinae: Looks like `TFieldValues` is valid type. At first the `fieldErrors` variable declared as `FieldErrors<TFieldValue>` [at this line](https://github.com/react-hook-form/react-hook-form/blob/master/src/useForm.ts#L938).

[Also the `focusOnErrorField` function takes `FieldErrors<TFieldValues>` type as a second argument.](https://github.com/react-hook-form/react-hook-form/blob/master/src/useForm.ts#L1002)

adhrinae:
## Key changes

- Fixed Cypress + Typescript typing + Made cypress test simpler
- Renamed `handleSubmit`'s argument names (`onValid, onInvalid`)

src/types/form.ts
kotarella1110: 👍

src/useForm.ts
kotarella1110: @adhrinae I thought that generic is necessary when input values and submit value are different as below.
https://github.com/react-hook-form/react-hook-form/blob/master/examples/nativeMultipleInput.tsx

kotarella1110 (approved):
LGTM!

src/useForm.ts
adhrinae: @kotarella1110 Hmm... interesting.
I just wondered about the generic of `onInvalid` function because I got the error below.
Should I just wrap the `fieldErrors` with `as FieldErrors<TSubmitFieldValues>` ?

image

kotarella1110: @adhrinae yes. but for now let's leave it as it is 👍

bluebill1049:
merging 🥳

bluebill1049 (approved):

keiya01 (approved):

bluebill1049:
oh wow @abihf u still didn't branch off master, lol consider you are in the team.

adhrinae:
I thought someone else should merge this PR and I should wait for #2377 [to be merged](https://github.com/react-hook-form/react-hook-form/pull/2377#issuecomment-671678153) 😅.

bluebill1049:
you can branch off from master, you don't' need to fork. :)

bluebill1049:
They all merged. we will release a nice minor update next week 🥳

adhrinae:
Ahh, I see. I just wanted to be safe for my first PR 🦺.
I don't have to fork the repo from next time.

Issue #2581, opened 2020-08-12
[Request for Feature] Support integer values for nested form field keys.


RookY2K:
**Describe the bug**
dirtyFields and isDirty are inconsistent with expected values. dirtyFields will zero out after blur and change events when there are still dirty fields. This is for version 6.3.3. This seems to only happen with nested fields with number properties. Nested Fields with string properties work as expected.

**To Reproduce**

1. Go to codesandbox linked below
1. Enter values into Left <span>#</span>1, Left <span>#</span>2
1. Validate that IsDirty === true
1. Delete value from Left <span>#</span>2

Additional repros are in codesandbox

**Codesandbox link (Required)**
https://codesandbox.io/s/react-hook-form-dirty-repro-tedbi

**Expected behavior**
IsDirty should be true
Left DirtyFields should equal ['left.1']

**Desktop (please complete the following information):**

- OS: MacOS (10.15.4)
- Browser [Chrome, Firefox]
- Version [Chrome: 84.0, FF: 79.0]

RookY2K:
I've tracked the problem down to this snippet:
```javascript
// src/utils/unset.ts - L45
if (
    currentPathsLength === index &&
    ((isObject(objectRef) && isEmptyObject(objectRef)) ||
      (isArray(objectRef) &&
        !objectRef.filter((data) => isObject(data) && !isEmptyObject(data))
          .length))
) {
    previousObjRef ? delete previousObjRef[item] : delete object[item];
}
```

objectRef is an array of boolean values representing `isDirty`. So it fails the `isObject` filter and results in the entire top level key from object... In the reproduction, that would be `left`. Looking to see if I can grok the purpose of the if statement. The naive fix is to verify that data isn't a boolean before deleting the key. But I don't have enough context into this project to be sure there are no unintended consequences of that check.

bluebill1049:
Thanks, heaps for the issue report, investigation, and solution!

bluebill1049:
Looking a quick look at your example, i don't think we support integer as `key` name.

bluebill1049:
<img width="931" alt="Screen Shot 2020-08-13 at 9 46 20 am" src="https://user-images.githubusercontent.com/10513364/90078956-d8ed4180-dd49-11ea-926b-2d4d01d82bf7.png">
actually we do.

bluebill1049:
I will try to reproduce the problem with correct `name` before assigning a bug as this issue 👍

RookY2K:
[...]

PR #2585, opened 2020-08-12
Fix inappropriate parent key removal from dirtyFields for nested fields with number keys.

RookY2K:
Fix for #2581 

Parent keys were being inappropriately deleted from dirtyFields if the array still contained boolean values.


bluebill1049 (approved):
LGTM.

PR #2616, opened 2020-08-15
Update Yarn Lock and improve cypress tests

pmaier983:
Bulky PR, I'll break it down for you: 

- This bumps the app to the latest version of as many deps as possible (This makes sure we are on top of bugs asap)
- Added the plugin:cypress/recommended to make sure we are doing things correctly
- Fixed some `prop={"why brackets?"}` -> `prop="why brackets?"` 
- Removed any object checks using cypress.contains("{}"). Replaced them with expect().deepEqualityCheck(). This significantly improves the error message we receive, while also making it less likely to have diffs like [this](https://github.com/react-hook-form/react-hook-form/pull/2600#discussion_r470920922) one

Important Note:
After bumping the app to the latest version of all its deps, cypress had some expectations change. I'm not really sure why. Most of the changes were good (decreased re-renders), but some of the changes were odd... take a look and tell me what you think?

I tried for a while to get cypress commands working, but was unable to within my timebox. If anyone has experience on this please give it a try. Cypress commands are super helpfull!


bluebill1049:
This is awesome! thanks @pmaier983

cypress/integration/controller.ts
```diff
@@ -50,30 +50,30 @@ describe('controller basic form validation', () => {
...
 
-    cy.get('p').should('have.length', 0);
-    cy.get('#renderCount').contains('9');
+    cy.get('p').should('have.length', 1);
+    cy.get('#renderCount').contains('8');
   });
 
-  it('should validate the form with onChange mode and reset the form', () => {
-    cy.visit('http://localhost:3000/controller/onChange');
+  // it('should validate the form with onChange mode and reset the form', () => {
```
bluebill1049, line 57: This is disabled because it's not working?

cypress/integration/manualRegisterForm.ts
```diff
@@ -1,7 +1,6 @@
 describe('manual register form validation', () => {
   it('should validate the form', () => {
     cy.visit('http://localhost:3000/manual-register-form');
-    cy.wait(200);
```
bluebill1049, line 4: nice!~

bluebill1049 (commented):
looks great!

cypress/integration/controller.ts
pmaier983: Short answer yes. 
Long answer is I couldn't understand what the test was supposed to be for, and most of the .contains('XXX Error') were failing. If you explain why clicking the input's is supposed to fail then i can fix it! If it would be easier to do yourself thats fine too, I'll look at the diff when you're done.

bluebill1049: The test was just basically testing Controller's 'onChange'  does trigger errors.

PR #2630, opened 2020-08-17
Resolver examples

fea17e86:
Adds two custom resolver examples:

1. How to use a custom resolver that is created in an asynchronous process.
2. How create a custom resolver based on a Open API 3 documentation


bluebill1049:
Thanks a lot for the PR @fea17e86 perhaps we can have one single example: https://codesandbox.io/s/react-hook-form-customresoliver-v6-h4ti6

call it: **custom resolver**

PR #2634, opened 2020-08-18
improve errors types to enable auto-complete

kotarella1110:
Improved `type` and `types` field in errors type.
As shown below, auto-complete is enabled and DX is improved!!

スクリーンショット 2020-08-18 12 22 28
スクリーンショット 2020-08-18 12 20 14
スクリーンショット 2020-08-18 12 40 54

Of course, as before, we can use keys other than completion suggestions.

スクリーンショット 2020-08-18 12 24 47
スクリーンショット 2020-08-18 12 41 08

### DIFF
--- a/src/types/utils.ts
+++ b/src/types/utils.ts
@@ -14,6 +14,10 @@ export type LiteralToPrimitive<T extends any> = T extends string
   ? boolean
   : T;
 
+export type LiteralUnion<T extends U, U extends Primitive> =
+  | T
+  | (U & { _?: never });
+
 export type Assign<T extends object, U extends object> = T & Omit<U, keyof T>;
 
 export type DeepPartial<T> = {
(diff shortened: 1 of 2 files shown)


src/types/utils.ts
```diff
@@ -14,6 +14,10 @@ export type LiteralToPrimitive<T extends any> = T extends string
   ? boolean
   : T;
 
+export type LiteralUnion<T extends U, U extends Primitive> =
+  | T
+  | (U & { _?: never });
```
bluebill1049, line 19: could you explain what this does?

bluebill1049 (approved):
Really impressive stuff for better DX.

src/types/utils.ts
kotarella1110: @bluebill1049 please see here: https://github.com/microsoft/TypeScript/issues/29729

kotarella1110: string literal inherits string, but it does not inherit `string & {}`, so string literal and `string & {}` are treated as distinguishable types.

bluebill1049: nice! thanks for the info @kotarella1110

PR #2635, opened 2020-08-18
improve NestedValue type to support class type

### LINKED ISSUE #2466: Custom types/classes passed into useForm get mangled
**Describe the bug**
Following advice from https://github.com/react-hook-form/react-hook-form/issues/1414 I tried using a controller to parse a string into a custom type. Unfortunately it seems that when I pass that type into `useForm` and later try to read the values in `handleSubmit`, the values are not of the same type. In the Codesandbox bellow I created a silly `Price` class with a custom getter to illustrate the problem. The getter is missing in the value passed to the `handleSubmit` callback.

https://codesandbox.io/s/react-hook-form-useform-template-liv7m?file=/src/index.tsx:1093-1134

Wait a few seconds until `@types/react` loads and check the error on line 55. Codesandbox transpiles the code anyway but TypeScript with `strict: true` won't.

**Expected behavior**
The type of the value passed to the `handleSubmit` callback should be the same type as the one passed to `useForm`

kotarella1110:
close #2466

CSB: https://codesandbox.io/s/react-hook-form-useform-template-forked-65zmz

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -26,7 +26,11 @@ export type FieldValue<
   TFieldValues extends FieldValues
 > = TFieldValues[InternalFieldName<TFieldValues>];
 
-export type NestedValue<TValue extends any[] | object = any[] | object> = {
+export type NestedValue<
+  TValue extends any[] | Record<string, unknown> =
+    | any[]
+    | Record<string, unknown>
+> = {
   [$NestedValue]: never;
 } & TValue;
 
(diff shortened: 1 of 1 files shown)


bluebill1049 (approved):
LGTM.

amcsi:
@kotarella1110 I believe this change made is so it's not possible to wrap `Map<>` in `NestedValue` anymore.

PR #2670, opened 2020-08-21
improved mutation observer

bluebill1049:
- improve to use single MutationObserver


maxsbelt (dismissed):
LG2M

pmaier983 (commented):
LGTM! Just wondering why one test is gone, and if we can simplify an if statement!

pmaier983 (approved):
Looks great!

PR #2699, opened 2020-08-24
Add schema in cypress config

devrnt:
Add `$schema` in `cypress.json`. This adds some juicy autocompletion for the cypress config file.


bluebill1049:
That's awesome. Thanks @devrnt.

PR #2702, opened 2020-08-25
improve rollup config to support IE11

### LINKED ISSUE #2583: TypeError: Incompatible receiver, Set required
**Describe the bug**
Starting from version `6.3.1` until the current version `6.3.3` I'm not able to use RHF in production. I use parcel-bundler to compile my project and when I start it it'll throw with the message `TypeError: Incompatible receiver, Set required`. It throws from the line https://github.com/react-hook-form/react-hook-form/blob/752d7efce18d6b2e26cbec7c9a1c24aa1ae5eb56/src/useForm.ts#L95

**To Reproduce**
Reproduction is real hard since I don't know what differs from the development runtime where it works as expected. But I use this alias to support IE11 out of the box:
```
  "alias": {
    "react-hook-form": "react-hook-form/dist/index.ie11.production.min.js"
  },
```

- OS: Windows 10 Version 10.0.20185 Build 20185
- Browser: Chromium Edge Version 86.0.605.0

**Additional context**
Quick google says maybe this helps https://github.com/zloirock/core-js/issues/514#issuecomment-561118374

kotarella1110:
The problem seems to have caused after transpiling with Babel.
Probably I think it's enough to transpile with es5 to support IE11.

close #2583

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "@babel/plugin-transform-runtime": "^7.10.4",
     "@changesets/changelog-github": "^0.2.6",
     "@changesets/cli": "^2.9.2",
-    "@rollup/plugin-babel": "^5.1.0",
     "@rollup/plugin-commonjs": "^13.0.0",
     "@rollup/plugin-node-resolve": "^8.4.0",
     "@rollup/plugin-replace": "^2.3.3",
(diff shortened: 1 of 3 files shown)


kotarella1110:
@bluebill1049 Let's release the beta version and see if the issue is resolved.

PR #2780, opened 2020-08-31
Removes redundant dependencies in append and remove of useFieldArray

### LINKED ISSUE #2773: Methods of use useFieldArray triggers hooks
**Describe the bug**

*Issue since version 6.5.3, version 6.5.2 works as expected (for `append`).*

I need to call `append` in `useEffect` hook, but unfortunately it's not working as I would expect.
To use `append` in hook I need to pass it as dependency and it causes problem. If I am not adding it to dependency list all works as expect, but then I am getting warning in console, that it should be a dependency.

I can have a workaround for that easily with additional state value, but I consider it as a bug, as `setValue` from the form works as expected - see below.

**To Reproduce**
See codesandbox below

https://codesandbox.io/s/react-hook-form-usefieldarray-forked-wrmyj?file=/src/index.js:424-438

Click one of "append", ~"prepend"~ or "delete" and check console output.

**Expected behavior**
All hooks are triggered only once - on loading, but should not be called when field is appended/prepended/removed.

I've checked also how it works with `setValue` from the form:
https://codesandbox.io/s/react-hook-form-v6-setvalue-forked-inyki?file=/src/index.jsx
and here all is fine - when we click "Set First Name Value", which triggers `setValue` the hook is not called, as I would expect.

- OS: macOS
- Browser Safari/Chrome
- Version latest stable

### LINKED ISSUE #2778: Fields inside useFieldArray are marked as dirty when only touched!
**Describe the bug**
The input text fields of a field array are getting marked as dirty when touched. This seems to be happening when providing a defaultValues object to useForm.

It also seems to only be happening when having a field array nested inside an object, but I'm a 100% sure as I need to test it a bit more!

https://codesandbox.io/s/react-hook-form-field-array-dirty-s9xg5?file=/src/App.js

**To Reproduce**

1. Go to the CodeSandbox provided above
2. click on add
3. press Tab to switch to the second field
4. first field is marked dirty

**Expected behavior**
Fields should not be marked as dirty.

- OS: macOs
- Browser chrome
- Version 85

**Additional context**
The issue is resolved by removing the defaultValues from the useForm parameters!

michalbundyra:
Fixes #2773

In `append` method `fillEmptyArray` cares just about the length of the array, so the `fields` dep was not needed there at all.

In `remove` dependency `fields` was not used.

All tests are still green. Probably we should add some test cases, so these dependencies are not re-added to the methods in `useFieldArray`.

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -247,7 +247,7 @@ export const useFieldArray = <
         ...get(
           dirtyFields,
           name,
-          fillEmptyArray(new Array(allFields.current.length)),
+          fillEmptyArray(allFields.current),
         ),
         ...filterBooleanArray(value),
       ]);
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -244,11 +244,7 @@ export const useFieldArray = <
       readFormStateRef.current.isDirty
     ) {
       set(dirtyFields, name, [
-        ...get(
-          dirtyFields,
-          name,
-          fillEmptyArray(allFields.current),
-        ),
+        ...get(dirtyFields, name, fillEmptyArray(allFields.current)),
         ...filterBooleanArray(value),
       ]);
       updateFormState({
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -244,7 +244,11 @@ export const useFieldArray = <
       readFormStateRef.current.isDirty
     ) {
       set(dirtyFields, name, [
-        ...get(dirtyFields, name, fillEmptyArray(fields)),
+        ...get(
+          dirtyFields,
+          name,
+          fillEmptyArray(new Array(allFields.current.length)),
+        ),
         ...filterBooleanArray(value),
       ]);
       updateFormState({
@@ -426,8 +430,8 @@ export const useFieldArray = <
     swap: React.useCallback(swap, [name, errors]),
     move: React.useCallback(move, [name, errors]),
     prepend: React.useCallback(prepend, [name, errors]),
-    append: React.useCallback(append, [name, errors, fields]),
-    remove: React.useCallback(remove, [name, errors, fields]),
+    append: React.useCallback(append, [name, errors]),
+    remove: React.useCallback(remove, [name, errors]),
     insert: React.useCallback(insert, [name, errors]),
     fields,
   };


src/useFieldArray.ts
```diff
@@ -244,7 +244,11 @@ export const useFieldArray = <
       readFormStateRef.current.isDirty
     ) {
       set(dirtyFields, name, [
-        ...get(dirtyFields, name, fillEmptyArray(fields)),
+        ...get(
+          dirtyFields,
+          name,
+          fillEmptyArray(new Array(allFields.current.length)),
```
bluebill1049, line 250:
```suggestion
          fillEmptyArray(allFields.current),
```

hmm, honestly I am trying to move away with that ref object :(

bluebill1049: I am happy to get this PR over the line, but I can't guarantee future change by completely removing that ref object. 🙏

bluebill1049:
hmmm actually I am going to hold this PR, I think to remove that ref object is the correct approach.

src/useFieldArray.ts
michalbundyra: The thing that it is still used here anyway - line 256 (260). That works to eliminate `fields` dep.
If you have better solution I'd be happy to see it :)

michalbundyra:
@bluebill1049 

> hmmm actually I am going to hold this PR, I think to remove that ref object is the correct approach

Ok, I believe you will find a better way to eliminate this dependency, as it is not needed in `prepend` as well. Thanks.

src/useFieldArray.ts
bluebill1049, line 250:
```suggestion
          fillEmptyArray(allFields.current),
```

src/useFieldArray.ts
```diff
@@ -426,8 +430,8 @@ export const useFieldArray = <
     swap: React.useCallback(swap, [name, errors]),
     move: React.useCallback(move, [name, errors]),
     prepend: React.useCallback(prepend, [name, errors]),
-    append: React.useCallback(append, [name, errors, fields]),
-    remove: React.useCallback(remove, [name, errors, fields]),
+    append: React.useCallback(append, [name, errors]),
+    remove: React.useCallback(remove, [name, errors]),
```
bluebill1049, line 434: please fix remove as well with `ref`.

bluebill1049 (changes requested):
Thanks for the PR.

src/useFieldArray.ts
michalbundyra: It was not used, as you see all tests are still green. Unless I am missing something?

I am not sure also where `errors` dep is used in these functions? Maybe it can be removed as well?

bluebill1049: you are right, that was my bad!

bluebill1049:
dame lint failed, you may have to pull the latest and run `yarn lint`. thanks

src/useFieldArray.ts
michalbundyra: Do you want me to remove `errors` dep as well. then?

michalbundyra:
@bluebill1049 

> dame lint failed

Fixed! :)

src/useFieldArray.ts
bluebill1049: it's used in `batchStateUpdate`.

bluebill1049: I will clean this update to only have `name` as dep. you can leave that to me, will sort out tonight.

PR #2786, opened 2020-09-01
enhance Controller props type

kotarella1110:
enhanced Controller props type to caused a type error when passing both `as` and `render` props 👍 

スクリーンショット 2020-09-01 12 03 29

### DIFF
--- a/src/controller.test.tsx
+++ b/src/controller.test.tsx
@@ -405,6 +405,7 @@ describe('Controller', () => {
   it('should be null if as and render props are not given', () => {
     const Component = () => {
       const { control } = useForm();
+      // @ts-ignore
       return <Controller defaultValue="" name="test" control={control} />;
     };
 
@@ -655,6 +656,7 @@ describe('Controller', () => {
       const Component = () => {
         const { control } = useForm();
         return (
+          // @ts-ignore
           <Controller
             as={'input' as const}
             render={() => <input />}
@@ -684,6 +686,7 @@ describe('Controller', () => {
       const Component = () => {
         const { control } = useForm();
         return (
+          // @ts-ignore
           <Controller
             as={'input' as const}
             render={() => <input />}
@@ -725,6 +728,7 @@ describe('Controller', () => {
           <form>
             {fields.map(({ id }, index) => {
               return (
+                // @ts-ignore
                 <Controller
                   name={`test[${index}].data`}
                   control={control}
@@ -767,6 +771,7 @@ describe('Controller', () => {
           <form>
             {fields.map(({ id }, index) => {
               return (
+                // @ts-ignore
                 <Controller
                   name={`test[${index}].data`}
                   control={control}
--- a/src/controller.tsx
+++ b/src/controller.tsx
@@ -181,7 +181,7 @@ const Controller = <
   return as
     ? React.isValidElement(as)
       ? React.cloneElement(as, props)
-      : React.createElement(as as string, props)
+      : React.createElement(as as string, props as any)
     : render
     ? render({
         onChange,
--- a/src/types/props.ts
+++ b/src/types/props.ts
@@ -34,19 +34,26 @@ export type ControllerProps<
     | 'textarea',
   TControl extends Control = Control
 > = Assign<
-  {
+  (
+    | {
+        as: TAs;
+        render?: undefined;
+      }
+    | {
+        as?: undefined;
+        render: (data: {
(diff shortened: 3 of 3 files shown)


bluebill1049:
oh nice!

bluebill1049 (approved):
Awesome thanks @kotarella1110

xiaoyu-tamu:
@bluebill1049 
@kotarella1110 

After this enhancement, typescript complain `<Controller name="firstName" />`, it  expect to pass either `as` or `render`. 

But read the code, the implementation does not set this constraint. 

see https://github.com/react-hook-form/react-hook-form/blob/master/src/controller.tsx#L182-L193


`<Controller name="firstName" />` is useful when field value is needed in the final form value but set programmatically,  especially for field array 

```
      {attributesField.fields.map((field, index) => {
        const arrayName = `${name}[${index}]`;
        return (
          <div>
            <Controller name={`${arrayName}.firstName`} required defaultValue={field.attribute} />
            <SomeOtherField required name={`${arrayName}.attributeValue`} />
          </div>
        );
      })}
```

I think this is a bug b/c the PR description said `enhanced Controller props type to caused a type error when **passing both as and render props**`

bluebill1049:
thanks, @xiaoyu-tamu oh, I never know users actually do that... I have to say pretty smart.

xiaoyu-tamu:
@bluebill1049 Do you think this is a bug? I can fix it if u want

bluebill1049:
@xiaoyu-tamu I don't think it's a bug, Controller should be easier, but then it's actually a pretty smart usage... I am in the middle. I think this will make custom register working: https://github.com/react-hook-form/react-hook-form/pulls i need to double-check, what're your thoughts on the type above @kotarella1110

xiaoyu-tamu:
@bluebill1049 the link does not point to a specific PR

bluebill1049:
oops @xiaoyu-tamu https://github.com/react-hook-form/react-hook-form/pull/2691

bluebill1049:
@xiaoyu-tamu would like to get a PR ready for the revert? looks like Kotaro is busy... I like the use case you provided above, perhaps we can live with the old type until Kotrao have objection we can discuss futher.

kotarella1110:
@xiaoyu-tamu @bluebill1049 As the error message below shows, I think Controller should be easier.

https://github.com/react-hook-form/react-hook-form/blob/a5197c1fabdfb8cb13a67cfd0717403ed44b3819/src/controller.tsx#L76

Such the use case are useful , but I think you should use custom register (the use case is few users).
I think this is bug. we should fix Controller.
What do you think?

mrlubos:
This throws for me as well @kotarella1110, so it feels more like a breaking change. What's the best way to import Controller props from the library by the way?

bluebill1049:
> @xiaoyu-tamu @bluebill1049 As the error message below shows, I think Controller should be easier.
> 
> https://github.com/react-hook-form/react-hook-form/blob/a5197c1fabdfb8cb13a67cfd0717403ed44b3819/src/controller.tsx#L76
> 
> Such the use case are useful , but I think you should use custom register (the use case is few users).
> I think this is bug. we should fix Controller.
> What do you think?

Thanks, @kotarella1110 for your reply, it's a pretty smart usage, but yea it can be considered as a bug. At the same time, it's like a custom register without running at the `useEffect`, which is kinda handy, but at the same time hard to document such behavior and may confuse normal users (except advance user like @xiaoyu-tamu) Happy to hear your guys' options as well.

@mrlubos are you using it with `useFieldArray` as well?

mrlubos:
@bluebill1049 Yes, I use it with `useFieldArray` too. I have a wrapper component implementing Controller though, so it fails for every instance, not with `useFieldArray` only.

bluebill1049:
I think i can come up with a "receipt"/Component which is not Controller but does the custom register inside. I think what @kotarella1110 did is correct and we should keep as it is.

bluebill1049:
```ts
function CustomRegister({ name, register, rules }) {
  React.useEffect(() => {
    register(name, rules);

    return () => unregister(name)
  })
  return <></>
}
```

What about the above solution to replace `Controller` without render and as?

xiaoyu-tamu:
LGTM, will this fn included in the lib or the user need  to define in their code?

bluebill1049:
> LGTM, will this fn included in the lib or the user need to define in their code?

user-defined code.

PR #2798, opened 2020-09-02
feature: is successfully submitted

### LINKED ISSUE #1452: Check if form has been submitted successfully
Currently `isSubmitted` is set to true, even though because of validation the `onSubmit` handler hasn't been triggered which is kind of confusing.

Both formik and final-form set `isSubmitted` to true only after the `onSubmit` handler has successfully ran

bluebill1049:
**Description:**

We are missing a piece of a state which is for successful submit, and this PR is going to address the problem. 

- [x] isSubmitSuccesful
- [x] update units
- [x] update cypress
- [x] update doc

poll result: https://twitter.com/HookForm/status/1301289633272086528

fix #1452

**Consider:**

To include status in the code and derived the boolean value form them, problem to be considered as well, for a parallel state like `isVald`, `isDirty` and etc


pmaier983 (dismissed):
LGTM!

bluebill1049:
Thanks @pmaier983 for the review.

PR #2807, opened 2020-09-03
add MutationObserver to interface stubs for React Native

Brendan-csel:
Resolve Typescript build error in React Native projects by adding stub for MutationObserver interface.

https://github.com/react-hook-form/react-hook-form/discussions/2805


bluebill1049:
Thanks

PR #2814, opened 2020-09-03
fix #2812 with isSubmitting

### LINKED ISSUE #2812: Strange async submit bug introduced on 6.5.0
Hi!

On my app I identified a bug where "isSubmitting" doesn't respect async functions on submitting form.

I will later identify where exactly this is happening to create a proper CodeSandBox repro.

But I couldn't so far... Async examples here are working fine.
I know this is on 6.5.0 because I was testing each version to find the one that breaks the behavior.

May I raise this issue til I figure out how to reproduce this?

This still happens on latest version

Version 6.4.1 with correct behavior
rhf 6-4-1

Version 6.5.0
rhf 6-5-0

bluebill1049:
fix #2812

https://codesandbox.io/s/happy-grass-xij0y?file=/src/src/useForm.ts

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -1050,6 +1050,7 @@ export function useForm<
         ) {
           updateFormState({
             errors: {},
+            isSubmitting: true,
           });
           await onValid(
             fieldValues as UnpackNestedValue<TSubmitFieldValues>,


src/useForm.ts
```diff
@@ -1050,6 +1050,7 @@ export function useForm<
         ) {
           updateFormState({
             errors: {},
+            isSubmitting: true,
```
JeromeDeLeon, line 1053: Hmm 🤔 so we are repeating this value to avoid bring replaced by it's previous one?

JeromeDeLeon: Shouldn't we just use the previous value from `setFormState` instead?

bluebill1049: this is because we are reading the value from the `ref` with Proxy, so the other way to fix it is `formStateRef.isSubmitting = true`, but this is one situation when we invoke updateFormState 3 times in a row. it's fine to bring another callback to the `updateState` for `prevState`, but I think it's a overkill unless we can find another scenario which we need it.

JeromeDeLeon: Oh yeah, seems good to me. We will have one more source of truth if we add that prevState 😆 my bad.

JeromeDeLeon (approved):

src/useForm.ts
bluebill1049: thanks for the review! @JeromeDeLeon

PR #2816, opened 2020-09-04
feature: setError focus the input

### LINKED ISSUE #949: focusErrorField while using setError
Would it be possible to focus first field with error while using `setError`? The functionality already exists in the code - there is `focusErrorField` being called in the `handleSubmit` function, maybe the same thing could be called inside `setError` function? (possibly as an opt-in feature, triggered by an additional option / argument e.g. `setError([...], { focusErrorField: true })`).

My case:
I am using api-platform with hydra jsonld format as a backend. When I submit a form, I may expect some errors coming from the api. I would like to set the errors to the form and have the first field with an error focused for better UX. Having that handled by `setError` would greatly reduce boilerplate code handling it, especially on large forms. (here I'm referring to handling refs manually for all the fields as per https://github.com/react-hook-form/react-hook-form/issues/230#issuecomment-527685852)

Example:
```typescript
  const onSubmit = async (data: any) => {
    try {
      await axios.post('/api/clients', {
        ...data,
      });
    } catch (error) {
      let message = 'An error occurred';

      const e: AxiosError = error;
      if (e.isAxiosError && e.response?.status && e.response.status < 500) {
        const { data } = e.response;
        message = data['hydra:title'];

        // i am setting possibly multiple errors within one setError call.
        // it would be great if this could focus the first field with an error,
        // e.g. with `setError([...], { focusErrorField: true })`
        setError(
          data?.violations.map((violation: any) => ({
            type: 'api',
            name: violation.propertyPath,
            message: violation.message,
          }))
        );
      } else {
        console.log(e);
      }

      alert(message);
    }
  };

```

PS. thanks for this awesome lib!

bluebill1049:
```ts
setError('test', { message: 'This is required', shouldFocus: true })
```

- [x] update doc

close #949


JeromeDeLeon (dismissed):
LGTM! :metal:

bluebill1049:
Thanks @JeromeDeLeon for the review.

vadimshvetsov:
Thanks, thats hot. I was ready to implement it after voting ends. Is there any way to get notified when feature is accepted and it's possible to start a PR?

bluebill1049:
> Thanks, thats hot. I was ready to implement it after voting ends. Is there any way to get notified when feature is accepted and it's possible to start a PR?

oh, next time I will make sure leave a message in the actual issue 👍

bluebill1049:
we will release this update this weekend.

Issue #1761, opened 2020-06-03
Support warnings instead of just errors in validation


JReinhold:
**Is your feature request related to a problem? Please describe.**

I have a complex form, where the user is allowed to enter and submit input, that is not in the desired shape. A simple example, is that the user can enter a date range with a start and end date (two separate input fields). The user is allowed to submit and save the date range with only one of the dates present (start OR end), but they are discouraged to do so. In that case, we want to show a warning to the user, but not an error.
The warning behaves like errors, and can be validated with the same logic (yup) as errors, but they should not block the form submission. They have another appearance in the UI than errors, and possibly also other strings.

**Describe the solution you'd like**
I would like an API that allows me to set warnings to fields, in the same way that I can set errors on them. I should of course also be able to read them in the same way. I see different ways of doing it:

1. add the same functionality that we have for errors, but with warnings instead. So `setWarning`, `clearWarning`, `ErrorMessage`, and make the `useForm` hook return a `warnings` object. I'm not sure how this would actually work with validation, because AFAIK, Yup doesn't support severity levels, so I don't know how we would tell from the result of the validation that the field is in a warning state and not an error state. we could have a `warningSchema` prop next to the `validationSchema` prop, but I don't think that is a very ergonomic way of doing it. However, this closely mimics the [warning-API in `redux-form`](https://redux-form.com/6.5.0/examples/syncvalidation/).
2. Add a `type` or `status` field to errors, that can be `"ERROR" | "WARN"`, and maybe even `"INFO"` in the future, if there's a use case for that. We could then use that field to determine if a warning or error should be shown, and if the form is valid or not (it should be invalid if at least one of the error types are `"ERROR"`). The naming might be confusing though, since `errors` would then both contain warnings and errors. And still, how do we actually specify in validation, that this should be a warning, and not an error? The only thing I can come up with, is a separate property for `warningSchema` and `warningResolver`, but maybe someone else is smarter than me.

**Describe alternatives you've considered**
My current workaround is to achieve this with a custom `validationResolver`. The `validationResolver` would first check for any errors, and then check for any warnings. The warnings would not actually make the `validationResolver` return any errors, to make sure the form is still considered valid. Instead, it would set a `warning` state in the component. This state would then be passed down to the fields with a custom React Context. So basically implement warnings completely outside `react-hook-form`. This however is only in theory, I haven't actually tried it out in practice yet.

**Additional context**
[...]

PR #2817, opened 2020-09-04
poc on warning message

bluebill1049:
A POC with Warning message

```tsx
register({
  warn: (value) => value < 3 && 'This is a warning message for you'
})
```

### DIFF
--- a/src/logic/validateField.ts
+++ b/src/logic/validateField.ts
@@ -24,7 +24,9 @@ import {
   FieldError,
   InternalFieldName,
   FlatFieldErrors,
+  FieldWarning,
 } from '../types';
+import isArray from '../utils/isArray';
 
 export default async <TFieldValues extends FieldValues>(
   fieldsRef: React.MutableRefObject<FieldRefs<TFieldValues>>,
@@ -40,9 +42,14 @@ export default async <TFieldValues extends FieldValues>(
     max,
     pattern,
     validate,
+    warn,
   }: Field,
   unmountFieldsStateRef: React.MutableRefObject<Record<string, any>>,
-): Promise<FlatFieldErrors<TFieldValues>> => {
+): Promise<{
+  error: FlatFieldErrors<TFieldValues>;
+  warning?: FieldWarning;
+}> => {
+  let warning;
   const fields = fieldsRef.current;
   const name: InternalFieldName<TFieldValues> = ref.name;
   const error: FlatFieldErrors<TFieldValues> = {};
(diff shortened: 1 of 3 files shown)


JeromeDeLeon:
Uhm, may I know what are we trying to achieve on this one?

bluebill1049:
> Uhm, may I know what are we trying to achieve on this one?

here is the feature request: https://github.com/react-hook-form/react-hook-form/issues/1761 nothing concrete, just giving it a spin.

JeromeDeLeon:
I'm not yet convinced of the idea because I believe a helper text could solve the problem but seems like a valid reason to add to this lib since it's a general approach for heaver form usage.

bluebill1049:
> I'm not yet convinced of the idea because I believe a helper text could solve the problem but seems like a valid reason to add to this lib since it's a general approach for heavy form usage.

honestly, I agree with you, I just want to do a POC and test the water. I do raise my concern over exploding the API and adding much more weight to the lib itself.

JeromeDeLeon:
Yeah, sure POC is fine 👍 , and also, yes, that couldn't be prevented from happening unless we use the same API like `clearErrors` to remove all warnings as well. I also don't like the idea of catching up with other libraries like just because they have it, we should have it too. What we are doing is leveraging tracking of form values, other than that are the spice to make it globally acceptable by users.

bluebill1049:
> Yeah, sure POC is fine 👍 , and also, yes, that couldn't be prevented from happening unless we use the same API like `clearErrors` to remove all warnings as well. I also don't like the idea of catching up with other libraries like just because they have it, we should have it too. What we are doing is leveraging tracking of form values, other than that are the spice to make it globally acceptable by users.

Totally agree. Hook form is special because it's unique and "light" weight, we shouldn't skew that easily.

JeromeDeLeon:
> ```ts
> const { warnings, setWarn, clearWarn } = useWarnings({
>   control,
> });
> ```

This one could work too or they could have wrapped the component with a warning and do `useWatch`. Might as well add that custom hook to make it more readable and testable.

bluebill1049:
I am going to close this PR. 👍

bluebill1049:
> > ```ts
> > const { warnings, setWarn, clearWarn } = useWarnings({
> >   control,
> > });
> > ```
> 
> This one could work too or they could have wrapped the component with a warning and do `useWatch`. Might as well add that custom hook to make it more readable and testable.

Yea, I agree. I think this can be a separate custom hook but not in the core.

PR #2830, opened 2020-09-05
fix reset updateFormState replacing isSubmitting

ejose19:
Fixes a variant of #2812 when reset gets called before onSubmit finishes.

In this case the proposed solution is for reset to not replace the isSubmitting property, so only handleSubmit controller is the one that manipulates isSubmitting.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -1101,7 +1101,7 @@ export function useForm<
     updateFormState({
       isDirty: isDirty ? formStateRef.current.isDirty : false,
       isSubmitted: isSubmitted ? formStateRef.current.isSubmitted : false,
-      isSubmitting: formState.isSubmitting,
+      isSubmitting: false,
       submitCount: submitCount ? formStateRef.current.submitCount : 0,
       isValid: isValid ? formStateRef.current.isValid : true,
       dirtyFields: dirtyFields ? formStateRef.current.dirtyFields : {},
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -1101,6 +1101,7 @@ export function useForm<
     updateFormState({
       isDirty: isDirty ? formStateRef.current.isDirty : false,
       isSubmitted: isSubmitted ? formStateRef.current.isSubmitted : false,
+      isSubmitting: formState.isSubmitting,
       submitCount: submitCount ? formStateRef.current.submitCount : 0,
       isValid: isValid ? formStateRef.current.isValid : true,
       dirtyFields: dirtyFields ? formStateRef.current.dirtyFields : {},


ejose19:
Same scenario as the one posted on issue, can confirm now it works correctly. [sandbox](https://codesandbox.io/s/react-hook-formreact-hook-form-app-forked-pz8c1?file=/src/form.tsx)

src/useForm.ts
```diff
@@ -1101,6 +1101,7 @@ export function useForm<
     updateFormState({
       isDirty: isDirty ? formStateRef.current.isDirty : false,
       isSubmitted: isSubmitted ? formStateRef.current.isSubmitted : false,
+      isSubmitting: formState.isSubmitting,
```
bluebill1049, line 1104:
```suggestion
      isSubmitting: false,
```

bluebill1049:
Thank you very much for the find 🔥

ejose19:
@bluebill1049 I think the state should be left unchanged instead of setting it to false, else a reset that runs while an async submit is processing may unset isSubmitting prematurely.

bluebill1049:
> @bluebill1049 I think the state should be left unchanged instead of setting it to false, else a reset that runs while an async submit is processing may unset isSubmitting prematurely.

It's a reset API :)

ejose19:
sure, should a comment regarding that be mentioned in the docs or it's redundant?

bluebill1049:
> sure, should a comment regarding that be mentioned in the docs or it's redundant?

https://react-hook-form.com/api#reset

`omitResetState` doesn't include isSubmitting, so I think it's safe to reset the isSubmitting as intended for now.

bluebill1049:
@ejose19 shall we get it in?

ejose19:
@bluebill1049 Sure, no other concerns.

bluebill1049:
> @bluebill1049 Sure, no other concerns.

Thank you very much for reporting and finding the bug. really appreciated it here.

bluebill1049 (approved):

ejose19:
Happy to contribute, this is the best react form library. Thank you as well for this amazing lib

alexssandrog:
Hi, I tested version 6.8.0, but the problem still occurs. I saw that isSubmitSuccessful is set to true, but isSubmitted is still false, isSubmitting is still true and submitCount is still zero. 

In my case, I use Redux Saga, so in submit I'm call redux actions and wait for status update by redux on an useEffect hook. Something like that

`const submit = (values) => {
  dispatch(updating({ values }));
};`

bluebill1049:
> Hi, I tested version 6.8.0, but the problem still occurs. I saw that isSubmitSuccessful is set to true, but isSubmitted is still false, isSubmitting is still true and submitCount is still zero.
> 
> In my case, I use Redux Saga, so in submit I'm call redux actions and wait for status update by redux on an useEffect hook. Something like that
> 
> ```
> const submit = (values) => {
>   dispatch(updating({ values }));
> };
> ```

can you reproduce in a csb?

alexssandrog:
On CSB it worked perfectly, I will investigate further and try to find out what it can be. Thank you for now.

alexssandrog:
Find the issue, [https://codesandbox.io/s/crazy-moon-dd3lk?file=/src/App.js](https://codesandbox.io/s/crazy-moon-dd3lk?file=/src/App.js), for test just fill those fields and click on _Save post_

I'm create an custom error for messages become from API, when I clear the error on useEffect, the isSubmitting still true.

Maybe it's the case to rethink my code.

ejose19:
@alexssandrog Seems the issue actually is with clearErrors

https://github.com/react-hook-form/react-hook-form/blob/e71f5d7a5bd659c410ee3e8d248fe4627f283b72/src/useForm.ts#L696-L709

since you're using it on useEffect, it's racing with handleSubmit and overwritting isSubmitting, leaving it set to true.

@bluebill1049 Maybe there should be another function like ```updateFormState``` but an actual partial update one, so it merges with ```formState``` instead of ```formStateRef.current```, that way we don't overwrite properties in race conditions like this one. If you're ok with it I can submit a PR

PR #2831, opened 2020-09-05
parseFormatInputValues example fixes

adrid:
fixes errors:
- handling new lines on text area - before it added semicolon after convert and didn't work
- number formatting - in the v6 the onChange event was moved to the render method and it didn't work anymore

fixes warnings on the example:
- emails is missing in the 'defaultValue' prop of either its Controller
- number is missing in the 'defaultValue' prop of either its Controller


bluebill1049:
Thank you very much.

PR #2833, opened 2020-09-06
Minor Test Improvements

pmaier983:
Very Minor changes to improve some tests. 

- One test was flakey (useForm) so I fixed the issue by having it wait for the element to be rendered. Mb keiya should poke in on this to double check.
- The coverage report said we were missing one line of coverage in deepMerge, so I created a test that covers that.

### DIFF
--- a/src/useForm.test.tsx
+++ b/src/useForm.test.tsx
@@ -237,7 +237,9 @@ describe('useForm', () => {
         render(<Component />);
       });
 
-      expect(screen.getByRole('button')).not.toBeDisabled();
+      await waitFor(() => {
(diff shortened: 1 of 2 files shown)


src/useForm.test.tsx
```diff
@@ -237,7 +237,9 @@ describe('useForm', () => {
         render(<Component />);
       });
 
-      expect(screen.getByRole('button')).not.toBeDisabled();
+      await waitFor(() => {
```
bluebill1049, line 240: Thanks for fixing this! FYI @keiya01

bluebill1049 (approved):
Looks great! thank you.

PR #2854, opened 2020-09-08
fix to output error message when as and render props are not given in Controller

kotarella1110:
fix for https://github.com/react-hook-form/react-hook-form/pull/2786#issuecomment-688387206


bluebill1049:
Thanks for the PR @kotarella1110 🙏

bluebill1049 (approved):

PR #2908, opened 2020-09-15
docs(README): add discord badge

neighborhood999:
Currently will see the **Widget Disabled**, you can check below discussion and setting:

ref: https://github.com/badges/shields/issues/1659


bluebill1049:
awesome! thanks

PR #2913, opened 2020-09-15
improve type with controller

bluebill1049:


src/controller.tsx
```diff
@@ -165,7 +165,7 @@ const Controller = <
     }
   };
 
-  const onChange = (...event: any[]) =>
+  const onChange = (...event: IArguments[]) =>
```
kotarella1110, line 168: Is this correct?

```ts
const t1: IArguments = 1; // type error
const t2: IArguments = 'test'; // type error
const t3: IArguments = true; // type error
```

bluebill1049: hmmm, you are right. looks like this has to be any

bluebill1049: thanks @kotarella1110

PR #2991, opened 2020-09-24
close #2990 improve dev mode warning

### LINKED ISSUE #2990: Remove message: "field" is missing in the 'defaultValue' prop of either its Controller
**Describe the bug**
I read from your docs that its _recommended_ to add all fields to the defaultValues object. I do get that that's the better pattern, but especially when working with non-vanilla form elements like react-select it often makes sense for a dev to work with undefined values.
Working with undefined values later on also helps with API requests since no empty strings will get communicated.

**Codesandbox link**
Here's an example that should work perfectly fine with undefined values
https://codesandbox.io/s/angry-euclid-yezfv?file=/src/Form.tsx

**Expected behaviour**
The above example should not throw annoying warnings (currently the console get A LOT of spam with a larger form).

If there is a REALLY good reason to why this warning should exist, at least optimize it to be one single warning that lists all the missing fields and only pops up once instead of spamming the console on every render. Also if this is the case, please explain and maybe put it into the docs.

bluebill1049:


tafelito:
@bluebill1049 what was the need to add the defaultValues warnings? Some places I don't need default values and now I have warnings all over the place

PR #3065, opened 2020-10-02
Fix 3058

bluebill1049:


src/useFieldArray.test.tsx
```diff
@@ -652,50 +709,6 @@ describe('useFieldArray', () => {
     );
   });
 
-  describe('with should unregister false', () => {
-    const Component = () => {
```
bluebill1049, line 656: almost killed me with this one... I was like why `.only` not working...

PR #3077, opened 2020-10-04
fixed jest config

keiya01:
I fixed jest config because coverage can not be outputted correctly 📝
And I fixed test script in package.json. `runInBand` options is removed because this is feature for debugging(https://jestjs.io/docs/en/cli#--runinband).


bluebill1049:
thank you 🧙 @keiya01 ❤️

PR #3115, opened 2020-10-07
fix: adjust isFormDirty function and set to exclude undefined

ejose19:
If setting undefined is indeed a valid value, then ```isFormDirty``` needs to use another function, since ```isEmptyObject``` returns false when the ```dirtyFields``` is in the format ```{"field": undefined}``` (as seen in two useFieldArray tests)

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -273,13 +273,7 @@ export function useForm<
     [],
   );
 
-  const isFormDirty = () =>
-    !deepEqual(
-      getValues(),
-      isEmptyObject(defaultValuesRef.current)
-        ? defaultValuesAtRenderRef.current
-        : defaultValuesRef.current,
-    ) || !isEmptyObject(formStateRef.current.dirtyFields);
+  const isFormDirty = () => !isEmptyObject(formStateRef.current.dirtyFields);
 
   const updateAndGetDirtyState = React.useCallback(
     (
@@ -483,10 +477,7 @@ export function useForm<
             );
 
             updateFormState({
-              isDirty: !deepEqual(
-                { ...getValues(), [name]: value },
-                defaultValuesRef.current,
-              ),
+              isDirty: isFormDirty(),
               dirtyFields: formStateRef.current.dirtyFields,
             });
           }
--- a/src/utils/set.ts
+++ b/src/utils/set.ts
@@ -27,8 +27,12 @@ export default function set(
           ? []
           : {};
     }
-    object[key] = newValue;
-    object = object[key];
+    if (newValue === undefined) {
+      delete object[key];
+    } else {
+      object[key] = newValue;
+      object = object[key];
+    }
   }
   return object;
 }


ejose19:
@bluebill1049 let me know regarding the question on the original message, if it's desired to have a field with undefined as value then ```set``` cannot be used on dirtyFields or it may end with ```{"field": undefined}``` which is not a good dirty state.

bluebill1049:
> @bluebill1049 let me know regarding the question on the original message, if it's desired to have a field with undefined as value then `set` cannot be used on dirtyFields or it may end with `{"field": undefined}` which is not a good dirty state.

`undefined` is possible with input and array fields as well.

ejose19:
@bluebill1049 if that's the case, should I create another function ```setDirtyFields``` to update ```dirtyFields``` only? that one will do exclude undefined values from the result (because having this ```{"field": undefined}``` as a ```dirtyFields``` value doesn't seem correct)

bluebill1049:
> @bluebill1049 if that's the case, should I create another function `setDirtyFields` to update `dirtyFields` only? that one will do exclude undefined values from the result (because having this `{"field": undefined}` as a `dirtyFields` value doesn't seem correct)

oops, sorry I am missing the context... it's going to be tricky here because you will have to take care of the case like: 
```
{
  test: {
     deep: [undefined, undfeind, 2]
  }
}
``` 
and after 2 gets removed, you have to remove the entire object, almost a reverse lookup. I think I remember why the deepEqual is better for the source of truth now.

by the way, thanks a lot for the PR and helping out.

ejose19:
I see, however we only need to check at the top level if there's any property with value !== undefined, so maybe extending https://github.com/react-hook-form/react-hook-form/blob/cdbbab8bc008ec0b9164ff000c1dfb83977b25e1/src/utils/isEmptyObject.ts#L1-L5 with a second argument (allowUndefined = false) and then using
```ts
isObject(value) && !allowUndefined
    ? !Object.keys(value).length 
    : !Object.entries(value).filter(([, v]) => v !== undefined).length; 
```
would solve the issue.

There's also some cases where isDirty = true doesn't have an equivalent in dirtyFields (like in this [sandbox](https://codesandbox.io/s/festive-hofstadter-2q7ci?file=/src/App.tsx:396-450), click delete on one of the rows, notice how isDirty is true but dirtyFields is empty) is this the correct behaviour? should dirtyFields be ```{data: [undefined, undefined, true]}``` assuming we deleted the third element?

bluebill1049:
> There's also some cases where isDirty = true doesn't have an equivalent in dirtyFields (like in this [sandbox](https://codesandbox.io/s/festive-hofstadter-2q7ci?file=/src/App.tsx:396-450), click delete on one of the rows, notice how isDirty is true but dirtyFields is empty) is this the correct behaviour? should dirtyFields be `{data: [undefined, undefined, true]}` assuming we deleted the third element?

This is a separate bug, which I will investigate for a fix.

ejose19:
I see the PR improved the behaviour with appending and removing, but when I do click on prepend what should be the correct behaviour? Supposing I have 3 elements [1,2,3] and I prepend a 0 => [0,1,2,3] shouldn't all 3 original elements be dirty? because when we look at each original index all their values differs (1 to 0, 2 to 1, 3 to 1). Current behaviour only marks dirty the first element, let me know if that's indeed correct.

Also, if ```dirtyFields``` is improved, we can rely on it to know if the form is dirty, hence proceeding with this PR making the last adjustment I proposed (either improving ```EmptyObject``` or making a different function). The idea is to make ```dirtyFields``` a robust foundation on which ```isDirty``` can rely on, without having 2 separate logics (unless you tell me you indeed want to have a ```isDirty``` form with an empty ```dirtyFields```.

bluebill1049:
> I see the PR improved the behaviour with appending and removing, but when I do click on prepend what should be the correct behaviour? Supposing I have 3 elements [1,2,3] and I prepend a 0 => [0,1,2,3] shouldn't all 3 original elements be dirty? because when we look at each original index all their values differs (1 to 0, 2 to 1, 3 to 2). Current behaviour only marks dirty the first element, let me know if that's indeed correct.

validate case, I will take a look at it.

bluebill1049:
> Also, if `dirtyFields` is improved, we can rely on it to know if the form is dirty, hence proceeding with this PR making the last adjustment I proposed (either improving `EmptyObject` or making a different function). The idea is to make `dirtyFields` a robust foundation on which `isDirty` can rely on, without having 2 separate logics (unless you tell me you indeed want to have a `isDirty` form with an empty `dirtyFields`.

I would still rely on deepEqual with `defaultValues`, I can't remember the detail, but it did serve a purpose.

ejose19:
I think you didn't got my point, I'm not against ```deepEqual``` with ```defaultValues```, the issue I see is that ```dirtyFields``` and ```isDirty``` uses different logic and that will make them prone to be out of sync (thus hiding bugs), if ```isDirty``` is always derived from ```dirtyFields``` them we can focus on making ```dirtyFields``` more robust (ie, porting the deepEqual to ```dirtyFields``` in a ```deepDiff``` like way)

bluebill1049:
> I think you didn't got my point, I'm not against `deepEqual` with `defaultValues`, the issue I see is that `dirtyFields` and `isDirty` uses different logic and that will make them prone to be out of sync (thus hiding bugs), if `isDirty` is always derived from `dirtyFields` them we can focus on making `dirtyFields` more robust (ie, porting the deepEqual to `dirtyFields` in a `deepDiff` like way)

That will go back to the original statement which I had: https://github.com/react-hook-form/react-hook-form/pull/3115#issuecomment-705245430 it's ok to have `undfined` in dirtyFields, and certainly i am not looking forward to change logic inside `set`. 

`const isFormDirty = //only do the deepEqual and remove dirtyFields check`

ejose19:
Seems we're not understanding each other, I'm not against having undefined in dirtyFields, if that's desired it's completely fine by me. The problem I see is having those 2 fields to act independent when in practice they aren't, will make users confused when they show opposite results.

I agree that using deepEqual on isFormDirty will give the correct result, but since dirtyFields is not using the same logic there will be cases that make them differs, that's why I can't agree with that solution.

bluebill1049:
> I agree that using deepEqual on isFormDirty will give the correct result, but since dirtyFields is not using the same logic there will be cases that make them differs, that's why I can't agree with that solution.

agree with your statement above, what's your solution or how would you solve this problem?

PR #3135, opened 2020-10-09
refactor: let SubmitHandler return any instead of void

pke:
This helps in code situations where your handler returns a promise with a non-void value.

## Before

```ts
onSubmit(data) {
  return updateUser(data).then(() => undefined)
}
```

## After

```ts
onSubmit(data) {
  return updateUser(data)
}
```

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -54,12 +54,12 @@ export type Mode = keyof ValidationMode;
 export type SubmitHandler<TFieldValues extends FieldValues> = (
   data: UnpackNestedValue<TFieldValues>,
   event?: React.BaseSyntheticEvent,
-) => void | Promise<void>;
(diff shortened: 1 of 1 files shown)


bluebill1049:
quick question: what's your usage like? because `handleSubmit` will be wrapped with your onSubmit right?

pke:
> quick question: what's your usage like? 

Yes `handleSubmit(onSubmit)` currently complains about the return type.

bluebill1049:
But why do u need to return anything from the onSubmit?

pke:
My button onPress handler actually can work with the response from onSubmit. The entity returned changed the buttons style. So the handleSubmit should return whatever type the caller wants because it’s used as a onPress callback which can be async

pke:
Thanks for merging!

PR #3141, opened 2020-10-10
reduce the complexity of stringToPath

bluebill1049:
- simplify stringToPath
- change the name to compact (https://lodash.com/docs/4.17.15#compact)


razh:
LoDash's `_.uniq` is more akin to `[...new Set(value)]` than `.filter(Boolean)`?

```tsx
_.uniq([true, false, 0, 1])
// [true, false, 0, 1]
```

The equivalent function looks to be `_.compact`:

https://lodash.com/docs/4.17.15#compact

bluebill1049:
Thanks @razh 👍

PR #3170, opened 2020-10-13
setError has completely optional properties for ErrorOptions, making it reasonable to have optional second arg

m3fawner:
* Update setError to accept 1 or 2 arguments, providing ErrorOptions default

I noticed setError has a union type, in which the first type has only one required property, and the second type definition is entirely optional properties.

We had a breaking change in version 6 which was caused by setError being used to block the form submission and display a broad message.

Given:
```
export type ErrorOption =
  | {
      types: MultipleFieldErrors;
      shouldFocus?: boolean;
    }
  | {
      message?: Message;
      type?: LiteralUnion<keyof ValidationRules, string>;
      shouldFocus?: boolean;
    };
```

I think it's safe to provide a default parameter to the second arg of an empty object.


Tymek:
reasonable. `FieldError` type in used indirectly in `UseFormMethods['errors']` has `type` property required. I'm curious what it the use case and consequences of not setting it.

m3fawner:
The case exists in the current implementation, too, as you can pass it an empty object with no `types` or `type` properties. We have used it for about 8 months now without providing type without issue. from what I can tell.

Issue #3002, opened 2020-09-25
Typescript error with Control


rusefendi:
**Describe the bug**
Typescript error with Control type from example (https://react-hook-form.com/ts#Control)

**To Reproduce**
1. Open https://codesandbox.io/s/objective-raman-jyt0g?file=/src/App.tsx
2. Look at the control prop from IsolateReRender component.
3. See error

**Codesandbox link (Required)**
https://codesandbox.io/s/objective-raman-jyt0g?file=/src/App.tsx

**Expected behavior**
Not having any errors

bluebill1049:
This issue is occured at 6.8.5

PR #3171, opened 2020-10-13
refactor watch, setError and clearErrors signatures in UseFormMethods to arrow functions

Tymek:
Part of the solution for #2887

Refactor for only methods unrelated to `Control` type + test for a bug in v6.8.5 from #3002


bluebill1049:
Thank you so much @Tymek for still try to investigate and find a solution. ❤️

PR #3236, opened 2020-10-22
Add package check

stramel:
Adds a package check and covers the rest of the outstanding issues. 

https://www.skypack.dev/blog/2020/10/skypack-quality-score-actionable-feedback-to-build-better-packages/
https://docs.skypack.dev/package-authors/package-checks

### DIFF
--- a/.github/workflows/unitTest.yml
+++ b/.github/workflows/unitTest.yml
@@ -10,6 +10,9 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install dependencies
         run: yarn install --frozen-lockfile
+      - name: Package Check
+        run: |
+          npm run check:package
       - name: Lint
         run: |
           npm run lint:types
--- a/package.json
+++ b/package.json
@@ -14,6 +14,11 @@
   "files": [
     "dist"
   ],
+  "exports": {
+    ".": {
+      "import": "./dist/react-hook-form.es.js"
+    }
+  },
   "scripts": {
     "clean": "rimraf dist",
     "prebuild": "npm run clean",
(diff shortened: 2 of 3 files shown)


.github/workflows/unitTest.yml
```diff
@@ -10,6 +10,9 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install dependencies
         run: yarn install --frozen-lockfile
+      - name: Package Check
+        run: |
+          npm run check:package
```
drwpow, line 15: 🙂

bluebill1049:
Thanks @stramel

klarstrup:
just FYI @stramel @bluebill1049 this breaks other left-out exports like `/dist/index.ie11.js` in some bundlers

stramel:
> just FYI @stramel @bluebill1049 this breaks other left-out exports like `/dist/index.ie11.js` in some bundlers

Hmmm can you provide anymore details on the setup you're referring to?

klarstrup:
I should've said "environment" rather than "bundler", as this also affects plain node. 

I'm not at a computer to reproduce right now but you can read about `exports` here: https://nodejs.org/api/packages.html#packages_package_entry_points

With this being a very relevant paragraph:
> Warning: Introducing the "exports" field prevents consumers of a package from using any entry points that are not defined, including the package.json (e.g. require('your-package/package.json'). This will likely be a breaking change.

stramel:
> I should've said "environment" rather than "bundler", as this also affects plain node. 
> 
> I'm not at a computer to reproduce right now but you can read about `exports` here: https://nodejs.org/api/packages.html#packages_package_entry_points
> 
> With this being a very relevant paragraph:
> > Warning: Introducing the "exports" field prevents consumers of a package from using any entry points that are not defined, including the package.json (e.g. require('your-package/package.json'). This will likely be a breaking change.

It looks like we can define a custom environment for the ie11 and umd exports and define a good default. I will put out a pr later today and do some more testing with this in various environments. 

Thanks for bringing it to our attention

bluebill1049:
I think this is the fix for that: https://github.com/react-hook-form/react-hook-form/pull/3274

PR #3241, opened 2020-10-23
Fix package names in exports

stramel:
Doh! I forgot to check the new build's file names!

This should be correct now and I added the `require` key as well.


bluebill1049:
oops... i didn't think twice as well.

PR #3262, opened 2020-10-25
fix #3261 dirtyField with object or array value compare

### LINKED ISSUE #3261: When a form element's value is an array, dirtyFields behaviour is different
**Describe the bug**
When a form element's value is an array, dirtyFields behaviour is different.

**To Reproduce**

1. Go to sandbox
2. Remove 'Afghanistan' option. **dirtyFields will set 'true' for countries.**
3. Re-select 'Afghanistan' option. **dirtyFields of countries is still 'true'.**

- [Material UI Autocomplete Multiple](https://codesandbox.io/s/react-hook-form-6-controller-forked-mow4i?file=/src/MuiAutoComplete.js) Template

**Expected behavior**
**dirtyFields** behaviour should be same for all types of values. i.e (string, number, array). Ideally, the dirtyFields should not include countries in the above case as the selected option (Afghanistan) is the same when the form was initally loaded and at the end of the third step. 

- OS: [macOs Catalina]
- Browser [chrome]
- Version [e.g. 81]

**Additional context**
I'm have'nt gone through how the dirtyFields are being built in react-hook-form. But for reference types values like arrays, I think we need to compare it with its values instead of their references.

bluebill1049:


vamshi-tg:
@bluebill1049, thanks for the super fast response on this 👍

When can we expect the v6.9.7?

PR #3258, opened 2020-10-24
Fix #3257 - Fix exports for react-native

cabiad:
See issue #3257 for background and details.

PR #3274, opened 2020-10-27
Fix exports for IE11

Qrymy:
## Problems
- After updating to v6.9.6, I could not build due to `ERR_PACKAGE_PATH_NOT_EXPORTED` with Next.js.

## Causes
- Defined base export path with `.`, so `./dist/index.ie11` is not defined by `exports` in `package.json`.  #3258 

## Fixes
- I've added `exports` paths to  `./dist/index.ie11`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,10 @@
     ".": {
       "import": "./dist/index.esm.js",
       "require": "./dist/index.js"
+    },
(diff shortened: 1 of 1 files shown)


stramel (approved):

bluebill1049:
Thanks @Qrymy and @stramel 's review.

PR #3299, opened 2020-10-29
Fix #3296 - Log warning if conflicting keyName is used

### LINKED ISSUE #3296: Calling reset does not reset controlled inputs within a useFieldArray, when rows have the same ID
**Describe the bug**
When calling reset, all inputs, both controlled and uncontrolled should be reset. However if I provide an id, rather than react-hook-form generating one, and it matches a controlled input that has already been rendered, calling reset will not reset the controlled input.

**To Reproduce**

See example

1. Create a form with a field array
2. Set some default values, which contain an  id: `{ test: [ {"id": "123", "firstName: "Bill", lastName: "Luo"} ] }` 
3. Assign properties to controlled inputs within the field array
4. Call `reset({  test: [{"id": "123", "firstName": "Milo", "lastName": "Hill" }]})`
5. Note that the controlled inputs are not reset to the new values

Include a codesandbox will help us to investigate the issue quicker.

https://codesandbox.io/s/react-hook-form-usefieldarray-template-forked-gs2xw?file=/src/index.js

**Expected behavior**
The controlled input should be reset, just like the uncontrolled inputs

Safari 14
Chrome 86

**Smartphone (please complete the following information):**

Not tested

**Additional context**

milo-:
Add some warning logs if a keyName used by the fieldValues. Based on the discussion here https://github.com/react-hook-form/react-hook-form/issues/3296

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -34,24 +34,22 @@ const mapIds = <
   values: Partial<TFieldArrayValues>[] = [],
   keyName: TKeyName,
 ): Partial<ArrayField<TFieldArrayValues, TKeyName>>[] => {
-  let reservedKeyUsed = false;
-  const result = values.map((value: Partial<TFieldArrayValues>) => {
-    if (keyName in value) {
-      reservedKeyUsed = true;
-    }
-    return {
-      [keyName]: generateId(),
-      ...value,
-    };
-  });
+  if (process.env.NODE_ENV !== 'production') {
+    for (const value of values) {
+      if (keyName in value) {
+        console.warn(
+          `📋 useFieldArray fieldValues contain the keyName \`${keyName}\` which is reserved for use by useFieldArray. https://react-hook-form.com/api#useFieldArray`,
+        );
 
-  if (process.env.NODE_ENV !== 'production' && reservedKeyUsed) {
-    console.warn(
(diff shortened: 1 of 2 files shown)


src/useFieldArray.ts
```diff
@@ -33,11 +33,26 @@ const mapIds = <
...
+  });
+
+  if (process.env.NODE_ENV !== 'production' && reservedKeyUsed) {
+    console.warn(
+      `📋 useFieldArray fieldValues contain the keyName \`${keyName}\` which is reserved for use by useFieldArray. https://react-hook-form.com/api#useFieldArray`,
+    );
+  }
+
+  return result;
+};
```
bluebill1049, line 55:
```suggestion
): Partial<ArrayField<TFieldArrayValues, TKeyName>>[] => {
  if (process.env.NODE_ENV !== 'production') {
    for (const value of values) {
      if (keyName in value) {
        console.warn(
          `📋 useFieldArray fieldValues contain the keyName \`${keyName}\` which is reserved for use by useFieldArray. https://react-hook-form.com/api#useFieldArray`,
        );

        break;
      }
    }
  }

  return values.map((value: Partial<TFieldArrayValues>) => ({
    [keyName]: generateId(),
    ...value,
  }));
};
```

Thank you for the quick PR! Love it! Just a small suggestion above.

milo-:
Great suggestion, thanks!

bluebill1049:
Thank you very much!

PR #3324, opened 2020-11-01
Memoize control so it does not change across renders

aaronjensen:
This is a big change, but it gives us a stable instance of `control` across re-renders. Most of the diff is moving functions around and wrapping them in `useMemo` or `useCallback`. All original tests pass with only one "hack" (adding a ref value to a dependency list, which shouldn't be necessary but is the way that defaults are dealt as far as I can tell)

I updated the example app `/controller/onSubmit` with a "Rerender" button and memoized the `Select` component. Now, when clicking rerender, the `Select` will not rerender.

### DIFF
--- a/app/src/controller.tsx
+++ b/app/src/controller.tsx
@@ -40,6 +40,8 @@ type Form = {
   RadioGroup: string;
 };
 
+const PureReactSelect = React.memo(ReactSelect)
+
 export default function Field(props: any) {
   const methods = useForm<Form>({
     defaultValues,
@@ -47,7 +49,10 @@ export default function Field(props: any) {
   });
   const { handleSubmit, errors, reset, control } = methods;
 
+  const [, setRerender] = React.useState(0)
   renderCount++;
+
+  const rerender = () => setRerender(Math.random())
 
   return (
     <form onSubmit={handleSubmit(() => {})}>
@@ -150,7 +155,7 @@ export default function Field(props: any) {
           <label>React Select</label>
           <Controller
             render={(props) => (
-              <ReactSelect isClearable options={options} {...props} />
+              <PureReactSelect isClearable options={options} {...props} />
             )}
             name="ReactSelect"
             control={control}
@@ -162,6 +167,8 @@ export default function Field(props: any) {
       </div>
 
       <span id="renderCount">{renderCount}</span>
+
+      <button type="button" onClick={rerender}>Rerender</button>
 
       <button
         type="button"
--- a/src/controller.test.tsx
+++ b/src/controller.test.tsx
@@ -946,4 +946,35 @@ describe('Controller', () => {
 
     expect(currentErrors.test).toBeUndefined();
   });
+
+  it('does not trigger rerenders for memoized inputs', () => {
+    let renderCount = 0;
+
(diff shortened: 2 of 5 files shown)


cschwebk:
Just came to see if this was in the works, as I need it for my own project, and found your PR. Thanks @aaronjensen, I hope it gets merged smoothly.

bluebill1049:
Thanks for the PR I will do some minor clean up and get it merged in.

aaronjensen:
That's great, thanks. Please clean away 😄

bluebill1049:
could you please test this beta version: `react-hook-form@6.10.2-beta.1`?

aaronjensen:
@bluebill1049 I'm trying to use it in the example app and I'm still seeing the rerender on the ReactSelect, but maybe I'm doing something wrong. What code is that package based off of?

aaronjensen:
I rebased my branch and retested and this PR works. If you could push a branch with your beta build I could maybe see what's going wrong.

aaronjensen:
From looking at the built code, 6.10.2-beta.1 is missing a lot of the required memoization (e.g. `onBlur`/`onChange` in `controller` and lots of stuff in `useForm`). Are you attempting to reduce memoization to less than is in this branch? The two tests I added in this branch should be pretty reliable for detecting re-renders.

bluebill1049:
> From looking at the built code, 6.10.2-beta.1 is missing a lot of the required memoization (e.g. `onBlur`/`onChange` in `controller` and lots of stuff in `useForm`). Are you attempting to reduce memoization to less than is in this branch? The two tests I added in this branch should be pretty reliable for detecting re-renders.

Yes, I haven't applied with Controller (i will just merge with your change), I am trying to reduce the memorization with `useForm`, yes, I am using your test at `useForm` it's passing.

aaronjensen:
> I am trying to reduce the memorization

Okay, were there performance impacts? With partial memoization, I worry about introducing subtle bugs and/or maintenance headaches in the future. I wonder if things could be refactored so that the things that do not need to be memoized don't appear to need it at all?

Said another way, how does a contributor to this repo know what's safe to leave off a memoization dependency list and what isn't?

bluebill1049:
Will do @aaronjensen valid points above, I am avoiding anything that is `ref`, or `useCallback` with empty `dependency`. I will release another beta today with your Controller update as well.

aaronjensen:
> I am avoiding anything that is `ref`

The eslint rules should prevent that assuming they were refs declared in the component... for the ones passed along control, it doesn't trust the `Ref` syntax. I wonder if that's configurable.



> `useCallback` with empty `dependency`

Good point! That'd be a nice thing for the lint rule to respect as well. It's too bad the lint rule is all or nothing and you can't say "No, I've considered this one and it's OK".

I'll look forward to the new beta, thank you.

bluebill1049:
here is the beta version with your Controller change: `6.10.2-beta.2`, let me know how did you go with the testing. Thanks.

aaronjensen:
@bluebill1049 it appears to work for me, at least in the demo app. Thanks!

bluebill1049:
Thanks @aaronjensen will sort out the rest this weekend! 🙏

PR #3356, opened 2020-11-04
Add verification if value is defined on FieldArray

djpremier:
When initializing a field array that is inside another field array via append, the initialization value is equivalent to 'undefined', with this an error is raised in React when in Development mode.

I propose checking that the variable containing the field value is initialized (it is true)


bluebill1049:
Thanks @djpremier 🙏

bluebill1049 (approved):

PR #3397, opened 2020-11-09
#2990 remove warning with defaultValue at Controller

### LINKED ISSUE #3329: watch vs. useWatch array length inconsistency when using useFieldArray
**Describe the bug**
Not sure if this is a bug/can be improved but... when using `useFieldArray` to insert/remove item to the fields, the resulting output (array) of that `useWatch` doesn't have a consistent length with the array length of updated fields after the insertion/removal.

**To Reproduce**

1. Go to the CSB link, clear the console after render. 
2. Click on "insert" or "remove".
3. See the console. Check the length of the `fields` vs. the length of the `useWatchedItems` on the first render loop.

https://codesandbox.io/s/watch-vs-usewatch-field-array-length-inconsistency-bi8it?file=/src/App.jsx:1365-1650

**Expected behavior**
The length of the `fields` and the length of the `useWatchedItems` should be the same on each render cycle.

**Additional context**
When `watch` is used instead of `useWatch` though, the watched array length seems to be consistent with the `fields` array length on each render. 👍

bluebill1049:
close #3329 useWatch return value to skip extra re-render

### DIFF
--- a/cypress/integration/useFieldArrayUnregister.ts
+++ b/cypress/integration/useFieldArrayUnregister.ts
@@ -43,6 +43,8 @@ describe('useFieldArrayUnregister', () => {
       ]),
     );
 
+    cy.get('input[name="data[0].name"]').blur();
+
     cy.get('#swap').click();
 
     cy.get('input[name="data[1].conditional"]').should('not.exist');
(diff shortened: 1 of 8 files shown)


xiaoyu-tamu:
Hi @bluebill1049. Is this PR reverted? 

I saw that you removed missing defaultValue warning in this PR. but it added back to `useController` hook.

bluebill1049:
> Hi @bluebill1049. Is this PR reverted?
> 
> I saw that you removed missing defaultValue warning in this PR. but it added back to `useController` hook.

Yea, it's reverted back.

xiaoyu-tamu:
Thanks. I will open a discussion ask for best practice to avoid this warning.

Issue #3350, opened 2020-11-04
Suggestion: rename "ref" prop in Controller render()


dadouam:
Hello,

**Is your feature request related to a problem? Please describe.**
During my migration from V5 to V6, I used the following syntax to simplify moving from the `as` prop to the `render` Controller prop, since most of the time, there was a 1-1 match between the names of the passed props and the component props (value, onChange...)
```jsx
<Controller name={"myName"} render={p => <MyComponent {...p}/>}/>
```
Note that the following syntax is described in the doc, so I thought it would be an easy and convenient way to make the migration easier.

However, I found out that 6.10 recently added a `ref` prop, and while I think it's a good idea to be able to pass down a ref, it causes the following error to show up on the console for most of my controlled functional components : `Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

I certainly don't want to wrap all of my functional form components with forwardRef (sometimes tricky to do in TypeScript, and not always relevant) and of course, I can choose to not include the `ref` prop in what I pass down to the render() component, but I wonder if maybe it would be a good idea to rename this `ref` prop?

I think it makes some sense because :

- `ref` are unlike other props so passing down a `ref` prop has some gotchas (can trigger errors in console, doesn't work out of the box with functional components...)
- Most components used with RHF should be functional, since class components are kinda legacy now. So it would mean we would have to be extra careful when passing down the `render()` props, and even more also when using the `as` prop (I've checked that with the `as` prop, it also passes down the ref prop, and unlike `render()`, there doesn't seem to be a way out of this problem, although `as` is no longer the recommended way) 

**Describe the solution you'd like**
Considering that `ref` is a special prop "reserved" by React, I suggest renaming the `ref` passed down by `Controller` to something else, like `controllerRef`. Then it would be up to the developer to bind this ref to either the component `ref` or other prop (`inputRef`, etc...)

dadouam:
A possible workaround is to wrap `Controller` into a custom Controller that renames the `ref` prop to something else. Not ideal of course since I'd need to keep the types consistent with future RHF releases.

```typescript
// eslint-disable-next-line @typescript-eslint/ban-types
type AsProps<TAs> = TAs extends undefined ? {} : TAs extends React.ReactElement ? Record<string, any> : TAs extends React.ComponentType<infer P> ? P : TAs extends keyof JSX.IntrinsicElements ? JSX.IntrinsicElements[TAs] : never;

[...]

PR #3407, opened 2020-11-09
Add warning on ref.current.focus when it is not a function

dadouam:
Following our discussion in this issue https://github.com/react-hook-form/react-hook-form/issues/3350.

I added a console.warn when users pass the `ref` from `Controller` to components that don't expose a `focus` function. The goal is to prevent an error log and to inform users that a `focus()` function must be present in order for this `ref` prop to work as intended.


bluebill1049:
thank you very much @dadouam 🙏

src/controller.tsx
```diff
@@ -67,7 +68,18 @@ const Controller = <
...
-  const onFocusRef = React.useRef(onFocus || (() => ref.current.focus()));
+  const onFocusRef = React.useRef(
+    onFocus ||
+      (() => {
+        if (isFunction(ref.current.focus)) {
+          ref.current.focus();
+        } else {
+          console.warn(
+            `📋 'ref' from Controller render prop must be attached to a React component or a DOM Element whose ref provides a 'focus()' method`,
+          );
```
bluebill1049, line 79:
```suggestion
          if (process.env.NODE_ENV !== 'production') {
            console.warn(
              `📋 'ref' from Controller render prop must be attached to a React component or a DOM Element whose ref provides a 'focus()' method`,
            );
          }
```

Issue #3342, opened 2020-11-04
defaultValues with field arrays and yup stopped working in 6.3.1


mortenk:
**Describe the bug**
Form fields are not populated with default values when using field arrays and yup resolver

**To Reproduce**
Click on the sandbox link
Try react-hook-form 6.3.0 to see default values being populated
Change it to 6.3.1 to see it stop working

**Codesandbox link (Required)**
https://codesandbox.io/s/react-hook-form-array-default-values-696-jigu1?file=/src/index.js

**Expected behavior**
I expect form fields to be populated

**Desktop (please complete the following information):**
Any os/browser

**Smartphone (please complete the following information):**
Any device/broser

bluebill1049:
Have a look at the changelog just like we discussed: https://github.com/react-hook-form/react-hook-form/blob/master/CHANGELOG.md

bluebill1049:
reference for others: https://github.com/react-hook-form/react-hook-form/discussions/3280

bluebill1049:
For more information in terms of defaultValue with `useFieldArray`, please refer to the doc: 
https://react-hook-form.com/api/#useFieldArray

mortenk:
This has been working as expected since at least 5.x, now it is broken. 
How can it not be a bug? 
It breaks in a patch version and you say it's intentional?

mortenk:
Commit 634bb0815c7e450c61c30a6c274ea68327e6af91 is the cause of this breakage

bluebill1049:
same reply as in the discussion:

This is the best opportunity for you to contribute if you truly believe without defaultValue will not introduce bug with all useFieldArray action. I would recommend you fork the repo, have some deep investigation before making a judgment that I can work without defaultValue. On the other hand, if it truly works and passes all unit tests and integration I am more than happy to review your PR. This is in the doc from the very beginning of V6, you can go watch the video tutorial. It may work (but with a bug) before 6.3, but it's always required to have defaultValue to avoid issues while shuffling the array field if may work for your use case, but this is not made only work single use case.

PR #3420, opened 2020-11-11
defaultValues with field arrays and yup stopped working in 6.3.1 #3342

mortenk:


src/controller.test.tsx
```diff
@@ -772,7 +772,7 @@ describe('Controller', () => {
       const Component = () => {
         const { control } = useForm({
           defaultValues: {
-            test: [{ data: '' }],
+            test: [{}],
```
bluebill1049, line 775: why is this changed?

src/useFieldArray.ts
```diff
@@ -482,6 +480,7 @@ export const useFieldArray = <
   React.useEffect(() => {
     const resetFunctions = resetFieldArrayFunctionRef.current;
     const fieldArrayNames = fieldArrayNamesRef.current;
+    fieldArrayNames.add(name);
```
bluebill1049, line 483: nice.

src/useFieldArray.test.tsx
```diff
@@ -3708,6 +3708,38 @@ describe('useFieldArray', () => {
       ]);
     });
 
+    it('should populate all fields with default values', () => {
```
bluebill1049, line 3711: Thanks for the test.

bluebill1049 (commented):
Thanks for the PR 👍 and your effort. I think we will still need to use defaultValue, because all the cypres testing still using defaultVlaue and it's going to raise issues when you actually remove all the default values (which combine the usage with all the other actions from useFieldArray). Happy to merge this PR tho.

src/controller.test.tsx
mortenk: The test verifies that a warning is printed to the console if defaultValue is missing from either the input or useForm(), but the my fix makes defaultValue no longer required for the input field. The value is populated from the defaultValues passed to useForm() and no warning is printed to the console anymore.

bluebill1049: It works with default populate, but not when you manipulate the fields (those actions). I will merge the PR anyhow, thanks for your effort.

PR #3461, opened 2020-11-17
🐞 fix #3460 issue with nested schema

### LINKED ISSUE #3460: The validation error disappears when schema is nested.
# Problem
Sorry if this is not an issue in this repository.

sandbox
https://codesandbox.io/s/react-hook-form-with-yup-forked-1ihqf?file=/src/index.js

I use yup's schema validation.
The schema is nested as follows.

```js
const schema = yup.object().shape({
  user: yup.object().shape({
    firstName: yup.string().required(),
    lastName: yup.string().required(),
    age: yup.string().required()
  })
```

I want the initial render to be in an error state, so I run trigger in useEffect.
And The mode of validation is onBlur.

```js
  // initial validate
  React.useEffect(() => {
    trigger();
  }, [trigger]);
```

The initial render shows the error as expected, but when I finish entering the first item, the error disappears even though there are items that have not been entered.

screencast 2020-11-16 21-40-23

Without nesting the schema, it behaved as expected.

bluebill1049:


bluebill1049:
e2e pass will investigate why failing after.

PR #3483, opened 2020-11-20
use `useField` internally for Controller

bluebill1049:
- expose logic into `useField` custom hook
- fox remove function with an array payload
- `useField` will not expose until API confirmed: https://github.com/react-hook-form/react-hook-form/discussions/3487


mattsputnikdigital:
Are you working on a useField hook similar to Formik's?

That would be the missing piece for me.

bluebill1049:
@mattsputnikdigital join the RFC if you interested: https://github.com/react-hook-form/react-hook-form/discussions/3487

Issue #2984, opened 2020-09-23
control type incompatibility in v6.5.0-6.8.4


delesseps:
**Describe the bug**

When specifying types for controls via `T extends Record<string, any>` typing errors are raised that exist from versions 6.5.0 through to 6.8.4.  I've demonstrated the issue in sandboxes below for the error as well as the working version in 6.4.1.

Note: that it doesn't seem to impact the execution on codesandbox, but under certain TypeScript configurations will break the build.

![image](https://user-images.githubusercontent.com/5125315/94095756-bfb9d380-fe65-11ea-8118-a86cc6f3ea18.png)

```typescript
Type 'Control<T>' is not assignable to type 'Control<Record<string, any>> | undefined'.
  Type 'Control<T>' is not assignable to type 'Control<Record<string, any>>'.
    Type 'Control<T>' is not assignable to type '{ removeFieldEventListener: (field: Field, forceDelete?: boolean | undefined) => void; mode: Readonly<{ isOnBlur: boolean; isOnSubmit: boolean; isOnChange: boolean; isOnAll: boolean; isOnTouch: boolean; }>; ... 19 more ...; renderWatchedInputs: (name: string, found?: boolean | undefined) => void; }'.
      Types of property 'updateFormState' are incompatible.
        Type '(args?: Partial<FormState<T>> | undefined) => void' is not assignable to type '(args?: Partial<FormState<Record<string, any>>> | undefined) => void'.
          Types of parameters 'args' and 'args' are incompatible.
            Type 'Partial<FormState<Record<string, any>>> | undefined' is not assignable to type 'Partial<FormState<T>> | undefined'.
              Type 'Partial<FormState<Record<string, any>>>' is not assignable to type 'Partial<FormState<T>>'.
                Types of property 'dirtyFields' are incompatible.
                  Type 'DeepMap<Record<string, any>, true> | undefined' is not assignable to type 'DeepMap<T, true> | undefined'.
                    Type 'DeepMap<Record<string, any>, true>' is not assignable to type 'DeepMap<T, true>'.ts(2322)
controller.d.ts(20, 5): The expected type comes from property 'control' which is declared here on type '(IntrinsicAttributes & { as: "textarea" | "input" | "select" | FunctionComponent<any> | ComponentClass<any, any> | ReactElement<...>; render?: undefined; } & { ...; } & Pick<...>) | (IntrinsicAttributes & ... 2 more ... & Pick<...>)'
```

[...]

PR #3501, opened 2020-11-22
fix controller type argument inference

### LINKED ISSUE #3500: Controller component's TControl type parameter does not respect subtyping
**Describe the bug**
Controller and a few other components/hooks/prop types are parameterized on `TControl` and, while this is not an issue in development (e.g. if you import types locally from `react-hook-form/src`), in production this ends up as following
```ts
// dist/controller.d.ts:4
declare const Controller: <..., TControl extends Control<Record<string, any>> = Control<Record<string, any>>...
```
This leads to Controller's `control` prop being incompatible with type parameters on userland component with similar type constraint `T extends Record<string, any>`

**To Reproduce**

0. Make sure your imports are from the production bundle (e.g. `node_modules` or `./dist` for local)
1. Create a component which has a generic parameter (note: this is the same as `Control`s type parameter) and render a `Controller` wrapper with its `control` property taken from parameterized props, as follows:
```tsx
import {Control, Controller, FieldName, useForm} from 'react-hook-form'

type FieldProps<FV extends Record<string, any>> = {
  control: Control<FV>
  name: FieldName<FV>
}

const Field = <FV extends Record<string, any>>(props: FieldProps<FV>) => (
  <Controller control={props.control} name={props.name} as="input" />
)
```
2. See type error at `<... control={props.control} .../>` which boils down to this:

```sh
Type 'DeepMap<Record<string, any>, true>' is not assignable to type 'DeepMap<FV, true>'. ts(2322)
```

https://codesandbox.io/s/stoic-dawn-xx41j

**Expected behavior**
No errors, types of Controller's `control` prop and passed in value's generic constraints match fully

**Additional context**
This is a long-standing issue. First, I faced this error back in July 2020. We were building re-usable components with integrations of `@material-ui` and `react-hook-form` in order to minimize the code needed to build large and complex forms (project requirements). And, on top of that, I spent a lot of time to make all these components and hooks typed as strictly as possible. Once this problem occurred I tried to discuss it here #2700 but got no response. Also, I had some priorities changed at work so I `@ts-expect-error`ed this and only came back to this recently, so I'm happy to help!

Also, there is a similar issue closed but not completely fixed I believe: #2984 

I found a partial fix for this 3d713b310d5ccb303313ccad82585ea81a068c4e in git logs but it was reverted due to some other error.

IMO that fix was in right direction since the main "type variable" here is the user's form data shape (i.e. `FieldValues`). So, it makes sense if we make all other type constraints based on this. And, this way we don't need `TControl` type param anywhere since it can be replaced by `Control<FV>` directly unless the `Control` type itself is not a static type (e.g. if there are some descendants of it, then it also needs to be mentioned in generic parameters, so far I didn't find such cases in the codebase).

amankkg:
This is my attempt to fix #3500 (there is a motivation behind these changes)

* replaced all `TControl` with `FieldValues` + `Control<FV>` combination
* removed `FieldValuesFromControl` type since now `FieldValues` are used directly and no need to infer them from `(T)Control`

TODO (need help or yes/no)

- [ ] how to test these changes? Since this bug doesn't appear in dev, but it can be reproduced locally if types are imported from `./dist` instead of `./src`
- [x] there are multiple bugs fixed by the TControl solution, if anyone remembers the issues I'd check them out for possible regression

**Update:** I found 3 more mentions of this bug so far: #2700 #2984 #3002
All of them with no such error on this revision: https://codesandbox.io/s/crazy-moon-kh79k

PS: branch name should be `controller-generics-contravariance` :P


bluebill1049:
Thanks, heaps for the PR 🙏 .

amankkg:
@bluebill1049 no problems, RHF is a great tool with very interesting bugs 😄 

Have you guys considered type tests (`dtslint` or similar tool?) since such things are not normal unit tests.

I personally used to write dummy tests with `@ts-expect-error` flags on assertion/reproduction line (great for regression checks). And then run `tsc` on such files so it fails on type errors.

Something like this

```ts
/* fn.ts */
export const fn = (n: number) => n + 1

/* fn.test.ts */
it('should check argument types', () => {
  const n: string = '1'

  // @ts-expect-error
  fn(n)

  expect(true).toBe(true)
})
```

amankkg:
Added similar bugs reproduction sandbox. All work as expected

bluebill1049:
we did consider adding tests for types, but just out of hand... if you are keen, please help and contribute. (no pressure tho)

kotarella1110 (approved):
LGTM 🎉
Thank you very much 🙏

bluebill1049:
Thank you very much for your work @amankkg and @kotarella1110's review.

amankkg:
Hey @bluebill1049 @kotarella1110 
I'm slowly getting familiar with the codebase, and I think there is another use of `TControl`
```ts
// src/types/fieldArray.ts
export type FieldArrayName = string;

export type UseFieldArrayOptions<
  TKeyName extends string = 'id',
  TControl extends Control = Control
> = {
  name: FieldArrayName;
  keyName?: TKeyName;
  control?: TControl;
};
```

Since I've not used this API yet I'm not sure if there are any bugs about it. However, the pattern is quite similar.

I quickly tried to replace `TControl` here with `Control<FieldValues` and got an error that led me to `FieldArrayName` type being just a `string` (kinda strange, isn't it?).

So, if you have some context on this line, can you shed some light on it. I'll take a deeper look later too.

Another Q: what is https://github.com/react-hook-form/strictly-typed and how it corelates with this repo? Thanks!

PR #3527, opened 2020-11-26
rename methods to formCtx/data to ctx

mwmcode:
Hello,

I think calling it `formCtx`/`ctx` better describes the value it holds since it holds other types/primitives like `formState` and `isValid` which are not methods (but objects & booleans)


bluebill1049:
Thanks, @mustafawm, I think we going to pass with this PR, we try to not have abbreviation in the example folder.

PR #3553, opened 2020-11-28
submitHandler forwards successful return values

jrobber:
If a submitHandler's callback function returns a value and there are no errors, return that value.

Why:  I encountered a scenario where I had to invoke the handleSubmit manually.  I needed to know the results of the operation to know how to proceed.

Tests:
Updated existing unit test.
All tests pass & build.


bluebill1049:
Thanks for the PR. How we deal with the return value is a tricky one for us (eg, the user may want to do something with return errors as well), we will not introduce this feature update at the moment until we have a better understanding and solution.

PR #3571, opened 2020-11-30
Make cloneObject work in both Node & browsers

### LINKED ISSUE #3570: ReferenceError: File is not defined when using useFieldArray
**Describe the bug**
In a Next.js project, in an SSR-ed page, I'm not able to use `useFieldArray`.

**To Reproduce**

1. Use a starter template of Next.js
2. Create a [server-side rendered page](https://nextjs.org/docs/basic-features/pages#server-side-rendering)
3. Use a `useFieldArray` in the page's default component.
4. See error —

```
ReferenceError: File is not defined
    at cloneObject (/my-project/node_modules/react-hook-form/dist/index.cjs.development.js:620:90)
    at useFieldArray (/my-project/node_modules/react-hook-form/dist/index.cjs.development.js:1518:71)
    at RowContainer (webpack-internal:///./pages/bookings/add/custom.tsx:134:69)
    at processChild (/my-project/node_modules/react-dom/cjs/react-dom-server.node.development.js:3353:14)
    at resolve (/my-project/node_modules/react-dom/cjs/react-dom-server.node.development.js:3270:5)
    at ReactDOMServerRenderer.render (/my-project/node_modules/react-dom/cjs/react-dom-server.node.development.js:3753:22)
    at ReactDOMServerRenderer.read (/my-project/node_modules/react-dom/cjs/react-dom-server.node.development.js:3690:29)
    at renderToString (/my-project/node_modules/react-dom/cjs/react-dom-server.node.development.js:4298:27)
    at Object.renderPage (/my-project/node_modules/next/dist/next-server/server/render.js:53:851)
    at Function.getInitialProps (webpack-internal:///../node_modules/next/dist/pages/_document.js:135:19)
```

I don't have one at the moment, terribly sorry, but will add in a day (see https://github.com/react-hook-form/react-hook-form/pull/3571 because I have a suggested fix as well).

**Expected behavior**
The error shouldn't happen.

**Screenshots**
Screenshot 2020-11-30 at 10 42 41 PM

- OS: macOS Catalina
- Browser Microsoft Edge 87.0.664.47
- Version RHF 6.12.1 (unsure if this works in earlier versions)

**Additional context**
I've managed to fix this issue locally by patching the [`cloneObject`](https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/cloneObject.ts) utility method by including a test for `File` before it is used. This fixes the issue. I've also raised a PR with this fix - https://github.com/react-hook-form/react-hook-form/pull/3571

paambaati:
This fixes #3570 by ensuring we check for `File` to be available as a global before we do an instance check. This is required as `File` isn't available on Node.js environments.

@bluebill1049 I'm not sure how to test for this, so any help there is appreciated.

### DIFF
--- a/src/utils/cloneObject.ts
+++ b/src/utils/cloneObject.ts
@@ -1,8 +1,5 @@
 import isPrimitive from './isPrimitive';
-import { UNDEFINED } from '../constants';
-import isUndefined from '../utils/isUndefined';
-
-const isWeb = typeof window !== UNDEFINED && !isUndefined(HTMLElement);
+import isWeb from './isWeb';
 
 export default function cloneObject<T extends unknown>(data: T): T {
   let copy: any;
--- a/src/utils/cloneObject.ts
+++ b/src/utils/cloneObject.ts
@@ -9,7 +9,7 @@ export default function cloneObject<T extends unknown>(
   if (
     isPrimitive(data) ||
     (isWeb && typeof File === 'function' && data instanceof File)
-    ) {
+  ) {
     return data;
   }
 
--- a/src/utils/cloneObject.ts
+++ b/src/utils/cloneObject.ts
@@ -6,7 +6,10 @@ export default function cloneObject<T extends unknown>(
 ): T {
   let copy: any;
 
-  if (isPrimitive(data) || (isWeb && typeof File === 'function' && data instanceof File)) {
+  if (
+    isPrimitive(data) ||
+    (isWeb && typeof File === 'function' && data instanceof File)
+    ) {
     return data;
   }
 
--- a/src/utils/cloneObject.ts
+++ b/src/utils/cloneObject.ts
@@ -1,15 +1,12 @@
 import isPrimitive from './isPrimitive';
+import { UNDEFINED } from '../constants';
 
-export default function cloneObject<T extends unknown>(
-  data: T,
-  isWeb = true,
-): T {
+const isWeb = typeof window !== UNDEFINED;
+
+export default function cloneObject<T extends unknown>(data: T): T {
   let copy: any;
 
-  if (
-    isPrimitive(data) ||
-    (isWeb && typeof File === 'function' && data instanceof File)
-  ) {
+  if (isPrimitive(data) || (isWeb && data instanceof File)) {
     return data;
   }
 
@@ -29,7 +26,7 @@ export default function cloneObject<T extends unknown>(
   if (data instanceof Map) {
     copy = new Map();
     for (const key of data.keys()) {
-      copy.set(key, cloneObject(data.get(key), isWeb));
+      copy.set(key, cloneObject(data.get(key)));
     }
     return copy;
   }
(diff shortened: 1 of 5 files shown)


paambaati:
@bluebill1049 On closer analysis, perhaps adding the `isWeb` parameter to the `cloneObject()` call here too would be a more robust fix? Thoughts?

https://github.com/react-hook-form/react-hook-form/blob/4407116f36df21de2247b839802028aa8cb0efe5/src/useFieldArray.ts#L152

src/utils/cloneObject.ts
```diff
@@ -6,7 +6,10 @@ export default function cloneObject<T extends unknown>(
 ): T {
   let copy: any;
 
-  if (isPrimitive(data) || (isWeb && data instanceof File)) {
+  if (
+    isPrimitive(data) ||
+    (isWeb && typeof File === 'function' && data instanceof File)
```
bluebill1049, line 11: I think the issue is probably caused due to isWeb is not passing down to the function.

shall we just keep `typeof window === UNDEFINED` inline and remove that extra isWeb argument?

bluebill1049:
Thanks, heaps for the PR 🙏

src/utils/cloneObject.ts
bluebill1049: actually, I think `isWindowUndefined` will do.

bluebill1049: `const isWindowUndefined = typeof window === UNDEFINED;`

paambaati: @bluebill1049 Yeah using `isWindowUndefined` seems good enough. Why was `isWeb` introduced in the first place? My initial guess was that it made it easier to test/mock browsers & Node in tests.

bluebill1049: yes.

paambaati: @bluebill1049 Done, just pushed. I decided to call it `isWeb` (instead of using `const isWindowUndefined = typeof window === UNDEFINED;` like you suggested) in `cloneObject.ts` because it made its purpose more explicit.

bluebill1049: hmmm looks like there is react native as well to take care.

src/utils/cloneObject.ts
```diff
@@ -1,9 +1,9 @@
 import isPrimitive from './isPrimitive';
+import { UNDEFINED } from '../constants';
 
-export default function cloneObject<T extends unknown>(
-  data: T,
-  isWeb = true,
-): T {
+const isWeb = typeof window !== UNDEFINED;
```
bluebill1049, line 4:
```suggestion
const isWeb = typeof window !== UNDEFINED && !isUndefined(HTMLElement);
```

bluebill1049: see if this fix react-native tests.

src/utils/cloneObject.ts
paambaati: Not sure how `window` globals work in RN; should we then go back to checking if `File` exists? Like `typeof File === 'function'`?

bluebill1049: hmmm, let me do some quick testing.

bluebill1049:
could you please try this beta? `6.12.2-beta.1`

paambaati:
@bluebill1049 Can confirm that `6.12.2-beta.1` works! 🥳

bluebill1049:
> @bluebill1049 Can confirm that `6.12.2-beta.1` works! 🥳

Thank you so much for your help!

PR #3607, opened 2020-12-05
change from deepClone to shallowClone with defaultValues's reset

bluebill1049:
- [x] include information in the doc

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-hook-form",
   "description": "Performant, flexible and extensible forms library for React Hooks",
-  "version": "6.12.2",
+  "version": "6.12.3-beta.1",
   "main": "dist/index.js",
   "module": "dist/index.esm.js",
   "umd:main": "dist/index.umd.production.min.js",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-hook-form",
   "description": "Performant, flexible and extensible forms library for React Hooks",
-  "version": "6.12.3-beta.1",
+  "version": "6.12.2",
   "main": "dist/index.js",
   "module": "dist/index.esm.js",
   "umd:main": "dist/index.umd.production.min.js",
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -1218,7 +1218,7 @@ export function useForm<
     }
 
     fieldsRef.current = {};
-    defaultValuesRef.current = cloneObject(values || defaultValuesRef.current);
+    defaultValuesRef.current = { ...(values || defaultValuesRef.current) };
     values && renderWatchedInputs('');
 
     Object.values(resetFieldArrayFunctionRef.current).forEach(


RPereiraRedspark:
i agree with this change, the deep clone broke the refs from my multiple selects, i use reset to bulk load all my form after a async get, i recently migrated from version 5 to 6 and noticed this new behavior, now i have to use setValue just for this field and trigger a redraw because of it

bluebill1049:
> i agree with this change, the deep clone broke the refs from my multiple selects, i use reset to bulk load all my form after a async get, i recently migrated from version 5 to 6 and noticed this new behavior, now i have to use setValue just for this field and trigger a redraw because of it

It's very sad people doing the following.

```
const defaultValues = {xxxx}
reset(defaultValues)
```

hence the deepClone, I think this is a minor change will need to be included in the doc as well, so people aware of that, it's safer to use reset with not the same reference.

RPereiraRedspark:
> > i agree with this change, the deep clone broke the refs from my multiple selects, i use reset to bulk load all my form after a async get, i recently migrated from version 5 to 6 and noticed this new behavior, now i have to use setValue just for this field and trigger a redraw because of it
> 
> It's very sad people doing the following.
> 
> ```
> const defaultValues = {xxxx}
> reset(defaultValues)
> ```
> 
> hence the deepClone, I think this is a minor change will need to be included in the doc as well, so people aware of that, it's safer to use reset with not the same reference.

The problem is, how do you reset fields such as Select without relying on the ref of the value? As setValue does not accept multiple values anymore, people will rely more the reset method to do such things

bluebill1049:
> > > i agree with this change, the deep clone broke the refs from my multiple selects, i use reset to bulk load all my form after a async get, i recently migrated from version 5 to 6 and noticed this new behavior, now i have to use setValue just for this field and trigger a redraw because of it
> > 
> > 
> > It's very sad people doing the following.
> > ```
> > const defaultValues = {xxxx}
> > reset(defaultValues)
> > ```
> > 
> > 
> > hence the deepClone, I think this is a minor change will need to be included in the doc as well, so people aware of that, it's safer to use reset with not the same reference.
> 
> The problem is, how do you reset fields such as Select without relying on the ref of the value? As setValue does not accept multiple values anymore, people will rely more the reset method to do such things

can you share a codesandbobx?

RPereiraRedspark:
> can you share a codesandbobx?

sure, https://codesandbox.io/s/sad-dust-vcgry?file=/src/App.js

in this case im simulating a fetch then using reset to populate o Controlled multi select, with setValue it works, with reset it  doesnt

bluebill1049:
> > can you share a codesandbobx?
> 
> sure, https://codesandbox.io/s/react-hook-form-material-ui-select-multiple-vcgry
> 
> in this case im simulating a fetch then using reset to populate a Controlled multi select, with setValue it works, with reset it doesnt

You should compare with the `id` instead of the entire object.

RPereiraRedspark:
the checkbox is only a visual help, the problem is the Select uses reference, updated the checkbox to not use reference check
https://codesandbox.io/s/react-hook-form-material-ui-select-multiple-forked-s6brs

RPereiraRedspark:
as stated in the mui select docs, if i use anythig diferent to a string, it will use reference check
image

PR #3616, opened 2020-12-07
fix NestedValue's Map type

bluebill1049:


src/types/form.ts
```diff
@@ -19,7 +19,7 @@ import { FieldArrayDefaultValues } from './fieldArray';
 declare const $NestedValue: unique symbol;
 
 export type NestedValue<
-  TValue extends unknown[] | Record<string, unknown> =
+  TValue extends unknown[] | Record<string, unknown> | Map<string, any> =
```
kotarella1110, line 22: Do we need to support `Set` as well?

bluebill1049: probably until someone reports another issue?

kotarella1110 (approved):
LGTM 🎉

bluebill1049:
thank you for the review @kotarella1110 👍

src/types/form.ts
amcsi: Also I don't understand the need to restrict the keys of Map to string, when NestedValue means the Map is to be treated as a leaf for the form value object tree. Shouldn't it be `Map<unknown, unknown>` or `Map<any, any>` ?

PR #3685, opened 2020-12-15
fixes #3684

### LINKED ISSUE #3684: deafaultValue={null} is causing pattern rule to trigger an error
**Describe the bug**
If defaultValue={null} is used instead of defaultValue={''} the pattern will cause a form error.

**To Reproduce**
Open codesandbox and press submit to show the difference between defaultValue={null} and defaultValue={''} 

https://codesandbox.io/s/react-hook-form-controller-template-forked-0j2nw?file=/src/index.js

**Expected behavior**
Expected defaultValue={null} to work as defaultValue={''} and not cause a form error. The required rule should cause an error here but not the pattern.

**Screenshots**
image

**Desktop**
- OS: OSX Catalina
- chrome

[EDIT] By the way thanks for the most awesome form lib I have ever used!

andidev:
Fixes deafaultValue={null} is causing pattern rule to trigger an error

it is verified with one test

### DIFF
--- a/cypress/integration/manualRegisterForm.ts
+++ b/cypress/integration/manualRegisterForm.ts
@@ -41,6 +41,6 @@ describe('manual register form validation', () => {
     cy.get('input[name="checkbox"]').check();
 
     cy.get('p').should('have.length', 0);
-    cy.get('#renderCount').contains('25');
+    cy.get('#renderCount').contains('26');
   });
 });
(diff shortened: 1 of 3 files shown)


bluebill1049 (dismissed):
Awesome work! Thanks ❤️

bluebill1049:
leave to me i will sort out the E2E test and what cause the issue 👍

PR #3738, opened 2020-12-20
V7 Register

bluebill1049:
- `register`

We really want to guarantee type-safe with `register` API, the current implementation does have difficulty to achieve this because there is no simple way for typescript to extract name attribute inside input element as `ref`, so we plan to move away from the current way of register native input. here is the proposal:

```diff
- <input ref={register, { required: true }} name="test" />
+ <input {...register('name', { required: true })} />
```

- register name now only support dot syntax.

-----

- `shouldUnregister` 

This option is getting removed, as `register` will no longer unregister input with mutation observer. 

-----

- `trigger`

`trigger` will no longer return validation result (as Promise), instead of subscribing to `errors` formState, each time calling trigger will trigger a re-render as well.

-----

- Improved performance with getting entire form values

-----

- `Controller`

  - `onFoucs` prop is removed and use `ref` from `render` prop.
  - simplify logic and reuse logic within useForm + lose weight.
  - `Controller` wrapped input ref will be stored and forward with the error object. 

-----

- `unregsiter`

**Important:** `unreigster` will no longer set dirty` input`

-----

- `useFieldArray`

optimize internal logic and performance

-----

- `watch`

watch `array` will return `array` instead of `object`

```diff
- watch(['test', 'test1']); // { test: '1', test1: '2' }
+ watch(['test', 'test1']); // ['1', '2']
```

-----

- remove ie11 build

-----

**Issues:** 

- `register(fieldArray${index})` still an issue.
- `control` prop
- nested `useFieldArray` and `name` with variable
- issue with useFormContext
- `SetValueWithSchema` re-render issue with schema
- `@ts-ignore`
- `as any` in the codebase


bluebill1049:
`register(`fieldArray${index}`)` still an issue.

bluebill1049:
@kotarella1110 I am planning to merge this one first, there are still quite a few issues but would like to merge it into the root branch and then can branch off to solve individual API types. Let me know your thoughts.

kotarella1110:
@bluebill1049 Sorry for the late reply! Today I will start working on V7 as well!
I think that the problem should be solved to some extent before merging, since minor and patch upgrades will not be possible.

bluebill1049:
> @bluebill1049 Sorry for the late reply! Today I will start working on V7 as well!
> Today and tomorrow I will be reviewing this PR. Then we'll merge!

Thanks, @kotarella1110 no worries. Hope you had a good break and a happy 2021.

bluebill1049:
> @bluebill1049 Sorry for the late reply! Today I will start working on V7 as well!
> Today and tomorrow I will be reviewing this PR. Then we'll merge!

By the way, I think you will love V7's change, the type is getting so much better with string template literal.

src/useForm.test.tsx
```diff
@@ -1031,36 +1059,61 @@ describe('useForm', () => {
...
+          test: {
+            test: string;
+          }[];
+        }>(),
+      );
 
-      result.current.register('test[0].test');
-      result.current.register('test[1].test');
-      result.current.register('test[2].test');
+      result.current.register('test.0.test');
```
bluebill1049, line 1114: see this @kotarella1110 ❤️

src/useForm.test.tsx
```diff
@@ -83,62 +80,66 @@ describe('useForm', () => {
...
-        ref.remove();
-
-        await waitFor(() => expect(mockListener).toHaveBeenCalled());
-        expect(screen.getByRole('alert').textContent).toBe('false');
-        await wait(() =>
-          expect(renderCount.current.Component).toBeRenderedTimes(2),
-        );
-      },
-    );
+    // test.each([['text'], ['radio'], ['checkbox']])(
```
bluebill1049, line 83: To replace with some other tests later on.

src/useForm.test.tsx
```diff
@@ -2905,21 +2981,6 @@ describe('useForm', () => {
         expect(screen.queryByRole('alert')).toBeInTheDocument();
       });
 
-      it.skip('should output error message when formState.isValid is called in development environment', () => {
```
bluebill1049, line 2908: This required to pass extra `formState` to the function itself, let's leave it to the doc to info the user.

src/useFieldArray.ts
```diff
@@ -280,28 +242,40 @@ export const useFieldArray = <
...
       cleanup(fieldsWithValidationRef.current);
     }
 
     formStateSubjectRef.current.next({
-      isDirty: isFormDirty(name, omitKey(updatedFormValues)),
+      isDirty: isFormDirty(
+        name as InternalFieldName,
+        omitKey(updatedFormValues),
+      ),
+      // @ts-ignore
```
bluebill1049, line 269: known issue.

src/useFieldArray.ts
```diff
@@ -316,37 +290,29 @@ export const useFieldArray = <
 
       formStateSubjectRef.current.next({
         isDirty: true,
+        // @ts-ignore
```
bluebill1049, line 293: known issue.

src/useForm.test.tsx
kotarella1110: amazing! 🎉

kotarella1110 (dismissed):
Awesome! Let's merge this PR once!

src/types/utils.ts
```diff
@@ -65,3 +65,120 @@ export type IsFlatObject<T extends object> = Extract<
...
+type ArrayChild = string | boolean | number | symbol | bigint;
+
+type Pred = [never, 0, 1, 2, 3, 4, 5, 6, 7];
+
+export type FieldPath<
+  TFieldValues,
+  TKey extends keyof TFieldValues = keyof TFieldValues,
+  TMaxRecursive extends number = 7
+> = [TMaxRecursive] extends [0]
+  ? 'RECURSIVE_EXCEED'
```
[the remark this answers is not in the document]
bluebill1049: yes, it's better, but I think this PR is fixing this recursive issue 😭  https://github.com/react-hook-form/react-hook-form/pull/3847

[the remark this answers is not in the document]
bluebill1049: I cried so much over the Xmas break with max amount recursion error from typescript, lol

bluebill1049:
merge into the root branch.

PR #3745, opened 2020-12-21
Add example of persisted input cache after unmount

lfarrel6:
[Re: This Discussion](https://github.com/react-hook-form/react-hook-form/discussions/3470)

Added a simple example for the `useInputCache` hook as discussed above.

Example provides a form with a checkbox which will cause the mount/unmount of a neighbouring form element. The hook will track the value of the checkbox, and use it to either store or dispatch the cached value of the effected element.


bluebill1049:
Thank you @lfarrel6! ❤️

PR #3781, opened 2020-12-25
fix #3232 issue with Controller registration order

### LINKED ISSUE #3232: Incorrect order on focus
**Describe the bug**
When I try step by step fix form errors, I got incorrect focus order

**To Reproduce**

1. Click on submit button
2. fill firstName field
3. Click again on submit button

**Codesandbox link**
https://codesandbox.io/s/react-hook-form-get-started-forked-jdx40?file=/src/index.js

ejose19:
Fixes: #3232

This should address the issue with Controller not behaving like ```register``` regarding order.

### DIFF
--- a/src/controller.test.tsx
+++ b/src/controller.test.tsx
@@ -578,7 +578,7 @@ describe('Controller', () => {
 
       render(<Component />);
 
-      expect(console.warn).toBeCalledTimes(2);
+      expect(console.warn).toBeCalledTimes(3);
 
       process.env.NODE_ENV = env;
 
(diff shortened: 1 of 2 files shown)


ejose19:
@bluebill1049 let me know what you think before I adjust the cypress tests, the only notable difference is that when using ```useFieldArray``` there will be an extra render as we're registering before ```useEffect```, but that shouldn't be an issue.

bluebill1049:
hey @ejose19 we are fixing that in V7 :) This is probably going to break some usage. thanks for the PR tho 👍

PR #3818, opened 2021-01-02
Make value arg optional in useFieldArray's insert, append, and prepend functions

andrewbusch7:
I found that when using `append()` from `useFieldArray`, I couldn't use it without passing in a `value` arg even though I didn't want to set the model. So I've had to do `append{{}}`.

This small change sallow the three insert, append, and prepend functions to default `value` to `{}` if nothing is passed in for `value`.

As a side note, I found that if I passed in `undefined` or `null`, I didn't get a TS error but I did get a runtime error.

If this change is fine, I'll also do a PR to update the docs at https://react-hook-form.com/api/#useFieldArrayRef to show they and `shouldFocus` are optional.


bluebill1049:
hey @andrewbusch7 Thanks for the PR, i think let's postpone this (as leave the PR open). We are working on V7 right now, and try our best to avoid breaking change to the type. We will confirm what's going to be with v7, before merging or closing this PR.

bluebill1049:
One more point, it's designed with a not empty object, and I am not sure what is the side effect of such change, if you have a strong opinion on this, perhaps you want to leave your opinion here: https://github.com/react-hook-form/react-hook-form/discussions

andrewbusch7:
@bluebill1049 ok sure thing, I'll try find an existing discussion there or start a new one. I've reverting the changelog changes for now to avoid conflicts whilst this is on hold.

bluebill1049:
Thanks for understanding @andrewbusch7 🙏

bluebill1049:
Hey, we can not allow empty {}, because it will lead to defaultValue as `undefined` for `Controller`. Thanks for your contribution and patient.

andrewbusch7:
I'll need to step through the code and understand why. Perhaps if it's not valid, it'd be good to add a validation error if {} is passed in rather than have it fail later in certain cases. Feel free to close the PR if you'd prefer.

bluebill1049:
thanks for understanding @andrewbusch7 take a look at this file: https://github.com/react-hook-form/react-hook-form/blob/v7/src/useController.ts which we will need a `defaultValue` for `Controller`, when `append` with `{}` the `defaultValue` for `fieldArray` will become `undefined` which leads to unexpected behaviour.

andrewbusch7:
Hi @bluebill1049, I've tried out using a Controller in my form with `append({})` and it seems to be working fine. From tracing the code, it looks like it'd be the same behaviour as if you didn't define every field in the object, which again I can't find issues with.

Is there any particular unexpected or problematic behaviour that you expect to arise?

It is very convenient not having to define every field for the default values.

bluebill1049:
hey @andidev from memory controlled component with checkbox or any controlled component doesn't support `undefined` as default value.

PR #3847, opened 2021-01-07
V7 - useFieldArray type issue

bluebill1049:
Huge thanks go to https://github.com/phpnode 's help on type definition. 

- fix useFieldArray with register index with variable method
- improve watch API type
- improve getValues API type and functionality 
- getValues() will return and shallow merge with defaultValues

### DIFF
--- a/src/types/utils.ts
+++ b/src/types/utils.ts
@@ -219,4 +219,6 @@ export type FieldPathValue<
     : never
   : TPath extends keyof TFieldValues
   ? TFieldValues[TPath]
+  : TPath & number extends keyof TFieldValues
+  ? TFieldValues[TPath & number]
   : never;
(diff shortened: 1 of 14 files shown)


src/types/utils.ts
```diff
@@ -66,119 +67,156 @@ export type IsFlatObject<T extends object> = Extract<
...
+    : ArrayKey | `${ArrayKey}.${FieldPath<E>}`
+  : ValueOf<
+      {
+        [K in keyof Root]: IsPrimitive<Root[K]> extends true
+          ? K & string
+          : (K & string) | `${K & string}.${FieldPath<Root[K]>}`;
+      }
+    >;
+
+export type FieldPathValue<
```
bluebill1049, line 209: issue with `FieldPathValue` not getting the correct type

bluebill1049:
nested field array still a problem.

src/types/utils.ts
kotarella1110: @bluebill1049 I will fix this 👍

bluebill1049: I think I got it working last night and just committed now, please double check.

bluebill1049: nvm, it's not working...
Screen Shot 2021-01-08 at 12 38 15 pm

bluebill1049:
let's merge this in and the `register` branch then, we can branch off from the main branch.

Issue #2155, opened 2020-07-10
[useFieldArray] customize autoFocus index 


xiaoyu-tamu:
**Is your feature request related to a problem? Please describe.**

It would be nice developer can control  auto focus using `field name` or `field index`.

The use case is when the first field has default value, we may want to focus second/third input instead.

**Describe the solution you'd like**

`fieldArray.append(data, boolean | number | string )`

If second argument is boolean, same as before, focus first field

If second argument is number, focus field by index;
If second argument is string, focus field by name;  Maybe?

**Describe alternatives you've considered**

**Additional context**

https://github.com/react-hook-form/react-hook-form/blob/master/src/useFieldArray.ts#L436-L447

PR #3858, opened 2021-01-09
V7 - useFieldArray improvement

bluebill1049:
- [x] nested field array cypress test
- [x] nested field array and array watch unit tests
- [x]  nested field array and `useWatch` integration tests
- [x] should drity, focusIndex: https://github.com/react-hook-form/react-hook-form/issues/2155
- [x] state watch with Subject
- [x]  unit tests for `focusIndex`
- [x] improve internal structure


src/types/form.ts
```diff
@@ -74,9 +70,7 @@ export type SetValueConfig = Partial<{
   shouldDirty: boolean;
 }>;
 
-export type HandleChange = (
-  event: Event | NativeSyntheticEvent<any> | ControllerEvent,
-) => Promise<void | boolean>;
+export type HandleChange = (event: any) => Promise<void | boolean>;
```
bluebill1049, line 73: avoid the need to import `react native` while working with `react`

PR #3874, opened 2021-01-12
make DeepMap type support generic

ellemedit:
`DeepMap` related types didn't work with generic. [you can see here](https://codesandbox.io/s/cocky-newton-8wyd5?file=/src/index.tsx)

I solved this with [some little hack](https://bit.ly/35x2FK4).
With this approach, TypeScript cannot infer `{}` as DeepMap type.
I don't know there's better way to solve this.

But I think this is required for dividing interface and separation of concerns.
I don't want my component to know all dependency. (let me know if I divide interface without generic in component)

### DIFF
--- a/src/logic/focusOnErrorField.ts
+++ b/src/logic/focusOnErrorField.ts
@@ -1,8 +1,8 @@
 import get from '../utils/get';
 import isUndefined from '../utils/isUndefined';
-import { FieldErrors, FieldRefs } from '../types';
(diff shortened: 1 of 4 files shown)


Beingbook:
Oops, there is more type errors. let me fix.

bluebill1049:
Thanks for the PR, if you would like to improve TS for us take a look at our v7 branch.

Beingbook:
I'll checkout v7 branch before continuing this work

PR #3881, opened 2021-01-12
V7 - Resolver add field being validated

jorisre:
- Change `isValidateAllFieldCriteria ` to `{ criteriaMode }`
- Add current field being validated


bluebill1049 (dismissed):
LGTM!

bluebill1049:
looks like e2e failed. probably something minor.

jorisre:
> LGTM!

I don't have finished ;) I have to add the field being validated.
`{ criteriaMode, field }`
________________^

src/types/resolvers.ts
```diff
@@ -17,11 +17,16 @@ export type ResolverResult<TFieldValues extends FieldValues = FieldValues> =
   | ResolverSuccess<TFieldValues>
   | ResolverError<TFieldValues>;
 
+export interface ResolverOptions {
+  criteriaMode?: CriteriaMode;
+  fields?: Field['__field'][];
```
bluebill1049, line 22: instead of an array of `fields`, maybe we can just pass the entire `fieldsRef`?

```
async (data, context, { fields }) => {
  // when generate the errors object in resolver, you can look up the fields object and include the `refs`
}
```

src/useForm.ts
```diff
@@ -348,11 +348,25 @@ export function useForm<
...
 
   const executeSchemaOrResolverValidation = React.useCallback(
-    async (names: InternalFieldName[]) => {
+    async (
+      names: InternalFieldName[],
+      currentNames: FieldName<TFieldValues>[] = [],
+    ) => {
+      const currentFields: Field['__field'][] = [];
+      for (const name of currentNames) {
+        const field = fieldsRef.current[name];
```
bluebill1049, line 357: This is probably going to fail with deeply nested field input, perhaps the above suggestion is the simpler solution and leaves the logic at `resolver`.

src/types/resolvers.ts
bluebill1049: oops, sorry I think I have mixed up with two things..

`{ fields, currentField, criteriaMode }`

which you got is right with `currentFields` 👍

src/useForm.ts
jorisre: Right ! I'll use `get` thanks ! :)

cypress/integration/basicSchemaValidation.ts
```diff
@@ -42,7 +42,7 @@ describe('basicSchemaValidation form validation', () => {
     cy.get('input[name="checkbox"]').check();
 
     cy.get('p').should('have.length', 0);
-    cy.get('#renderCount').contains('24');
+    cy.get('#renderCount').contains('25');
```
jorisre, line 45: @bluebill1049 I'm not sure for this test update

bluebill1049: wouldn't worry too much if only increase by 1 ❤️

src/useForm.ts
```diff
@@ -587,7 +606,7 @@ export function useForm<
           const { errors } = await resolverRef.current(
             getValues(),
             contextRef.current,
-            isValidateAllFieldCriteria,
+            { criteriaMode, fields: [field.__field] },
```
bluebill1049, line 609: what about?

```
fields: {
  [name]: field.__filed
}
```

src/useForm.ts
```diff
@@ -348,11 +348,25 @@ export function useForm<
   );
 
   const executeSchemaOrResolverValidation = React.useCallback(
-    async (names: InternalFieldName[]) => {
+    async (
+      names: InternalFieldName[],
+      currentNames: FieldName<TFieldValues>[] = [],
+    ) => {
+      const currentFields: Field['__field'][] = [];
```
bluebill1049, line 355: I think this works for deep nested: 

```
let currentFields = {};

for (name in fieldsNamesRef.current) {
  const field = get(fieldsRef.current, name);
  
  field && currentFields[name] =  field.__field;
}
```

jorisre:
@bluebill1049 Ready to review again ! thanks :)

bluebill1049:
Exiciting!

scripts/jest/jest.config.js
```diff
@@ -3,6 +3,9 @@ const { defaults: tsjPresets } = require('ts-jest/presets');
 const jestPresets = require('@testing-library/react-native/jest-preset');
 
 const common = {
+  clearMocks: true,
+  resetMocks: true,
+  restoreMocks: true,
```
bluebill1049, line 8: holy shit! I have been writing those before after all those years, term out there is a bloody config... thank you @jorisre 🙏

src/logic/getFields.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  for (const name in fieldsNames) {
+    const field = get(fieldsRefs, name) as Field | undefined;
+
+    if (field) {
+      !isKey(name)
+        ? set(currentFields, name, field.__field)
+        : (currentFields[name] = field.__field);
+    }
+  }
```
bluebill1049, line 20:
```suggestion
    (field.__field) && set(currentFields, name, field.__field);
```

save some bytes? I haven't run the test with this.

bluebill1049 (approved):
Amazing work! worse to test out that script I got above, see if it's the same result, and save some bytes. Thanks, @jorisre for the awesome input.

src/logic/getFields.ts
bluebill1049: sorry, I think perhaps we don't need set?

jorisre: I don't know how to save more bytes, here is the output:

```ts
function oe(e, r) {
  const t = {};
  for (const n in e) {
    const e = u(r, n);
    e && (a(n) ? (t[n] = e.__field) : o(t, n, e.__field));
  }
  return t;
}
```

we need `set` for nested fields.

You can merge whether it's good ;)

bluebill1049: let's get it in. ❤️

PR #3911, opened 2021-01-15
V7 - watch & getValues

bluebill1049:
- [x] `watch` support callback api
- [x] `getValues` return default values before `render`


304s (commented):
Nice

PR #3920, opened 2021-01-17
improve type for nested field array

bluebill1049:
"try to" fix nested field array type with useFieldArray

don't think there is any good solution for ts right now, the user will have to cast for nested field array.

going to move forward with this...


src/types/utils.ts
```diff
@@ -216,6 +216,14 @@ export type FieldPathValue<
       : TFieldValues[Key] extends (infer U)[]
       ? U
       : never
+    : Key extends Digits
```
bluebill1049, line 219: issue with 

```
type FieldType = FieldPathValue<FormValues, `test.${number}.keyValue`>;
```

bluebill1049:
```
const { fields } = useFieldArray<FormValues, 'test.0.keyValue'>({ // give correct type `fields`
  name: `test.${index}.keyValue` as const, // type error here Type '`test.${number}.keyValue`' is not assignable to type '"test.0.keyValue"'.ts(2322)
  control,
});
```

src/__tests__/useFieldArray.test.tsx
```diff
@@ -552,15 +551,24 @@ describe('useFieldArray', () => {
...
         index,
         control,
       }: {
-        control: Control;
+        control: Control<FormValues>;
         index: number;
       }) => {
-        const { fields } = useFieldArray({
-          name: `test.${index}.nestedArray`,
+        const { fields } = useFieldArray<FormValues>({
```
bluebill1049, line 570: second generic will default to `FieldPath`, which results type at the root of the object, instead of the nested field array.

bluebill1049:
only solution found so far: 

```
        const { fields } = useFieldArray<FormValues, 'test.0.keyValue'>({
          name: `test.${index}.keyValue` as 'test.0.keyValue',
          control,
        });
```

src/__tests__/useFieldArray.test.tsx
```diff
@@ -781,8 +781,8 @@ describe('useFieldArray', () => {
         control: Control<FormValues>;
         index: number;
       }) {
-        const { fields } = useFieldArray<FormValues>({
-          name: `test.${index}.keyValue` as const,
+        const { fields } = useFieldArray<FormValues, 'test.0.keyValue'>({
```
bluebill1049, line 784: hack: had to use a cast to resolve the type error...

src/__tests__/useFieldArray.test.tsx
```diff
@@ -781,7 +781,7 @@ describe('useFieldArray', () => {
         control: Control<FormValues>;
         index: number;
       }) {
-        const { fields } = useFieldArray<FormValues, 'test.0.keyValue'>({
+        const { fields } = useFieldArray({
           name: `test.${index}.keyValue` as 'test.0.keyValue',
```
bluebill1049, line 785: This annoying cast seems to be required.

PR #3955, opened 2021-01-19
Fix compatibility with older web brows

GustavoEklund:


src/utils/onDomRemove.ts
```diff
@@ -10,7 +10,10 @@ export default function onDomRemove<TFieldValues>(
   ) => void,
 ): MutationObserver {
   const observer = new MutationObserver((): void => {
-    for (const field of Object.values(fieldsRef.current)) {
+    const fields = Object.keys(fieldsRef.current).map(
+      (value) => fieldsRef.current[value],
+    );
+    for (const field of fields) {
```
bluebill1049, line 16: i think we just need to do a `isObject` check before the `for` loop

GustavoEklund: it seems to be a good solution, but looking at `Object.values` browser compatibility [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Object/values#browser_compatibility), it says that supports since `Chrome 54`, on the issue I was running on a `Chrome 51`. (that's why it gives the error `Object.values is not a function`). I think it would be a better idea add an `isObject` before `Object.keys` then

bluebill1049: oh, so you should use ie build. it's not a library's issue.

bluebill1049: or add a polyfill to solve this problem at your app.

GustavoEklund: you are totally right, I have just installed `core-js` and imported it before React, and now I think it may be about 100% compatible with older Browsers, I didn't know about polyfills, thank you

bluebill1049: no worries! thanks for spending us a PR as well. 🙏

PR #3988, opened 2021-01-22
🐞 fix #3985 radio input default state to undefined instead of empty string

### LINKED ISSUE #3985: problem with zod validation of nativeEnum with optional
Hi, pls help me
i cant validate my schema with zod using z.nativeEnum(Enum).optional()

example:

https://codesandbox.io/s/react-hook-form-nestedvalue-forked-6gh4n?file=/src/App.tsx

optional field is invalid how i can fix it?
image

workaround - but it looks strange 
image

bluebill1049:
🐞 fix #3985 radio input default state to undefined instead of empty string

### DIFF
--- a/app/src/formState.tsx
+++ b/app/src/formState.tsx
@@ -40,7 +40,7 @@ const FormState = (props: {
       lastName: '',
       select: '',
       checkbox: false,
-      radio: '',
+      radio: undefined,
       'checkbox-checked': true,
     },
   });
--- a/app/src/formState.tsx
+++ b/app/src/formState.tsx
@@ -30,7 +30,7 @@ const FormState = (props: {
     firstName: string;
     lastName: string;
     select: string;
-    radio: string;
+    radio: string | null;
     checkbox: boolean;
     ['checkbox-checked']: boolean;
   }>({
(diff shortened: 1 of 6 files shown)


app/src/formState.tsx
```diff
@@ -40,7 +40,7 @@ const FormState = (props: {
       lastName: '',
       select: '',
       checkbox: false,
-      radio: '',
+      radio: undefined,
```
kerosan, line 43: @bluebill1049  Hi
can you please check if this bug reproduced for input type="hidden" ? out dropdown based on hidden input and have same problem 
thank you = )

bluebill1049: what do you mean @kerosan? do you have a codesandbox?

kerosan: sure 
i updated previous example just added Dropdown
https://codesandbox.io/s/react-hook-form-nestedvalue-forked-6gh4n?file=/src/App.tsx

bluebill1049: for input type hidden will stay at default empty `string`, because that's what it returns when you invoke `e.target.value`.

bluebill1049:
actually, I think it should be `null` instead of `undefined` what do you think @jorisre? because the input is defined, but there is no selection, undefined sounds to me, the whole input doesn't exists.

jorisre:
> actually, I think it should be `null` instead of `undefined` what do you think @jorisre? because the input is defined, but there is no selection, undefined sounds to me, the whole input doesn't exists.

Sounds good to me

jorisre (approved):

PR #4017, opened 2021-01-25
improve perf with resolver

bluebill1049:
I think it's too expensive to call `getFields` onChange, by a loop through the names of the entire fields. This PR I only keep it to the current name or names.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -580,7 +580,7 @@ export function useForm<
             contextRef.current,
             {
               criteriaMode,
-              fields: getFields(fieldsNamesRef.current, fieldsRef.current),
+              fields: getFields([name], fieldsRef.current),
               names: [name as FieldName<TFieldValues>],
             },
           );
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -342,15 +342,18 @@ export function useForm<
   const executeSchemaOrResolverValidation = React.useCallback(
     async (
       names: InternalFieldName[],
-      currentNames: FieldName<TFieldValues>[] = [],
+      currentNames?: FieldName<TFieldValues>[],
     ) => {
       const { errors } = await resolverRef.current!(
         getValues(),
         contextRef.current,
         {
           criteriaMode,
-          names: currentNames,
-          fields: getFields(fieldsNamesRef.current, fieldsRef.current),
+          names: currentNames || [],
+          fields: getFields(
+            currentNames || fieldsNamesRef.current,
+            fieldsRef.current,
+          ),
         },
       );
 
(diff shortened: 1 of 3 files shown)


jorisre:
Right! I think, it should be the same behavior for `updateIsValid` and `submit` otherwise it should be well documented.

- `onChange`: fields = only current fields
- `onSumit`: fields = all form's fields
- `trigger`: fields = all form's fields

bluebill1049:
> Right! I think, it should be the same behavior for `updateIsValid` and `submit` otherwise it should be well documented.
> 
> * `onChange`: fields = only current fields
> * `onSumit`: fields = all form's fields
> * `trigger`: fields = all form's fields

sounds good to me! also, optimize `trigger` as well. 

name[] or all form fields.

src/useForm.ts
```diff
@@ -342,15 +342,18 @@ export function useForm<
...
       const { errors } = await resolverRef.current!(
         getValues(),
         contextRef.current,
         {
           criteriaMode,
-          names: currentNames,
-          fields: getFields(fieldsNamesRef.current, fieldsRef.current),
+          names: currentNames || [],
+          fields: getFields(
+            currentNames || fieldsNamesRef.current,
```
jorisre, line 354: I think it's confusing for the user because sometimes he will get all either fields or only the fields being validated.
I think it should be more clear, it should be one or the other but not both. Do you see what I mean ?

I really like the idea to have all fields, it’s more powerful but I understand it’s expensive to call `getFields` so the only one solution we have is to pass flatten fields as we discussed.

Example:

```ts
const myResolver = async (values, _context, { names, fields }) => {
   // names are: `my-name` or `my.nested.0.name`
   // fields could be: `{ 'my-name': Field, `my.nested.0.name`: Field }`
   const field = fields[names[0]]

    // ... 🪄
}
```

And finally `getFields` _(I haven't tested)_:
```ts
getFields(
  fieldsNames: Set<InternalFieldName> | InternalFieldName[],
  fieldsRefs: FieldRefs,
) {
  const currentFields: Record<InternalFieldName, Field['_f']> = {};

  for (const name of fieldsNames) {
    const field = get(fieldsRefs, name) as Field | undefined;

    if (field) {
      currentFields[name] = field._f;
    }
  }

  return currentFields;
}
```

Do you think It's still too expensive to call `getFields` ?

jorisre: A new idea came to me, to improve perf, we can do some caching based on the current names. That's just an idea

bluebill1049: happy to revert back to the trigger part. I see your points there, but as fields size grows, it's going to be noticed on slow devices with each onChange doing that for a loop. I think it's essential to find a good balance between DX and UX.

bluebill1049:
ready to be reviewed @jorisre 👍 reverted the trigger change.

jorisre (approved):

PR #4148, opened 2021-02-10
exclude submission data with disabled

bluebill1049:
disabled input will be excluded from submission data.  I think it's made sense to follow the HTML standard for input disabled. https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#enabling-and-disabling-form-controls:-the-disabled-attribute

### DIFF
--- a/src/__tests__/logic/getFieldsValues.test.ts
+++ b/src/__tests__/logic/getFieldsValues.test.ts
@@ -289,6 +289,6 @@ describe('getFieldsValues', () => {
           },
         },
       ),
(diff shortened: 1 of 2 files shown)


xiaoyu-tamu:
Is this enabled by default or via flag? 

Disabled inputs could be useful when the value is readonly.

bluebill1049:
@xiaoyu-tamu still deciding... https://github.com/react-hook-form/react-hook-form/discussions/3714#discussioncomment-364145 let me know your thoughts.

PR #4150, opened 2021-02-10
Fixing issue of unmounted component trying to set state after reset for field arrays

tonycoco:
Fixes an issue where a component that contains a fieldArrray is unmounted and the `useEffect` hook is run, then tries to set state (via `setFields`), and creates an unmounted set state error.


bluebill1049:
Thanks for the PR, do you have reproducible CSB for this issue?

tonycoco:
I do not.

Issue #4219, opened 2021-02-17
formState.isValid behaviour changed between version 6.11.3 and 6.11.4 when using resolver.


micnil:
**Describe the bug**
In version 6.11.3 and before, the formState.isValid was true when the forms defaultValues are valid but not rendered in the DOM.
In version 6.11.4 and after, the formState.isValid is initially set to true, but is immediately set to false after the resolver has run, even if the defaultValues are valid.

**To Reproduce**

1. Go to https://codesandbox.io/s/rhf-incorrect-isvalid-wluyk?file=/src/App.tsx
2. Open console and see that: isValid is initially true, but later set to false after resolver has been run.
3. Click "Toggle form" and see that the Confirm button is disabled (and the isValid flag continues to be false)

**Codesandbox link (Required)**
https://codesandbox.io/s/rhf-incorrect-isvalid-wluyk?file=/src/App.tsx

**Expected behavior**
I expect the isValid flag to be true when clicking "Toggle form" to display the form.
Expected behavior can be reproduced by opening the above codesandbox (https://codesandbox.io/s/rhf-incorrect-isvalid-wluyk?file=/src/App.tsx) and changing the react-hook-form version to 6.11.3

**Desktop (please complete the following information):**

- OS: MacOS 10.15.7
- Browser Chrome  88.0.4324.146 
- Version  6.11.4 and later

leapful:
@micnil @bluebill1049 

There was a change on useForm.tsx that causes this unexpected behavior at commit: 

https://github.com/react-hook-form/react-hook-form/commit/9ccec798e7a976871ef3d8b9a742e533ed97cc40#diff-07a36f93b9f2954db217ab021a0444c89c5fabc9506b8c390db329e5e74b9f06L714-L720

where the "defaultValuesRef" is removed on values to be validated by resolver.

@bluebill1049 Could you please check again why we need to remove "defaultValuesRef" ?

bluebill1049:
Thanks for the investigation! @leapful you want to send a PR (since you find it out) to revert that code? I think I was trying to trim some extra logic.

leapful:
@micnil @bluebill1049 

PR is created at https://github.com/react-hook-form/react-hook-form/pull/4229

leapful:
@bluebill1049

There is one failed automation test for the use case show/hide required input. 

When a required input is hidden, the test case expected formState.isValid is false but now because "defaultValues" is added on validation resolver, formState.isValid is now true.

[...]

PR #4229, opened 2021-02-18
#4219 Add default values when validating by using resolver

leapful:
- Add default values to resolver validation in order to solve issue formState.isValid is set to false even default values exist.
- Write unit test for the specific case.

Related issue: https://github.com/react-hook-form/react-hook-form/issues/4219

### DIFF
--- a/src/useForm.test.tsx
+++ b/src/useForm.test.tsx
@@ -3323,6 +3323,62 @@ describe('useForm', () => {
 
       expect(resolverData).toEqual({ test: 'value' });
     });
+
+    it('should have formState.isValid equals true with defined default values after executing resolver', async () => {
+      const Toggle = () => {
+        const [toggle, setToggle] = React.useState(false);
+
+        const { register, formState } = useForm({
+          defaultValues: { test: 'Test' },
+          mode: 'onChange',
+          resolver: async (values) => {
+            if (!values.test) {
+              const result = {
+                values: {},
+                errors: {
+                  test: {
+                    type: 'required',
+                  },
+                },
+              };
+              return result;
+            }
+
+            return {
+              values,
+              errors: {},
+            };
+          },
+        });
+
+        return (
+          <>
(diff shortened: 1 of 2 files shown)


micnil:
A Cypress test failed:

> isValid
>        should showing valid correctly with schema validation and default value supplied:
>      AssertionError: Timed out retrying: Expected to find content: 'false' within the element: <div#isValid> but never did.
>       at Context.eval (http://localhost:3000/__cypress/tests?p=cypress/integration/isValid.ts:146:28)

https://github.com/react-hook-form/react-hook-form/blob/master/cypress/integration/isValid.ts
```
  it('should showing valid correctly with schema validation and default value supplied', () => {
    cy.visit('http://localhost:3000/isValid/schema/defaultValues');
    cy.get('#isValid').contains('true');

    cy.get('input[name="firstName"]').clear();
    cy.get('#isValid').contains('false');
    cy.get('#renderCount').contains('2');
    cy.get('input[name="firstName"]').type('test');
    cy.get('#isValid').contains('true');
    cy.get('#isValid').contains('true');
    cy.get('#toggle').click();
    cy.get('#isValid').contains('false');
    cy.get('#toggle').click();
    cy.get('input[name="firstName"]').type('t');
    cy.get('#isValid').contains('true');
  });
```
This is the test page: https://github.com/react-hook-form/react-hook-form/blob/master/app/src/isValid.tsx

Not sure exactly which assertion failed, but spontaneously I think that second last line `type('t')` shouldn't be necessary to get `isValid==true`. Toggling a valid fields visibility off and on should make the form valid without having to modify the input no?

bluebill1049:
Thansk @leapful i will help out on this one tmr for the cypress test.

leapful:
@micnil @bluebill1049

I've updated the handling in validateResolver to support both cases show/hide the whole form and only show/hide some inputs inside the form.

Could you please review it ?

bluebill1049:
Awesome! thanks for the awesome contribution! ❤️

PR #4296, opened 2021-02-26
Make isWeb work better on react native

cranberyxl:
# ~blocked on #4297 being merged~

One of the libraries we're using is creating document, so it's not undefined. This adds a more reliable way to check based on https://github.com/facebook/react-native/blob/6e6443afd04a847ef23fb6254a84e48c70b45896/Libraries/Core/setUpNavigator.js#L15

The consequences of this inaccurately reporting is that `useForm` tries to use `MutationObserver` which does not exists outside of the DOM

### DIFF
--- a/src/utils/isWeb.test.ts
+++ b/src/utils/isWeb.test.ts
@@ -16,6 +16,7 @@ describe('isWeb', () => {
   it('should return true when in a browser', () => {
     expect(isWeb()).toBeTruthy();
   });
+  
   it('should return false when navigator product is ReactNative', () => {
     productGetter.mockReturnValue('ReactNative');
     expect(isWeb()).toBeFalsy();
--- a/src/utils/isWeb.test.ts
+++ b/src/utils/isWeb.test.ts
@@ -1,0 +1,23 @@
+const isWeb = () => {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const isWeb = require('./isWeb');
+
+  return isWeb.default;
+};
+
+describe('isWeb', () => {
+  let productGetter: jest.SpyInstance;
+
+  beforeEach(() => {
+    jest.resetModules();
+    productGetter = jest.spyOn(window.navigator, 'product', 'get');
+  });
+
+  it('should return true when in a browser', () => {
+    expect(isWeb()).toBeTruthy();
+  });
+  it('should return false when navigator product is ReactNative', () => {
+    productGetter.mockReturnValue('ReactNative');
+    expect(isWeb()).toBeFalsy();
+  });
+});
--- a/src/utils/isWeb.test.ts
+++ b/src/utils/isWeb.test.ts
@@ -16,7 +16,7 @@ describe('isWeb', () => {
   it('should return true when in a browser', () => {
     expect(isWeb()).toBeTruthy();
   });
-  
+
   it('should return false when navigator product is ReactNative', () => {
     productGetter.mockReturnValue('ReactNative');
     expect(isWeb()).toBeFalsy();
--- a/src/utils/isWeb.ts
+++ b/src/utils/isWeb.ts
@@ -1,5 +1,5 @@
 import { UNDEFINED } from '../constants';
 
-export default typeof window !== UNDEFINED &&
-  typeof document !== UNDEFINED &&
-  navigator.product !== 'ReactNative';
+export default typeof window !== UNDEFINED && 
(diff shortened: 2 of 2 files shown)


cranberyxl:
Noticed that the build on circle is using npm instead of yarn and failing. I'm guessing it's due to no package-lock. Giving yarn a go to see if it passes. If so I'll pull that into a separate PR

cranberyxl:
@bluebill1049 I think I rebased over your change, apologies

bluebill1049:
no worries @cranberyxl thanks for the PR.

bluebill1049:
> One of the libraries we're using is creating document, so it's not undefined. 

can you share that library? ideally, I prefer not to have to check that string `ReactNative`.

cranberyxl:
We haven't identified which library does it. But any library could globally set this variable. Navigator.product is a reliable way to detect this. Does it also need to work in a node ssr context?

cranberyxl:
Isolated the problem to this library: https://github.com/cometchat-pro/react-native-chat-sdk/issues/26 

I've figured out a workaround: if we require `react-hook-form` first then the `isWeb` value is evaluated.

bluebill1049:
Thanks very much for the investigation @cranberyxl hopefully react-native-chat could resolve this issue so we can skip this string check 🙏 really appreciate your looking into this and sending the solution as well. let's wait a bit to see if they will respond.

s97712:
Can MutationObserver and File be detected directly? I am trying to run on [Mini Program](https://developers.weixin.qq.com/miniprogram/en/dev/framework/) , that has window and document but dose not has MutationObserver

bluebill1049:
@s97712 MutationObserver is removed in v7.

s97712:
What about ` window !== UNDEFINED && window.HTMLElement !== UNDEFINED && typeof document !== UNDEFINED;
`?

bluebill1049:
looks better @s97712, not sure if it solves the problem. @cranberyxl could you verify?

s97712:
> What about `window !== UNDEFINED && window.HTMLElement !== UNDEFINED && typeof document !== UNDEFINED;`?

Are any update? That  solves my problem, Whether to consider to merge?

bluebill1049:
@s97712 what's the problem that you are having? and affected by which lib?

s97712:
> @s97712 what's the problem that you are having? and affected by which lib?

I need to use react-hook-form on [Mini Program](https://developers.weixin.qq.com/miniprogram/en/dev/framework/). It like web and has document, but has no dom

bluebill1049:
@s97712 I will release a beta, so you can test it out and let me know if it's working or not.

bluebill1049:
Screen Shot 2021-03-16 at 8 27 25 pm
doesn't work @s97712

bluebill1049:
is there any place I can test with "mini" program?

s97712:
> Screen Shot 2021-03-16 at 8 27 25 pm
> 
> doesn't work @s97712

sorry, I forget `typeof`

`typeof window !== UNDEFINED && typeof window.HTMLElement !== UNDEFINED && typeof document !== UNDEFINED;`

bluebill1049:
let me know `6.15.5-beta.0`

s97712:
> let me know `6.15.5-beta.0`

`control` can work, but `register` can not work, that enough for me.

bluebill1049:
what do you mean `control` not working?

s97712:
> what do you mean `control` not working?
```
      <Controller 
        control={form.control}
        ...
      >
```

bluebill1049 (commented):
> > what do you mean `control` not working?
> 
> ```
>       <Controller 
>         control={form.control}
>         ...
>       >
> ```

what's the error?

s97712:
> > > what do you mean `control` not working?
> > 
> > 
> > ```
> >       <Controller 
> >         control={form.control}
> >         ...
> >       >
> > ```
> 
> what's the error?

That no error.
`register` does not work because the mini program change event does not have a `target.value` .

bluebill1049:
> > > > what do you mean `control` not working?
> > > 
> > > 
> > > ```
> > >       <Controller 
> > >         control={form.control}
> > >         ...
> > >       >
> > > ```
> > 
> > 
> > what's the error?
> 
> That no error.
> `register` does not work because the mini program change event does not have a `target.value` .

I will apply this change in v7 root branch, and keep PR as it is.

PR #4297, opened 2021-02-26
Use yarn for circleci build

cranberyxl:
I had a failing build on #4296 . It looks like the lack of `package-lock.json` is causing the problem. Switching to yarn fixes things.


bluebill1049:
LGTM! Thank you.

PR #4375, opened 2021-03-07
Fix 4368

bluebill1049:


justjoeyuk:
ie11 support got shot on this PR?

PR #4389, opened 2021-03-09
Fixing Field Array's remove function

fermion17:
Since JavaScript's sort function sorts elements alphabetically if compareFunction is omitted, we need to give it by hand. This leads to wrong result when Field Array's length is larger then 10 and `remove()` was given array containing indexes larger than 10.

This PR fixes it by giving compareFunction explicitly.

cf. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#parameters


bluebill1049:
thanks, @fermion17 ❤️

PR #4394, opened 2021-03-09
[V7] Correct option "keepValue" and "keepError" for method "unregister"

leapful:
- Switch option flag "keepError" and "keepValue" to correct handling for method "unregister".


bluebill1049:
thanks @leapful

PR #4402, opened 2021-03-10
Fix issue with setInternalValue() for array values

### LINKED ISSUE #4390: Issue with setInternalValue() for array values
There is a bug in`setInternalValue()` in `useForm.ts` if you pass an Array as `value`:

https://github.com/react-hook-form/react-hook-form/blob/c18ceee1975651c14e1d33fc420f5eece0824f1f/src/useForm.ts#L465

```
 !shouldUnregister &&
        !isPrimitive(value) &&
        set(shallowFieldsStateRef.current, name, { ...value });
```
If `value` is an array, `isPrimitive` return false so `set(/*...*/)` gets executed, which converts the array into an 'array object', like `["test"]` => `{0: "test"}`.

So the code block needs an additional check for an array like 
`set(shallowFieldsStateRef.current, name, Array.isArray(value) ? [...value] : { ...value });`

bernie43:
Fixes #4390

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -462,7 +462,7 @@ export function useForm<
     ) => {
       !shouldUnregister &&
         !isPrimitive(value) &&
-        set(shallowFieldsStateRef.current, name, { ...value });
+        set(shallowFieldsStateRef.current, name, Array.isArray(value) ? [...value] : { ...value });
 
       if (fieldsRef.current[name]) {
         setFieldValue(name, value);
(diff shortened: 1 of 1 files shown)


src/useForm.ts
```diff
@@ -462,7 +462,7 @@ export function useForm<
     ) => {
       !shouldUnregister &&
         !isPrimitive(value) &&
-        set(shallowFieldsStateRef.current, name, { ...value });
+        set(shallowFieldsStateRef.current, name, Array.isArray(value) ? [...value] : { ...value });
```
bluebill1049, line 465:
```suggestion
        set(
          shallowFieldsStateRef.current,
          name,
          Array.isArray(value) ? [...value] : { ...value },
        );
```

bluebill1049:
thank you @bernhardc 👍

PR #4516, opened 2021-03-25
improve useFieldArray type to fix ts recursion error

### LINKED ISSUE #4499: V7 Typescript issue - RC 7.0.0-rc.7 - useFieldArray append issue
**Describe the bug**
After upgrading to 7.0.0-rc.7, I'm noticing a typescript error with `useFieldArray's` `append` method when using a generic type for our form values. My form values are valid and have not changed between versions. The error is:

```
Argument of type '{ name: string; }' is not assignable to parameter of type 'never[]'.
  Object literal may only specify known properties, and 'name' does not exist in type 'never[]'
```

It seems related to having nested objects present in the form values type along with a field array. If you remove some of the other objects, the error should go away. 

**To Reproduce**

1. Go to the sandbox
2. View the error the append method

https://codesandbox.io/s/interesting-edison-sbbpy?file=/src/App.tsx

**Expected behavior**
There should be no error as my form values are a valid type

**Screenshots**
Screenshot from 2021-03-23 17-16-16

kotarella1110:
## Issue

#4499 

## Ref

- [CodeSandbox](https://codesandbox.io/s/tender-wozniak-kn51c?file=/src/App.tsx)

### DIFF
--- a/src/types/fieldArray.ts
+++ b/src/types/fieldArray.ts
@@ -16,19 +16,20 @@ export type UseFieldArrayProps<
   control?: Control<TFieldValues>;
 };
 
-type InferArrayType<T> = T extends (infer U)[] ? U : never;
-
 export type FieldArrayWithId<
   TFieldValues extends FieldValues = FieldValues,
   TFieldArrayName extends FieldArrayPath<TFieldValues> = FieldArrayPath<TFieldValues>,
   TKeyName extends string = 'id'
-> = InferArrayType<FieldArrayPathValue<TFieldValues, TFieldArrayName>> &
-  Record<TKeyName, string>;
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks @kotarella1110 🥃

wdfinch:
@bluebill1049 when do you think the next rc will be out? I couldn't use 7.0.0-rc.7 because of this and 7.0.0-rc.6 has a crash associated with appending a new field array element.

bluebill1049:
> Thanks for pr and merge! @bluebill1049 when do you think the next rc will be out? I couldn't use 7.0.0-rc.7 because of this and 7.0.0-rc.6 has a crash associated with appending a new field array element.

i will release another RC today.

PR #4606, opened 2021-04-03
update gitignore to include yarn 2 generated files

wdfinch:
This pr updates `gitignore` to include files generated when using yarn 2 as the package manager.


bluebill1049:
thanks @wdfinch

PR #4607, opened 2021-04-03
Polish README

pmaier983:
Bug fixes:
 - v6 translation switching. (previously unable to go from one translation to another)

Improvements:
 - Polished some grammar
 - I think the example should be in camel case just to show best practice, but this is also debatable.


bluebill1049:
for one second i thought a Polish 🇵🇱 readme lol.

Issue #4603, opened 2021-04-03
Chakra checkbox returns "" instead of true when is checked


violabg:
I don't know if is a react-hook-form or a chakra-ui bug, but after updating  to  react-hook-form version 7, I'm getting "" instead of true for checkbox value when selected. It only happens on Chakra checkbox, not input checkbox.

I'm attaching a codesandbox with both inputs to reproduce this bug.

https://codesandbox.io/s/react-hook-form-get-started-forked-gjcbs?file=/src/index.js

bluebill1049:
https://github.com/react-hook-form/react-hook-form/issues/2848

violabg:
Sorry but I didn't know ha to explain better than attaching a codesanbox. I'll try again

Describe the bug
I don't know if is a react-hook-form or a chakra-ui bug, but after updating to react-hook-form version 7, I'm getting "" instead of true for checkbox value when selected. It only happens on Chakra checkbox, not input checkbox.

To Reproduce
Steps to reproduce the behavior:

Go to https://codesandbox.io/s/react-hook-form-get-started-forked-gjcbs?file=/src/index.js
check both checkboxes 
open code sandbox console
press submit button
it prints:
`{test: "", test2: true}`


Codesandbox link (Required)
https://codesandbox.io/s/react-hook-form-get-started-forked-gjcbs?file=/src/index.js

Expected behavior
should print:
`{test: true, test2: true}`

j-dominguezp:
Can confirm this. 

The only difference I found between HTML native checkbox and Chakra UI checkbox is that onChange event returns different values in **e.target.value** HTML native returns **"on"** and chakra-ui returns **""**.

**Both e.target.checked have the right value.** I had to use Controller to make it work.

Hope that helps.

violabg:
I have also created e forked sample with react-hook-react version 6 that works:

https://codesandbox.io/s/react-hook-form-get-started-forked-6k5yc?file=/src/index.js

violabg:
> 
>  I had to use Controller to make it work.
>

I'll try that

manzoorwanijk:
I can confirm this with Chakra UI Switch, which is nothing but a checkbox.

violabg:
> > I had to use Controller to make it work.
> 
> I'll try that

Yes with the Controller is working, but as you can see from here:
https://codesandbox.io/s/react-hook-form-get-started-forked-6k5yc?file=/src/index.js

in version 6 was working without Controller.
[...]

PR #4614, opened 2021-04-04
#4603 Update getCheckboxValue for case empty value

leapful:


bluebill1049:
Thanks very much @leapful

Issue #4604, opened 2021-04-03
[Preact, RHF 7.0] Radio inputs not working 


r04423:
**Describe the bug**
Radio inputs aren't working correctly in Preact project after upgrading to the latest RHF version.
Only last selected input produces a value, other input values are shown as `null`.

**To Reproduce**

1. Go to the Codesandbox example
2. Click on 'One', click 'Submit'
3. `{test: null}` is logged to the console
4. Click on 'Three', click 'Submit'
5. `{test: "three"}` is logged to the console

**Codesandbox link (Required)**
https://codesandbox.io/s/funny-saha-1dr1e?file=/src/index.js

**Expected behavior**
All radio inputs produce values.

**Screenshots**
N/A

**Desktop:**

- OS: Windows
- Browser: Edge
- Version: 89

**Additional context**
Not sure if Preact compatibility is something that has any priority, just wanted to report this as it worked in RHF 6.

leapful:
@r04423 

Preact is currently handling "ref callback" different than React. At the time "ref callback" is triggered, radio inputs are not attached to the page yet. When collecting and merging DOM reference of multiple radio inputs in a radio group, method "register" of react-hook-form v7 performs a safe filtering to check if those radio inputs exists in the page or not by using "document.contains(ref)" therefore the previous DOM references are cleared and only the last "ref" is available.

Using Controller is a workaround for this issue at the moment:
https://codesandbox.io/s/admiring-bush-ht19u?file=/src/index.js

@bluebill1049 

Do you see any problem if we remove the filter condition "document.contains" in 

https://github.com/react-hook-form/react-hook-form/blob/17788756bbdabf3121edcb96ada1e8e465ba5076/src/useForm.ts#L866

bluebill1049:
> @r04423
> 
> Preact is currently handling "ref callback" different than React. At the time "ref callback" is triggered, radio inputs are not attached to the page yet. When collecting and merging DOM reference of multiple radio inputs in a radio group, method "register" of react-hook-form v7 performs a safe filtering to check if those radio inputs exists in the page or not by using "document.contains(ref)" therefore the previous DOM references are cleared and only the last "ref" is available.
> 
> Using Controller is a workaround for this issue at the moment:
> https://codesandbox.io/s/admiring-bush-ht19u?file=/src/index.js
> 
> @bluebill1049
> 
> Do you see any problem if we remove the filter condition "document.contains" in
> 
[...]

PR #4615, opened 2021-04-04
#4604 Remove checking ref in page when registering checkbox or radio …

leapful:
…group

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -865,8 +865,8 @@ export function useForm<
           ? {
               ...field._f,
               refs: [
-                ...compact(field._f.refs || []).filter(
-                  (ref) => isHTMLElement(ref) && document.contains(ref),
+                ...compact(field._f.refs || []).filter((ref) =>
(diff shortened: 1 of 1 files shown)


bluebill1049:
hmmm looks like one of the e2e failed.

src/useForm.ts
```diff
@@ -865,8 +865,8 @@ export function useForm<
           ? {
               ...field._f,
               refs: [
-                ...compact(field._f.refs || []).filter(
-                  (ref) => isHTMLElement(ref) && document.contains(ref),
+                ...compact(field._f.refs || []).filter((ref) =>
```
bluebill1049, line 868: I see this is made to remove unmounted `ref`, so we will need this.

leapful: @bluebill1049 You're right, "document.contains" is used for case show/hide checkbox (un-mounting checkbox).

It seems we cannot make it compatible with Preact because the state of a "not attached" DOM element is the same with "detached" DOM element.

PR #4669, opened 2021-04-07
Improve app routing

Lincerossa:
**WHAT and WHY**
When i run **yarn run start:app** i did not see anything on localhost:3000, so for a while i felt lost.
I am a newbie on this project, so maybe it was just my fault :)

I suggest you this new commit in order to render something at that route.
It could be a kind of shortcut page for keeping track of all the available e2e routes

**HOW**
Screenshot 2021-04-07 at 16 23 27

### DIFF
--- a/app/README.md
+++ b/app/README.md
@@ -1,5 +1,17 @@
 # App for cypress automation
+## What?
+This app is used by cypress for e2e testing purpose
 
-## Install
+## Why?
+We simply want to test each react-hook-form functionality in a separated environment.
+The current app folder is used for this purpose.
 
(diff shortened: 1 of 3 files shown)


bluebill1049:
oh wow, that's awesome! ❤️✌🏻

Lincerossa:
oh thanks!! 😄
I will continue on this topic to finish the welcome page for the devs that for the first time run *yarn run start:app*

PS: How could I suggest new things without creating new pull requests directly in master? are there any discussions, discord channels or whatever?

PR #4714, opened 2021-04-10
🐛 fix can not reuse a regexp with g flag as pattern

UncleBill:
reproduce demo, try inputing `a` and repeat hitting <kbd>enter</kbd> in the form:

https://codesandbox.io/s/react-hook-form-get-started-forked-y8q3x


bluebill1049:
thank you Uncle Bill (my name is Bill) 😆

PR #4721, opened 2021-04-10
fix typings for watch

lynxtaa:
Fixes typings for watch when using array of names. 

Before:
```ts
const { watch } = useForm<{ name: string; age: number }>()
watch(['name', 'age']) // (string | number)[]
```

After:
```ts
const { watch } = useForm<{ name: string; age: number }>()
watch(['name', 'age']) // [string, number]
```

It uses a solution similar to `Promise.all`: https://github.com/microsoft/TypeScript/blob/131875bb849c0a9c56c26d09881453aace0cbfed/lib/lib.es2015.promise.d.ts#L41

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -804,7 +804,7 @@ export function useForm<
   const watch: UseFormWatch<TFieldValues> = (
     fieldName?:
       | FieldPath<TFieldValues>
-      | FieldPath<TFieldValues>[]
+      | ReadonlyArray<FieldPath<TFieldValues>>
       | WatchObserver<TFieldValues>,
     defaultValue?: unknown,
   ) =>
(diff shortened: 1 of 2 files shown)


bluebill1049:
oh! thanks buddy for the fix.

bluebill1049 (approved):

src/useForm.ts
```diff
@@ -804,7 +804,7 @@ export function useForm<
   const watch: UseFormWatch<TFieldValues> = (
     fieldName?:
       | FieldPath<TFieldValues>
-      | FieldPath<TFieldValues>[]
+      | ReadonlyArray<FieldPath<TFieldValues>>
```
bluebill1049, line 807: actually i think we can use `FieldPathValues`?

bluebill1049: nvm, i think what you got is good.

bluebill1049:
thank you very much @lynxtaa

PR #4722, opened 2021-04-11
close #4716 support name prop with useFormState to isolate re-render by name

### LINKED ISSUE #4716: Add a useWatchError or equivalent way to watch errors for a single field in a nested component

I haven't found an easy way to watch errors of a single field in a nested component without re-rendering it each time _any_ other field error changes (or other events that change the form state).

`useWatch` works well to return the value of the provided field name and don't cause re-renders when other fields have a value change (assuming we pass the control as prop instead of relying on the FormProvider). I'd love to have the same to track errors instead of values.

The idea is to have a nested component for each field. Each component is responsible for displaying its own errors, and should _not_ render on events that are not related to this field (e.g. other fields' errors or value changes). Example use case: make a form builder (based on a JSON model).

Passing specific errors as props. But it's not ideal for all use cases.

I have checked the useWatch implementation to get some inspiration, but I haven't found any easy way to watch error changes from control methods. Maybe `formStateSubjectRef`? But it might require to loop over all errors, for each component that is subscribing.

**Additional context**

N/A

bluebill1049:
## New feature

- `useFormState({ name: 'fieldName' })` API

### Example 

```ts
const { errors } = useFormState({ name: 'fieldName' });
console.log(errors) // should only re-render when fieldName error updated.
```

### Feature request

- close #4716

### Functional Test

- https://codesandbox.io/s/strange-panini-z9cub?file=/src/App.js

### Todos

- [ ] doc update


bluebill1049:
@jorisre your function saved my day!

bluebill1049:
Screen Shot 2021-04-11 at 9 46 42 pm

I lost my shit on this...

src/logic/shouldRenderFormState.ts
```diff
@@ -1,16 +1,22 @@
...
   isRoot?: boolean,
-) =>
-  isEmptyObject(formState) ||
-  Object.keys(formState).length >= Object.keys(readFormStateRef).length ||
-  Object.keys(formState).find(
-    (key) =>
-      readFormStateRef[key as keyof ReadFormState] ===
-      (isRoot ? VALIDATION_MODE.all : true),
+) => {
+  const formState = omit(formStateData, 'name');
```
bluebill1049, line 11: savior!

PR #4723, opened 2021-04-11
👷‍♂️ simplify the build process

bluebill1049:
## Description

we are no longer doing env check, so we can safely remove those extra builds, and make our build simpler and faster. Maybe we should consider moving to the micro build package which used in the resolver? hopefully, it doesn't blow up the size tho.


scripts/rollup/createRollupConfig.js
```diff
@@ -56,12 +30,8 @@ export function createRollupConfig(options, callback) {
         commonjs({
           include: /\/node_modules\//,
         }),
-      options.env !== undefined &&
-        replace({
-          'process.env.NODE_ENV': JSON.stringify(options.env),
-        }),
       sourcemaps(),
-      shouldMinify &&
+      options.format !== 'esm' &&
```
jorisre, line 34: I think you can minify for all bundles

jorisre (approved):
LGTM ! I'll try again microbundle

scripts/rollup/createRollupConfig.js
bluebill1049: Thanks for the review @jorisre i remember last time @kotarella1110 pointed me to a PR that was recommended not minify esm.

jorisre: Ah ! I want to know why :)
By default, microbundle minify esm.
In all case, when user import React Hook Form esm, the bundler of its project will minify it.

kotarella1110: @bluebill1049 @jorisre Please see this https://github.com/formium/tsdx/issues/155 👍

kotarella1110: As joris said, in the case of esm, the project bandler can minify, and the decision to minify should be left to the user.

kotarella1110 (approved):
LGTM 🎉

bluebill1049:
thanks for the review guys.

PR #4758, opened 2021-04-13
support unregister with register and useController

### LINKED ISSUE #4355: Feature Request - auto unregister inputs on unmount
Yes. In V7 and going forward, inputs are not unregistered on unmount. So in other words, we must manually unregister an input if we want its value removed from form values. While this is a great design choice and in most cases makes things much easier, there are cases where we do want form values removed on unmount. 

For example, take a check box that adds additional fields. If the user unchecks the box, then in most cases we can assume that the fields are no longer relevant and we don't want to keep the values around. While it's totally doable to unregister the fields via `watch` and `useEffect`, it adds extra code and logic. It gets especially bad when we have a chain of related fields. For example show C if B and show B if A. We then need to add a bunch of checks to make sure we are not unregistering the field when we shouldn't be. 

I propose adding a prop to Controller and an option to register that will unregister the input when on unmount. So essentially the V6 behavior. This would give us complete control over how to manage our inputs regarding mounting and unmounting. 

Using `useEffect`, `watch`, and `unregister`.

bluebill1049:
close #4355

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -944,7 +944,9 @@ export function useForm<
             onChange: handleChange,
             onBlur: handleChange,
             ref: (ref: HTMLInputElement | null) =>
-              ref && registerFieldRef(name, ref, options),
+              ref
+                ? registerFieldRef(name, ref, options)
+                : options && options.shouldUnregister && unregister(name),
           };
     },
     [defaultValuesRef.current],
@@ -1145,6 +1147,7 @@ export function useForm<
         formStateRef,
         defaultValuesRef,
         fieldArrayDefaultValuesRef,
+        unregister,
       }),
       [],
     ),
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -948,8 +948,8 @@ export function useForm<
             ref: (ref: HTMLInputElement | null) =>
               ref
                 ? registerFieldRef(name, ref, options)
-                : shouldUnregister ||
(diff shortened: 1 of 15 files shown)


src/useForm.ts
```diff
@@ -947,7 +948,8 @@ export function useForm<
             ref: (ref: HTMLInputElement | null) =>
               ref
                 ? registerFieldRef(name, ref, options)
-                : options && options.shouldUnregister && unregister(name),
+                : shouldUnregister ||
+                  (options && options.shouldUnregister && unregister(name)),
```
iamacook, line 952: If `shouldUnregister` is `true`, it won't call `unregister(name)` here.

```js
(shouldUnregister || options && options.shouldUnregister) && unregister(name),
```

bluebill1049: yep, that's intentional. It's either.

iamacook: Oh, right - sorry!

bluebill1049: all good, thanks for the review. If you don't mind, i will ping you for the final review.

iamacook: That would be great - thank you!

bluebill1049: Hey, iamacook. Ready to be reviewed. 👍

iamacook: All looks good to me. I'm excited for this feature!

Issue #4793, opened 2021-04-14
V7: Trigger validation with return


bjarkehs:
**Is your feature request related to a problem? Please describe.**
We have a blocker before we can upgrade to v7. We currently have some code that looks like this:

```javascript
const onClick = async () => {
  const valid = await inputForm.trigger();
  if (!valid) {
    return;
  }
  const data = inputForm.getValues();
  // use data
}
```

We are currently unable to move to v7 since I cannot figure out how to do this in v7.

**Describe the solution you'd like**
Some way to trigger validation and get the result back.

**Describe alternatives you've considered**
None

bluebill1049:
You can subscribe to the `errors` object with `useFormState` as an alternative.

bjarkehs:
Subscribing to errors is unfortunately not enough, since the changes to errors don't occur until next render. So I can't trigger validation, wait for it, check if everything is valid, and then perform some action.

My use case is I have some button that uses the form data outside of submission.

lynxtaa:
@bjarkehs I think you can use `handleSubmit`:

```ts
const onClick = handleSubmit(
  values => {
    // ...
  }
)
```

bjarkehs:
@lynxtaa That could be a solution, but I need to also be able to only trigger validation on certain inputs.

So my example above is more like:

```javascript
const onClick = async () => {
  const valid = await inputForm.trigger(paths);
  if (!valid) {
    return;
  }
  const data = inputForm.getValues();
  // use data
}
```

PR #4802, opened 2021-04-15
Return type of trigger should be Promise<void>

bjarkehs:
Currently Typescript says that `await` is not necessary. But in order to wait for validation to have kicked in, we need to `await`. This makes Typescript aware that `trigger` returns a `Promise`.

This is necessary to handle the case mentioned in #4793 where we trigger validation and then check the errors object.


bjarkehs:
Unfortunately this is not enough, since the formstate is not updated until next render anyway.

riccardobertolini:
I agree with bjarkehs, I'm facing exactly the same issue. Thanks for preparing the PR as well

riccardobertolini (approved):

bluebill1049:
no longer required. thanks for the PR.

PR #4856, opened 2021-04-19
return trigger result

### LINKED ISSUE #4793: V7: Trigger validation with return

### LINKED ISSUE #4903: Errors properties off by one render outside render
**Describe the bug**
When we need to do something after the validation (in another function) is done, it seems like `errors`-properties are off by one render on their values.

**To Reproduce**

1. Open console (to see output)
2. Go to firstName input-field
3. Remove John
4. Insert a single space-character ' '
5. Click outside input-field (to trigger onBlur)
6. Look at console-output
7. Go to firstName input-field
8. Remove the single space and leave input blank
9. Click outside input-field (to trigger onBlur)
10. Look at console-output

https://codesandbox.io/s/small-thunder-nq0t4

**Expected behavior**
Expected that `errors.firstName` contains the error as the same as `error` does in the console.

**Screenshots**
image
(Image above) This one shows that `errors` contains the correct error, but `errors.firstName` is undefined. Should not say `Do something`. The UI-error is correct.

image
(Image above) This one shows that `errors` contains the correct error, but `errors.firstName` is `errors` previous value. Should say `Don't do something`. The UI-error is correct.

**Additional context**
Have tried the same in JS (not TS) same result.
Tested on version: 7.2.2 and 7.1.1

bluebill1049:
close #4793 
close #4903


src/useForm.ts
```diff
@@ -451,18 +451,20 @@ export function useForm<
           ),
         );
       } else {
-        isUndefined(name)
+        isValid = !!(isUndefined(name)
           ? await validateForm(fieldsRef.current)
           : await Promise.all(
               fields.map(async (data) => await executeValidation(data, null)),
-            );
+            ));
```
KirchoKirev, line 458: @bluebill1049 Is this correct? `await Promise.all(...)` will return an array with boolean values which is always truthy. If we want the correct value for `isValid` in this case we would want to do `(await Promise.all(...)).every(Boolean)` just like in [v6](https://github.com/react-hook-form/react-hook-form/blob/v6.15.7/src/useForm.ts#L429)?

bluebill1049: oh wow, good pick! want to send a PR, credit goes to you.

bluebill1049: I will wait till end of today, hopefully you can make it and should be another minor release.

KirchoKirev: Hey, I saw that you already made the change, thank you :)

bluebill1049: no worries! thank you for point this out!

Issue #4837, opened 2021-04-17
`useWatch` method does not watch form values when `defaultValue` has a value.


YuitoSato:
**Describe the bug**
`useWatch` method does not watch form values when the `defaultValue` has a value.
When `defaultValue` is undefined, form values are watched.

**To Reproduce**
The expected value A is not displayed even though `defaultValues` has a value  in the codesandbox page below.

```ts
import "./styles.css";
import { useForm, useWatch } from "react-hook-form";

export default function App() {
  const { control } = useForm({
    shouldUnregister: false,
    defaultValues: {
      select: "A"
    }
  });

  return (
    <span>
      <SelectedValue control={control} />
    </span>
  );
}

const SelectedValue = function SelectedValue({ control }) {
  const selectedValue = useWatch({
    control,
    name: "select",
    defaultValue: ""
  });

  // Value A is not displayed.
  return <>selectedValue: {selectedValue}</>;
};
```

**Codesandbox link (Required)**
https://codesandbox.io/s/immutable-browser-09fxk?file=/src/App.js

V7

- [JS](https://codesandbox.io/s/elated-dubinsky-vol10) Template
- [TS](https://codesandbox.io/s/frosty-wave-f1pco) Template

V6

- [useForm](https://codesandbox.io/s/react-hook-form-useform-template-lsm61) Template
- [useFieldArray](https://codesandbox.io/s/react-hook-form-usefieldarray-template-q0jek) Template
- [Controller](https://codesandbox.io/s/react-hook-form-controller-template-yu0z5) Template

**Expected behavior**
form values are watched by the useWatch method

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

YuitoSato:
I'm willing to create pull request if you like. Thank you for your great OSS

bluebill1049:
hey, @YuitoSato thanks for the kind words and willingness to fixing this issue. I think this got overwrite by inline `defaultValue: ""`

bluebill1049:
I am open to discussion on this right behavior too.

YuitoSato:
I think `useWatch` should return a value in a form if they exist, not are undefined.

When a form value is undefined → it should return a value in `defaultValue`.
When a form value is defined (including empty string) → it should return a form value.

https://github.com/react-hook-form/react-hook-form/blob/a6586260727a746770045db4cf4e4f795b4a03e8/src/useWatch.ts#L49

[...]

PR #4952, opened 2021-04-25
useWatch method should watch form values when defaultValue has a value.

YuitoSato:
This pr is for this [issue ](https://github.com/react-hook-form/react-hook-form/issues/4837)

https://github.com/react-hook-form/react-hook-form/blob/770f7240158ce7e544ead553702648488118b7a5/src/useWatch.ts#L49

When `useWatch` has `defaultValue` property, `defaultValue` is used for the first value of `useState`, not values in a form.
I changed its behavior to use the return value of watchInternal as the first value.

### DIFF
--- a/src/useWatch.ts
+++ b/src/useWatch.ts
@@ -44,10 +44,10 @@ export function useWatch<TFieldValues>(props?: UseWatchProps<TFieldValues>) {
   nameRef.current = name;
 
   const { watchInternal, watchSubjectRef } = control || methods.control;
+
+  const firstValue = watchInternal(name as InternalFieldName);
   const [value, updateValue] = React.useState<unknown>(
-    isUndefined(defaultValue)
-      ? watchInternal(name as InternalFieldName)
-      : defaultValue,
+    isUndefined(firstValue) ? defaultValue : firstValue,
   );
 
   React.useEffect(() => {


src/useWatch.ts
```diff
@@ -44,10 +44,10 @@ export function useWatch<TFieldValues>(props?: UseWatchProps<TFieldValues>) {
   nameRef.current = name;
 
   const { watchInternal, watchSubjectRef } = control || methods.control;
+
+  const firstValue = watchInternal(name as InternalFieldName);
   const [value, updateValue] = React.useState<unknown>(
-    isUndefined(defaultValue)
-      ? watchInternal(name as InternalFieldName)
-      : defaultValue,
+    isUndefined(firstValue) ? defaultValue : firstValue,
```
bluebill1049, line 50: Thanks @YuitoSato we specifically don't want to fall back to the default value, because `defaultvalue` should only return on the first invoke before the render.

YuitoSato: Thanks. Sorry I'm late.
Do you mean defaultValue should  return only once before the target form control render?
I need to understand what "render" means, because after the target form control render, watchInternal no longer return undefined value, right?

YuitoSato: @bluebill1049

bluebill1049: Yes, so the idea behavior should be `watch` only return default value once before the `render`, once `render` inputs, it should what's in the view.

YuitoSato: Thank you.

So how should I check if the target control is finished or not? like the code below 

```
 const [value, updateValue] = React.useState<unknown>(
    !isAlreadyRendered() & isUndefined(firstValue) ? defaultValue : firstValue,
);
```

YuitoSato: @bluebill1049 please check my comment when you have time 👀

bluebill1049: sorry for the late reply, I guess the problem is inline defaultValue will always overwrite the form value for the first render. There are two approached to solve the problem, but it's not without a trade-off. 

1. usage level: don't supply `defaultValue`
2. check `formValues` for watched value for undefined or not before using the `defaultValue`

personally, I prefer the first option as the second option does use more performance budget for the hook.

YuitoSato: I prefer the first option, too. So should we close this issue? It is possible to delete `defaultValue` property but it will be disruptive change.

Or just fix the documentation?

bluebill1049: I think we can improve on the doc to reflect on this. I wouldn't delete the `defaultValue` as some use case do actually make good use of it, eg no defaultValues is supplied and etc.

bluebill1049: Screen Shot 2021-05-17 at 7 02 35 pm

YuitoSato: Thank you for your fix! (Or did the doc already have an explanation?)
 I think this issue should be closed

bluebill1049: Thank you for ur help and sending through the PR.

Issue #2791, opened 2020-09-01
Focus error field on manually triggered validation


kreftk:
**Is your feature request related to a problem? Please describe.**
Currently there is a different behaviour whether the error field is focused on validation. If validation is triggered by `onSubmit` the field is focused but if validation is triggered manually by `trigger()` the error field is not focused.

**Describe the solution you'd like**
I assume that this is the desired behaviour so it would be awesome to have an opt-in option in a config object as an argument:
```ts
trigger: (payload?: string | string[], config?: { shouldFocusError: boolean = false }) => Promise<boolean>
```

**Example**
https://codesandbox.io/s/mutable-dust-tglj1?file=/src/App.js

Would like to work on this feature if desired 😉

Awesome project 🙂

bluebill1049:
Thanks @kreftk I will tag it as a feature request and waiting for more users to vote before including the feature.

bluebill1049:
proposal API:

```ts
trigger('test', { shouldFocus: true })
```

by default `trigger('test', { shouldFocus: false })`

kreftk:
@bluebill1049 Your proposed API LGTM 🙂 
Should I work on this feature?

bluebill1049:
if you interested, yes please @kreftk.

pablohenriquedev:
Hello guys. Is this feature ready to use?

PR #4960, opened 2021-04-26
feature: focus with trigger

bluebill1049:
## usage

```ts
trigger('test', { shouldFocus: true })
```

## Feature Request

#2791


gsaandy:
is this ready to release?

kaisermann (dismissed):
API looks great and simple enough 👍 

(not that my review means anything, just want to show support for this feature as well (I read the code of course  👀))

bluebill1049:
Thank your ur review. It definitely means a lot from the community. We are at the refinement for 7.6, plan to release this 7.7 in the following weeks.

kaisermann:
@bluebill1049 Thanks for merging this! Forgive me for asking, but 

> This only works when the input's reference is registered, it will not work for custom register as well.

Does this mean that it won't work with inputs inside `<Controller>...</Controller>`?

Issue #4356, opened 2021-03-05
Feature request - provide access to schema level validation errors


wdfinch:
**Is your feature request related to a problem? Please describe.**
Yes. If we are using schema-based validation like joi or yup, then we currently only have access to field-level error messages. However, it's common to have field array level error messages and multi-field error messages that don't relate to one field. For example, a very common use case would be to require that a field array contains at least one element.

While the validation works fine, we have no easy way of showing the user an error message. Instead, we must write our own resolver or do the validation ourselves to get this information. 

**Describe the solution you'd like**
I would like to find a way to add schema level error to `formState.errors`. This does break the rule of having each error relate to a field but I think it's necessary given how common a use case is. I think we could modify the object shape to be `formState.errors.schema` and `formState.errors.inputs`. This would mean that `inputs` could stay the same as it is now and `schema` could take on a new shape. 

**Describe alternatives you've considered**
Custom resolver and running the validation myself or again.

bluebill1049:
https://github.com/react-hook-form/react-hook-form/discussions/6348#discussioncomment-1246393

wdfinch:
I replied. I think we should have the discussion here rather than on the issue since the issue is related to an invalid bug report.

bluebill1049:
https://github.com/react-hook-form/react-hook-form/discussions

bluebill1049:
updated.

PR #4986, opened 2021-04-28
feature: Ts custom errors

### LINKED ISSUE #4408: Feature request: Add ability to set errors unrelated to inputs
This is related to #4356 but more focused on the `setError` function. Right now we can only set errors that relate to a specific input. However, there are many cases where there is an error that doesn't relate to a specific input. A simple example is requiring that a field array have at least one element.

I would like to expand `setError` or add a new function (perhaps called `setFormError` or something), that allows us to set arbitrary errors that don't need to relate to a specific field. We could have these errors be under a specific key in `formState.errors` or in another part of `formState`.

My solution right now is to add hidden inputs and use `setError` to set arbitrary errors on it. This works, but it feels wrong as I'm using an input just for the purpose of settings errors. Its value is meaningless.

bluebill1049:
## Context

To allow custom server error.

```ts
useForm<
  TContext extends object = object,
  TCustomErrors extends object = object
>()
```

## Usage

```ts
type ServerError = {
  test: string
}

useForm<FormState, {}, ServerError>();

setError('test', { type: '' })
```

## Feature Request

close https://github.com/react-hook-form/react-hook-form/issues/4408

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "bundlesize": "^0.18.0",
     "copyfiles": "^2.4.1",
     "coveralls": "^3.1.0",
-    "cypress": "6.4.0",
+    "cypress": "^7.2.0",
     "cypress-parallel": "^0.1.9",
     "eslint": "^7.20.0",
     "eslint-config-prettier": "^6.15.0",
--- a/src/__tests__/useForm/setError.test.tsx
+++ b/src/__tests__/useForm/setError.test.tsx
@@ -44,6 +44,7 @@ describe('setError', () => {
       },
     ],
   ];
+
   test.each(tests)('%s', (_, input, output) => {
     const { result } = renderHook(() => useForm<{ input: string }>());
     act(() => {
@@ -51,5 +52,23 @@ describe('setError', () => {
     });
     expect(result.current.formState.errors).toEqual(output);
     expect(result.current.formState.isValid).toBeFalsy();
+  });
+
+  it('should be able to set server error', () => {
+    const { result } = renderHook(() =>
+      useForm<{ input: string }, {}, { server: string }>(),
+    );
+
+    result.current.setError('server', {
+      type: 'server',
+      message: 'something is wrong',
+    });
+
+    expect(result.current.formState.errors).toEqual({
+      server: {
+        type: 'server',
+        message: 'something is wrong',
+      },
+    });
   });
 });
--- a/src/types/errors.ts
+++ b/src/types/errors.ts
@@ -29,8 +29,9 @@ export type ErrorOption =
     };
 
 export type FieldErrors<
-  TFieldValues extends FieldValues = FieldValues
-> = DeepMap<TFieldValues, FieldError>;
+  TFieldValues extends FieldValues = FieldValues,
+  TCustomErrors extends object = object
+> = DeepMap<TFieldValues, FieldError> & DeepMap<TCustomErrors, FieldError>;
 
 export type InternalFieldErrors = Partial<
   Record<InternalFieldName, FieldError>
--- a/src/types/fields.ts
+++ b/src/types/fields.ts
@@ -26,6 +26,8 @@ export type FieldValue<
 
 export type FieldValues = Record<string, any>;
 
+export type CustomFieldErrors = Record<string, any>;
+
 export type FieldElement<TFieldValues extends FieldValues = FieldValues> =
   | HTMLInputElement
   | HTMLSelectElement
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -6,6 +6,7 @@ import { ErrorOption, FieldErrors } from './errors';
 import { EventType } from './events';
 import { FieldArrayDefaultValues } from './fieldArray';
 import {
+  CustomFieldErrors,
   FieldRefs,
   FieldValue,
   FieldValues,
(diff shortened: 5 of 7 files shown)


wdfinch:
Questions: 

1. Would the custom errors be present on `formState.errors`?
2. Now that we have this feature is there any chance we could close out #4356 as well? They are closely related and for consistency it would be great to support the custom errors when using schema validation.

bluebill1049:
> Would the custom errors be present on formState.errors?

Yes, however, still under debate inside the team on whether we should let users take care of `useState` instead within hook form.

kotarella1110:
@wdfinch 

> Now that we have this feature is there any chance we could close out #4356 as well? They are closely related and for consistency it would be great to support the custom errors when using schema validation.

Wouldn't #4356 be solved by using the [`NestedValue`](https://react-hook-form.com/jp/ts/#NestedValue) type?

kotarella1110:
@wdfinch I don't think React Hook Form should take care of the server side.
If you want to customize the error, you can also do so as follows:

```ts
type FormState = {
  name: string
}
type CustomError = {
  test: string
}

const { setError } = useForm<FormState & CustomError>();

setError('test', { type: '' });
```

wdfinch:
Hi @kotarella1110, thanks for the reply. Sorry but I think I'm a bit confused here. This PR was linked in my feature request #4408, so I assumed it added the functionality specified there.

It appears from the notes above that this adds a new generic argument that we can use to specify server errors or arbitrary errors unrelated to specific inputs. Additionally, it expands the `setError` method to support setting these non-input errors and a means to access them via `formState.errors`.  Hence the `setError('test', { type: '' })`. Is this correct or am I missing something?

Therefore, my other question above regarding #4356 was could there be a way to access schema level errors unrelated to inputs as well assuming we are adding this functionality to `setError`. It's not a typescript issue so I don't think `NestedValue` is relevant.

I guess what I and other users are facing is there are some very common errors we would want to have access to that don't relate to a specific input field. A prime example being a field array that requires at least one element. While we can validate this, there are issues with both methods of validation (`setError` and schema-based) when it comes to retrieving the error message. 

For `setError` we can only tie an error to a input so we must choose some input to add the error to or add a hidden input to hold the error.

For schema-based validation, we can do the validation for this, but we cannot get the error message if the error is not associated with an input in the schema. So when presenting the error to the user (you need at least one element in your field array) we have no way to do this easily.

I believe the PR above gives this functionality to `setError` but it seems like since schema validation is fully supported as an alternative, we should have this functionality present in both methods.

kotarella1110:
@wdfinch Thanks for your reply! This PR only changes the type of the TS. It just adds a third Generics (`TCustomErrors`).

> For setError we can only tie an error to an input so we must choose some input to add the error to or add a hidden input to hold the error.

What do you mean?
I think it is possible to register errors other than those related to a specific input with `setError`.

https://codesandbox.io/s/flamboyant-voice-q0ckx?file=/src/App.tsx

kotarella1110:
@wdfinch Is the codesandbox you provided correct?

wdfinch:
@kotarella1110 my bad. I've got a bunch of sandboxes going and I didn't fork properly and overwrote what I had :(. 

I've updated my post above to have the correct link. Here's it again.

https://codesandbox.io/s/stoic-archimedes-85zx4?file=/src/App.js

bluebill1049:
> @kotarella1110 my bad. I've got a bunch of sandboxes going and I didn't fork properly and overwrote what I had :(.
> 
> I've updated my post above to have the correct link. Here's it again.
> 
> https://codesandbox.io/s/stoic-archimedes-85zx4?file=/src/App.js
> 
> So what you'll notice is `formState.isValid` is false initially because we don't have 2 array elements, but `formState.errors` is empty meaning we can't tell the user why the form is invalid. So my hope would be that we could schema errors that don't relate to inputs to `formState.errors`.

hmmm, that's a concern for me, it's making tracking valid even more difficult, I am saying this not just from schema's perspective but build invalidation as well.

bluebill1049:
> @bluebill1049 would it be possible to just forward what the schema outputs somewhere we can access in `formState` and then we can sort through it? So you don't need to try and parse it and whatnot. Not having details on the error is a major issue for using schema validation.

you can build a custom resolver right? which you have access what the schema returned.

wdfinch:
I could but are there rules about what I can pass to errors? Can I just use some arbitrary object? Otherwise, I would need to use redux or context to make the error available. I'm guessing I would need to follow some strict formate for the error object or typescript won't be happy. 

```
const useYupValidationResolver = validationSchema =>
  useCallback(
    async data => {
      try {
        const values = await validationSchema.validate(data, {
          abortEarly: false
        });

        return {
          values,
          errors: {}
        };
      } catch (errors) {
        return {
          values: {},
          errors: errors.inner.reduce(
            (allErrors, currentError) => ({
              ...allErrors,
              [currentError.path]: {
                type: currentError.type ?? "validation",
                message: currentError.message
              }
            }),
            {}
          )
        };
      }
    },
    [validationSchema]
  );
```

wdfinch:
I guess I was hoping this could be built in since it seems like a pretty common need.

> hmmm, that's a concern for me, it's making tracking valid even more difficult, I am saying this not just from schema's perspective but build invalidation as well.

Could you explain why it's difficult? Could we not just add whatever the resolver spits out that doesn't align with an input name to some area in `formState.errors` or `formState.resolverErrors` (a new key)?

kepejas:
hi, any plans to merge this?

PR #5000, opened 2021-04-29
fix use form trigger type

faouzioudouh:
To fix this issue:

```
Type 'Boolean' is not assignable to type 'boolean'.
'boolean' is a primitive, but 'Boolean' is a wrapper object. Prefer using 'boolean' when possible.
```


bluebill1049:
thanks

PR #5005, opened 2021-04-29
Typing for watch & getValues

### LINKED ISSUE #4994: [v7] watch method return types when over 10 keys watched
**Describe the bug**
When using `watch` with an array of more than 10 keys, the types of the returned tuple fallback to the behaviour observed before #4721 was merged  I believe in `v7.0.6` (aka: a `union` of each possible types of the retrieved `FieldValue`)

**To Reproduce**

1. Go to [this sandbox](https://codesandbox.io/s/cranky-kapitsa-3sd8m?file=/src/App.tsx)
2. Hover variables [a1, ..., j1] from the first `watch` call (The one with 10 keys) types are correct
3. Hover variables [a2, ..., k2] from the second `watch` call. See the variables being falsely typed.

**Expected behavior**
The types should be correct in both cases.

___
<details>

<summary> Potential solutions </summary>

I see two solutions to get rid of this typing issue.

- "Manually" type the watch call
```ts
  const [key1, ..., key11] = watch<["key1", ..., "key11"]>(["key1", ..., "key11"]);
```

***Pros***: It doesn't require any changes to the current API. (and this issue can be closed right away)
***Cons***: Even tho it might be a niche use case, it's a bit verbose and the DX (that I found wonderful since v7) is less enjoyable.

- Make use of [the variadic tuple typing](https://github.com/microsoft/TypeScript/pull/39094) of TS 4.0
For what I have tested this could be done by changing [this definition](https://github.com/react-hook-form/react-hook-form/blob/4cb6579b03e5f2a5b22ed2c04d5b6df037b33965/src/types/form.ts#L167-L346) to:
```ts
  <TFieldNames extends Array<FieldPath<TFieldValues>>>(
    fieldNames: readonly [...TFieldNames],
    defaultValue?: UnpackNestedValue<DeepPartial<TFieldValues>>,
  ): [...FieldPathValues<TFieldValues, TFieldNames>];
```

***Pros***: It provides a better DX imho.
***Cons***:  It could potentially be a breaking change if people have typed their `watch` like so (even tho I cannot see a use case to do so):
```ts
watch<"k1", "k2">(["k1", "k2"])
```
as it would need to be define like this afterward:
```ts
watch<["k1", "k2"]>(["k1", "k2"])
```

</details>

If, by chance, you would like to see this implemented, I'll be more than happy to create a PR.

BTiers:
Fix #4994

For reference, [here is the sandbox that shows the issue](https://codesandbox.io/s/cranky-kapitsa-3sd8m?file=/src/App.tsx:500-549)

As explained in the issue mentioned above, this PR make use of [TS variadic tuple types](https://github.com/microsoft/TypeScript/pull/39094) to extract the proper type of the tuple returned by `watch` & `getValues` 

To avoid any potential breaking changes on `watch` type definition, I've kept the [previously defined signatures](https://github.com/react-hook-form/react-hook-form/pull/4721).
____

#### The behaviour if this gets merged would be as follow

- `getValues`:
```ts
type MyFormValues = {
    k1: string;
    ...,
    kN: number;
}

// Previously resolved to (string | ... | number)[]
const ['k1', ..., 'kN'] = getValues(['k1', ..., 'kN']);

// With this PR, resolved to [string, ..., number]
const ['k1', ..., 'kN'] = getValues(['k1', ..., 'kN']);

``` 

- `watch`:
```ts
type MyFormValues = {
    k1: string;
    ...,
    k10: boolean,
    ...,
    kN: number;
}

// Previously resolved to (string | number | ...)[]
const [k1, ..., kN] = watch(['k1', ..., 'kN']);

// With this PR, all of the following calls resolved to [string, ..., number]
const [k1, ..., kN] = watch(['k1', ..., 'kN']);
const [k1, ..., kN] = watch<['k1', ..., 'kN']>(['k1', ..., 'kN']);

// To avoid any breaking changes, this call resolved to [string, ..., boolean] as prior to this PR
// The limit for this syntax still being at 10 keys
const [k1, ..., k10] = watch<'k1', ..., 'k10'>(['k1', ..., 'k10']);
``` 
____

**_Ps_**: First time contributing to an open source project, happy to take any advices :)

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -158,13 +158,13 @@ export type UseFormGetValues<TFieldValues extends FieldValues> = {
     fieldName: TFieldName,
   ): FieldPathValue<TFieldValues, TFieldName>;
   <TFieldNames extends FieldPath<TFieldValues>[]>(
-    fieldNames: TFieldNames,
-  ): FieldPathValues<TFieldValues, TFieldNames>;
+    fieldNames: readonly [...TFieldNames],
+  ): [...FieldPathValues<TFieldValues, TFieldNames>];
 };
 
 export type UseFormWatch<TFieldValues extends FieldValues> = {
   (): UnpackNestedValue<TFieldValues>;
-  <TFieldNames extends Array<FieldPath<TFieldValues>>>(
+  <TFieldNames extends FieldPath<TFieldValues>[]>(
     fieldNames: readonly [...TFieldNames],
     defaultValue?: UnpackNestedValue<DeepPartial<TFieldValues>>,
   ): [...FieldPathValues<TFieldValues, TFieldNames>];
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -164,6 +164,10 @@ export type UseFormGetValues<TFieldValues extends FieldValues> = {
 
 export type UseFormWatch<TFieldValues extends FieldValues> = {
   (): UnpackNestedValue<TFieldValues>;
+  <TFieldNames extends Array<FieldPath<TFieldValues>>>(
+    fieldNames: readonly [...TFieldNames],
+    defaultValue?: UnpackNestedValue<DeepPartial<TFieldValues>>,
+  ): [...FieldPathValues<TFieldValues, TFieldNames>];
   <
     T1 extends FieldPath<TFieldValues>,
     T2 extends FieldPath<TFieldValues>,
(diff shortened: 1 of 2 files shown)


kotarella1110 (approved):
Awesome 🎉
Thank you 🙌

bluebill1049:
Thanks very much @BTiers for the PR and @kotarella1110's review.

Issue #4033, opened 2021-01-27
Support useController hook to be called "conditionally" (in a safe manner)


dwiyatci:
**Is your feature request related to a problem? Please describe.**
There are some (edge) cases in my code where I cannot simply use `<Controller>` wrapper component. That is why I resort to `useController` hook. As far as I notice, calling this hook means registering the field with the given `name` regardless the corresponding control will be rendered or not.

**Describe the solution you'd like**
Since [the rules of hooks](https://reactjs.org/docs/hooks-rules.html) says that we shouldn't call hooks conditionally, I'd like to be able to call `useController()` with `undefined`, `null`, `false`, or any falsy argument so that I can always call the hook without registering a field/control. Please see my hypothetical code below in the additional context.

**Describe alternatives you've considered**
As workaround, I'm currently considering: https://unsplash.com/blog/calling-react-hooks-conditionally-dynamically-using-render-props/

**Additional context**
```js
export default function App() {
  const { handleSubmit, control } = useForm();
  const onSubmit = (data) => console.log(data);

  const firstNameControllerProps = useController({ name: 'firstName', control });

  const showLastName = false;
  // currently this is not supported, code will break here
  const lastNameControllerProps = useController(showLastName && { name: 'lastName', control });

  return (
    <form onSubmit={handleSubmit(onSubmit)}>
      <CustomControl {...firstNameControllerProps} />
      {showLastName && <CustomControl {...lastNameControllerProps} />}
      <input type="submit" />
    </form>
  );
}
```

getTobiasNielsen:
You don't want to make <CustomControl> and have the useController in those, and just pass the name and control props?

getTobiasNielsen:
https://codesandbox.io/s/vibrant-ride-7tll8?file=/src/App.js

getTobiasNielsen:
I guess, after rereading your question, I understand what you want, but that is already done automatically if you hide the component.

getTobiasNielsen:
What you are looking for already happens automatically:

https://codesandbox.io/s/vibrant-ride-7tll8?file=/src/CustomControl.js

Hope this helps.

dwiyatci:
@getTobiasNielsen Thank you but unfortunately I don't have control over `<CustomControl>`. 🙂  My `<CustomControl>` there is more like a component from some UI widget library, just like the `<TextField>` in the API docs that can only accepts `value`, `onChange`, `onBlur`, and `ref`.

[...]

PR #5015, opened 2021-05-03
[Feature Request] Make useController save to return undefined if no c…

dohomi:
…ontext or control present

Skip initialization of useController if no control or context is present. This is an attempt to solve https://github.com/react-hook-form/react-hook-form/issues/4033

The main use case would be, that users can use the hook inside of their custom component and the functionality will be silently ignored in case its not needed.

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -29,6 +29,9 @@ export function useController<
   TName
 > {
   const methods = useFormContext<TFieldValues>();
+  if(!methods && !control) {
+    return undefined;
+  }
   const {
     defaultValuesRef,
     register,


bluebill1049:
Thanks for the PR, however, i don't think this is going to be type-safe for what `useController` returns.

dohomi:
@bluebill1049 the `useFormContext` works in the way that it returns undefined/null if no context is given. It would be ideal if `useController` could do the same, at least I get the feeling that people would expect that it works in the same way.

bluebill1049:
I know what do you mean, but It's case by case.

matinzd:
Any updates?!

Issue #5070, opened 2021-05-07
Support array as well as object in register return value


kotarella1110:
**Is your feature request related to a problem? Please describe.**

It is a bit cumbersome to customize prop names because supports only object in `register` return value.

```ts
const { name, ref: inputRef, onChange: onChangeText, onBlur } = register("name");
```

**Describe the solution you'd like**

We can support array as well as object.
If this feature request is accepted, I will create a PR 👍🏻 

**Describe alternatives you've considered**

```ts
// object destructing
const { name, ref: inputRef, onChange: onChangeText, onBlur } = register("name");
// array destructing, making it easy to customize the prop names 
const [name, inputRef, onChangeText, onBlur] = register("name");
```

**Additional context**

https://codesandbox.io/s/react-hook-form-5070-fy2kd?file=/src/index.tsx

bluebill1049:
@kotarella1110 that's related to https://github.com/react-hook-form/react-hook-form/projects/1 oh yeah, I really like the array idea! please go ahead. I will add this into project board.

Issue #4629, opened 2021-04-05
Allow users to specify the `ref` prop for `register`


ricokahler:
**Is your feature request related to a problem? Please describe.**

Congrats on v7! I haven't been watching the pre-releases but all of the changes make sense. There is just one piece that makes the API less appealing than v6 and I would like to propose a solution.

My issue is with the new `register` API. All of our input components are wrapped in divs so their `ref` property refers to that `div` vs the nested `input` component. The migration guide gives [an example of how to work around this](https://react-hook-form.com/migrate-v6-to-v7/#main:~:text=Missing%20ref) however that example isn't as appealing when you have a lot of inline register calls (therefore requiring multiple declarations).

**Describe the solution you'd like**

I think it makes sense to create an option in the second argument of `register` called `refProp`. By default that could be just `ref` but for those of us using components libs like [material-ui](https://material-ui.com/api/text-field/#textfield-api), it would be great to change this to `inputRef`.

```tsx
<InputField {...register('foo', { refProp: 'inputRef' })} />
```

**Describe alternatives you've considered**

I can't think of any alternate APIs at the moment but the goal would be to keep the same inline usage as v6 but using custom ref prop names.

**Additional context**

I believe typescript would allow capturing this prop name so it can be in the return type. [Here is a demo](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgPIAczAPYgM4A8AKshAB6QgAmeyeYUoA5gHzIDeAUMj8lBDAAKUbOgBcyIgG5OAX06cYAVxAIsuPhCbB60YqQoRqteoxCsAFGAkZ1+YiwCUE9shjZsE082SzkAMg5kAG1BZFBJAF0JOBAAT18Obl5+MCUoECTebLcPCQByACM4KHyAGhCwADp+IRF0aOQQJQAbFuQ4Wli45J4-TqD3TzoGHz9A11DwzKJG7t8ZeU4EXHpNPFawZABeTW1dKAtXWuFRAtB0JTAAJQEAQXiwAAtmAE1sJQB1WLB830cZAogQowHF0ChbhsWltdqDwdgYOtNlIgA)

![image](https://user-images.githubusercontent.com/10551026/113615657-1e84b480-9622-11eb-9d92-8044688bdbdc.png)

Notice how `Result` reflects the string literal.

I would be willing to work on a PR for this if you are accepting. Thank you!

yudielcurbelo:
[...]

PR #5071, opened 2021-05-07
support register return array value

kotarella1110:
## Context

make it easy to customize the prop names.

```ts
// object destructing
const { name, ref: inputRef, onChange: onChangeText, onBlur } = register("name");
// array destructing, making it easy to customize the prop names 
const [name, inputRef, onChangeText, onBlur] = register("name"); // <- support!
```

## Feature Request

#5070
#4629

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -982,28 +982,38 @@ export function useForm<
       fieldsNamesRef.current.add(name);
       isInitialRegister && updateValidAndValue(name, options);
 
-      return isWindowUndefined
-        ? ({ name: name as InternalFieldName } as UseFormRegisterReturn)
-        : {
-            name,
-            onChange: handleChange,
-            onBlur: handleChange,
-            ref: (ref: HTMLInputElement | null): void => {
-              if (ref) {
-                registerFieldRef(name, ref, options);
-              } else {
-                const field = get(fieldsRef.current, name) as Field;
-                field && (field._f.mount = false);
+      const callbackRef = (ref: HTMLInputElement | null): void => {
+        if (ref) {
+          registerFieldRef(name, ref, options);
+        } else {
+          const field = get(fieldsRef.current, name) as Field;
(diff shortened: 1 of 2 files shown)


src/useForm.ts
```diff
@@ -982,28 +982,38 @@ export function useForm<
...
+        result.name = result[0];
+        return result;
+      }
+      const result = [
+        name as InternalFieldName,
+        callbackRef,
+        handleChange,
+        handleChange,
+      ] as UseFormRegisterReturn;
+      result.name = result[0];
```
bluebill1049, line 1012: can we please change the order to `[name, onChange, onBlur, ref]`?

bluebill1049 (commented):
i think lint failed 👍

kotarella1110:
@bluebill1049 I hadn't checked the console 🙏🏻 
Since I can't solve this, let's close this Feature Request!

スクリーンショット 2021-05-07 18 56 31

bluebill1049:
I have been thinking about this issue for the past couple of weeks. 

**option A:** https://twitter.com/HookForm/status/1381918850753568773
**option B:** bring `ref={regsiter('test')}` back, however, It's difficult, and will probably need a different method instead make `register` a mess.

kotarella1110:
> option B: bring ref={regsiter('test')} back, however, It's difficult, and will probably need a different method instead make register a mess.

Yes, OptionB is also a good choice, since name, onChange, and onBlur can be bound to ref.

How about binding only onChange and onBlur to ref without changing the API?

```ts
const { name, ref } = register("name");
```

bluebill1049:
> but I think it is less frequently customized than onChange or onBlur.

You will be surprised with those external libraries, eg: `onTextChange` 

I think from lib's package size point of view, option A is probably the preferred choice. 

```
register((props) => ({
  name: ‘test’
  inputRef: props.ref,
}));
```

a callback will solve all the rename problems, but it does make API not clean.

PR #5092, opened 2021-05-09
5089 Correct setValue in case update value for field array name

### LINKED ISSUE #5089: Bug - using setValue on field array breaks future setValue call
**Describe the bug**
**If** we use `setValue` to initialize a field array. EX: 

```
React.useEffect(() => {
  setValue("names", [{ name: "will" }]);
}, []);
```

**And** then later on we try and update an index of the same field array using `setValue`:

```
const changeValue = () => {
  setValue(`names.${index}`, {
    name: "updated"
  });
}; 
```

**Then** formValue will **not** update

**To Reproduce**

1. Go to the sandbox
2. Click on the change value button
3. Notice that the value does not change
4. Comment out line 48 or replace line 48 with lines 51-53.
5. Notice that it works now.

https://codesandbox.io/s/pensive-hofstadter-chm2i?file=/src/App.js:302-310

**Expected behavior**
I would expect a call of `setValue` to not affect future calls of `setValue`. While we can use reset here as a workaround, I think this should work. Additionally, I know in this example it would be better to use default values, but in my project, I need to use either `setValue` or `reset` to initlize the form.

leapful:


bluebill1049:
Thank you very much @leapful!

bluebill1049:
hey, @leapful would you like to set up a GitHub sponsor? I would like to show some of my appreciation. If that's not working for you, you can message me privately with your PayPal address as well.

leapful:
@bluebill1049 Thanks for your suggestion. I'm just a FOSS developer as a hobby and try to find some fun when solving problems, that's enough to me :D

bluebill1049:
Thank you @leapful again, I am really appreciated all your help and works. 🙏 You have my respect.

PR #5113, opened 2021-05-10
Make properties in controller.field stable

Noitidart:
I was trying to use direct manipulation so the text input never renders in react-native, but onBlur and onChange kept changing referntially. I think it's safe to make it stable though. Can you please check.

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -73,7 +73,8 @@ export function useController<
 
   return {
     field: {
-      onChange: (event: any) => {
+      onChange: React.useCallback(
+        (event: any) => {
         const value = getControllerValue(event);
         setInputStateValue(value);
 
@@ -85,17 +86,19 @@ export function useController<
           type: EVENTS.CHANGE,
         });
       },
-      onBlur: () => {
+        [onChange, name],
+      ),
+      onBlur: React.useCallback(() => {
         onBlur({
           target: {
             name: name as InternalFieldName,
           },
           type: EVENTS.BLUR,
         });
-      },
+      }, [onBlur]),
       name,
       value,
-      ref: (elm) => elm && ref(elm),
+      ref: React.useCallback((elm) => elm && ref(elm), [ref]),
     },
     formState,
     fieldState: Object.defineProperties(


bluebill1049:
what's the use case which you have to memo those methods? it's actually slower for memo than not.

Noitidart:
> what's the use case which you have to memo those methods? it's actually slower for memo than not (when your form scale up).

Example is when memo'ing the controlled input, for example the react-native input that uses direct manipulation to not do any re-renders - https://snack.expo.io/@noitidart/react-hook-form-v7---controller

Noitidart:
I'm thinking maybe, I should detect if it's react-native and do this direct manipulation technique on the register! What do you 
think? Like around here - https://github.com/react-hook-form/react-hook-form/blob/master/src/useForm.ts#L986

PaulCordo:
Totally on brand with this pull request, I have the same issue where a unstable onChange is causing infinite loop.
I don't have any writing rights on this repo but agree with these changes, I would only maybe suggest to rebase on master and make a fixup of last commit?
Thanks for making this patch!

Noitidart:
Thanks @PaulCordo! That's totally true this issue is not just limited to react-native. I had the unnecessary re-render too of an expensive controlled component. My memo wasn't working on it so I was confused.

leapful:
@PaulCordo Could you please provide CSB sample for the issue that causes infinite loop as a test case for this pull request ?

@Noitidart Could you please explain more detail what is the issue in your use case when onBlur and onChange are kept changing ? How it affects the use case direct manipulation on text input in react-native ?

Noitidart:
@leapful in this example, every character I type re-renders the component:

```
function ClassicHookedTextInput(props) {
  const controllerProps = {........};
  const textInputProps = {.......};

  const controller = useController(controllerProps);

  return (
      <AdornedTextInput
        {...textInputProps}
        onChangeText={controller.field.onChange}
        onBlur={controller.field.onBlur}
      />
  );
}
```

So I then use direct manipulation in React Native to not re-render the `<AdornedTextInput>` like this:

```
function DirectHookedTextInput(props) {
  const controllerProps = {........};
  const textInputProps = {.......};

  const controller = useController(controllerProps);

  const lastChangedText = React.useRef(controllerProps.defaultValue);
  const handleChangeText = React.useCallback(text => {
    lastChangedText.current = text;
    controller.field.onChange(text);
  }, [controller.field.onChange]);

  const textInputRef = React.useRef();

  React.useEffect(() => {
    if (lastChangedText.current !== controller.field.value) {
      lastChangedText.current = controller.field.value;
      textInputRef.current?.setNativeProps({ text: controller.field.value });
    }
  }, [controller.field.value]);

  return (
      <DirectAdornedTextInput
        {...textInputProps}
        inputRef={textInputRef}
        onChangeText={handleChangeText}
        onBlur={controller.field.onBlur}
      />
  );
}

const DirectTextInput = React.memo((props) => <AdornedTextInput {...props} ref={props.inputRef} />);
```

However `DirectTextInput continues to re-render despite the `React.memo`. It turns out the prop that keeps changing is `controller.field.onBlur` and `controller.field.onChange`. The `controller.field.onChange` makes my `handleChangeText` keep re-creating the callback. And exhaustive deps makes me put `controller.field.onChange` into the deps. This is not expected. All other form libs keep `onBlur` `onChange` stable.

Noitidart:
In the end, how much does a `useCallback` degrade the peformance? I don't think it should be anymore than we already do in other places of react-hook-form. Like this place in react-hook-form already extensively makes use of `useCallback`:

https://github.com/react-hook-form/react-hook-form/blob/master/src/useForm.ts#L986

leapful:
@Noitidart 

The concern of using React.useCallback for "useController" hook is because this one is field level, it is different than "useForm" hook which is parent form level.

In other use case, "useController" hook or "Controller" component will be related to a custom field which can be used as multiple fields in a form. For example: 

```
const ControlledField = () => {} // Component uses "useController" hook or wrap "Controller" component

const Form = () => {

    return <>
<ControlledField label="Text 1"/>
<ControlledField label="Text 2"/>
...
<ControlledField label="Text 20"/>
</>;
}

```

While "useForm" only creates one callback , "useController" will create many callback, this will lead to more allocation memory.

Referrer about pros and cons of memo: https://kentcdodds.com/blog/usememo-and-usecallback

If it's about memo component "DirectAdornedTextInput" with changed props "onBlur" and "onChangeText", it can be achieved by using comparison callback of "useMemo" : https://snack.expo.io/edl4c9wb6 (line 84)

Noitidart:
> The concern of using React.useCallback for "useController" hook is because this one is field level, it is different than "useForm" hook which is parent form level.
> 
> In other use case, "useController" hook or "Controller" component will be related to a custom field which can be used as multiple fields in a form. For example:

Oh I see!

So we are comparing:

* Every render for every field it creates a new callback. So memory has 1 callback always, be we always reallocating.
* VERSUS single callback per field no matter how many renders. No reallocating.

It seems that in both cases you always have 1 callback in memory, just in second case it is stable. Is reallocating bad for memory? In that case that would be another reason to useCallback field level.

Is my thoughts above accurate?


> If it's about memo component "DirectAdornedTextInput" with changed props "onBlur" and "onChangeText", it can be achieved by using comparison callback of "useMemo" : https://snack.expo.io/edl4c9wb6 (line 84)

Yes correct, this is the work around I use for now. I actually do useCallback onBlur and onChangeText and silence the exhaustive deps to keep it stable. I was concerned as I didn't know if there was a reason for the non-stable. Maybe at the least, if adding useCallback is indeed worse, we should document that these are not stable, but at same time you can treat it as stable.

leapful:
@Noitidart 

Actually, using React.useCallback is not bad unless you have a lot of them just like re-rendering is not bad too unless you have  complex computation inside your component. It depends on volume of data and computation, that's why we always needs to measure first before optimizing performance.

I also agree to update the documentation to make it more clear when using the API. Thanks for your contribution on this issue.

Noitidart:
Oh a note though, can `name` change? Then if it does, wouldn't memo'ing be issue here? It will have memoed the wrong thing?

So in docs we should say if we want to memo instead of this:

```
(prevProps, nextProps) => isEqual(omit(prevProps, ['onBlur', 'onChangeText']), omit(nextProps, ['onBlur', 'onChangeText']))
```

we should add `|| isEqual(prevProps.name, nextProps.name)`?

leapful:
@Noitidart 

The "name" attribute of form element is usually unchanged. In case using field array names, the "name" is index name which can be changed after removing of inserting, and value will be updated for each field array element, in that case re-rendering should be performed so isEqual(omittedPrevProps, omittedNextProps) is enough to check change on "name".

zyasir:
So when I do implementation of `useCallback` here, I should include name right? Is this correct?

```
function DirectHookedTextInput(props) {
  const controllerProps = {........};
  const textInputProps = {.......};

  .....

  const stableBlur = useCallback(controller.field.onBlur, [controller.field.name]);

  return (
      <DirectAdornedTextInput
        {...textInputProps}
        inputRef={textInputRef}
        onChangeText={handleChangeText}
        onBlur={stableBlur}
      />
  );
}
```

zyasir:
> * Every render for every field it creates a new callback. So memory has 1 callback always, be we always reallocating.
> 
> * VERSUS single callback per field no matter how many renders. No reallocating.

You actually do re-allocate `useMemo` as that has a function that wraps. But useCallback yes I don't think you have re-allocation, and it avoids the GC and the extra rendering that would happen. So I am also seeing improved performance with a useCallback here. I also see improved developer experience as it is not clear that I have to add `name` as a dep in the `useCallback` in my last post for `stableBlur`.

The Kent Dodds article says:

image

He is saying comparing elements in the array is expensive. I think I disagree. It's a `===` of one to two extra values. Comparing `===` of two a single render of a react function is always cheaper guaranteed.


Here is also the case where Coinbase memo's everything and they have measured to justify the performance tradeoff.

https://attardi.org/why-we-memo-all-the-things/#memory-cost-of-reactmemo

> There is a memo meme that’s going around: they say that React.memo has a memory cost, because – like other memoizations – you have to keep around the old values in case you need them again later. Makes sense? Unfortunately, that’s not quite right. Because of how React works, the result of previous renders needs to be kept around anyway – to diff it against subsequent renders. That’s the basis of React’s reconciliation algorithm. It can’t work without it.

src/useController.ts
```diff
@@ -85,17 +86,19 @@ export function useController<
...
+      ),
+      onBlur: React.useCallback(() => {
         onBlur({
           target: {
             name: name as InternalFieldName,
           },
           type: EVENTS.BLUR,
         });
-      },
+      }, [onBlur]),
```
zyasir, line 98: You need `name` here as well.

zyasir (changes requested):

PaulCordo:
> 
> 
> @PaulCordo Could you please provide CSB sample for the issue that causes infinite loop as a test case for this pull request ?
> 
> @Noitidart Could you please explain more detail what is the issue in your use case when onBlur and onChange are kept changing ? How it affects the use case direct manipulation on text input in react-native ?
> 
> It would be great if you can provide a unit test for this pull request to keep track on the context of this code change.

Hey @leapful!
Would [that ](https://codesandbox.io/s/react-hook-form-v7-controller-forked-2dzhq?file=/src/index.js)works for you?
I hope I didn't oversimplify it! My main issue is that I have a custom Select component that can be both controlled and uncontrolled and that absolutely needs a stable onChange prop (otherwise it keeps calling it with a new reference to a selected values array)
It was working fine with react-hook-form onChange method until I needed to give my component a onChange callback that would both call react-hook-form onChange and do extra logic. The only way I found was to put react-hook-form's onChange into a state owned by its parent, then call that onChange in another callback with the extra logic...

Does that sound clear? (haha I hope so, lemme know if you need more details)

Have a good one

leapful:
@PaulCordo 

I think the problem in your CSB sample is related to "useEffect" handling with "onChange". The "Select" component is watching change from its prop "value" to trigger prop "onChange" but then "onChange" handling also updates that prop "value". Usually, this handling will lead into circle side effect.

In other word, "onChange" from react-hook-form or from any other form library, which can change prop "value" of your "Select" component should not be called inside "useEffect". Just pass your other "onChange" for extra logic directly to "Select" component with "useEffect" is enough.

It would be great if you can create a topic in Discussion to have continue discussing on this issue.

rubber-duck-software:
Another reason this is important is that it fits with our performance philosophy 

> Minimizes the number of re-renders, minimizes validate computation, and faster mounting.

and our claim about isolating re-renders 

> You have the ability to isolate components re-renders which leads to better performance on your page or app.

This is a case where we currently do not allow the consumer to isolate/minimize re-renders.

mattsputnikdigital:
A quick fix, in the component where you access control and get onChange from field.

```
const handleChange = useCallback((event) => handleChangeRef.current(event), []);
  const handleChangeRef = useRef((event) => onChange(event));
  useEffect(() => {
    handleChangeRef.current = (event) => onChange(event);
  }, [onChange]);
```

mgcrea:
Just encountered this as well with a `useEffect` that had the `onChange` callback from `useController` as a dependency

Maybe we could have an option to enable stable references for `useController`. Also there could be a way to have some kind of map of field-level callbacks memoized once at the form level to prevent excessive field-level memoization.

Noitidart:
onChange and onBlur are stable now is that correct? https://github.com/react-hook-form/react-hook-form/blob/09d51cf70073efd296eac71981097efe47d83cf8/src/useController.ts#L105

I would be curious to see the results of the tests that showed useCallback is now safe to use without hurting performance at scale. I always thought it was safe but never proved it.

Noitidart:
@bluebill1049 @leapful  Gentle bump here. I'm not upset at all that my work to stabilize was not taken. I'm genuinely curious what was the data that led you guys to change your belief that is more performant? It would help me reinforce my hunch.

mattsputnikdigital:
> @bluebill1049 @leapful Gentle bump here. I'm not upset at all that my work to stabilize was not taken. I'm genuinely curious what was the data that led you guys to change your belief that is more performant? It would help me reinforce my hunch.

Somehow I got it into my head it was stable as well, maybe from another ticket being closed?

Noitidart:
> > @bluebill1049 @leapful Gentle bump here. I'm not upset at all that my work to stabilize was not taken. I'm genuinely curious what was the data that led you guys to change your belief that is more performant? It would help me reinforce my hunch.
> 
> Somehow I got it into my head it was stable as well, maybe from another ticket being closed?

Thanks @mattsputnikdigital - yeah its for sure stable now it seems based on this line of code here - https://github.com/react-hook-form/react-hook-form/blob/09d51cf70073efd296eac71981097efe47d83cf8/src/useController.ts#L105

Would be awesome if @leapful @bluebill1049 can verify if this is true. And if possible share the findings that led to this.

PR #5116, opened 2021-05-10
Correct array like object keys manipulating

artischockee:
### Context

To allow preserving initial array-like objects state.
At the current moment, the library handles array-like object an unexpected way.

Given this object as an example default value (where object keys are numbers):
```javascript
{ a: { 1: { 2: true }, 4: { 182: false } } }
```
any changes applied to, for example, checkbox named `"a.4.182"`, will transform the whole object model to an array, saving empty space between "indexes". This will look something like:
```javascript
{ a: [empty, [empty x 2, true], empty x 2, [empty x 182, false] ] }
```

### Expected behaviour

```tsx
// pseudo-code

const { register, watch } = useForm({
  defaultValues: {
    a: {
      1: {
        8: true,
        108: false,
      },
      4: {
        615: false,
      },
    },
});

render(<input type="checkbox" {...register("a.1.108")}  />);

watch(); // { a: { 1: { 8: true, 108: false }, 4: { 615: false } } }

checkbox.click();

watch(); // { a: { 1: { 8: true, 108: true }, 4: { 615: false } } }
```

_P.S._ I wasn't able to find proper solution for this case, neither in "Issues" tab, nor StackOverflow, nor the documentation, so  I decided to extend the library's existing code a little bit.

### DIFF
--- a/src/utils/set.ts
+++ b/src/utils/set.ts
@@ -15,7 +15,7 @@ export default function set(
   const lastIndex = length - 1;
 
   while (++index < length) {
-    const key = tempPath[index];
+    const key = tempPath[index].replace(/^\[/, '');
     let newValue = value;
 
     if (index !== lastIndex) {
@@ -23,7 +23,9 @@ export default function set(
       newValue =
         isObject(objValue) || Array.isArray(objValue)
           ? objValue
-          : !isNaN(+tempPath[index + 1])
+          : (tempPath[index + 1].startsWith('[') &&
+              !isNaN(+tempPath[index + 1].slice(1))) ||
+            !isNaN(+tempPath[index + 1])
           ? []
           : {};
     }
--- a/src/__tests__/utils/set.test.ts
+++ b/src/__tests__/utils/set.test.ts
@@ -56,5 +56,19 @@ describe('set', () => {
         ],
       },
     });
+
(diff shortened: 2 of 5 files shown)


src/utils/set.ts
```diff
@@ -15,15 +15,17 @@ export default function set(
...
+    const key = tempPath[index].replace(/^\[/, '');
     let newValue = value;
 
     if (index !== lastIndex) {
       const objValue = object[key];
       newValue =
         isObject(objValue) || Array.isArray(objValue)
           ? objValue
-          : !isNaN(+tempPath[index + 1])
+          : (tempPath[index + 1].startsWith('[') &&
```
bluebill1049, line 26: we no longer support `[` this is made to align with TS type check.

artischockee: Okay, I missed that in `v7`.
So, what could be the solution for the problem?

I'll try to dig around anyway, but any solutions proposed will be met warmly.

bluebill1049: I think that's a limitation, as the number is becoming array index. any chance to use character instead, by the way really appreciated sending a PR.

bluebill1049: Thank you very much! It really means a lot to me and the rest of the team.  (you made my day 🕺 )

Thank you again for your kind words, understanding, and even attempted to create a PR for a fix. ❤️ 

cc @kotarella1110 @jorisre @leapful

Issue #1840, opened 2020-06-11
V6: Remove `ErrorMessage` Component


stramel:
I'm proposing that we drop the `ErrorMessage` component.

The basis of the component is to grab the error message of a specific field and present that in a certain way normally using `as` or `render`.

```jsx
<ErrorMessage as="p" errors={errors} name="firstName" message="required!" />
```

The `as` or `render` props are essentially just wrapping a message (string/element). Wrapping a string/element in React is pretty trivial. So if we pull this out of the `ErrorMessage` component, we end up with something like this:

```jsx
<p>
	<ErrorMessage as="p" errors={errors} name="firstName" message="required!" />
<p>
```

After pulling out the wrapping component, the end-user has much more control over how the message is rendered. However, what are we left with? `ErrorMessage` is essentially just a getter for the `errors` object. With the optional chaining operator `?.`, this "getter" functionality is relatively trivial:

```jsx
<p>
	{errors?.firstName && 'required!'}
</p>
```

or 

```jsx
<p>
	{errors?.firstName?.message}
</p>
```

This again provides the end-user much more control over how they would like to render it. For example, if the user doesn't want to render the empty `<p>` tag if no error message:

```jsx
<>
	{errors?.firstName && <p>{errors.firstName.message}</p>}
</>
```

When it comes to handling multiple error messages, I believe that handling them via the `ErrorMessage` component almost complicates things more than just handling it outside.

```jsx
<ErrorMessage errors={errors} name="multipleErrorInput">
    {({ messages }) =>
        messages &&
		Object.entries(messages).map(([type, message]) => (
			<p key={type}>{message}</p>
		))
	}
</ErrorMessage>
```

vs 

```jsx
<>
	{errors?.multipleErrorInput?.types && Object.entries(messages).map(([type, message]) => (
		<p key={type}>{message}</p>
	))}
</>
```

By removing this, helper component, we could trim some size off our lib size without impacting the end-user experience of rendering errors.

Also, an end-user can still create this relatively simple abstraction that aligns closer to their needs if they wish. I have done this in previous projects that didn't have this abstraction and it worked out well.

Maybe we pull some of these non-critical helpers out into a separate library if people still feel there is value in us supporting them?

/cc @bluebill1049 @kotarella1110 @JeromeDeLeon

bluebill1049:
[...]

PR #5186, opened 2021-05-13
Message can be ReactElement or string

fabien-somnier:
**Currently**
Error messages can only be `string`

-----

**Now**
Error messages can also be `ReactElement`.
This is useful when you want to include (for example) a link inside your message.

-----

**Example**
example

where the "£5,000" can trigger a modal opening with more information, or automatically fill the input with the maximum amount.

```
<Controller
  ...
  rules={{
    max: {
      value: 5000,
      message: (
        <Trans components={[<Link as="a" onClick={() => console.log('action')} />]}>
          {"You cannot withdraw this amount. You can withdraw a maximum of <0>£5,000.00</0>"}
        </Trans>
      )
    }
  }}
/>
```

Note: `Trans` is a `react-i18next` component (see details [here](https://react.i18next.com/latest/trans-component))

### DIFF
--- a/src/types/errors.ts
+++ b/src/types/errors.ts
@@ -1,8 +1,10 @@
+import { ReactElement } from 'react';
+
 import { FieldValues, InternalFieldName, Ref } from './fields';
 import { DeepMap, LiteralUnion } from './utils';
 import { RegisterOptions, ValidateResult } from './validator';
 
-export type Message = string;
+export type Message = string | ReactElement;
 
 export type MultipleFieldErrors = {
   [K in keyof RegisterOptions]?: ValidateResult;


bluebill1049:
thanks for the PR, unfortunately i don't think we can support that: https://github.com/react-hook-form/react-hook-form/issues/1840

fabien-somnier:
> thanks for the PR, unfortunately i don't think we can support that: #1840

@bluebill1049 I don't see the relation between the `message` type (my request), and the `ErrorMessage` component from the PR you're indicating.

My request is just a typing problem. Currently, we have no choice on message type, it can only be pure string. After experimenting it in my project (thanks to `@ts-ignore`), we could also have a `ReactElement` message, giving more flexibility to it (for example: adding some link-action-tips inside it)

bluebill1049:
error message can't support multiple types, this will be a breaking change for existing users and it will not be compatible with a component which only supports `string`. I know it's convenient to support JSX, which is why I made it work for JS instead of TS. You can work around this issue by composing your own message with JSX.

fabien-somnier:
> error message can't support multiple types, this will be a breaking change for existing users and it will not be compatible with a component which only supports `string`. I know it's convenient to support JSX, which is why I made it work for JS instead of TS. You can work around this issue by composing your own message with JSX.

I don't really understand the work around you're proposing, as the main problem is that the whole RHF validation code is typed with `Message` which is a string :(. So it's not about the final render, it's mainly all the TS errors happening when setting the rules in the form.

PR #5238, opened 2021-05-18
fix-missing-register-options-generic-type

Rhodanthe1116:
## Currently

I am using Controller and Typescript. I found the `validate` function in props `rules` is not typed (`any`).

image

## Now

It has type.

image

## What's changed

I am using Controller and Typescript. I found the `validate` function in props `rules` is not typed (`any`).

image

And I trace the type definition and add types in `UseControllerProps`

image
(before)
image
(after)

And it works:

image

I am not sure it is correct to add it. If it is not need, please close this PR.

## Note

The type is correct (not `any`) in `register`. 

image

I also check that definition in `register`

<https://github.com/react-hook-form/react-hook-form/blob/102a60d3e0ed93142e7d863c3bab88ac283868a4/src/types/form.ts#L146>

### DIFF
--- a/src/types/controller.ts
+++ b/src/types/controller.ts
@@ -35,7 +35,10 @@ export type UseControllerProps<
   TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
 > = {
   name: TName;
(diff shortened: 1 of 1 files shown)


bluebill1049:
Thank you very much @Rhodanthe1116 ❤️

PR #5317, opened 2021-05-24
accept readonly field path arrays

willheslam:
I was trying to use `trigger` from `UseFormReturn` and found it wouldn't accept a readonly array (due to the array of trigger names being declared with `as const`).

A workaround would be to awkwardly apply a "mutable" type mapping to my array, but instead I've naively widened the signatures for everything that accepted
`FieldPath<TFieldValues>[]`
to now also accept
`readonly FieldPath<TFieldValues>[]`

In some cases that means adding an extra overload signature, in others just a union.

This fixed the issue in my project and, as far as I can tell, should be backwards compatible.
Please let me know if you have any questions!


bluebill1049:
Thanks very much for the PR!

bluebill1049 (approved):

PR #5350, opened 2021-05-27
5277 Update value when switch to enabled from initial disabled state

leapful:
Context:

When initialize field element with disabled state, "register" will not set internal value which is handled by react-hook-form (which is correct according to the convention to not submit disabled field value), but if the disabled state is switched to enabled, next "register" did not set the internal value correctly and leads to issue no value for "validate" callback which uses the internal value.


leapful:
@bluebill1049 It seems the new release 4.3.2 of Typescript caused errors on type checking (typescript was set "^ 4.2.3" in react-hook-form package.json).

I would like to specify typescript 4.2.4 for this pull request and create another pull request to analyze and fix type checking with newest typescript version. What do you think ?

bluebill1049:
> @bluebill1049 It seems the new release 4.3.2 of Typescript caused errors on type checking (typescript was set "^ 4.2.3" in react-hook-form package.json).
> 
> I would like to specify typescript 4.2.4 for this pull request and create another pull request to analyze and fix type checking with newest typescript version. What do you think ?

Yes, please. Thanks very much for the investigation and PR.

src/useForm.ts
```diff
@@ -998,7 +998,15 @@ export function useForm<
                 const field = get(fieldsRef.current, name) as Field;
                 const shouldUnmount =
                   shouldUnregister || (options && options.shouldUnregister);
-                field && (field._f.mount = false);
+                if (field && field._f) {
+                  field._f.mount = false;
+                  // If initial state of field element is disabled,
+                  // value is not set on first "register"
+                  // re-sync the value in when it switched to enabled
+                  if (!field._f.value && field._f.ref.value) {
```
bluebill1049, line 1006:
```suggestion
                  if (isUndefined(field._f.value) && isUndefined(field._f.ref.value)) {
```

I think check `undefined` is safer, in case of `boolean` value or negative number.

bluebill1049:
Thanks very much again @leapful let me know if you are happy with my minor modification. I will get a beta version out now as well.

leapful:
@bluebill1049 Thank you very much. Things got crazy in my side, I had a lot of critical tasks to do and forgot to update this pull request.

bluebill1049:
no worries at all, I really appreciate all your work and contribution to the library. you are an amazing "hidden" team member. thank you.

PR #5356, opened 2021-05-27
reduce logic with value check

bluebill1049:
@leapful I think we can reduce the logic if the value is undefined and set up the value. let me know if this is ok? try to save couple of bytes here.


leapful (approved):
@bluebill1049 It looks fine to me 👍

PR #5375, opened 2021-05-28
improve controller defaultValue prop type

kotarella1110:


bluebill1049:
thank you @kotarella1110

PR #5383, opened 2021-05-28
fix #5382 trigger() return incorrect valid state

### LINKED ISSUE #5382: trigger() issue with return state
Hi,

I have some nested fields in a paginated form and using `trigger` function to validate them when changing tabs.
While trying to update from version 5 to 7, I noticed that `trigger` function is returning `true` even with errors on nested fields.

I wrote a simple form to simulate it with two fields, `firstname` and `last.name`. With just the not nested field filled, the `trigger()` returns `true` but `trigger('last.name')` returns `false`, and the error object is like: 
```js
{
  last: {
    name: {
      type: "required",
      message: "", 
      ref: <input name="last.name" placeholder="Last name"></input>
    }
  }
}
```

[codesandbox example](https://codesandbox.io/s/react-hook-form-js-forked-hpp9l?file=/src/App.js)

Is `trigger()` supposed to validated all fields, even those registered like `{...register('last.name', { required: true })}`, or I got it wrong?
 
Thank you very much for your time.

_Originally posted by @andrepxa in https://github.com/react-hook-form/react-hook-form/discussions/5337_

bluebill1049:
ref: https://codesandbox.io/s/rhf-trigger-validation-example-forked-8r33k?file=/src/App.js:113-977

### DIFF
--- a/src/__tests__/useForm/trigger.test.tsx
+++ b/src/__tests__/useForm/trigger.test.tsx
@@ -465,4 +465,39 @@ describe('trigger', () => {
       screen.getByText('true');
     });
   });
+
+  it('should return correct valid state when trigger the entire form with build in validation', async () => {
+    let isValid;
+
+    function App() {
+      const { register, trigger } = useForm();
+
+      const onTrigger = async () => {
+        isValid = await trigger();
+      };
+
+      return (
+        <form>
+          <input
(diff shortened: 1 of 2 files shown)


whj1995:
v7.8.5, trigger() return incorrect valid state. @bluebill1049

PR #5391, opened 2021-05-28
bubble up error when thrown in submitHandler

bgandy94:
This PR is to address an issue where when an error is thrown during the `onValid` parameter passed in to `handleSubmit.

This issue is primarily caused when you want to call `handleSubmit` remotely and also `try/catch` within it.

Currently, if you call it remotely, it swallows the error and the caller has no insight as to what error occurred..

The code didn't use to have a `catch` block, but I believe [this change](https://github.com/react-hook-form/react-hook-form/commit/82f2e4be80a0a4e625b6492ca3df2e8a75707093) plus the modifications I've made here would allow for valid `onSubmitSuccessful` and also handling errors with a remote caller.

I'm open to other ways of doing this as well, possibly such as some kind of prop we pass in that would allow for the behavior to be changed in the event there's no need for the caller to have insight into the error.

Either way, it's mainly a discussion point and I wanted to gather other's thoughts.

Thanks in advance!

### DIFF
--- a/src/__tests__/useForm/formState.test.tsx
+++ b/src/__tests__/useForm/formState.test.tsx
@@ -226,7 +226,39 @@ describe('formState', () => {
     });
   });
 
-  it('should only set isSubmitSuccessful to false when there is a promise reject', async () => {
+  it('should bubble the error up when an error occurs in the provided handleSubmit function', async () => {
+    const errorMsg = 'this is an error';
+    const App = () => {
+      const [error, setError] = React.useState('');
+      const { register, handleSubmit } = useForm();
+
+      const rejectPromiseFn = jest.fn().mockRejectedValue(new Error(errorMsg));
+
+      return (
+        <form
+          onSubmit={() =>
+            handleSubmit(rejectPromiseFn)().catch((err) =>
+              setError(err.message),
+            )
+          }
+        >
+          <input {...register('test')} />
+          <p>{error}</p>
+          <button>Submit</button>
+        </form>
+      );
+    };
+
+    render(<App />);
+
+    await actComponent(async () => {
+      fireEvent.click(screen.getByRole('button'));
+    });
+
+    screen.getByText(errorMsg);
+  });
+
+  it('should set isSubmitSuccessful to false when there is a promise reject', async () => {
     const App = () => {
       const {
         register,
@@ -234,12 +266,12 @@ describe('formState', () => {
         formState: { isSubmitSuccessful, isSubmitted },
       } = useForm();
 
+      const rejectPromiseFn = jest
+        .fn()
+        .mockRejectedValue(new Error('this is an error'));
+
       return (
-        <form
-          onSubmit={handleSubmit(async () => {
-            throw new Error('something is wrong');
-          })}
-        >
+        <form onSubmit={() => handleSubmit(rejectPromiseFn)().catch(() => {})}>
           <input {...register('test')} />
           <p>{isSubmitted ? 'isSubmitted' : 'no'}</p>
           <p>
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -1072,8 +1072,9 @@ export function useForm<
               fieldsNamesRef.current,
             );
         }
-      } catch {
+      } catch (err) {
         hasNoPromiseError = false;
+        throw err;
       } finally {
         formStateRef.current.isSubmitted = true;
         formStateSubjectRef.current.next({


bluebill1049:
Thanks @bgandy94 for the PR, this sounds good to me. I had a similar concerns for the above as well.

mAAdhaTTah:
@bluebill1049 @bgandy94 This is a problem for our use case: We want to indicate to RHF that the submit was not successful, which we could do by throwing, but now that the error bubbles it up, it gets surfaced in ways we don't need it to be (because it's really already handled). Is there a canonical way to indicate that the submit was unsuccessful without bubbling unnecessary errors?

bgandy94:
> @bluebill1049 @bgandy94 This is a problem for our use case: We want to indicate to RHF that the submit was not successful, which we could do by throwing, but now that the error bubbles it up, it gets surfaced in ways we don't need it to be (because it's really already handled). Is there a canonical way to indicate that the submit was unsuccessful without bubbling unnecessary errors?
> 
> Edit: To explain my use case, I have several forms on multiple pages. I want to disable the button in the following scenarios:
> 
> 1. `!isValid` - obviously, don't want to allow submission if the form is invalid.
> 2. `isSubmitting` - prevent users from resubmitting an API request in flight.
> 3. `isSubmitSuccessful` - this is more unique to our setup. we're using NextJS so after completing a form, there's a slight delay before the page transitions. we want to use `isSubmitSuccessful` to disable the button so it doesn't reenable while NextJS is downloading the data for the next page.

Hi James, hope all is well.. So a couple thoughts that come up right off the top of my head would be:
1. When you say  "because it's really already handled", I wouldn't consider wrapping something in a try/catch by itself handling an error. Not only that, but also the original implementation would swallow the error, not allowing the submitHandler any ability or insight into what error occurred.
2. I do agree that this does allow for errors to be bubbled up in such a way that could cause issues, assuming your handler isn't called remotely and wrapped in a try/catch.

My goal for this specific PR was to get the code back in a state where it was, allowing callers to have insight into submit errors while also keeping the `isSubmitSuccessful` logic.

I think, for now, one option you do have until this can be hashed out further, would be to wrap the submit handler in your components in a try/catch and call your submitHandler remotely, as opposed to calling it directly from your submit button.

As far as a future state, the only thing I can think to do would be to possibly expose another value from the hook for `submitError` so that we can avoid bubbling the error, but also allow a component to be able to not only know that an error occurred (whether it's through checking the `isSubmitSuccessful` and/or evaluating the `submitError` value), but also be able to see the error that occurred. The thing I'd want to make sure we do is stay consistent with other values to make sure that this new value's lifecycle (ie resetting, clearing, etc...) is also thought through during the implementation.

I would suggest opening a PR with these things in mind, as I feel like you can get the best of both worlds here. I'm sure @bluebill1049 will have some helpful feedback to your concerns as well.

mAAdhaTTah:
> When you say "because it's really already handled", I wouldn't consider wrapping something in a try/catch by itself handling an error. Not only that, but also the original implementation would swallow the error, not allowing the submitHandler any ability or insight into what error occurred.

Yeah, at the time, I considered the swallowing a "feature" or expected behavior because I'm only throwing for the purpose of notifying RHF that the submission was not successful. We are explicitly handling all of the error cases in our program, currently, so I don't need RHF to deal w/ it or bubble it up; we've got it covered.

> I do agree that this does allow for errors to be bubbled up in such a way that could cause issues, assuming your handler isn't called remotely and wrapped in a try/catch.

Yeah, exactly, and I kind of expect the "remote call" behavior to be less common than simply passing it to `onSubmit` on a `form` element (although I could be wrong).

> I think, for now, one option you do have until this can be hashed out further, would be to wrap the submit handler in your components in a try/catch and call your submitHandler remotely, as opposed to calling it directly from your submit button.

I've got at least a dozen+ forms that I'd have to wrap in this way, which seems onerous, so while this is feasible, I'd prefer not to have to do that. It's no less hacky than my other approach of doing a `setTimeout` & `reset` `isSubmit` to `false` after the fact.

> As far as a future state, the only thing I can think to do would be to possibly expose another value from the hook for `submitError` so that we can avoid bubbling the error...

Would a config option `rethrowSubmitErrors` be acceptable? Might be the simplest implementation. @bgandy94 @bluebill1049? Would be happy to PR that if so.

bgandy94:
> Would a config option rethrowSubmitErrors be acceptable? Might be the simplest implementation. @bgandy94 @bluebill1049? Would be happy to PR that if so.

I actually think that'd be another option, main question is what it'll default to. I'm guessing to keep the change as non-breaking as possible, it might be good to default it to true.

I'd say open the PR, I'm not an authority figure by any means on the library, but I don't see that as being an unreasonable middle ground.

bluebill1049:
is this an option to consider? `handleSubmit().catch()` i think we already have too many configs, it would be the last config to introduce another.

mAAdhaTTah:
@bluebill1049 That "works", but I would have go from `onSubmit={handleSubmit(onSubmit)}` to `onSubmit={(e) => handleSubmit(onSubmit)(e).catch(() => {})}` on _every_ form, which is quite tedious.

bluebill1049:
have a reusable function, so it's not as tedious as above.

mAAdhaTTah:
@bluebill1049 I could do that, but it still seems weird to me that there's no way to prevent `isSubmitSuccessful` from turning to true besides throwing an error and wrapping `handleSubmit` like this. What's the purpose/use of that boolean?

bluebill1049:
The purpose is when u submit your form without an error and pass handleSubmit

bluebill1049:
The real question is how far hook form should reach for form success? should we reach all the way from server response? My understanding we want to leave at the form level, if we need to handle server response, then we probably need more work to get it right. at the moment, hook form set `isSubmitSuccessful` as long as `handleSubmit` gets invoked without build-in validation error or resolver error. I think bubble up the error seems to be a better option than swallow it at the moment, and like I said I prefer not to introduce another config, unless someone can convince all of us that swallow that error is preferred usage for more use cases.

mAAdhaTTah:
> The purpose is when u submit your form without an error and pass handleSubmit

One thing that might help here is distinguishing "expected" & "unexpected" errors. `isSubmitSuccessful` turns `true` if there are no unexpected errors, but there isn't currently a great way to indicate that the form failed to submit because of an expected error. We have `isSubmitted` & `isSubmitSuccessful`, but neither of them fully cover this case.

> My understanding we want to leave at the form level, if we need to handle server response, then we probably need more work to get it right. at the moment, hook form set `isSubmitSuccessful` as long as `handleSubmit` gets invoked without build-in validation error or resolver error.

... or errors from `onValid`. With `isSubmitSuccessful`, RHF is now managing state related to what happens after submit, whether that's server response, db saving, or whatever. It works by default on unexpected errors throw in `onValid` but there isn't any way to express "an expected error happened here, treat submit as unsuccessful". We originally tried to do it with `reset` but because it's called before the function completes, the post-`onValid` logic overwrites whatever we do. So now we're calling `reset` in a `setTimeout`, which is really kludgey. You could also set some logic

I don't know that not rethrowing is the best solution here, because in that case, you end up swallowing unexpected errors too, which isn't great, but there isn't a good way of expressing what I'm trying to with `isSubmitSuccessful`, and maybe solving that problems suggests an approach?

bluebill1049:
> One thing that might help here is distinguishing "expected" & "unexpected" errors. isSubmitSuccessful turns true if there are no unexpected errors, but there isn't currently a great way to indicate that the form failed to submit because of an expected error. We have isSubmitted & isSubmitSuccessful, but neither of them fully cover this case.

I think this is a more reasonable approach, if there is an error been bubbled, we could omit isSubmittedSucessful to be `true`, would you like to send us a PR? @bgandy94 let us know your thoughts as well.

mAAdhaTTah:
I'm not sure I follow. "bubbled" from where?

bgandy94:
Yeah I think if we can a little more clearly define the change being proposed, I could have more thoughts.. I do agree with what's being said for handled errors within the callback being provided by the caller, but I'm not sure I understand the proposed solution.

bluebill1049:
if we `catch` the `error`, then we do not set `isSubmittedSucessful` to be `true`.

bgandy94:
Isn't that how it currently works?

bluebill1049:
oops.. you are right @bgandy94 https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L1077

bgandy94:
So I think we shortfall being mentioned is being able to handle the error within your provided submit handler, but still be able to indicate that the submission was unsuccessful. I could be off base, but that's what I think I'm understanding

mAAdhaTTah:
> So I think we shortfall being mentioned is being able to handle the error within your provided submit handler, but still be able to indicate that the submission was unsuccessful.

Yes exactly. I don't have a proposed solution, tbh, but here's a couple ideas off the top:

1. Throw an error with a specific code/message/property/type in `onValid`, which RHF knows not to rethrow.
2. Return a specific string/symbol/object structure/reference, e.g. `return { success: false }`.
3. Call a function to indicate submit failure (note: `setError` doesn't work here because that also marks `isValid == false`, which isn't the case. The form is valid, but the submission failed.)
4. Modify RHF such that an error of type `submit` or `server` doesn't mark the form as invalid and use `setError`. (This seems likely to break someone's usage though).

Thoughts?

bgandy94:
> Throw an error with a specific code/message/property/type in onValid, which RHF knows not to rethrow, e.g. throw new SubmitError('server failure!').

I'm not typically a fan of conventions such as these, mainly because it's implicit and its implications aren't entirely obvious on its face.

> Return a specific string/symbol/object structure/reference, e.g. return { success: false } or return Symbol('rhf#submitUnsuccessful').

This one seems more explicit in its implications, but I'm just seeing a Typescript need to define some sort of new type that wraps your handler's return type to make typescript recognize that your handler could return either the non-error return of your function, or this new shape. (this is assuming you define a return type). Not necessarily a deal-breaker, but something to consider.

> Call a function to indicate submit failure (note: setError doesn't work here because that also marks isValid == false, which isn't the case. The form is valid, but the submission failed.), e.g. setSubmitFailed().

I was kind of leaning towards this one, as I think it would be the most explicit and obvious. I think the only thing I was thinking was trying to only expose the function within the handler, by passing it into the handler defined by the caller. The thing I worry about this is having to define a 3rd prop into the `handleSubmit` function, thereby forcing you to define the `SubmitErrorHandler` prop. And idk about most people, but my use case for this second function is very rare. I guess one way around this would be to just return it as a usable function from `useForm`, I just worry about race conditions and also over exposure of the function in this case, but it would get rid of the need to pass a 3rd function into the `handleSubmit` function returned by `useForm`

> Modify RHF such that an error of type submit or server doesn't mark the form as invalid and use setError. (This seems likely to break someone's usage though).

I have similar concerns to yours on this approach as well.

mAAdhaTTah:
> I'm not typically a fan of conventions such as these, mainly because it's implicit and its implications aren't entirely obvious on its face.

Fair enough. I will mention this is redux-form's approach, although I don't know if that's a point for or against...

> This one seems more explicit in its implications, but I'm just seeing a Typescript need

The plus side is the return type of the SubmitHandler is `any | Promise<any>`, so I don't think we'd break anyone by adding `| { success: false } | Promise<{ success: false }>` (or `submitError: true` or whatever). I do think the use of `any` there may swallow things a bit but I'm not 100% sure.

> I guess one way around this would be to just return it as a usable function from `useForm`, I just worry about race conditions and also over exposure of the function in this case,

This was my thought when I wrote it, it would provide full control over `isSubmitSuccessful` but 1. is _another_ API and 2. there would be 2 mechanisms for manipulating that state, which may be less than ideal.

My current preference is for 2, as I think that's the minimal new API surface area to solve this problem, but I'm still open.

bgandy94:
> The plus side is the return type of the SubmitHandler is any | Promise<any>, so I don't think we'd break anyone by adding | { success: false } | Promise<{ success: false }> (or submitError: true or whatever). I do think the use of any there may swallow things a bit but I'm not 100% sure.

Valid points! This would be the least breaking.. We'd just have to very clear in the docs about the convention (which I'm sure is obvious lol).. 

Let's let @bluebill1049 get caught up, and see what he thinks..

bluebill1049:
Thanks for all the suggestions. All those solutions above make things more complicated and hard to document for this specific behavior. I will update the doc to make the current behavior more clear to everyone. If you have concerns over the bubble up for the error, I would suggest to fork this repo and publish your version to solve the problem.

mAAdhaTTah:
@bluebill1049 I respect that this is your library and ultimately, you decide what goes into it, so I'll make one last case before taking my own approach internally: given that RHF provides state for several "post-submission" actions (`isSubmitting`, `isSubmitted` & `isSubmitSuccessful`), the lack of a mechanism for indicating "expected failure" makes the feature incomplete, as it covers several "post submit" cases except that one. The library already reaches into server submission by providing these states as it is, so I do think we need a solution for this case as well (or honestly, remove the post-submit states, which could be considered out of scope for a form lib).

Entirely understandable if you disagree, so I don't want to be pushy, but I ask you to reconsider one more time before we close this out.

bluebill1049:
any thoughts on the approach

```
catch (err) {
  if (onInvalid) {
    onInvalid(_formState.errors, err);
  } else {
    throw err;
  }
  hasNoPromiseError = false;
}
```

if user provided a invalid callback then, we pass that error, otherwise throw to the user?

mAAdhaTTah:
According the docs, the second value passed to `onInvalid` is an event object currently, so I think that's breaking, unfortunately.

PR #5415, opened 2021-05-30
Improve isValid form State by removing extra internal formState

bluebill1049:
## Description

`isValid` formState is currently driven by two pieces of state:

- `fieldsWithValidation`
- `validFields`

why? This is purely considering on the aspect of performance, as react hook form track single input at the time. By doing a deep comparison with the two states. we can work out the `isValid` state efficiently. 

however, this has its own trade-off as we are tracking multiple states to drive the result of isValid. so summarised what's current in the lib.

-----

## Changes

We will remove those extra pieces of states, and tracking the isValid by validate the entire form and abort early. This is going to be less performant, however, it pretty much covers reverse the props and cons above, which personally I believe is the better direction for a single source of truth for `isValid` state.

### Cons

- Less performant 

### Props

- Less error-prone
- Reduce extra states for tracking isValid
- Weight lost

----

## Improvements

- consistent fields names and fields for `resovler`
- isValid will always return `false` to start with instead of `true`


src/useForm.ts
```diff
@@ -1123,15 +1071,16 @@ export function useForm<
...
         dirtyFields: keepDirty ? formStateRef.current.dirtyFields : {},
         touchedFields: keepTouched ? formStateRef.current.touchedFields : {},
         errors: keepErrors ? formStateRef.current.errors : {},
         isSubmitting: false,
         isSubmitSuccessful: false,
       });
+
+      if (!keepIsValid) {
+        isMountedRef.current = false;
+      }
```
JamesWong1999, line 1083:
```suggestion
      isMountedRef.current = !keepIsValid;
```

Issue #2659, opened 2020-08-20
Display automatically an error message


sonatard:
**Is your feature request related to a problem? Please describe.**

I don't want to add `ErrorMessage` Component in my React Component.

**Describe the solution you'd like**
HTML 5 input attribute validation.
https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation

HTML5 form validation display error message automatically.

![error-firefox-win7](https://user-images.githubusercontent.com/1586910/90758480-10478980-e31a-11ea-8df1-9e562e2d7925.png)

If I set validation and message to React Hook Form like following code, I want to display HTML5 form validation error message like the above image.

```tsx
      <input
        id="password"
        name="password"
        ref={register({
          required: "required",
          minLength: {
            value: 5,
            message: "min length is 5"
          }
        })}
        type="password"
        setCustomValidity // Option: Display automatically an error message or not
      />
```

**Additional context**
setCustomValidity Function enable to change HTML5 form validation error message.

bluebill1049:
we did that in V3: https://react-hook-form.com/v5/api#Browserbuiltinvalidation but the feature never pick up in the community, so probably not going to reimplement it anytime soon, but will tag it with a feature request.

sonatard:
I see.  Thank you for tagging it as a feature request.

bluebill1049:
no worries @sonatard once the upvotes is increased, we will bring it back.

bluebill1049:
Please cast your vote: https://twitter.com/HookForm/status/1402189525057638413

CSFlorin:
@bluebill1049 shouldUseNativeValidation seems to not use the validationRule's message. This could be implemented using the setCustomValidity API @sonatard originally mentioned. It's odd because it seems you do use that API in https://github.com/react-hook-form/react-hook-form/blob/dd547079dc12b55fa44bd0135a94a8f5d595dfcc/src/logic/validateField.ts#L54 but I don't see it working in version 7.32.1

CSFlorin:
@bluebill1049 shouldUseNativeValidation seems to trigger even when there isn't a validation error yet. Even if I change mode and revalidateMode to onSubmit. Opening the form for the first time already shows the native validation message whereas I only intend to show it when there is an error. Since there's no error, it just shows the default message. Also it seems to apply some new rule that doesn't allow me to submit even when it is correct. Only removing shouldUseNativeValidation fixes it.

PR #5503, opened 2021-06-06
feature: support browser native form control

bluebill1049:
## Description

This is an opt-in option, which allows users to apply native constrain API. This feature is scoped to web platform only, and the API is using under the hook is [Constraint validation API](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation).

What will this unlock:

- Browser native CSS selector for `:invalid`, `:valid`
- Browser native message pop up (this can be opt-out as well)
- Validation `mode` will still be applied and respected
 
```tsx
useForm({
  shouldUseNativeValidation: boolean
})
```

## Props

- one step close to HTML standard

## Cons

- Increate package size

## Limitations

- limitation on `onSubmit` mode to register those input validation message
- limitation on build-in validator only.
- Web only

## Feature request

#2659

## Codesandbox

ref: https://codesandbox.io/s/pensive-ritchie-5f91e

## Video

The following example is using the `:invalid` without any external CSS names.

https://user-images.githubusercontent.com/10513364/120920614-0b19c680-c703-11eb-9888-91320798137d.mov

## Tasks

- [x] unit tests
- [x] poll on config name: https://twitter.com/HookForm/status/1402189525057638413
- [ ] doc update

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -93,6 +93,7 @@ export function useForm<
   context,
   defaultValues = {} as DefaultValues<TFieldValues>,
   shouldFocusError = true,
+  shouldUseCustomValidity = false,
   shouldUnregister,
   criteriaMode,
 }: UseFormProps<TFieldValues, TContext> = {}): UseFormReturn<TFieldValues> {
@@ -356,6 +357,7 @@ export function useForm<
         await validateField(
           get(fieldsRef.current, name) as Field,
           isValidateAllFieldCriteria,
+          shouldUseCustomValidity,
         )
       )[name];
 
@@ -412,6 +414,7 @@ export function useForm<
           const fieldError = await validateField(
             field,
             isValidateAllFieldCriteria,
+            shouldUseCustomValidity,
           );
 
           if (shouldCheckValid) {
@@ -704,9 +707,13 @@ export function useForm<
 
           isValid = isEmptyObject(errors);
         } else {
-          error = (await validateField(field, isValidateAllFieldCriteria))[
-            name
-          ];
+          error = (
+            await validateField(
+              field,
+              isValidateAllFieldCriteria,
+              shouldUseCustomValidity,
+            )
+          )[name];
         }
 
         !isBlurEvent &&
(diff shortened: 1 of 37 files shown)


src/useForm.ts
```diff
@@ -92,6 +92,7 @@ export function useForm<
   context,
   defaultValues = {} as DefaultValues<TFieldValues>,
   shouldFocusError = true,
+  shouldUseCustomValidity = false,
```
jorisre, line 95:
```suggestion
  shouldUseNativeValidation = false,
```

Shouldn't be `shouldUseNativeValidation `?

jorisre (dismissed):
LGTM! Great work 👏🏻

jorisre:
Another question, with the use of resolvers, we can't use `criteriaMode: 'all'` and `shouldUseNativeValidation: true` at the same time right? Maybe we can make it typesafe, what's your thought?

bluebill1049:
> Another question, with the use of resolvers, we can't use `criteriaMode: 'all'` and `shouldUseNativeValidation: true` at the same time right? Maybe we can make it typesafe, what's your thought?

yes! 100% agree.

jorisre:
Ok got it, I'll open a PR for resolvers. Is that new feature only for the v7?

bluebill1049:
> Ok got it, I'll open a PR for resolvers. Is that new feature only for the v7?

yes.

bluebill1049:
I will release a beta tonight @jorisre

bluebill1049:
`v7.8.4-beta.0` beta

bluebill1049:
Screen Shot 2021-06-11 at 9 07 41 am

got a strange test case failure by `resolver.test.tsx`, hence the above commit. strange one...

bluebill1049:
beta released `7.8.4-beta.0`

jorisre:
> Screen Shot 2021-06-11 at 9 07 41 am
> 
> got a strange test case failure by `resolver.test.tsx`, hence the above commit. strange one...

Seems good to me 🤔 
Screenshot 2021-06-15 at 09 23 37

bluebill1049:
@jorisre strangley, i fixed the issue with this commit: 89660247724331d12f22872439d7e3904748cca6

PR #5592, opened 2021-06-13
Avoiding FormProvider performance implication on library hooks

orenklein:
Current FormProvider implementation brings performance implication to most of the hooks:
`useFormState`
`useController`
`useWatch`
`useFieldArray`

These hooks use `control` internally, either taken from the function property or from `useFormContext` which updates frequently. (whereas these hooks do not use any of `useFormContext` other properties)

The big question is - are there reasons to not split `control` from the other form context properties? It will isolate updates only to necessary ones (I'm assuming the answer is not, due to the fact you can use these hooks without form context)
Because `control` is always the same object reference it will never cause unnecessary render for the component using the needed hook.

Perhaps, it is planned for the future.

Implementation wise - obviously, there are many ways to achieve that. I added the `FormControlProvider` into `FormProvider`, since it will be as seamless as using previous versions, but with the performance improvement :)

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -12,7 +12,7 @@ import {
   UseControllerProps,
   UseControllerReturn,
 } from './types';
-import { useFormContext } from './useFormContext';
+import { useFormControl } from './useFormContext';
 import { useFormState } from './useFormState';
 
 export function useController<
@@ -28,7 +28,7 @@ export function useController<
   TFieldValues,
   TName
 > {
-  const methods = useFormContext<TFieldValues>();
+  const controlContext = useFormControl<TFieldValues>();
   const {
     defaultValuesRef,
     register,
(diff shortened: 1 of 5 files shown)


bluebill1049:
Thanks for the PR, i don't think we want to introduce another hook. Also react 18 is going to improve context perf, so we may get at for free on the next major release.

orenklein:
I don't think another hook should be introduced either. It meant to be an internal hook only for performance purposes. The performance gain is quite significant and seamless with minimum effort (either by the package or by the developer).
Currently, using FormProvider forces these performance implications on the other hooks. 
For example, if you "subscribe" in any component to `formState.errors` / `formState.isDirty` / etc.. the abovementioned hooks will unnecessarily rerender all the components using them. (by subscribe I mean - having `const { errors } = formState`)
I haven't found anywhere details about improvements React 18 going to bring to the context API, do you have any resource?

bluebill1049:
thanks for the reply, let's keep this PR open for awhile until I have more time to test against react 18.

smashercosmo:
This looks like a reasonable change

bluebill1049:
pump.

bluebill1049:
7.13.0 should covert this PR.

Issue #4751, opened 2021-04-13
V7: Add onChange and onBlur options to register()


kensukesaito:
**Is your feature request related to a problem? Please describe.**

In V6, I was able to use the following, but in V7, it is no longer available, which is inconvenient.

There is a workaround as follows, but it is tedious and not beautiful.

It would be nice to be able to use the onChange and onBlur options in register().

```jsx
// V6
<input type="text" name="example" ref={register({ required: true })} onChange={(event) => { ... }} onBlur={(event) => { ... }} />
```

**Describe the solution you'd like**

```jsx
// Wish
<input type="text" {...register('example', {
  required: true,
  onChange: (event) => { ... },
  onBlur: (event) => { ... },
})} />
```

**Describe alternatives you've considered**

```jsx
// Workaround
const { onChange, ...restProps } = register('example')

<input {...restProps} onChnage={(event) => {
  onChange(event)
  // do something
}} />
```

bluebill1049:
Good feature request!

msheakoski:
I ran into this issue as well after upgrading from 6 to 7. I did not realize that my input's onChange was conflicting with register()'s. @kensukesaito's API suggestion looks good to me! 👍

Sannnao:
I think it could be possible to also add a possibility to pass a custom ref

Issue #5428, opened 2021-06-01
custom onChange not working for ts


gankedbybus:
**Describe the bug**
when using the register explicitly, the inbuilt onChange will not work and update the state

**To Reproduce**

1. explicitly register element via `const firstName = register("firstName");`
2. add ref and onchange to element 
` <input
      onChange={firstName.onChange}
      ref={firstName.ref}
      placeholder="First Name"
/>`
3. type in any input into the element e.g "test"
4. hit submit and see firstName is still empty

**Codesandbox link (Required)**
Include a codesandbox will help us to investigate the issue quicker.
https://codesandbox.io/s/dark-field-cy9jo?file=/src/App.tsx

**Expected behavior**
expected firstName to equal "test" when submitted

**Screenshots**

**Desktop (please complete the following information):**

- OS: Windows
- Browser Chrome

gankedbybus:
figure out the problem. `name={firstName.name}` needs to be added to the prop list. probably should be added in the documentation here  
![image](https://user-images.githubusercontent.com/42853427/120251625-77409880-c2c5-11eb-992d-08bfe86e0320.png)

PR #5620, opened 2021-06-15
feat: Ability to provide custom onBlur and onChange handlers

michalbundyra:
Related to #4977 and #4751

If we want to provide custom `onBlur` or `onChange` we can just do:

```jsx
<input type="text" {...register('test', {
    onChange: (e) => {
        // custom onChange handler
    },
    onBlur: (e) => {
        // custom onBlur handler
    },
})}/>
```


src/__tests__/useForm/register.test.tsx
```diff
@@ -1014,4 +1014,68 @@ describe('register', () => {
...
+    });
+
+    expect(onChange).toBeCalledTimes(1);
+    expect(onChange).toBeCalledWith(
+      expect.objectContaining({
+        bubbles: true,
+        cancelable: false,
+        currentTarget: null,
+        type: 'change',
+      }),
```
michalbundyra, line 1048: I am not sure what is the better way to check if we are getting the correct event

src/types/validator.ts
```diff
@@ -39,4 +39,6 @@ export type RegisterOptions<
   value: FieldPathValue<TFieldValues, TFieldName>;
   setValueAs: (value: any) => any;
   shouldUnregister?: boolean;
+  onChange?: (event: any) => void;
```
michalbundyra, line 42: wasn't sure if it should be void or maybe `Promise<boolean | void>`

src/useForm.ts
```diff
@@ -928,8 +928,22 @@ export function useForm<
         ? ({ name: name as InternalFieldName } as UseFormRegisterReturn)
         : {
             name,
-            onChange: handleChange,
-            onBlur: handleChange,
+            onChange: options.onChange
+              ? (event: any) => {
+                  const result = handleChange(event);
```
michalbundyra, line 933: per provided examples, I am calling react-hook-form handler first, and then the custom one, not sure if it makes any difference?

Maybe we can just do:
```typescript
options.onChange(event);
return handleChange(event);
```

src/useForm.ts
```diff
@@ -928,8 +928,22 @@ export function useForm<
...
+                  options.onChange && options.onChange(event);
```
michalbundyra, line 934: imo this condition is redundant as it is already inside `options.onChange` but linter and tests were complaining, that `options.onChange` might be undefined.

michalbundyra (commented):
@bluebill1049 This is my attempt of adding custom handlers for onBlur and onChange. 

See my comments below and please let me know what do you think. Thanks

bluebill1049:
Thanks for the PR, I am still keeping an eye on the feature request before the implementation itself. If you need this feature badly for your app, I suggest to fork the repo and publish your own for now. The current state of the lib is pretty stable so I wouldn't be so worried about updates.

bluebill1049:
pump.

cwagner22:
Any update? I spent hours migrating to v7, but now I realized I'd have to spend more time moving the register out of my components to be able to use my custom onChange callbacks. Btw see https://github.com/react-hook-form/react-hook-form/issues/5428, we also need to add `name={firstName.name}` which is missing in the doc for custom onChange

bluebill1049:
thanks for the contribution and patient @michalbundyra will push that into the next minor update.

bluebill1049:
@michaelbirchler FYI: got few bugs to roll out before I merge this PR since I just saw you in the issue.

michalbundyra:
@bluebill1049 no problem, thanks 👍

bluebill1049:
doc update: https://github.com/react-hook-form/documentation/pull/668

PR #5659, opened 2021-06-17
refactor formMethods and control object (use more js and less react for performance reasons)

mh-alahdadian:
split one of custom hook logics and use pure-js for better performance and better control of breaking logics down

next changes will be split all logics and move core logics to a single file and other hooks could be wrappers around it 
and also writing less `.current` for refs will decrease bundle size 
even it could be possible to completely break logic into separate repo named `@hookform/core` and some hook utilities around core named `@hookform/react`

also in future maybe we could export `createForm` function to be used out of react components

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import { useEffect, useRef, useState } from 'react';
 
 import focusFieldBy from './logic/focusFieldBy';
 import getFieldsValues from './logic/getFieldsValues';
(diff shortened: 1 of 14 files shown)


MHA15:
in third commit there is a braking change in using control object
control object doesn't have many ref objects inside itself instead we can read latest value by getting it from control 
it means these are equal `control.x.current` is now `control.x` and also for accessing last value you should pass control itself and we can not pass `control.x` as a ref object anymore

MHA15:
I just realized that one of the most important breaking changes of this PR is that useForm methods is equal per renders and this is a huge performance improvement

MHA15:
@bluebill1049 
I would be happy if you review this PR and let me know

MHA15:
It could be a good idea to remove control setters which are not used/should not used , but I'm not sure about them
And also in future whole core logics could be integrated in createForm and can change whole control type (mostly have internal methods instead of exporting private values)

bluebill1049:
I will get back to you.

MHA15:
any update ?

bluebill1049:
still got a few more PR ahead of this one, will get to it.

MHA15:
> still got a few more PR ahead of this one, will get to it.

I think that would be a good idea to read PR description and feedback to it before really reviewing code of it

bluebill1049:
Thanks very much for the amazing work! I finally get time to look into it a bit more this time. I would recommend having a discussion next time before a big change like this.

MHA15:
Yeah I stop continuing branch to have discussion first, and as I said third commit has all braking changes and if we don't want them we could merge only first two commits

bluebill1049:
Definitely some interesting and great ideas here! Once this [PR](https://github.com/react-hook-form/react-hook-form/pull/5958) is solid and merged and I will come back to this. Maybe it's worth doing some performance comparison.

>  it could be possible to completely break logic into separate repo named @hookform/core and some hook utilities around core named @hookform/react

Love this idea above as well.

bluebill1049:
> all braking changes

yea, probably avoid breaking changes for now.

src/useForm.ts
```diff
@@ -82,21 +82,23 @@ import {
...
 const isWindowUndefined = typeof window === 'undefined';
 
-export function useForm<
+const defaultOptions = {
+  mode: VALIDATION_MODE.onSubmit,
+  reValidateMode: VALIDATION_MODE.onChange,
+  shouldFocusError: true,
+} as const;
+
+export function createForm<
```
bluebill1049, line 91: I really like this approach.

bluebill1049:
Do you mind rever back to the second commit? or else I can do that later as well.

bluebill1049:
never mind, i will sort it out later 👍

MHA15:
But this control refactor is a good approach and can help us to receive the vision I told in first commit 
Is there any problem with this breaking change ? let's edit other usages too. 
Or  do you want to let this refactor to merge right before V8  ?

bluebill1049:
Don't worry i will sort it out.

bluebill1049:
If you want to review this PR: https://github.com/react-hook-form/react-hook-form/pull/5987

MHA15:
should we close this merge request ?

bluebill1049:
> should we close this merge request ?

yes, plz

Issue #5555, opened 2021-06-10
V7 watch + fieldArray + controlledField unable to set defaultValues on append


Raphaww:
**Describe the bug**
When appending to a fieldArray, defaultValues are set to null for SOME controlledFields only when using watch in the same component. If watch is removed the fields are initialized correctly. It looks like the problem happens only when using nested names like `name.${index}.attribute.foo`

**Codesandbox link (Required)**
V7
- https://codesandbox.io/s/youthful-williamson-rxnzo?file=/src/TextForm.js

**To Reproduce**
1. click 'add component'

**Expected behavior**
the radiobutton group named 'Estilos' should be initialized with 'Default' 
the input named 'Nombre' should be initialized with 'style name'

**Screenshots**

Bug
<img width="266" alt="Screen Shot 2021-06-10 at 4 30 20" src="https://user-images.githubusercontent.com/7801116/121501257-ae3d3b00-c9a4-11eb-9df3-c9f9a87ca11c.png">

Expected
<img width="272" alt="Screen Shot 2021-06-10 at 4 30 03" src="https://user-images.githubusercontent.com/7801116/121501925-4a674200-c9a5-11eb-94b2-df9ee3130659.png">

**Desktop (please complete the following information):**

- OS: [Mac Os Big Sur]
- Browser [Chrome]
- Version [Version 91.0.4472.77 (Official Build) (arm64)]

**Additional context**
- I include a uncontrolled component (Color) to show that uncontrolled are not affected
- I include a controlled component (Nombre Estilos) with a not nested name (name${index}.attribute) to show that those are not affected either
- switching from watch to useWatch solves this problem

bluebill1049:
`type` and `name` is reserved `key` word for input registration.

bluebill1049:
<img width="911" alt="Screen Shot 2021-06-10 at 9 31 11 pm" src="https://user-images.githubusercontent.com/10513364/121517906-39651200-ca33-11eb-9884-22912bcfce39.png">

doc is updated.

Raphaww:
Thank you for your quick response, however, I changed all the uses of those reserved words and I still get the same results

https://codesandbox.io/s/youthful-williamson-rxnzo?file=/src/TextForm.js

bluebill1049:
my statement above is invalid, try to fix the bug. (going to delete it to stop confusing users)

PR #5705, opened 2021-06-22
don't spread non-primitive values in nested array fields while using `useFieldArray`

tdzienniak:
I found a problem when using file fields in nested array. There was that line, that was spreading value if it is was not primitive, which caused `File` object to become just a plain object with one prop `path`. I don't know why this behaviour exists. I think it's a bug. After the change, all tests are passing.

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -221,7 +221,7 @@ export const useFieldArray = <
               ? registerFieldArray(value, valueIndex, inputName)
               : (register as UseFormRegister<TFieldValues>)(
                   inputName as Path<TFieldValues>,
-                  { value: isPrimitive(value) ? value : { ...value } },
+                  { value },
                 );
           });
     });


bluebill1049:
thanks! can you share a use case that causes this issue? are you trying to `append({ test: new File() })`

tdzienniak:
Sure. I'm using `Dropzone` component that fires `onChange` event. I have callback like this:
```ts
const appendNewFiles = useCallback(
  (files: File[]) => {
    append(files.map(file => ({
      id: generateUniqueId(),
      file: file,
    })));
  },
  [append],
);
```
that is responsible for handling that event.

soanvig:
I guess this can be easily reproduced with Date object, right?

tdzienniak:
> I guess this can be easily reproduced with Date object, right?

Yes, any instance of some class suffers from this, but `Date` and `File` are the most common in the context of forms.

bluebill1049:
Thanks, guys. I think if that's the case we should skip `Date` and `File` for now, and yet shallow clone the other object.

bluebill1049:
this is the relevant bug: https://github.com/react-hook-form/react-hook-form/issues/5555 hmmm looks like my unit test didn't really fix or cover the use case...

bluebill1049:
looks like it works with the latest PR to deprecate defaultValue: https://codesandbox.io/s/sweet-mestorf-pofve?file=/src/App.js
so I am going to merge your PR.

PR #5748, opened 2021-06-27
Update README.md

smitd21:
V7 Link to examples


bluebill1049:
thanks for the PR, actually the doc is made that way so users don't click v7 when they r already on the v7 page. you can refer to v6 page as well.

PR #5892, opened 2021-07-11
Add constants to exports

eberens-nydig:
Allows for:

```
import { VALIDATION_MODE, useForm } from 'react-hook-form';

const methods = useForm<TFieldValues>({
  mode: VALIDATION_MODE.onTouched
});
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,4 @@
+export * from './constants';
 export * from './controller';
 export * from './logic';
 export * from './types';


bluebill1049:
Screen Shot 2021-07-11 at 9 40 33 pm
Screen Shot 2021-07-11 at 9 40 56 pm

thanks very much for the PR @eberens-nydig 🙏 however, I am probably not going to merge this PR due to the package size increase to the lib, and honestly I care so much about the package size if this is going to in incase 1%, perhaps it's better to create a constant inside your app instead. again thanks a lot for the PR.

eberens-nydig:
@bluebill1049 understand the thoughts around the bundle size. My initial rationale for this PR was around similar considerations. Looking at the bundles it appears the string constants are already contained so my thinking was exporting these to make them available to consumers.

Screen Shot 2021-07-12 at 10 22 59 AM

Regardless this is super minor so defer to you.

🙏🏻 for the library. It's been very nice to work with.

PR #5941, opened 2021-07-14
V8 (LazyPath to be reverted)

bluebill1049:
# V8

RFC: https://github.com/react-hook-form/react-hook-form/discussions/7433


kotarella1110:
@bluebill1049 Will the React Hook Form work in React 18? I'm interested in this PR 😎

bluebill1049:
> @bluebill1049 Will the React Hook Form work in React 18? I'm interested in this PR 😎

Yea, it should be compatible with 18. I am try to test it out, last time the automation actually all pass except the onBlur even has some issue with useEffect which I had raised an issue in react 18 discussion group. So far looks promising.

bluebill1049:
The issue: https://github.com/reactwg/react-18/discussions/40 which i have posted on React-18 group discussion.

carlosagsmendes:
Hi @bluebill1049 so is it safe to use it with React 18 or do we have to wait for this PR? Thanks in advance

bluebill1049:
> Hi @bluebill1049 so is it safe to use it with React 18 or do we have to wait for this PR? Thanks in advance

Running and the react 18 beta version seems to be fine for now.

bluebill1049:
bump

bluebill1049:
@jorisre I am planning to remove `getResolverOptions`, I don't encounter users is using that API and it adds an extra runtime check to return all the fields' names. Let me know your thoughts, that would be a breaking change for resovler. so resolver needs to push to the next version as well.

jorisre:
> @jorisre I am planning to remove `getResolverOptions`, I don't encounter users is using that API and it adds an extra runtime check to return all the fields' names. Let me know your thoughts, that would be a breaking change for resovler. so resolver needs to push to the next version as well.

`getResolverOptions` is required for `toNestError` in the resolvers repo.

bluebill1049:
sorry, i mean this part:  (which provide current validated field names)

```tsx
for (const name of fieldsNames) {
  const field: Field = get(_fields, name);

  field && set(fields, name, field._f);
}
```

nycgavin:
Will the runtime performance improvement fix vscode intellisense performance issues? Its really slow for me using typescript 4.3 and above

app/src/main.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import App from './app';
+import { App } from './app';
 
-ReactDOM.render(<App />, document.getElementById('root'));
+// @ts-ignore
+ReactDOM.createRoot(document.getElementById('root')).render(<App />);
```
JuanM04, line 6: You can tell TypeScript to use types for v18 instead of 17 by doing `import type {} from 'react/next'` and `import type {} from 'react-dom/next'`. More info [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/next.d.ts)

felixschorer:
> Will the runtime performance improvement fix vscode intellisense performance issues? Its really slow for me using typescript 4.3 and above

Yes! See https://github.com/react-hook-form/react-hook-form/discussions/7389

app/src/main.tsx
bluebill1049: Oh cool that's awesome thank you.

src/__tests__/useWatch.test.tsx
```diff
@@ -98,6 +98,19 @@ describe('useWatch', () => {
...
+    const { result } = renderHook(() => {
+      const { control } = useForm<{ test: string; test1: string }>();
+      return useWatch({
+        control,
+        name: ['test', 'test1'],
+        defaultValue: ['test', 'test1'],
+      });
+    });
+
+    expect(result.current).toEqual(['test', 'test1']);
```
bluebill1049, line 111: @felixschorer this type is returned as `unknown[]`.

felixschorer: Hm, I can't find a solution for that issue. TS seems to infer `TFieldNames` from the `defaultValue` prop instead of the `name` prop.

bluebill1049: maybe leave as it is and skip the IntelliSense for now, at least the type checking is working at the current branch?

bluebill1049: Do you mean the current v8 branch?

felixschorer: It's only broken when **both** these conditions are met:
1. `name` is a tuple, e.g. `['foo', 'bar']`
2. `defaultValue` is present.

bluebill1049: I see 👍 I think we can consider that as a design limation. Honestly, i wish we can remove defaultValue altogether but this would be a large breaking change which i would like to avoid, but i will keep encourage users to include `defaultValues` instead inline `defaultValue`.

bluebill1049:
@romain-trotard I have tried to merge master back into the v8 branch, this caused a lot of test failures including a lot of warnings for `act`, would you be able to look into merging your change into this root branch for v8? thanks in advance.

romain-trotard:
> @romain-trotard I have tried to merge master back into the v8 branch, this caused a lot of test failures including a lot of warnings for `act`, would you be able to look into merging your change into this root branch for v8? thanks in advance.

I'll look into it as soon as possible ;)

bluebill1049:
thanks @romain-trotard no pressures ❤️

Moshyfawn:
There seems to be some config issue with the CSB CI integration still, as[ I've tried to do the same in my fork of RHF](https://github.com/Moshyfawn/react-hook-form/pull/5) and it works flawlessly

bluebill1049:
> There seems to be some config issue with the CSB CI integration still, as[ I've tried to do the same in my fork of RHF](https://github.com/Moshyfawn/react-hook-form/pull/5) and it works flawlessly

feel free to check old commits, i did made it working before :)

Moshyfawn:
The only diff is that the config contains the `buildCommand` and the `node` version _(which shouldn't influence anything, really)_:
```diff
 {
+ "sandboxes": [
+ "react-hook-form-v7-js-template-quqov",
+ "react-hook-form-v7-ts-template-4tn8i"
+ ],
- "buildCommand": "build",
+ "node": "14"
 }
```

Again, it works in RHF fork, which is interesting..

Do you mind checking the GH CSB CI app and maybe "reconnect" it with the repo? IDK why it would only work in a fork of **the same** lib with **the same** config other than because of some weirdness with the GH app

bluebill1049:
Did a re-config with CSB @Moshyfawn

Moshyfawn:
Yas, it worked! #8278

image

bluebill1049:
Plan to move into `beta` once `useEvent` releases.

bluebill1049:
update: currently blocked by TS 4.7.0 it brought a breaking change to the lazy type.

itsUndefined:
@bluebill1049 there is a PR open in TypeScript https://github.com/microsoft/TypeScript/pull/52997 to fix the issue with LazyPaths. They will work fine once again.

bluebill1049:
oh that's awesome to hear @itsUndefined cc @felixschorer

itsUndefined:
@bluebill1049 fix landed on TS 5.3 so this feature can return on the development of V8?

OreQr:
@bluebill1049 Are you planning to do something about this? So that reusable forms are finally possible? This has been a major blocker for quite some time.

Issue #2381, opened 2020-07-27
Ability to call trigger() on fieldArray fields


benswinburne:
It'd be useful to be able to call `trigger()` on field arrays.

A minimum API would be `trigger('myfield[i].property')`

Further potential APIs

`trigger('myfield')` iterates over `myfield[]` and `Object.values()`
`trigger('myfield[0]')` iterates over `Object.values()`

**Additional context**

Related discussion https://github.com/react-hook-form/react-hook-form/issues/2379

bluebill1049:
what about leverage regex `trigger(/^test/)`?

dwiyatci:
@bluebill1049 @benswinburne I'm sorry, here as far as my understanding goes, after following the discussion #2379, `trigger('fields[i].property')` is already supported because `fields[i].property` is basically a `fieldName`, right?

bluebill1049:
> @bluebill1049 @benswinburne I'm sorry, here as far as my understanding goes, after following the discussion #2379, `trigger('fields[i].property')` is already supported because `fields[i].property` is basically a `fieldName`, right?

right now, it only works for the input name.

dwiyatci:
🤔 Hm ok, just to clarify again ('cause I'm also gonna use this manual trigger for fieldArray elements in my project 😁): here in my CSB (https://codesandbox.io/s/react-hook-form-usefieldarray-forked-wu88d?file=/src/index.js), since `` `test[${index}].firstName` `` is an input name, `` trigger(`test[${index}].firstName`) `` works, right?

PR #5958, opened 2021-07-15
⏩  General perf enhancement and enhanced `useWatch`, `watch` and `getValues` 

bluebill1049:
# Improvement

We focus a lot on performance! This is one of the reasons hook form is getting popular, and we would like to push performance even further, we reduced re-render and now it's time to improve the internal when users subscribed to inputs updated. The current design tight user inputs reference with value, however, this also has an impact on deeply nested input value lookup, with the new value store. We can skip that nested value lookup and straightway access the value store. 

This PR has the following improvements:

- Dedicated value store
- Improve perf `getValues`, `watch`, `useWatch`
- enabled register ref input lookup https://github.com/react-hook-form/react-hook-form/issues/6098 
- Disabled inputs value will be updated during `register`
- `setValue` getting uplifted with value update instead rely on `register` inputs
- enable `trigger` for field array and object fields. #2381
- package size reduction **5%**! yay!
- `createFormControl` move logic into its own factory function #5659 credit and idea goes to https://github.com/MHA15
  -  PR for the above work: https://github.com/react-hook-form/react-hook-form/pull/5987

# Added

- register('name', { disabled: boolean })

Currently, the `disabled` attribute is working with a ref callback, this is not an ideal solution. The following prop will guarantee and as a prop solution for toggle `disabled` input.

- `disabled` prop to solve toggle disable for the input value 

```tsx
register('name', { disabled: true })
````

- useWatch({ disabled: boolean })

This new config will allow users to pause/unsubscribe the watch, and you can resume the subscription by toggle the `disable`

```tsx
useWatch({ disabled })
```

- useFormState({ disabled: boolean })

This new config will allow users to pause/unsubscribe formState, and you can resume the subscription by toggle the `disable`

```tsx
useFormState({ disabled })
```

# Result (Dev build)

The following test is tested with my personal laptop below and using chrome latest build with **CPU throttle with x 4**

### Flat Example: 

```tsx

const Component = () => {
  const { register, watch } = useForm();
  console.log(watch());

  return (
    <div>
      {[...Array(1000).keys()].map((item, i) => (
        <div key={i}>
          <input {...register(`nest${i}`)} />
        </div>
      ))}
    </div>
  );
};
```

### Nested Example: 

```tsx

const Component = () => {
  const { register, watch } = useForm();

  console.log(watch());

  return (
    <div>
      {[...Array(1000).keys()].map((item, i) => {
        const nest = i / 100;
        return (
          <div key={i}>
            <input {...register(`nest${'.test'.repeat(nest)}${i}`)} />
          </div>
        );
      })}
    </div>
  );
};
```

### Deeply Nested Example: (1000 * 2 * 1000)

```tsx
import { useForm, useWatch } from 'react-hook-form';

const Nest = ({ control }) => {
  useWatch({
    control,
  });
  return null;
};

const Component = () => {
  const { register, watch, control } = useForm();

  console.log(watch());

  return (
    <div>
      {[...Array(1000).keys()].map((item, i) => {
        const nest = i / 100;
        return (
          <div key={i}>
            <input {...register(`nest${'.test'.repeat(nest)}${i}`)} />
            <Nest control={control} />
          </div>
        );
      })}
    </div>
  );
};
```

## 🧾  Outcome:

`7.11.2-beta.0`

### Flat

Average: **18ms** each keystroke to watch input and render the output of watch result

Screen Shot 2021-07-17 at 11 02 38 am

Mouting:  **470ms** 

Screen Shot 2021-07-19 at 11 21 05 am

### Nested

Average: **20ms** each keystroke to watch input and render the output of watch result

Screen Shot 2021-07-17 at 11 05 06 am

### Deeply Nested (2000 * 2 * 2000)

Average: **550ms** each keystroke to watch input and render the output of watch result
Screen Shot 2021-07-17 at 11 37 52 am

with `createFormControl`

Average: **200ms** each keystroke to watch input and render the output of watch result
Screen Shot 2021-07-18 at 9 34 06 pm

`7.11.1`

### Flat

Average: **40ms** each keystroke to watch input and render the output of watch result

Screen Shot 2021-07-17 at 10 49 29 am

Mouting:  **470ms** 

mout 7 11 1

### Nested

Average: **43ms** each keystroke to watch input and render the output of watch result

Screen Shot 2021-07-17 at 11 44 54 am

### Deeply Nested (2000 * 2 * 2000)

Average: **2000ms** each keystroke to watch input and render the output of watch result

Screen Shot 2021-07-17 at 11 30 38 am

# Conclusion 

Dev Build + Throttle 4x CPU + Chrome + MAC M1 

| Version | Flat (1000) | Nested (1000)  | Deeply Nested (1000 * 2 * 2000) |  Flat (1000) Mount |
|:--- |:---:|:---:|:---:|:---:|
| `7.11.2-beta.0` | 18ms | 20ms | 550ms |  470ms|
| `7.11.2-beta.0` + `createFormControl` | || 200ms (huge improvement) |  470ms|
| `7.11.1` | 40ms | 43ms | 2000ms |  470ms|

The outcome is the new build is faster than the current build. Have a dedicated value store avoided the nested lookup value. The next step will release a beta version and let more users test out this build before release the final minor update.

# Todos

- [x] unit tests coverage
- [x] docs update
- [x] functional tests
- [x] improve `useFieldArray` perf


src/__tests__/useForm/register.test.tsx
```diff
@@ -1,3 +1,5 @@
+// @ts-nocheck
```
bluebill1049, line 1: This is a github bug??? Don't see that in code.

src/__tests__/useForm/setValue.test.tsx
```diff
@@ -965,7 +929,7 @@ describe('setValue', () => {
     expect(screen.getByTestId(inputId)).toHaveValue('updated value');
   });
 
-  it('should register deeply nested inputs correctly', () => {
+  it('should not register deeply nested inputs', () => {
```
bluebill1049, line 932: todo: add test for value.

bluebill1049:
Thank you @MHA15 we are moving foward!

Aryaman73:
@bluebill1049 thanks a lot for this PR! 🥳 
 
I noticed a possible bug with the change to `trigger()`. It seems that after upgrading to `7.13.0-next.0`, an error shows up when you pass something to trigger that doesn't exist (e.g `trigger("random-field-that-does-not-exist):`)

image

Looks like there isn't a catch for if there aren't any fields with that name. I could create an issue for this and code sandbox if necessary, but unless I'm mistaken this should be easy to reproduce in any project. We are relying on useFormContext, if that matters.

Thanks again!

bluebill1049:
It's better to crash your app at dev before it breaks your prod. `trigger` `name` should be part of your form.

app/src/useFieldArrayUnregister.tsx
```diff
@@ -24,7 +24,7 @@ const ConditionField = <T extends any[]>({
   index: number;
   fields: T;
 }) => {
-  const output = useWatch<FormInputs>({
+  const output = useWatch({
```
barrymay, line 27: is this purposefully removed (is the typing assumed?)

bluebill1049: yes, `useWatch` doesn't need the generic. 

https://github.com/react-hook-form/react-hook-form/blob/master/src/useWatch.ts#L19

PR #5987, opened 2021-07-18
create form control

bluebill1049:
# Context

This PR is to making createFormControl as pure js and more out from the `useForm`. It maybe merged to this PR: https://github.com/react-hook-form/react-hook-form/pull/5659 

❤️ original **idea** and **credit** goes to @MHA15

# Result: 

### Deeply Nested (2000 * 2 * 2000) 

**Context:** Dev Build + Throttle 4x CPU + Chrome + MAC M1

From 
Average: **550ms** each keystroke to watch input and render the output of watch result

To 
Average: **300ms** each keystroke to watch input and render the output of watch result

Screen Shot 2021-07-18 at 3 01 22 pm


MHA15:
I have two questions 
- I had tried to create FormControl in a way that `useForm` return value be the same reference in multiple renders (less child and provider rerender) , didn't you like it ? 
-  why did you rename current to val , let's just remove them and use field as value not as an object with field val

MHA15:
and let's use createForm in useForm with that if and else for first render and others, that way createForm will be called once only and it's options will be updated in other renders but in this way options will be same as first call and createForm will be called in every render

bluebill1049:
> why did you rename current to val , let's just remove them and use field as value not as an object with field val

what do you mean by that? reason for the `val` is to diff from `React.MutableRef`

MHA15:
> > why did you rename current to val , let's just remove them and use field as value not as an object with field val
> 
> what do you mean by that? reason for the `val` is to diff from `React.MutableRef`.
> 
> I think I know what you do refer to the `_fields` object.

why shouldn't we just write control as I did in that PR like this 

```
control: {
get _fields() {},
set _fields() {}, 
get _names() {},
set _names() {}
}
```

and use them like `control._names` instead of `control._names.val` ?

bluebill1049:
> and use them like control._names instead of control._names.val ?

yep! @MHA15 don't need `get` as well. 👍  updated.

bluebill1049:
> and use them like control._names instead of control._names.val ?

oh, I got you...  will get the rest of them updated! to move `.val`. thanks.

bluebill1049:
I am going to merge this in @MHA15  feel free to leave more feedback and comments. I am a bit tired to get into your PR, hope you understand. thanks very much for the idea and the origin PR. Will make sure to give you credit when we release this. Thank you <3

MHA15:
> I am going to merge this in @MHA15 feel free to leave more feedback and comments. I am a bit tired to get into your PR, hope you understand. thanks very much for the idea and the origin PR. Will make sure to give you credit when we release this. Thank you <3

Thank you for your works. 
It's great we finally merged this, later we will improve it even more with same idea <3

bluebill1049:
Thank you @MHA15 this is amazing!

bluebill1049:
Do you have a Twitter account @MHA15? Let's tag you to let everyone hook form is about to get even faster.

MHA15:
> Do you have a Twitter account @MHA15? Let's tag you to let everyone know hook form is about to get even faster.

No unfortunately I don't have an active twitter account , I have a [linkedin](https://www.linkedin.com/in/mohamad-hosein-alahdadian/) account if you want.

Issue #5981, opened 2021-07-17
Auto focus on the first input with error when both native and custom inputs are used


tedliang:
Auto focus on the first input with error when both native and custom inputs are used

**Describe the solution you'd like**
https://codesandbox.io/s/react-hook-form-v6-auto-focus-on-the-first-input-with-error-8s50w?file=/src/index.js
When submit button is clicked, it should focus on the input for first name.

**Describe alternatives you've considered**
N/A

**Additional context**
We can use fieldErrors and form.elements to find out the first focusable error input.
https://erikras.com/blog/focus-on-errors

bluebill1049:
We can't guarantee the focus order at v6, the order is based on the `register` order. If you want to get it consistent, I would recommend to use all `Controller` at v6.

bluebill1049:
consider to use v7: https://codesandbox.io/s/goofy-elbakyan-nety1?file=/src/App.js

tedliang:
v7 has the same problem: https://codesandbox.io/s/zen-hertz-2ftkt
The workaround would be using Controller for all inputs. 

Thanks for your prompt reply.

jeina7:
Hi, I am also struggling with getting an auto focus on input which has error with `<Controller>`.

It seems that now `"react-hook-form": "7.11.1"` version works fine with simple `<input>` tag, and also works fine when `<Controller>` renders simple `<input>` tag. However, it doesn't work for 3rd party components. 

I have a demo with material-ui: https://codesandbox.io/s/lucid-christian-2bqqx

It takes auto focus when "firstName" field or "lastName" field is empty, however it doesn't take auto focus on material-ui field when it is still empty.

Is this on-going issue on react-hook-form v7?

tedliang:
MUI is using inputRef instead of ref.

PR #5988, opened 2021-07-18
Access input from CustomElement

tedliang:
With this change, user can implement custom logic to manually focus on the first error input instead of using Controller for all inputs. See #5981 

In order to determine the focus order for error inputs, 'ref' from Controller render prop is required in the CustomElement, so that it can be accessible from FieldErrors.

### DIFF
--- a/src/types/fields.ts
+++ b/src/types/fields.ts
@@ -25,6 +25,7 @@ export type CustomElement<TFieldValues extends FieldValues> = {
   options?: HTMLOptionsCollection;
   files?: FileList | null;
   focus?: () => void;
+  input?: any;
 };
 
 export type FieldValue<
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -126,6 +126,7 @@ export function useController<TFieldValues extends FieldValues = FieldValues>({
             {
               name,
               focus: onFocusRef.current,
+              input: ref.current,
             },
             {
               value: {


bluebill1049:
thanks for the PR @tedliang 🙏 for V6 it's on maintain mode, prefer not to introduce more props/feature at this stage.

tedliang:
Do you consider this for v7?

bluebill1049:
This is already in V7, basically the ref prop from the render props.

tedliang:
The ref prop from the render props is not accessible from FieldErrors in v7 as well.
https://codesandbox.io/s/blissful-montalcini-p457g?file=/src/App.js
When submit button is clicked, it's focused on the last input instead of the first one in the form.
The input ref for native field is accessible via errors.native.ref. However, the input ref for controller field is not accessible. 
image

bluebill1049:
you are referring to the focus order, if that's the cause you will have to either use register/controller, feel free to submit a feature request in which we can come up with a better API but it's a difficult one. especially when there are things like media queries and different break points.

tedliang:
With the input ref for both native and controller fields in errors, I can then use form.elements to determine the order for the error inputs. 

Ideally, errors.materialui.ref should refer to the ref we passed to the TextField.
Screen Shot 2021-07-19 at 7 48 38 pm

But if it's too hard, can we make it available via errors.materialui.ref.input?

bluebill1049:
This was made so the controller is diff from `register`, and will not use ref to update inputs. Don't think we want to change anything at this stage.

tedliang:
Will this be too hard?
Screen Shot 2021-07-19 at 7 53 04 pm

bluebill1049:
not with the current design, yes it's not simple.

tedliang:
Can we make the elm available in ref? eg. errors.materialui.ref.elm? I can see it from the log. It's right there. so close!🤤
image

PR #5999, opened 2021-07-19
Remove proxy api

mh-alahdadian:
use basic getter api of js instead


bluebill1049:
This looks great! Thanks.

PR #6013, opened 2021-07-19
fix(useFieldArray): key.startsWith is not a function

### LINKED ISSUE #6012: Error on focus after nested object being removed and field array element being added
**Describe the bug**
This bug is super hard to spot and hard to reproduce (at least for me) ... so I'll try my best describing it:

In some weird edge cases, [this function call](https://github.com/jsbrain/react-hook-form/blob/ea5eceefd32ca322793376b00d04d53a97ddab9a/src/useFieldArray.ts#L417) causes the crash. After removing some nested object and then trying to add another element to a field array, my app crashes with:

```
e.startsWith is not a function
```

I spend basically the whole day trying to figure out what's going on, tried multiple different versions of my code but somehow I always get the same result. After digging deeper and logging the iterations of [focusFieldBy](https://github.com/jsbrain/react-hook-form/blob/ea5eceefd32ca322793376b00d04d53a97ddab9a/src/logic/focusFieldBy.ts#L7) it seems that in my case in one iteration `field` looks like `{ ..., _f: {} }` with `_f` being an empty object. This leads to the callback being called with `_f.name` which is undefined, leading to the error that `.startsWith` cannot be called on undefined (obviously).

I tried to reproduce it consistently in a Codesandbox but I just can't get it to (not) work. On my local machine it crashed every time, in the sandbox it only crashed very rarely but it does from time to time.

Not sure if the error is actually in my code because I'm doing some recursive nested stuff there but everything else works just fine, the submitted values are all good and so on. Just this focus error happens in some cases and I guess it's pretty safe to introduce an additional undefined check there.

While you probably won't be able to reproduce the error with my sandbox, [here it is](https://codesandbox.io/s/stoic-ishizaka-bcyf9?file=/src/Form.tsx) anyway, so you could at least get a glimpse of what I'm trying to achieve :)

The error (normally) happens after I
1. add a Notification
2. add a User to the Notification
3. add another Notification inside
4. add another User
5. remove the User
6. Add another Notification to the 1. level notifications array
7. -> Crash

jsbrain:
Fix #6012

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -414,7 +414,7 @@ export const useFieldArray = <
 
     focusNameRef.current &&
       focusFieldBy(fieldsRef.current, (key: string) =>
-        key.startsWith(focusNameRef.current),
+        key?.startsWith(focusNameRef.current),
       );
 
     focusNameRef.current = '';
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -413,8 +413,9 @@ export const useFieldArray = <
     });
 
     focusNameRef.current &&
-      focusFieldBy(fieldsRef.current, (key: string) =>
-        !!key && key.startsWith(focusNameRef.current),
+      focusFieldBy(
+        fieldsRef.current,
+        (key: string) => !!key && key.startsWith(focusNameRef.current),
       );
 
     focusNameRef.current = '';
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -414,7 +414,7 @@ export const useFieldArray = <
 
     focusNameRef.current &&
       focusFieldBy(fieldsRef.current, (key: string) =>
-        key?.startsWith(focusNameRef.current),
+        key && key.startsWith(focusNameRef.current),
       );
 
     focusNameRef.current = '';
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -414,7 +414,7 @@ export const useFieldArray = <
 
     focusNameRef.current &&
       focusFieldBy(fieldsRef.current, (key: string) =>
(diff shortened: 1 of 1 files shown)


src/useFieldArray.ts
```diff
@@ -414,7 +414,7 @@ export const useFieldArray = <
 
     focusNameRef.current &&
       focusFieldBy(fieldsRef.current, (key: string) =>
-        key.startsWith(focusNameRef.current),
+        key?.startsWith(focusNameRef.current),
```
bluebill1049, line 417:
```suggestion
        key && key.startsWith(focusNameRef.current),
```

bluebill1049: save bytes for the transpired code.

src/useFieldArray.ts
```diff
@@ -414,7 +414,7 @@ export const useFieldArray = <
 
     focusNameRef.current &&
       focusFieldBy(fieldsRef.current, (key: string) =>
-        key.startsWith(focusNameRef.current),
+        key && key.startsWith(focusNameRef.current),
```
bluebill1049, line 417:
```suggestion
        !!key && key.startsWith(focusNameRef.current),
```

markcnunes:
I had the same error, but the issue is now closed [here](https://github.com/react-hook-form/react-hook-form/issues/6012). I hope you don’t mind me sharing what worked for me, in case someone else has the same problem.

Here are the steps that caused the issue:
1. Add multiple itinerary items.
2. Delete one of the recently added items.
3. Save.
4. Add a new itinerary item.
5. The app crashes with the error (e.startsWith is not a function).

For context, the useForm was used in a higher component, and useFieldArray was in a child component using useFormContext.

I fixed the issue by replacing `defaultValues` in useForm with `values`.

PR #6030, opened 2021-07-21
stricter type check on onChange event type

Newbie012:


Newbie012:
@bluebill1049 If I change it to

```ts
export type ChangeHandler = (event: {
  target: {
    name: string;
    value?: unknown;
    type?: unknown;
  };
  type?: any;
}) => Promise<void | boolean>;
```
then the build will pass. Although, I have a feeling that the fix has bubbled up potential bugs. What do you think?

bluebill1049:
yea... let's leave as it is for now, shall we?

Newbie012:
@bluebill1049 It's probably not perfect, but I believe it's better than the current implementation of `ChangeHandler`. But I guess it's up to you :)

bluebill1049:
Thanks, I am worried this may introduce breaking changes to existing users. Let's skip this one for now. 👍

bluebill1049:
thanks for the PR.

bluebill1049:
actually, let's roll with this if any issues coming from this PR we can looking further for solutions.

bluebill1049:
looks like it breaks for RN.

Newbie012:
@bluebill1049 I haven't dug deeper into the project, but I have managed to create a prototype that:

 - works both on react-dom and react-native
 - better type safety
 - type inference

Check it here - https://codesandbox.io/s/modern-shadow-0rj3o?file=/src/App.tsx

image
image

bluebill1049:
oh, nice! feel free to send another PR! that's awesome. Thanks.

Newbie012:
I'm afraid it would cause breaking changes

bluebill1049:
ok than, thanks for the investigation and perhaps we can look into this again in the next major version.

PR #6084, opened 2021-07-26
🐞 make sure the field is set before setting it to mounted

wtchnm:
Hi! Thanks for your awesome work!

This PR consists of verifying if the field is defined before accessing it in the body of useEffect, as we have in the clean up function.


bluebill1049:
thanks for the PR.

PR #6141, opened 2021-08-02
Feature: add dependent validation - run a dependent field's validation after another field's validation runs

rosemariesadler:
* Relevant Discussion: https://github.com/react-hook-form/react-hook-form/discussions/5597
* Register gets a new option: deps, which is an arry of field names. When these dependent fields run their validation, the registered field will also run its validation.
* This is useful for things like making sure passwords match and having one of three fields not be empty
* I added cypress tests instead of unit tests - following the pattern I saw for other validation testing
* I'm unfamiliar with TypeScript, so I may have messed up the types here - please forgive if something is missing or incorrect.


src/__tests__/useForm.test.tsx
```diff
@@ -1074,16 +1076,36 @@ describe('useForm', () => {
...
+            ref: {
+              name: 'test1',
+            },
+          },
+        };
+
         expect(resolver).toHaveBeenNthCalledWith(3, defaultValues, undefined, {
           criteriaMode: undefined,
-          fields,
+          fields: fields2,
```
rosemariesadler, line 1108: This is something I couldn't figure out with Typescript - I wanted to just add hasValidated to test1 here, but I was getting some kind of typescript error about it and I couldn't figure out how to resolve it. So I'm sure there's a better way to do this.

bluebill1049:
This is awesome! I will help on this PR tonight and making some minor adjustments 👍

bluebill1049:
I made the change so it only affects the `register` API without touching the `trigger` for now.

src/logic/createFormControl.ts
```diff
@@ -504,6 +505,10 @@ export function createFormControl<
           values: getValues(),
         });
 
+      if (field._f.deps) {
+        trigger(field._f.deps as FieldPath<TFieldValues>[]);
```
rosemariesadler, line 509: I see that you removed the `hasValidated` check that I had added. This means that even if the field hasn't been touched, its validation will still run and show up - is that what we want, or would it be better to let it run its own validation the first time around and then update it every time a dependency updates?

rosemariesadler:
> I made the change so it only affects the `register` API without touching the `trigger` for now.

Thanks so much for your help on this! 

I'm thinking about this - if we don't touch trigger, then we won't have the ability to cascade dependencies - which could get us stuck in an infinite loop if the consumer creates a loop of dependencies. And normally if a consumer is calling trigger, we can just call it for the dependencies as well.

bluebill1049:
thanks @rosemariesadler can you share an example: 

> I'm thinking about this - if we don't touch trigger, then we won't have the ability to cascade dependencies - which could get us stuck in an infinite loop if the consumer creates a loop of dependencies. And normally if a consumer is calling trigger, we can just call it for the dependencies as well. I like it!

bluebill1049:
If everything goes well, let's merge this after 7.13.0 is out and making this a new feature at 7.14.0

rosemariesadler:
> thanks @rosemariesadler can you share an example:
> 
> > I'm thinking about this - if we don't touch trigger, then we won't have the ability to cascade dependencies - which could get us stuck in an infinite loop if the consumer creates a loop of dependencies. And normally if a consumer is calling trigger, we can just call it for the dependencies as well. I like it!

Yes, so the situation you helped us avoid with your changes goes something like this:
```
<input {...register('firstName', { deps: ['title'] )} />
<input {...register('lastName', { deps: ['firstName'] })} />
<input {...register('title', { deps: ['lastName'] })} />
```
If we were to update `trigger` to also call dependent validation, when `firstName` changes, it will cause `lastName` to also validate, which would then cause `title` to validate, which would then cause `firstName` to valiate again into an infinite loop. I don't think a consumer would intentionally do this, but it could cause problems. I do think that if a user were to manually call `triggger`, they could call it with nothing to validate the entire form, or call it with all three fields.

rosemariesadler:
> If everything goes well, let's merge this after 7.13.0 is out and making this a new feature at 7.14.0

Sounds fantastic. Thank you!

bluebill1049 (approved):

src/logic/createFormControl.ts
bluebill1049: oh looks like I miss this comment, I think let's keep this one simple this will be just a simple `trigger` inside, if users want to have specific behaviour it's better to include logic at their app level.

bluebill1049:
```tsx
const { register, formState } = useForm({ mode: 'onTouched' })

<input {...register('test', { deps: 'test1' })} />
{formState.touchedFields.test1 && formState.errors.test1 && <p>Error</p>}

<input {...register('test1')} /> 
```

bluebill1049:
Amazing work @rosemariesadler! merged and let's roll it out in 7.14.0

rosemariesadler:
Thank you!

PR #6173, opened 2021-08-04
fix `TypeError: Cannot read property '_f' of undefined` bug in createFormControl.ts

leloykun:
`get()` in line 926 in `createFormControl.ts` can return `undefined` (making `field` `undefined`), but subsequent lines assume that `field` is never `undefined` - causing the `Cannot read property '_f' of undefined` type error


bluebill1049:
thanks for the PR. Do you have a reproducible CSB for the issue? I suspect this may lead to a misuse case.

Linus-Boehm:
We had the same Issue, but it was caused by an active DevTool from @hookforms

Issue #6037, opened 2021-07-21
Strange behaviour with insert


revich2:
**Describe the bug**
I have form with fieldArray and elements of fieldArray can add by modal with form for input the name of element. Elements are controlled inputs. In my case, when form with name submitted i need to insert new element in defined place.
When insert happened i see that some elements clones and values from handleSubmit have one more elements than expected.

**To Reproduce**

1. Go to CSB
2. Click button 'Add any element'
3. Type to input and click on 'save'
4. Do 2,3 points again with another value in input
5. You can see that, value which displaying in simple span correct, but value in controlled input isn't correct

**Codesandbox link (Required)**
https://codesandbox.io/s/upbeat-lovelace-2rvmk?file=/src/App.tsx

**Expected behavior**
Insert should be work correct with Controller and second form for this case

**Screenshots**
<img width="424" alt="Screenshot 2021-07-21 at 10 49 36 PM" src="https://user-images.githubusercontent.com/24573190/126550937-4d2b0e5a-559e-4930-a5a3-28c4fb4e7d50.png">

**Desktop (please complete the following information):**

- OS: Big Sur 11.4
- Browser chrome
- Version 91.0.4472.114

**Additional context**
All keys of elements is unique and valid.
I don't know why it happen, but i noticed some regularity. If to do one of this, then all looks like ok:
1) Not use `handleSubmit` in modal form(Simple to fire onSubmit from props)
2) Not use Controller with input

PR #6178, opened 2021-08-05
Fixes react state race condition for field array methods

revich2:
This PR solves issues from this topics - #6037, #6039

After deep research i found problem, why this strange behaviour happens. It is race condition of react state updates.
You can see issue here - https://codesandbox.io/s/delicate-flower-rgzq0?file=/src/App.js

1) If you will use external RHForm for adding new items to field array, then you will see collision - items will be added incorrect.
2) If you will use native html form for adding new items to field array, then all will be OK.

It really strange because i understood that it is race condition of react state updates, but i don't know that causes that.

All integration tests have done.

I have problem with unit tests of this PR, because some tests, which contains `useFieldArray`, fails due to `setTimeout(..., 0)`. which i used in `createFormControl` to schedule `_subjects.state.next` as early as it possible.
And i don't know how i can fix failed unit test, i tried to use `jest.useFakeTimers`, but it didn't help.

If who knows how to solve problem with `setTimeout(..., 0)` in unit tests, then i will be able to fix them.

@bluebill1049 check it, please


bluebill1049:
thanks for the PR and investigation. the codesandbox is a lot easier for me to disgest as well. please refer to the doc on the `reset`.

https://react-hook-form.com/api/useform/reset/

Screen Shot 2021-08-06 at 12 25 21 pm

bluebill1049:
also worth mentioning, handleSubmit onSubmit is async not sync.

revich2:
@bluebill1049 oh, it is unobviously thing, really.

what do you think about this PR? all integration tests have done and this logic works obviously

will it be merged? after fixes unit tests, of course

bluebill1049:
> @bluebill1049 oh, it is unobviously thing, really.

I guess it's a question of whether we support async onSubmit for users, I guess the answer is pretty simple: 'yes' 

-> onSubmit -> async -> setState

vs

-> onSubmit -> setState

revich2:
@bluebill1049 but what the problem with async onSubmit and changes from this PR? sorry, i'm not understand you will approve this changes or not

bluebill1049:
I will not have `setTimeout` in this lib, if you have a strong opinion on the `reset` API, please fork the repo and publish a version of your own.

revich2:
@bluebill1049 this changes not about `reset` API, it about field array methods. `reset` isn't influence on this case
while i was investigating this case i found that `control._bathFieldArrayUpdate` has to be before `setFields` in field array methods functions, because it's also influence on race condition and it without any `setStates`.
Screenshot 2021-08-06 at 10 25 28 AM

with `reset` i agree with you, but array methods can used only in external RHForm, not in onSubmit their form

src/logic/createFormControl.ts
```diff
@@ -706,12 +706,16 @@ export function createFormControl<
       unsetEmptyArray(_formState.dirtyFields, name);
     }
 
-    _subjects.state.next({
-      isDirty: _getIsDirty(name, omitKey(updatedFieldArrayValues, keyName)),
-      dirtyFields: _formState.dirtyFields,
-      errors: _formState.errors,
-      isValid: _formState.isValid,
-    });
+    setTimeout(
```
bluebill1049, line 709: what's this?

src/useFieldArray.ts
```diff
@@ -118,6 +113,15 @@ export const useFieldArray = <
...
 
+    setFields(
+      mapIds(
+        updatedFieldArrayValues as Partial<
+          FieldArray<TFieldValues, TFieldArrayName>
+        >[],
+        keyName,
+      ),
+    );
+
```
bluebill1049, line 124: I don't mind those changes, as long as there is a test case to prove this does fix an issue.

bluebill1049:
i will take over PR and remove the `setTimeout`, your effort is appreciated here.

src/logic/createFormControl.ts
bluebill1049: This is not accepted.

PR #6271, opened 2021-08-19
Implement option `shouldRegisterAbsentFields`

delijah:
Hello all

This PR is related to the following discussion https://github.com/react-hook-form/react-hook-form/discussions/6240

The `registerAbsentFields` feature is giving us troubles. Adding a simple option to enable/disable it, gives developers more freedom on how to work with react-hook-form.

This PR acts as a base for the above mentioned discussion. Of course we could add documentation and tests if this will be considered a viable way to go. It's currently based on version 7.12.2, not sure where to add for 7.13.x

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -104,6 +104,7 @@ export type UseFormProps<
   shouldFocusError: boolean;
   shouldUnregister: boolean;
   shouldUseNativeValidation: boolean;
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks for the PR, would you like to send a feature request first and let the community drive if this is required for another config.

delijah:
> community

Ok, thank you, i will do so 👍

PR #6304, opened 2021-08-22
[useFieldArray] new method `replace()`

OlehDutchenko:
This PR adds new method for fully replacement fieldArray values in one action.
_Resolves #6227_

Example:

```ts
const { control } = useForm({
  defaultValues: {
    test: [{ value: 'lorem' }, { value: 'ipsum' }]
  }
});
const { fields, replace } = useFieldArray({ 
  control, 
  name: 'test'
});

const handleFullyReplacement = (): void => {
  // remove old and set fully new values
  replace([{ value: 'dolor' }, { value: 'sit' }, { value: 'amet' }]);
}

...
```

### DIFF
--- a/app/src/useFieldArray.tsx
+++ b/app/src/useFieldArray.tsx
@@ -24,11 +24,20 @@ const UseFieldArray: React.FC = (props: any) => {
       : {}),
     mode: props.match.params.mode === 'formState' ? 'onChange' : 'onSubmit',
   });
-  const { fields, append, prepend, swap, move, insert, remove, update } =
-    useFieldArray({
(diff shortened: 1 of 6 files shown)


bluebill1049:
thank you very much for the PR. I will send an RFC on this feature and see if developers are required to have such a feature. each feature that we add, it does impact the size of the lib. will keep everyone updated on this.

bluebill1049:
https://github.com/react-hook-form/react-hook-form/discussions/6357

dutchenkoOleg:
Awesome! 🤟🏻

PR #6386, opened 2021-08-29
fix: TypeError: e.focus is not a function

jeina7:
`react-hook-form@7.14.0` version shows the type error when I run the jest test.

```
TypeError: e.focus is not a function

    at Object.options [as focus] (node_modules/react-hook-form/src/useController.ts:113:39)
    at next (node_modules/react-hook-form/src/logic/getFocusFieldName.ts:8:3)
    at Object.<anonymous> (node_modules/react-hook-form/src/logic/createFormControl.ts:1081:25)
```

The test passed on `react-hook-form@7.12.2` version, where [this](https://github.com/react-hook-form/react-hook-form/commit/b6ffb6e33a253e13df8f0569ed09be899dcf0174#diff-3f5bdfa7d822b4562b8e1c72dd1996197fe93478b039779bb78853ab3ae2a3feL108) change has not committed. 
I assume that `elm.focus &&` statement has to be added.

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -105,9 +105,9 @@ export function useController<
       ref: (elm) => {
         const field = get(control._fields, name);
 
-        if (elm && field) {
+        if (elm && field && elm.focus) {
           field._f.ref = {
-            focus: () => elm.focus && elm.focus(),
+            focus: () => elm.focus(),
             setCustomValidity: (message: string) =>
               elm.setCustomValidity(message),
             reportValidity: () => elm.reportValidity(),
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -107,7 +107,7 @@ export function useController<
 
         if (elm && field) {
           field._f.ref = {
-            focus: () => elm.focus(),
+            focus: () => elm.focus && elm.focus(),
             setCustomValidity: (message: string) =>
               elm.setCustomValidity(message),
             reportValidity: () => elm.reportValidity(),


bluebill1049:
Thanks Jeina. What your controller usage like? Perhaps your ref is not correctly supplied for controller.

bluebill1049 (dismissed):

src/useController.ts
```diff
@@ -107,7 +107,7 @@ export function useController<
 
         if (elm && field) {
```
bluebill1049, line 108: actually, Would you like to move the check to here instead? just so we can block any invalid `ref` object.
```suggestion
        if (elm && field && elm.focus) {
```

jeina7: I've moved :)

jeina7: I've moved it :)

jeina7:
This is my Controller usage. (Shorttened)

```javascript
import { useRef } from "react";
import { Controller, SubmitHandler, useForm } from "react-hook-form";
import { InputField } from "~/app/components";

export const LoginForm = () => {
  const inputRef = useRef(null)
  const { control, handleSubmit } = useForm({
    criteriaMode: "all",
    mode: "all",
  });

  return(
    <Controller
      control={control}
      name="email"
      render={({ field, fieldState: { error } }) => (
        <InputField
          {...field}
          ref={inputRef}
          labelText="email"
          errorMessage={error?.message}
        />
      )}
      rules={{
        required: "Please enter your email.",
      }}
    />
  );
};
```

I have custom Component `<InputField>`, and use it to be rendered by the render function of Controller.    
I pass the `inputRef` to `<InputField>` component, and it goes into `<input>` tag.    


Would there be any mistake of using the ref for controller?    
It worked well on `react-hook-form@7.12.2` version, as I said.. but doubt if I made any mistake.

src/useController.ts
bluebill1049: thanks!

bluebill1049:
```tsx
<InputField
  {...field} // this will pass ref to controller but it's not a actual input ref
  ref={inputRef}
  errorMessage={error?.message}
/>
```

bluebill1049:
Thank you for your contribution! ❤️

jeina7:
> ```tsx
> <InputField
>   {...field} // this will pass ref to controller but it's not a actual input ref
>   ref={inputRef}
>   errorMessage={error?.message}
> />
> ```

@bluebill1049 Does `but it's not a actual input ref` means I don't have to pass `inputRef` as `ref={inputRef}`?    
Would Controller handle ref automatically?

Thank you for your fast reply :)

bluebill1049:
do this instead:

```tsx
import { useRef } from "react";
import { Controller, SubmitHandler, useForm } from "react-hook-form";
import { InputField } from "~/app/components";

export const LoginForm = () => {
  const inputRef = useRef(null)
  const { control, handleSubmit } = useForm({
    criteriaMode: "all",
    mode: "all",
  });

  return(
    <Controller
      control={control}
      name="email"
      render={({ { onChange, onBlur, ref }, fieldState: { error } }) => (
        <InputField
          onChange={onChange}
          onBlur={onBlur}
          inputRef={ref} // if inputRef is the real input reference, so the focus management will work correctly
          errorMessage={error?.message}
        />
      )}
      rules={{
        required: "Please enter your email.",
      }}
    />
  );
};
```

PR #6404, opened 2021-08-31
Update watch observer types

joshkel:
The watch callback is only invoked with one field at a time.  See https://github.com/react-hook-form/react-hook-form/discussions/6392


bluebill1049:
thanks!

Issue #5055, opened 2021-05-06
useFieldArray (and useController) should allow specifying value type constraints


mayorandrew:
**Is your feature request related to a problem? Please describe.**
In our project we have several different forms which have a field represented by an array of objects with specific type.
We have created a reusable React component to edit this arrays, basically a custom field (in abstract sense).
This component accepts `control` and `name` properties and uses `useFieldArray`.

Because this component is generic, we can not infer the value type from `FieldValues` since the specific schema of `FieldValues` is unknown at that point. 
But our component expects `fields` returned from `useFieldArray` to have specific type.
Currently typings in `react-hook-form` do not provide sufficient capabilities, and we have to rely on manual type casting via `field as unknown as MyType`.

Note that this is only a type system problem, javascript code works fine.

[See the example in CodeSandbox](https://codesandbox.io/s/react-hook-form-use-field-values-generic-problem-oi5qz)

This problem also arises for simple custom fields when we use `useController` hook. 
For example, if we have a custom text field, we would like to restrict value type to be `string`.

[useController exampe in CodeSandbox](https://codesandbox.io/s/react-hook-form-use-controller-generic-problem-dcp3s?file=/src/App.tsx)

**Describe the solution you'd like**
The solution I envision consists of two parts:
1. We need to allow specifying type constraints in `FieldArrayPath` and `FieldPath` so that when defining custom field components we could restrict `TFieldArrayName` and `TFieldName` to extract only those paths which have the desired type. This would allow for type safety for users of our custom field component in concrete forms.
2. Hooks `useFieldArray` and `useController` should allow for custom type specification for cases where concrete shape of `FieldValues` is unknown. This could be a third type argument which defaults to `FieldArrayWithId` / `PathValue`. This would allow for type safety inside our custom field component.

These two parts work together to ensure complete type safety given that we correctly specify the same custom type constraint in both of them. 

If we only have the first part of the solution, typescript is not smart enough to extract the correct value type when using the `useFieldArray` or `useController` hooks.

Here is the CodeSandbox with some of the library types changed which kinda shows the complete solution:
https://codesandbox.io/s/react-hook-form-use-controller-generic-33yo5

**Describe alternatives you've considered**
No other API in `react-hook-form` seems to provide the desired functionality for custom field components.

bluebill1049:
cc @kotarella1110

MonsterDeveloper:
[...]

PR #6466, opened 2021-09-05
Make `Path` carry the expected result type to allow fully typed generic components

julienfouilhe:
I'm in no way a Typescript expert but I wanted to be able to create generic form components.
The way I thought it could work was if the `Path` type returns all paths for a compatible type instead of all possible paths.

So you can write components like this (where expected type would actually be infered from a yup/zod/iots... schema)

```typescript
type ExpectedType = { fileName: string; url: string };

const GenericFileUpload = <FormValues extends FieldValues>({
  name,
  control,
}: {
  name: FieldPath<FormValues, ExpectedType>;
  control: Control<FormValues>;
}) => {
  useController({
    name,
    control,
    defaultValue: { fileName: 'My test file', url: 'https://localhost/image.jpg' }, // this is typed and will fail if you pass it something that does not extend ExpectedType
  });

  return null;
};
```

Used like this:

```typescript
interface FormValues {
  invalid: { 
    subpath: string;
  };
  file: {
    fileName: string;
    url: string;
  };
}

// this will pass
<GenericFileUpload name="file" control={control} />
// while this will not, as expected
<GenericFileUpload name="invalid" control={control} />
```

⚠️ This does not handle all use cases yet and I may have went a little overboard where I put it but I wanted to check if you thought this could be a good idea before continuing?

### DIFF
--- a/app/src/setError.tsx
+++ b/app/src/setError.tsx
@@ -45,7 +45,7 @@ const SetError: React.FC = () => {
         message: 'Minlength is 10',
       },
     ].forEach(({ name, type, message }) =>
-      setError(name as FieldPath<FormInputs>, { type, message }),
+      setError(name as FieldPath<FormInputs, any>, { type, message }),
     );
     setError('username', {
       types: {
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -8,7 +8,7 @@ import {
 } from '@testing-library/react';
 
 import { Controller } from '../controller';
-import { Control } from '../types';
+import { Control, FieldPath, FieldValues } from '../types';
 import { useController } from '../useController';
 import { useForm } from '../useForm';
 
@@ -27,6 +27,37 @@ describe('useController', () => {
       });
 
       return null;
+    };
+
+    render(<Component />);
+  });
+
+  it('should render generic component correctly', () => {
+    type ExpectedType = { test: string };
+
+    const Generic = <FormValues extends FieldValues>({
+      name,
+      control,
+    }: {
+      name: FieldPath<FormValues, ExpectedType>;
+      control: Control<FormValues>;
+    }) => {
+      useController({
+        name,
+        control,
+        defaultValue: { test: 'test' },
+      });
+
+      return null;
+    };
+
+    const Component = () => {
+      const { control } = useForm<{
+        test: string;
+        key: ExpectedType[];
+      }>();
+
+      return <Generic name="key.0" control={control} />;
     };
 
     render(<Component />);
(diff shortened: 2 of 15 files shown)


bluebill1049:
Thanks for the contribution! I will take a look at it. cc(@kotarella1110)

bluebill1049:
looks like this break the build. 

Screen Shot 2021-09-06 at 8 17 39 pm

julienfouilhe:
@bluebill1049 Yes I know, this PR is not finished, I just wanted to know if you thought this was a good idea or if you had some context that would make this a no go.

bluebill1049:
I like this idea, I think the reusable component is one of the weaknesses of the current strict type with template literal. @kotarella1110 let me know your thoughts.

related issue: https://github.com/react-hook-form/react-hook-form/issues/5055

src/controller.tsx
```diff
@@ -3,9 +3,13 @@ import { useController } from './useController';
 
 const Controller = <
   TFieldValues extends FieldValues = FieldValues,
-  TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
+  TResult = unknown,
```
bluebill1049, line 6: `TFieldValue` sounds better?

src/__tests__/useController.test.tsx
```diff
@@ -26,6 +26,37 @@ describe('useController', () => {
...
   });
 
+  it('should render generic component correctly', () => {
+    type ExpectedType = { test: string };
+
+    const Generic = <FormValues extends FieldValues>({
+      name,
+      control,
+    }: {
+      name: FieldPath<FormValues, ExpectedType>;
```
bluebill1049, line 36: I guess the only cons/pros is the user will be able to overwrite/skip the type check.

slbls:
@julienfouilhe, is there anything I can do to help with this PR? The ability to type generic components would be a great addition to React Hook Form and directly useful in projects of my own.

bluebill1049:
> @julienfouilhe, is there anything I can do to help with this PR? The ability to type generic components would be a great addition to React Hook Form and directly useful in projects of my own.

Thanks 👍 will let u know if we need some help. Give us sometimes on this one.

bluebill1049:
hey @barrymay may i get your thoughts on this PR as well?

app/src/setError.tsx
```diff
@@ -45,7 +45,7 @@ const SetError: React.FC = () => {
         message: 'Minlength is 10',
       },
     ].forEach(({ name, type, message }) =>
-      setError(name as FieldPath<FormInputs>, { type, message }),
+      setError(name as FieldPath<FormInputs, any>, { type, message }),
```
barrymay, line 48: can we use unknown instead of any here?

barrymay:
@julienfouilhe Agreed this would be a great add to the library, although you really want to make sure it has good defaults for everyone, while doing what you say it's gonna do.  It's certainly something I've been interested in!

I'll have to play with this branch a bit before I comment more ;)

kotarella1110:
I like this suggestion.
If it is difficult to fix, I can help you 👍🏻

bluebill1049:
I think this is one of the missing pieces right now for reuseable components, would be great to push this one moving forward since both @barrymay and @kotarella1110 agrees too.

julienfouilhe:
Closing in favor of #6562

PR #6531, opened 2021-09-12
fix typos

frankie303:


bluebill1049:
thanks a lot for those corrections.

src/logic/getResolverOptions.ts
```diff
@@ -10,14 +10,14 @@ import set from '../utils/set';
 
 export default <TFieldValues>(
   fieldsNames: Set<InternalFieldName> | InternalFieldName[],
-  _fieldss: FieldRefs,
```
bluebill1049, line 13: 🤦🏻‍♂️

PR #6550, opened 2021-09-15
Support UTF-8 Email

kiznick:


bluebill1049:
thanks for the PR, looks like there is some security issue been found.

bluebill1049:
in fact, would like to have this as a separate test?

PR #6562, opened 2021-09-16
✨  Add support for generic components using `FieldPathWithValue`

julienfouilhe:
This seems like a better and much simpler API for the support of generic components than #6466

If the expected type is not specified to useController/`Controller`, then all possible types from the `FieldValues` are returned.

You can look at the test in `useController.test.tsx` for how it can be used.


julienfouilhe:
cc @bluebill1049 @kotarella1110 and @barrymay since you were tagged as reviewers on my other PR.

I could not find a way to keep the tests the same, I had to explicitly pass the types to `userController`/`Controller`. I think there may be a way to infer everything but I couldn't find a way.

If someone better at Typescript than me knows of a way, please tell me 🙏

src/useController.ts
```diff
@@ -17,10 +17,14 @@ import { useFormState } from './useFormState';
 
 export function useController<
   TFieldValues extends FieldValues = FieldValues,
-  TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
+  TResult = any,
```
julienfouilhe, line 20: This needs to be `any` if unspecified because we want `TName` to be able to be one of all the possible field paths of `TFieldValues`.
There's a check where it's actually used so TResult should never really be used as `any`

src/types/controller.ts
```diff
@@ -24,49 +25,69 @@ export type ControllerFieldState<
...
   name: TName;
   rules?: Omit<
     RegisterOptions<TFieldValues, TName>,
     'valueAsNumber' | 'valueAsDate' | 'setValueAs' | 'disabled'
   >;
   shouldUnregister?: boolean;
-  defaultValue?: UnpackNestedValue<FieldPathValue<TFieldValues, TName>>;
+  defaultValue?: IsAny<TResult> extends true
+    ? FieldPathValue<TFieldValues, TName>
+    : TResult;
```
julienfouilhe, line 59: If the expected type is `any`, we try to resolve the type of the field using the field values type and the path.

If it's specified, we use that type.

That allows for components using:

```typescript
type FormValues = { test: string; test: string[] };

const { control } = useForm<FormValues>;

useController({ control, name: 'test', defaultValue: 'str' });
```
to resolve `TResult` to `FormValues[name]` because it knows both the FormValues and the name key.

But it also allows generic components, for which we do not know the FormValues and the path, to actually fix the expected type.

bluebill1049: Thank you so much for this amazing contribution ❤️

bluebill1049 (dismissed):
LGTM.

bluebill1049:
let's wait for @barrymay and @kotarella1110's review feedback 🙏

src/useController.ts
barrymay: Superb work - thanks for the explanation.  I look forward to leveraging this!

barrymay (dismissed):

src/types/controller.ts
```diff
@@ -24,49 +25,69 @@ export type ControllerFieldState<
...
 > = {
   onChange: (...event: any[]) => void;
   onBlur: () => void;
-  value: UnpackNestedValue<FieldPathValue<TFieldValues, TName>>;
+  value: IsAny<TResult> extends true
+    ? FieldPathValue<TFieldValues, TName>
+    : TResult;
   name: TName;
   ref: RefCallBack;
 };
```
kotarella1110, line 41: Does `ControllerRenderProps` and `UseControllerProps` type need `TResult` ?
I think that `TResult` can be infer from `TName` 👀 

As follows: 

```ts
export type ControllerRenderProps<
  TName extends FieldPathWithValue<TFieldValues, TResult> = FieldPathWithValue<
    TFieldValues
  >,
> = {
  onChange: (...event: any[]) => void;
  onBlur: () => void;
  value: TName extends FieldPathWithValue<TFieldValues, infer TResult>
    ? IsAny<TResult> extends true
      ? FieldPathValue<TFieldValues, TName>
      : TResult
  : FieldPathValue<TFieldValues, TName>;
  name: TName;
  ref: RefCallBack;
};
```

src/types/controller.ts
```diff
@@ -6,9 +6,10 @@ import {
   FieldErrors,
   FieldPath,
   FieldPathValue,
+  FieldPathWithValue,
   FieldValues,
+  IsAny,
   RefCallBack,
-  UnpackNestedValue,
```
kotarella1110, line 11: Why did you remove the `UnpackNestedValue` type?

kotarella1110 (commented):
Amazing!!! 🙌
Please check my comments.

src/types/controller.ts
julienfouilhe: In the context of a generic component, TFieldValues is just another generic variable, just as TName. Therefore the TResult type cannot be infered.

The goal of a generic component is to define itself the type it expects, so it can't infer it, it has to define it itself.

kotarella1110: I see. Thanks!

src/types/controller.ts
```diff
@@ -24,49 +25,69 @@ export type ControllerFieldState<
...
+  TResult = any,
+  TName extends FieldPathWithValue<TFieldValues, TResult> = FieldPathWithValue<
+    TFieldValues,
+    TResult
+  >,
 > = {
-  field: ControllerRenderProps<TFieldValues, TName>;
+  field: ControllerRenderProps<TFieldValues, TResult, TName>;
   formState: UseFormStateReturn<TFieldValues>;
   fieldState: ControllerFieldState<TFieldValues, TName>;
```
kotarella1110, line 73: Can you fix the `fieldState.error` type in the same way? 🙏🏻 

```ts
export type ControllerFieldState<
  TResult = any,
  TName extends FieldPathWithValue<TFieldValues, TResult> = FieldPathWithValue<
    TFieldValues,
    TResult
  >,
> = {
  invalid: boolean;
  isTouched: boolean;
  isDirty: boolean;
  error?: FieldErrors<
    IsAny<TResult> extends true ? FieldPathValue<TFieldValues, TName> : TResult
  >;
};
```

src/types/controller.ts
kotarella1110: `NestedValue` type is still used for `errors`.

```ts
type FormValues = {
  test: NestedValue<{ custom: string }>;
}
const { fomState: { errors } } = useForm<FormValues>();
```

It is still needed 👍🏻

src/types/controller.ts
julienfouilhe: @kotarella1110 You're right, just did! Thanks!

src/types/controller.ts
julienfouilhe: I don't get what it's for, and therefore how to fix it.

Could you maybe provide a test that doesn't pass? 
Why would someone want to use NestedValue in your example. What is the benefit?

julienfouilhe: I added it back on `FieldPathValue`, but I don't know if I should use it for `TResult` too?

kotarella1110: please see docs: https://react-hook-form.com/ts#NestedValue
The `NestedValue` type is used when the field value is an object.

The following is an example where field name is `test` and field value is `{ custom: string }`.

```ts
type FormValues = {
  test: NestedValue<{ custom: string }>;
}
const { fomState: { errors } } = useForm<FormValues>();
// errors
// {
//    test?: FieldError;
// }
```

If `NestedValue` is not used, it is not possible to derive the correct `errors` type.

```ts
type FormValues = {
  test: { custom: string };
}
const { fomState: { errors } } = useForm<FormValues>();
// errors
// {
//    test?: {
//      custom?: FieldError;
//  }
// }
```

kotarella1110: > I added it back on `FieldPathValue`

Thanks! looks good!

>  don't know if I should use it for `TResult` too?

This is a tough one...
@bluebill1049 What do you think?

julienfouilhe: @kotarella1110 I'll add tests. It will show if unpacking is needed :)

src/types/controller.ts
bluebill1049: thanks @julienfouilhe 👍

src/types/controller.ts
julienfouilhe: @kotarella1110 Sorry for the delay, I added a test! Definitely needed to use `UnpackNestedValues` for `defaultValue` and `value`. It would not pass otherwise! Thanks for catching that!

bluebill1049:
@julienfouilhe looks like my last merge from master break your buid. could you take a look at it? 🙏

julienfouilhe:
@bluebill1049 Fixed. Seems like `FieldErrors` behaviour changed and did not like very much primitives anymore.

```typescript
type t = FieldErrors<string>; // fails because string is not compatible with FieldValues instead of returning FieldError
type v = FieldErrors<NestedValue<{ test: string}>>; // { [$NestedValue]?: undefined; test?: FieldError | undefined; } instead of FieldError
```

bluebill1049:
I think that make sense, thanks for the update.

src/__tests__/useController.test.tsx
```diff
@@ -33,6 +38,82 @@ describe('useController', () => {
...
+    const Component = () => {
+      const { control } = useForm<{
+        test: string;
+        key: ExpectedType[];
+      }>();
+
+      return <Generic name="key.0" control={control} />;
+    };
+
+    render(<Component />);
```
barrymay, line 76: thanks for the test, but can you add an expectation of the result?

bluebill1049:
@kotarella1110 would you like to re-review this PR?

src/__tests__/useController.test.tsx
```diff
@@ -33,6 +38,82 @@ describe('useController', () => {
...
+  });
+
+  it('should be able to access values and error in generic components using NestedValue', () => {
+    type ExpectedType = NestedValue<{ test: string }>;
```
kotarella1110, line 80: Awesome!

kotarella1110 (dismissed):
LGTM 🎉

julienfouilhe:
@bluebill1049 Done! Sorry for the force-push which dismissed @kotarella1110 review! I usually rebase instead of merge so I was confused for a moment. But all good now!

The diff of the merge commit is weird but I just moved a type I had called `ControllerFieldErrors` and that was below `ControllerFieldState` above it, and renamed it `ControllerFieldError` which was the name used in one of the latest commits. 

I also added expectations in my tests as @barrymay suggested. Since they were tests that just ensured everything was typed correctly, I just put basic expectations.

bluebill1049 (approved):

bluebill1049:
Thanks for everyone's contribution and review. Let's do another patch release for all the bug fixes before release this as a new minor update. I will merge that once the patch goes out, there are still few small issues yet to be resolved.

barrymay (approved):

bluebill1049:
Just some thoughts, if this is a pattern that we all agree on, perhaps we can introduce something similar to `useFieldArray`? any thoughts.

julienfouilhe:
@bluebill1049 Definitely something that can be applied to it too!
I don't know how people create generic components. I mainly use Controller thus why I implemented it there, but if others use extensively `useFieldArray` (which I haven't used much for now) it could be useful to make that change there later too!

src/types/controller.ts
```diff
@@ -1,80 +1,115 @@
...
-  UnionLike,
   UnpackNestedValue,
   UseFormStateReturn,
 } from './';
 
-type ControllerFieldError<T> = DeepMapImpl<
-  DeepPartialImpl<UnionLike<T>>,
-  FieldError
->;
+type ControllerFieldError<
```
bluebill1049, line 20: I didn't pay close enough attention here, @kotarella1110 this was completely changed after the merge.

bluebill1049: which potentially lead to this: https://github.com/react-hook-form/react-hook-form/issues/6679

julienfouilhe: @bluebill1049 I don't understand the linked issue is talking about versions before this was merged right?

bluebill1049: Screen Shot 2021-10-05 at 6 41 06 pm

i am just digging around at the moment, see if i can find a fix for it.

bluebill1049: ` DeepMapImpl<DeepPartialImpl<UnionLike<TFieldValues>>, FieldError>`

vs

`DeepMap<DeepPartial<UnionLike<TFieldValues>>, FieldError>`

bluebill1049: still try to get my head around it.

bluebill1049: ok confirmed the latest definitely causing the issue, it could be related to the user's typescript version as well, as i couldn't reproduce the issue at the lib level but codesandbox with CRA, after omit TResult and the issue went away.

bluebill1049: https://codesandbox.io/s/pedantic-elbakyan-s04xv?file=/src/App.tsx

bluebill1049:
I have to revert this one back guys, i don't have quick solution yet and it probably impact a lot of users.

bluebill1049:
we can come back to this, I am sorry to roll this back but it's better not impact more users.

bluebill1049:
I think we still want this awesome feature.

- [ ] reproduce the issue locally (the one in codesadnbox) https://codesandbox.io/s/pedantic-elbakyan-s04xv?file=/src/App.tsx switch to 17.0.0, it's strange behavior as it wasn't having the issue within all the test local lib has...
- [ ] re-open the PR and publish again

barrymay:
> switch to 17.0.0

For clarity, switch RHF to 7.17.0 to see the issue.  As you said, very strange this didn't come up in the test run.  I agree this feature is very much worth trying to get right (and also worth ensuring our test runs can catch this too)

bluebill1049:
`react-hook-form@7.18.0-next.0`

PR #6614, opened 2021-09-22
refactor subscription logic into useSubscribe 

### LINKED ISSUE #6613: [Bug]:  watch(from useFormContext) does not subscribe changes in Child Component
### Version Number

7

### Codesandbox/Expo snack

https://codesandbox.io/s/react-hook-form-v7-watch-bug-uv54e?file=/src/App.js

### Steps to reproduce

1. Go to the sandbox
2. During initialisation, first name & last name form field stay empty

### Expected behaviour

During initialisation, first name & last name form field should be update correctly (value: a & b)

If I downgrade to v6, the watch works (subscribe the change properly)
Or
If I call watch (from useForm) in the parent component(App), it also works.

I investigated a little bit and find out that the `_subjects.state` in `createFormControl` (https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts) is not set correctly(always empty) in the buggy case.
And when setValue, it try to emit changes to all subscribers from `_subjects.state`, but find no subscribers(https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L749).

Firefox, Chrome

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
## Change

Subscription with `formState` & `formValue` will initialised at function invocation instead `useEffect`. 

close #6613


src/useWatch.ts
```diff
@@ -49,50 +50,44 @@ export function useWatch<TFieldValues>(props?: UseWatchProps<TFieldValues>) {
...
+    callback: ({ name: fieldName }) => {
+      if (
+        !name ||
+        !fieldName ||
+        convertToArrayPayload(name).some(
+          (fieldName) =>
+            fieldName &&
+            fieldName &&
+            (fieldName.startsWith(fieldName as InternalFieldName) ||
+              fieldName.startsWith(fieldName as InternalFieldName)),
```
byuji, line 66:
```suggestion
          (currentName) =>
            fieldName &&
            currentName &&
            (fieldName.startsWith(currentName as InternalFieldName) ||
              currentName.startsWith(fieldName as InternalFieldName)),
```

bluebill1049: thanks.

PR #6605, opened 2021-09-21
fix Mapped types for defined value type with interface

kotarella1110:
## Overview

Improved to not map types defined with `interface`.
We can determine the type defined with `interface` as follows:

```ts
interface Interface {
    test: string;
}
type TypeAlias = {
    test: string;
}

type IsInterface<T> = T extends Record<any, unknown> ? false : true;

type Result1 = IsInterface<Interface>; // true
type Result2 = IsInterface<TypeAlias> // false
```

[TypeScript Playground](https://www.typescriptlang.org/play?ts=4.4.0-beta&ssl=11&ssc=47&pln=1&pc=1#code/JYOwLgpgTgZghgYwgAgJLmvJyDeAoZQ5SAZzAC5kypQBzAbjwF88wBPABxQBVOIBBADbA4JZAF5cBIqQpUwNEA2Z5WfNCXSRYiCAB5uAPgnJuyCAA9IIACZiAShAQB7KDb1wQbADTIAriAA1iDOAO4gxgD8yPCCJCiUCn4QjGpcyI4kfoJgAIwmqJoYOkh6Wpi6hvTIAPQ1xFDJaSiZ2WAATAVF2lj6vFxCIiTGdTFwcRDNGRBZOQDMXeUl+gAicJAj9UmT7OmtOQAsi8W9egBiwIIQmw1Nuy0zbQCsxz2655cQADLAZDfbU32YAA7K8KqUABQASgkxhwTH+jR26iBAA4wctzgEEGBgM4IrUtkjAY8cgBODGnACycA4emodF8DKUhkRdxRpLyAAZKe8AMoQMD0hR0VmE24QIA)

if a field value is a nested value (object), users only needs to define the field value type with `interface` instead of using the `NestedValue` type.

using `NestedValue`:

```ts
import { useForm, NestedValue } from "react-hook-form";

type FormValues = {
  fieldName: NestedValue<{ custom: string }>;
};

const { formState: errors } = useForm<FormValues>({
  defaultValues: { fieldName: { custom: "value" }}
});

// `defaultValues` type:
// {
//   fieldName?: { custom: string } | undefined;
// }
// `errors` type:
// {
//   fieldName?: FieldError | undefined;
// }
```

using `interface`:

```ts
import { useForm } from "react-hook-form";

interface CustomValue {
  custom: string;
}

type FormValues = {
  fieldName: CustomValue
};

const { formState: errors } = useForm<FormValues>({
  defaultValues: { fieldName: { custom: "value" }}
});

[...]

PR #6626, opened 2021-09-23
 fix Path and ArrayPath types for defined value type with interface 

kotarella1110:
ref: https://github.com/react-hook-form/react-hook-form/pull/6605


bluebill1049:
build pass 👁️

src/types/utils.ts
```diff
@@ -73,13 +75,15 @@ export type Path<T> = T extends ReadonlyArray<infer V>
...
-type ArrayPathImpl<K extends string | number, V> = V extends Primitive
-  ? never
+type ArrayPathImpl<K extends string | number, V> = IsAny<V> extends true
+  ? `${K}` | `${K}.${ArrayPath<V>}`
   : V extends ReadonlyArray<infer U>
-  ? U extends Primitive
-    ? never
-    : `${K}` | `${K}.${ArrayPath<V>}`
-  : `${K}.${ArrayPath<V>}`;
+  ? U extends ReadonlyArray<any> | Record<any, unknown>
```
bluebill1049, line 81: I guess this is the part checking if it's `interface`? any chance we could just add a smile unit test? with interface usage, I can help if you are busy as well.

bluebill1049 (approved):
LGTM! Thank you @kotarella1110

src/types/utils.ts
kotarella1110: I will add unit test!

bluebill1049:
this pr is related to this right: https://github.com/react-hook-form/react-hook-form/issues/6693

bluebill1049:
let's merge this in, we can follow up with the test later on.

Issue #6498, opened 2021-09-08
[Bug]: TypeScript fieldState.error is not an array when the relative data type is an array


peter-oms:
### Version Number

7.14.2

### Codesandbox/Expo snack

https://codesandbox.io/s/react-hook-form-nestedvalue-forked-usecontroller-z72sg?file=/src/App.tsx

### Steps to reproduce

1. Go to [https://codesandbox.io/s/react-hook-form-nestedvalue-forked-usecontroller-z72sg?file=/src/App.tsx](https://codesandbox.io/s/react-hook-form-nestedvalue-forked-usecontroller-z72sg?file=/src/App.tsx)
2. Add a Todo
3. Check the console for the error object output
3. Open `Todos.tsx` and hover on the destructured `error` property
4. See that its type is `FieldError`

### Expected behaviour

Expecting that because the type of todos is an array (defined in `TFormValues.tsx`), any errors that occur when validating the schema would be returned as a FieldError[] array type (it can be seen in the console they are they are actually an array).

This makes getting a particular item in the errors array difficult.

Error type:
![image](https://user-images.githubusercontent.com/83673746/132585750-af88397f-7038-4ffa-8196-14f20a01e66d.png)

Error object:
![image](https://user-images.githubusercontent.com/83673746/132585777-ec2daba7-92f9-48a2-a2ff-b771ae7a9993.png)

Example trying to get the zeroth item of the error array:
![image](https://user-images.githubusercontent.com/83673746/132587187-dd19522c-3977-4aad-bf25-ccce83e7d4a8.png)

### Other attempts

1. I have read a few other issue reports where `NestedValue` was suggested to be used when dealing with arrays or nested structures.  I tried doing this as well, but I haven't been able to achieve any success because I hit another issue in `yupResolver`:<br />
         ```
Property '[$NestedValue]' is missing in type 'TTodo[]' but required in type '{ [$NestedValue]: never; }'.
        ```<br/>
I think this is already covered in [#234](https://github.com/react-hook-form/resolvers/issues/234) of the resolvers package, though. I am not entirely sure that `NestedValue` will actually achieve the behaviour I am looking for though.

   ![image](https://user-images.githubusercontent.com/83673746/132586798-74182516-3542-48e6-a46e-eceb7a42a610.png)

2. Another approach I tried is using `useFieldArray` instead of `useController`, but this wouldn't even register my validation schema errors.<br/>
[...]

PR #6650, opened 2021-09-26
fix #6646

### LINKED ISSUE #6646: [Bug]: 7.16.0 breaks types of fieldState in render Controller
### Version Number

7.16.0

### Codesandbox/Expo snack

https://codesandbox.io/s/clever-wind-e90us?file=/src/App.tsx

### Steps to reproduce

Property 'message' does not exist on type 'string'

### Expected behaviour

Types to work without issues.

Chrome

### Relevant log output

```shell
Property 'message' does not exist on type 'string'.
   31 |                 error={fieldState.error !== undefined}
   32 |                 label="Name"> 
> 33 |                 helperText={fieldState.error?.message}
     |                                               ^
```

### Code of Conduct

bluebill1049:
`fieldState`'s error is not all errors, only contains errors related to the field. I believe the fix was introduced last time was due to the union object.

close #6646

related: https://github.com/react-hook-form/react-hook-form/issues/6498

### DIFF
--- a/src/__tests__/controller.test.tsx
+++ b/src/__tests__/controller.test.tsx
@@ -1238,6 +1238,6 @@ describe('Controller', () => {
 
     render(<App />);
 
-    expect(screen.getAllByRole('textbox'));
+    expect(screen.getAllByRole('textbox').length).toEqual(3);
   });
 });
--- a/src/__tests__/controller.test.tsx
+++ b/src/__tests__/controller.test.tsx
@@ -1189,6 +1189,9 @@ describe('Controller', () => {
   it('should not throw type error with field state', () => {
     type FormValues = {
       firstName: string;
+      deepNested: {
+        test: string;
+      };
     };
 
     function App() {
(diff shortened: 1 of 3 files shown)


bluebill1049:
sorry for keep assign and remove... found an issue as I test.

src/__tests__/controller.test.tsx
```diff
@@ -1185,4 +1185,59 @@ describe('Controller', () => {
...
+            control={control}
+            name="todos"
+          />
+        </form>
+      );
+    }
+
+    render(<App />);
+
+    expect(screen.getAllByRole('textbox'));
```
barrymay, line 1241: should this be expecting check for 3 textboxes?  (just wondering if this might catch a textbox from somewhere else. 

The code itself looks good, just checking to see if we can strengthen this new test.

src/types/controller.ts
```diff
@@ -3,23 +3,31 @@ import * as React from 'react';
...
+
 export type ControllerFieldState<
   TFieldValues extends FieldValues = FieldValues,
   TFieldName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
 > = {
   invalid: boolean;
   isTouched: boolean;
   isDirty: boolean;
-  error?: FieldErrors<FieldPathValue<TFieldValues, TFieldName>>;
+  error?: ControllerFieldError<FieldPathValue<TFieldValues, TFieldName>>;
```
bluebill1049, line 30: 👍 better

src/__tests__/controller.test.tsx
bluebill1049: Will update to include number of inputs check as well 👍

barrymay:
@kotarella1110 great change 😊

bluebill1049:
thanks for the review @kotarella1110 & @barrymay

bluebill1049:
going to release this if u guys are happy with the PR before it impact too many users from upgrade to the latest.

PR #6658, opened 2021-09-28
fix fieldState error type with NestedValue

### LINKED ISSUE #6653: [Bug]: FieldState errors wrong typings
### Version Number

7.16.1

### Codesandbox/Expo snack

Latest version 7.16.1 does not seem to install on Codesandbox, see example repo instead

### Steps to reproduce

1. Clone example repo: https://github.com/StampixSMO/react-hook-form-fieldstate-errors-type
2. Install dependencies
3. Run `npm run compile`
4. Watch the TS error.

### Expected behaviour

The type of the errors object should not reflect the nested objects but only expose the `message` property.
This works fine in 7.15.4.

### Relevant log output

```
error TS2339: Property 'message' does
not exist on type 'DeepMap<DeepPartial<NestedValue<PhoneNumber>>, FieldError>'.
```

### Code of Conduct

bluebill1049:
revert back to this commit: 3985e5bc308b473534c8d44621500c5ad5e4b0d5 and looks like fix the issue. @kotarella1110 I am sure you have better solution than mine 🤣

### DIFF
--- a/src/__tests__/controller.test.tsx
+++ b/src/__tests__/controller.test.tsx
@@ -1254,6 +1254,6 @@ describe('Controller', () => {
 
     render(<App />);
 
-    expect(screen.getAllByRole('textbox').length).toEqual(3);
+    expect(screen.getAllByRole('textbox').length).toEqual(4);
   });
 });
--- a/src/__tests__/controller.test.tsx
+++ b/src/__tests__/controller.test.tsx
@@ -1239,12 +1239,7 @@ describe('Controller', () => {
             name="todos"
           />
           <Controller
-            render={({ field, fieldState }) => (
-              <>
-                <input {...field} />
-                <p>{fieldState.error?.message}</p>
-              </>
-            )}
+            render={({ fieldState }) => <p>{fieldState.error?.message}</p>}
             control={control}
             name="nestedValue"
           />
--- a/src/__tests__/controller.test.tsx
+++ b/src/__tests__/controller.test.tsx
@@ -1239,7 +1239,12 @@ describe('Controller', () => {
             name="todos"
           />
           <Controller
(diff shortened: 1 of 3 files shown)


src/__tests__/controller.test.tsx
```diff
@@ -1193,11 +1193,17 @@ describe('Controller', () => {
         test: string;
       };
       todos: string[];
+      nestedValue: NestedValue<{ test: string }>;
```
bluebill1049, line 1196: include test for `NestedValue` as well

src/types/controller.ts
```diff
@@ -16,7 +16,7 @@ import {
 } from './';
 
 type ControllerFieldError<T> = DeepMapImpl<
-  DeepPartial<UnionLike<T>>,
+  DeepPartialImpl<UnionLike<T>>,
```
bluebill1049, line 19: Awesome!

bluebill1049:
Thank you @kotarella1110!

src/types/utils.ts
```diff
@@ -18,7 +18,7 @@ export type LiteralUnion<T extends U, U extends Primitive> =
   | T
   | (U & { _?: never });
 
-type DeepPartialImpl<T> = T extends NestedValue
+export type DeepPartialImpl<T> = T extends NestedValue
```
barrymay, line 21: this is not a concern but just a curiosity (I need to spend some time wrapping my head around all the latest typings) - is there a reason that NestedValue is in the form.ts file?

bluebill1049: Sounds good to me, i tried to break those type into categories so they are a bit easier to manage.

bluebill1049: by the way, would you like to add yourself to this page? https://react-hook-form.com/about-us

barrymay: can do!

PR #6662, opened 2021-09-28
Add Example for Dirty Field Only Submission

nicholascm:
I found this to be a helpful example for a common use-case - interacting with a PUT endpoint which only wants changed fields. Credit for utility goes to @dandv and others here: https://github.com/react-hook-form/react-hook-form/discussions/1991#discussioncomment-1393156


bluebill1049:
Thank you.

PR #6692, opened 2021-10-01
code cleanup by removing unnecessary formState from useForm

kotarella1110:


kotarella1110:
I thought I could reduce the package size... hahaha

kotarella1110:
reduce code size: -5 B 👍🏻 
Also, the code is now more readable than before.

bluebill1049:
Thanks @kotarella1110 for the PR. However, this will break the `useEffect` at `deps` with `formState`. In fact, with react 18 we should break all states into individual `useState`, so `useEffect` can take individual form state. The original reason to have all form state in a single object is so we can do a batch update alike state update, however, there is a side effect that leads to `useEffect` won't be able to differentiate the different individual state update, but it's not the case anymore with react 18.

kotarella1110:
Oh, I see! I understood it.
close this 👍

Issue #6703, opened 2021-10-04
[Bug]: Not updating isDirty form state on date comparison


garthtee:
### Version Number

7.15.3

### Codesandbox/Expo snack

https://codesandbox.io/s/react-hook-form-v7-controller-forked-pfs0q?file=/src/index.js

### Steps to reproduce

1. Go to the [code sandbox](https://codesandbox.io/s/react-hook-form-v7-controller-forked-pfs0q?file=/src/index.js).
2. Select a different date and see the dirty field change as expected. (e.g. 27th)
3. Select the default date again (28th) and see the dirty state not change as expected
 

### Expected behaviour

IsDirty should update when the date object returns to the same date. This issue persists when using setValue with a `shouldDirty` option.

### What browsers are you seeing the problem on?

Firefox, Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bluebill1049:
hi there, thanks for the detailed issue report. currently, this is expected behavior, please refer to this issue/feature request for the reason why this is the case: https://github.com/react-hook-form/react-hook-form/issues/6657

bluebill1049:
going to close this and address it in the feature request above, looks like there are quite a few upvotes.

PR #6706, opened 2021-10-04
close #6657 support equality check on date time object

### LINKED ISSUE #6657: Support equality checks on form values of Date objects
The input field of the project I work on works with native Date objects. We also allow the user to clear any changes to an individual input field back to its default value. However, since RHF's `deepEqual` function only does a simple reference equality check on Date objects, manually setting the value back to its default value doesn't satisfy the check of if the field `isDirty`, since `(new Date()) !== (new Date())`

The `deepEqual` utility that RHF uses to check against default values already checks whether either of the values are instances of `Date`. Rather than handle `Date` objects with a simple `===` check, I'd love for there to be a separate case to handle the comparison of two `Date` objects in a way like:
```JS
if (isDateObject(object1) && isDateObject(object2)) {
  return object1.valueOf() === object2.valueOf();
}
```

The most obvious alternative is for my team to refactor our project's use of Date objects to instead work with Dates in some normalized primitive representation like Epoch string, UTC, etc. This would be a non-trivial amount of work for us, so I'm hoping that supporting equality checks on Date objects would be a welcome enhancement to RHF, since you already make an effort to handle more typical object types of n-depth.

bluebill1049:
close #6657 
related #6703

thanks @jviall's suggestion on the implmemantion.

Screen Shot 2021-10-04 at 10 41 15 pm

switched to `getTime` instead of `valueOf`

### DIFF
--- a/src/__tests__/utils/deepEqual.test.ts
+++ b/src/__tests__/utils/deepEqual.test.ts
@@ -89,4 +89,10 @@ describe('deepEqual', () => {
       ),
     ).toBeTruthy();
   });
+
+  it('should compare date time object valueOf', () => {
+    expect(
+      deepEqual({ test: new Date('1990') }, { test: new Date('1990') }),
+    ).toBeTruthy();
+  });
 });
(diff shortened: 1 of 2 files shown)


bluebill1049:
Screen Shot 2021-10-04 at 10 47 26 pm
issue with github action, so run automation locally and passed.

Issue #6679, opened 2021-09-30
[Bug]: Regression: TS2589: Type instantiation is excessively deep and possibly infinite


StampixSMO:
### Version Number

7.16.2

### Codesandbox/Expo snack

Codesandbox still doesn't allow me to install the latest version

### Steps to reproduce

1. Upgrade from 7.16.1 to 7.16.2, no other changes
2. Cue error

### Expected behaviour

TypeScript compilation should succeed without any warnings.
I believe #6658 is the cause of introducing this error, this is a regression of #4441

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bluebill1049:
> Codesandbox still doesn't allow me to install the latest version

you can provide an old version as well.

StampixSMO:
> > Codesandbox still doesn't allow me to install the latest version
> 
> you can provide an old version as well.

The bug was introduced in 7.16.2 for me, I can only install up until 7.15.3 in CSB

bluebill1049:
> The bug was introduced in 7.16.2 for me, I can only install up until 7.15.3 in CSB

You can provide the same old with an old version in a codesanbox.

StampixSMO:
> > The bug was introduced in 7.16.2 for me, I can only install up until 7.15.3 in CSB
> 
> You can provide the same code with an old version in a codesanbox. I need the code so i can reproduce.
Sorry but I don't succeed in getting a reproducible scenario with a small codebase.
It's something that pops up in large codebases, see TS issue on it: https://github.com/react-hook-form/react-hook-form/issues/6679#issuecomment-931269260

bluebill1049:
then I am afraid it would be very difficult for me to investigate if I don't have a codesanbox to reproduce the problem, i will keep this issue open for couple days, if other have similar issue.

sebastian-kricke:
We´re migrating from 6.15.8 to latest 7.16.1 or today 7.16.2 but having hard issues on that.

It´s big industry code base with tree of nested domain interfaces and causes problems even on start. It hang on typecheck, killing IDE- functionality also.

This is error message after some actions:

> Files successfully emitted, waiting for typecheck results...
> 
> <--- Last few GCs --->
> 
[...]

PR #6730, opened 2021-10-07
Simplify controller field state error type

bluebill1049:
JS: https://codesandbox.io/s/xenodochial-sound-qlqfh?file=/src/App.js
TS: https://codesandbox.io/s/serene-goldstine-4dx4x?file=/src/App.tsx

related #6679

- Controller with array type should still return FieldError instead of an array of errors, this should help with the recursive issue that we are facing right now.
- Controller is always a single unit

Screen Shot 2021-10-07 at 12 25 18 pm

### DIFF
--- a/src/__tests__/controller.test.tsx
+++ b/src/__tests__/controller.test.tsx
@@ -1232,7 +1232,7 @@ describe('Controller', () => {
             render={({ field, fieldState }) => (
               <>
                 <input {...field} />
-                <p>{fieldState.error?.[0]?.message}</p>
+                <p>{fieldState.error?.message}</p>
               </>
             )}
             control={control}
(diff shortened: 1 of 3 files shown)


kotarella1110 (approved):
LGTM! Thank you very much!

#6498 will occur again, but #6679 seems to have a larger impact on users. Therefore, I think this PR should be merged and released first.
We'll come up with a different solution for #6498 later.

bluebill1049:
> #6498 will occur again,

I think so, ideally, `Controller` should only host a single node of error, otherwise, it's overcomplicated, I am happy to take that as design limation for typescript, after all, there is only so much we can achieve especially with recursive limitation. I will attach #6498 as a design limation for now until we have a possible solution.

PR #6717, opened 2021-10-05
Revert "✨  Add support for generic components using `FieldPathWithVal…

bluebill1049:
…ue` (#6562)"

This reverts commit f5a25199c27b4b0622556ca634cd869b271117eb.

fix #6679

PR #6747, opened 2021-10-08
⏮ revert interface & mapped type update

bluebill1049:
close #6679
close #6723

PR #6753, opened 2021-10-08
Revert "Revert "✨  Add support for generic components using `FieldPat…

bluebill1049:
…hWithValue` (#6562)" (#6717)"

This reverts commit 08883bb4d5bfe6898c0dfd82815afaeafb5e8531.


bluebill1049:
`react-hook-form@7.18.0-next.0`

barrymay:
I found something very interesting about this problem when playing around between code sandbox and vscode tonight, and I think I have an answer as to why this is happening:

First a note
- With the codesandbox we've noted about the problem (here's my copy https://codesandbox.io/s/polished-waterfall-srdmv?file=/src/App.tsx), it's easy to reproduce, but note that you CAN'T clearly control  what version of TypeScript code sandbox uses - there's issues noted about this (https://github.com/codesandbox/codesandbox-client/issues/942) but they're not resolved or answered.

So here's what I did:
- On my local machine, Ran a brand new `yarn create react-app my-app --template typescript` to create an instance of the demo project
- Put all the code in from the codesandbox, and the issue doesn't show.
- I then revert my codebase in the package.json to 4.1.2 (which from the issue looks like the ts version codesandbox uses (at least potentially), and run yarn install (or npm install if you so like, doesn't matter what you use)
- Then I force VSCode to use the version from the project
  - To do this, open App.tsx, click on the "{}" next to where it says "TypeScript React" on the bottom right of VSCode, in the popup there hit "select version" and in the dropdown on top hit "Use Workspace Version", which will make VSCode USE 4.1.2 as its visual compiler instead of the default which is going to be a newer version, like 4.4.3
- Once I do this I see this in my VSCode, for the first time in this codebase

image

So in summary, I don't think you can trust codesandbox to show what version of TS ppl are using, and we should encourage our RHF users to use recent version of TypeScript

Lastly, in his recent TSConf keynote, Anders Hejlsberg mentioned they've made great advances in the very recent versions to avoid this "excessively deep instantiation problem" which they used to admit they used in order to keep TS compiling fast.  Therefore I believe the problem just doesn't exist any more in the latest TS versions, and we should (potentially) restrict RHF usage to recent versions of TS (4.3/4.4) so that users get our best new typing benefits.

I'm happy to see this proven wrong in a GitHub repo if needed, but this what I believe for now.

bluebill1049:
that's an awesome discovery 👍 thanks very much for looking closer look into it. We did mention this in the doc for combine TS4.3 above.

Screen Shot 2021-10-09 at 4 23 02 pm

from memory, the recursive issue even impacted the user with 4.3 above. https://github.com/react-hook-form/react-hook-form/issues/6679#issuecomment-932202692

barrymay:
yeah you're right - in that case it's just a note to be careful with codesandbox reproductions of typing issues, as I'm not sure when that version issue will be fixed.  

I saw the repro related to Mui 5.0 (which has been a significant update) - I'll check it out to see if this is reproducible there

bluebill1049:
thanks, @barrymay I think v7.17.0 is the version that pushing the recursive limit even further 

combination of the following:

- https://github.com/react-hook-form/react-hook-form/pull/6562/files/40d5fd34fe9e6ed3626a300ef2305049627dea39#diff-c087cbd3569794c8873e99551d7bf95c6ac3bfeb6693830ff592149e0ac2279fR32
- https://github.com/react-hook-form/react-hook-form/pull/6747
- https://github.com/react-hook-form/react-hook-form/pull/6730/files#diff-070b8e435c80a2965d5d5bcd7df56033f7ecd901f0bb6bd44b4a581c41d70a44L36

each time we are doing a deep type check, it pretty much `n*n*n` type check which potentially leads to the recursive issue at the end, we just gotta be careful to play with the limit, I am yet to find a solution to early bail out with type deep nested check. probably something is out there which I didn't know.

barrymay:
that's fair @bluebill1049 I agree being careful with our nesting types is important.  

I guess my point here is that if ppl were having the problem were specifically having with 7.17.0, then the issue MAY be a VSCode issue as I've noted here, and code sandboxes could be a red herring.  

Another thing I'll note (and I had previously a note about Anders Heljsberg's tsconf talk earlier which I misstated, which I'll correct here). 

In the upcoming TS 4.5, there's this https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#tail-recursion-elimination-on-conditional-types - this essentially states that they've taken a looping trick from functional programming to allow in certain cases for essentially **much** deeper, if not seemingly infinite type recursion in cases where the type is "tail-recursive".  In the example on the page they also show how a non-tail-recursive type could be turned into a tail-recursive type to get the same treatment.

So while the VSCode trick I mentioned above seems to help to some extent, the upcoming 4.5 change could help to a much larger one!

bluebill1049:
> In the upcoming TS 4.5, there's this https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#tail-recursion-elimination-on-conditional-types - this essentially states that they've taken a looping trick from functional programming to allow in certain cases for essentially much deeper, if not seemingly infinite, type recursion in cases where the type is "tail-recursive". In the example on the page they also show how a non-tail-recursive type could be turned into a tail-recursive type to get the same treatment.

So while the VSCode trick I mentioned above seems to help to some extent, the upcoming 4.5 change could help to a much larger one!

that's really exciting stuff!

bluebill1049:
just a thought: I guess we can potentially roll back to what we had, but locked with ^4.5?

barrymay:
> just a thought: I guess we can potentially roll back to what we had, but locked with ^4.5?

Yes, once 4.5 comes out of beta and we test it, I think it's a very fair thing to try.  If ppl still see problems with it, I'd very much like to get a Github Codespace example (which doesn't have the Codesandbox issues) or a zoom/google meet-based demo on the issue if we can.

bluebill1049:
update on number of users are the `next`: Screen Shot 2021-10-11 at 5 54 41 pm

bluebill1049:
plan on merge this back in next week if no issue is getting reported.

slbls:
Is there any way to use the older types when using an older version of RHF? Right now I have a project depending on generic components and reinstalling `7.18.0` brings in the new type definitions for `7.18.1` with generic component support removed. I understand this feature is not working as intended and needed to be rolled back, but the project I'm working on now requires the old type definitions., albeit that they may be incomplete or incorrect.

bluebill1049:
thanks for the feedback @slbls We don't have a good plan at the moment so ensure this feature doesn't bring other breaking issues for existing applications. This is definitely something we want to support.

slbls:
Of course, @bluebill1049. I understand needing to roll it back to ensure others don't experience issues. I'm just curious as to if there's a way to install `7.18.0` with the definitions for that version.

In other words: my project has `7.18.0` installed, but is pulling type definitions for `7.18.1`. Is there a way that I can use the `7.18.0` type definitions while `7.18.0` is installed?

bluebill1049:
I don't think that's possible as far as I know.

slbls:
Okay, thank you for the quick reply. I'll find a way to use the old types for the time being and report back with a workaround in case anyone else needs them as a temporary fallback until the feature is reintroduced.

bluebill1049:
thanks for understanding, let's try to find a solution for this. 👍

slbls:
For projects using `7.18.0`'s generic controller, but missing the old type definitions:

1. Install `react-hook-form@7.18.0`
2. Copy the files [located here](https://github.com/react-hook-form/react-hook-form/tree/v7.18.0/src/types) into `node_modules/react-hook-form/types`

It's important to note that this won't fix the issues mentioned in #6933 or #6934, but if a project depends on the implementation of generic controllers in `7.18.0` then this will allow them to continue being used.

PR #6762, opened 2021-10-11
Add support for extended generics in path type

### LINKED ISSUE #6726: [Bug]: `Path<T>` can not handle generics
### Version Number

7.17.1

### Codesandbox/Expo snack

https://codesandbox.io/s/rhf-generic-path-report-iwvwy?file=/src/App.tsx

### Steps to reproduce

1. Create a component, that takes a generic, which extends a minimum set of field values
2. Take the `useForm`-return from a parent component through a prop typed as `UseFormReturn<T>`
3. Try to register a field with one of the prop names defined in the minimum field set
4. Error: Argument of type '<field-name>' is not assignable to parameter of type 'Path<T>'

This becomes way clearer if you look at the codesandbox.

### Expected behaviour

I should be able to register a field with the paths, that can be derived from the extended shape of the generic.

Firefox, Chrome, Safari, Edge

### Relevant log output

_No response_

### Code of Conduct

hpohlmeyer:
With this change the `Path<T>` type can take a generic as `T` and honors the existing keys on said generic. This solution does not support nested generic shapes though.

Fixes #6726

### DIFF
--- a/src/types/utils.ts
+++ b/src/types/utils.ts
@@ -73,15 +73,17 @@ type PathImpl<K extends string | number, V> = V extends Primitive
   ? `${K}`
   : `${K}` | `${K}.${Path<V>}`;
 
-export type Path<T> = T extends ReadonlyArray<infer V>
-  ? IsTuple<T> extends true
-    ? {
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks 🌷

Issue #5038, opened 2021-05-05
Typescript Structural Typing does not work on `UseFormReturn` type


SimonSelg:
**Describe the bug**
Typescript is based on [structural typing](https://en.wikipedia.org/wiki/Structural_type_system). This allows to use the following pattern:
```tsx
type Big = { foo: string, bar: string}
type Small = { foo: string } // or Pick<Big, 'foo'>

const big: Big = { foo: "foo", bar: "bar"}
const small: Small = big // <-- no error, since "big" has all the properties that are required by Small
```

I find this pattern very useful when working with react components - have them only require the data which they absolutely need (to have a clear interface, and to make tests easier), but if someone passes data of the bigger type to the component, that's no issue - neither for the component, nor typescript.

Prior to react-hook-form, this approach worked well for the `UseFormReturn` type (it was renamed in v7, but the principle remained the same - it's the data returned by `useForm`): 

```typescript
type FormValues = { foo: string, bar: string}
type FormValuesSubset = Pick<FormValues, 'foo'> // or { foo: string}

const methods: UseFormReturn<FormValues> = useForm<FormValues>()
const smallMethods: UseFormReturn<FormValuesSubset> = methods // <----- no error in v6, error in v7
```

However, with `react-hook-form` v7, I am now getting a type error:
```
Type 'UseFormReturn<FormValues>' is not assignable to type 'UseFormReturn<Pick<FormValues, "foo">>'.
  Types of property 'getValues' are incompatible.
    Type 'UseFormGetValues<FormValues>' is not assignable to type 'UseFormGetValues<Pick<FormValues, "foo">>'.
      Type 'Pick<FormValues, "foo">' is not assignable to type 'FormValues'.ts(2322)
```

**To Reproduce**
Compile the code above, you will get a typescript error

**Codesandbox link**
minimal example: https://codesandbox.io/s/serene-https-zqmh4
real world example: https://codesandbox.io/s/react-hook-form-methods-type-error-kruqg

**Expected behavior**
I would expect `methods` can be assigned to `smallMethods` without type errors. My reasoning behind this is that if the form state includes more fields then the component expects this is no problem. And due to the structural typing system of typescript, this should also be possible to implement. 

**Additional context**
Here I describe why the pattern above is useful and why have used it in the past. A concrete example is in the "real world example" codesandbox.

In the past, I have followed the following pattern
- Have a form, which has a typed FormState
[...]

PR #6807, opened 2021-10-14
Enable sharing form sections between forms in typescript

SimonSelg:
Hi, 

thanks for your work on this library. We are enjoying it a lot, especially the good typescript support and the API design.

This PR fixes an [issue](https://github.com/react-hook-form/react-hook-form/issues/5038) which prevented sharing parts of a form between forms when using typescript. I created an issue for this in the past: https://github.com/react-hook-form/react-hook-form/issues/5038. However, back then neither I nor the maintainers were able to tackle this, and the issue was closed as design limitation (good choice). 

Since my team now has a need to share parts of a form between forms, I took a second take at the issue, and managed to debug it and to come up with a proposed solution.

This PR is meant to gather feedback on whether you'd be willing to accept it. If yes, I'd properly document it, and add examples to the documentation of the library.

## Problem Description

Intuitively, it should be possible to create components which contain form fields, and use this component in different forms. 
I have created a codesandbox which demonstrates this: https://codesandbox.io/s/hook-form-subform-1qfwg. 

One can do this by passing the `UseFormReturn` value as a prop to this component, and then using it inside the component to register the fields. However, when using typescript, and therefore defining which fields are available in the `UseFormReturn` value, we run into a problem here:

```typescript
type FormValuesBig = { foo: string, bar: string }
type FormValuesSmall = { foo: string } // those fields should be inside a shared component

const formMethods = useForm<FormValuesBig>()
// now, pass this value as a prop to a component, which knows only abouts the fields defined in FormValuesSmall
const formMethodsSmall: UseFormReturn<FormValuesSmall> = formMethods // <-- typescript error! not possible to assign
```

Hence, it was not possible to make the shared form component only aware of the fields which it actually used, hence preventing the creation of such a component. 

## Analysis

The root cause was that the `UseFormReturn` type includes two functions, which are not compatible (where compatible means that one is a subtype of the other) when initializing `UseFormReturn` as shown above. One is the `watch` method, the other is part of the `resolver` option, which is contained in the `_options` key of the `UseFormReturn` value.

This is due to the way how typescript (correctly) checks whether a function is a subtype of another function. It does this by comparing the parameters *contravariantly*, and the return type *covariantly*. In a nutshell for a function to be a subtype of another function, the parameters have to be supertypes of the original paramters and the return type has to be a subtype of the original return type. For background, see these two links:
- https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-6.html#strict-function-types
- https://www.stephanboyer.com/post/132/what-are-covariance-and-contravariance

This was not the case for the two functions: They included the `FormField` as a parameter - and  `FormValuesSmall` is certainly not a supertype of `FormValuesBig` (it's the other way around).

### Proposed Solution

I fixed the problem by relaxing the interface of the `Controller` and the `FieldArray` components such that they don't require those functions anymore (they don't use them anyways), and by introducing a third template parameter `TExcludeSubtypeIncompatibleData` on the `UseFormReturn` type which excludes those functions:

```typescript
const formMethods = useForm<FormValuesBig>()
const formMethodsSmall: UseFormReturn<FormValuesSmall, object, true> = formMethods // <-- works now
```

This is a non-breaking change, and does not change existing behavior. There are other ways of getting the same result - for example by introducing a new type instead of adding a new template parameter - I'm happy to discuss them with you.

Thank you for taking the time to look at this PR/issue. You may use [my codesandbox](https://codesandbox.io/s/hook-form-subform-1qfwg) to test my proposed fix - you will probably have to download it, so you can use my PR as a base.

### DIFF
--- a/src/__tests__/type.test.tsx
+++ b/src/__tests__/type.test.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 
-import { Path } from '../types';
+import { Path, UseFormReturn } from '../types';
 import { useFieldArray } from '../useFieldArray';
 import { useForm } from '../useForm';
 
@@ -23,6 +23,28 @@ test('should not throw type error with path name', () => {
   ];
 
   test;
+});
+
+test('all APIs except watch are compatible with a superset of FormFields', () => {
+  type FormFieldsSmall = {
+    gender: 'female' | 'male';
+    firstName: string;
+    lastName: string;
+  };
+
+  type FormFieldsBig = {
+    foo: string;
+    bar: 'test1' | 'test2';
+    test2: boolean;
+  } & FormFieldsSmall;
+
+  function App() {
+    const form = useForm<FormFieldsBig>();
+    const test: Omit<UseFormReturn<FormFieldsSmall>, 'watch'> = form;
+    test;
+    return null;
+  }
+  App;
 });
 
 test('should not throw type error with optional array fields', () => {
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -325,7 +325,7 @@ export type Control<
     action: boolean;
     watch: boolean;
   };
-  _options: UseFormProps<TFieldValues, TContext>;
+  _options: Pick<UseFormProps<TFieldValues, TContext>, 'shouldUnregister'>;
   _getDirty: GetIsDirty;
   _formState: FormState<TFieldValues>;
   _updateValid: Noop;


SimonSelg:
Hey both of you, 

thank you for your very fast replies. I do understand your concerns, and they are valid. 

> you can handle the Provider/Context hook dependency concern in a simple unit test (or localhost validation).

Yes. However, I like to make dependencies explicit / visible in the interfaces. However, this is the personal style of me and my team; not everyone prefers that. It's a tradeoff. 

>  on the concern of the type checking: it's an valid issue for sure, but there's a tradeoff of guaranteed contravariance/covariance down the stack vs extra testing scenario complexity that would need to be accounted for here.

I agree. Perhaps we can agree on a simplified version of this PR, which does not suffer from this issue. I will work on a proposal.

> The new flag in this PR seems like something that could just cause confusion over time - if we could work out how to allow the covariance/contravariance through built-in TS conditional or utility types or one of RHF's many existing ones, that would be more palatable and may allow us to eliminate the need for this conditional 'unknown' fallout.

Yes, I agree. The flag as part of the public interface will probably cause confusion. - this was just my first draft - a POC. 

I will rework the PR towards a more simplified version which would still allow people to share form sections between forms (with type checking), while not making the library interface more complicated and adding additional burden for the maintenance. Maybe we can find a way to make my requirements possible while not making the libraries API more complicated. 

Thank you both for your valuable feedback.

barrymay (dismissed):

bluebill1049 (dismissed):

lundn:
I'm happily awaiting this addition! 👏

bluebill1049:
thanks @SimonSelg for the test case, looks like the test failed.

bluebill1049 (approved):

sergiogirado:
Hey @SimonSelg !

It have been a while but I need to ask: I see this was reverted, what is the current aprpoach to deal with form sub sections?

PR #6810, opened 2021-10-14
Allow Jest VSCode extension to run successfully on startup

barrymay:
## Changes

- Latest Jest VSCode extension has great support for the new Testing sidebar - easy detection without having to run the full test suite
- Without this change, the extension simply fails as it hasn't been configured to use the proper jest.config file

Addresses https://github.com/react-hook-form/react-hook-form/issues/6811

To test (which I ran to verify): 
- Install the Jest VSCode extension
- Restart VSCode, the suite will run and pass successfully
- Force one of the tests to fail: the VSCode status bar will show the problem
- Fix the issue: VSCode status bar will show success again

## Checklist:

- [n/a] I have added tests that prove my fix is effective or that my feature works (this is dependent on a VSCode extension)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+  "configurations": [
+    {
+      "type": "node",
+      "name": "vscode-jest-tests",
+      "request": "launch",
+      "console": "integratedTerminal",
+      "internalConsoleOptions": "neverOpen",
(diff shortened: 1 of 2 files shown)


barrymay:
Moving this to draft - (these changes don't cause an issue, but after testing the extension, I feel it may cause false positives so may be more trouble than its worth ;)

PR #6817, opened 2021-10-15
Return value from onValid in handleSubmit

petercpwong:
Returning the value from `onValid` in `handleSubmit` allows higher order functions to easily work with data returned from a successful submit without passing in extra callbacks or having to save the state with additional hooks (ie. `setState`). This will allow us to pass the callback created from `handleSubmit` back to a parent component or as a callback in another hook so they can inspect the data from a successful submit.

Example use case:

- passing in data of a newly created resource to an `onSuccess` callback in a modal

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1021,7 +1021,7 @@ export function createFormControl<
             errors: {} as FieldErrors<TFieldValues>,
             isSubmitting: true,
           });
-          await onValid(fieldValues, e);
+          return await onValid(fieldValues, e);
         } else {
           onInvalid && (await onInvalid(_formState.errors, e));
           _options.shouldFocusError &&


PopJoestar:
This is really useful !!!

Gazpa:
We had to do a very complex solution to make a form reusable just because this is not in place, is there any chance we can revisit this and add it? Seems like the logic behaviour to have in the library.

boomgetchopped:
This is a very important feature for us.

ipanasenko:
@bluebill1049 would so great to have this merged. The amount of times I need updated value after submit, so that I can do something with it is soo high, workarounds are very annoying. 

Thank you, hope you will be able to take a look into it and merge this PR. This will make react-hook-form even more easy to work with.

Let us know if something needs to be done with this PR in order for it to be merged, I'm happy to help

Issue #6505, opened 2021-09-09
[Bug]: useFieldArray returns never[] if path to name contains optional field 


darylf:
### Version Number

7.4.1

### Codesandbox/Expo snack

https://codesandbox.io/s/vibrant-hermann-2mu3l?file=/src/App.tsx

### Steps to reproduce

 If form includes optional array field, useFieldArray will return with type never[] for fields

This has been documented in two seperate discussions:
* https://github.com/react-hook-form/react-hook-form/discussions/5906
* https://github.com/react-hook-form/react-hook-form/discussions/5517

Changing the object definition to not be optional isn't always feasible. In my case, the objects are from generated code

### Expected behaviour

Fields should be an empty array of the appropriate type 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

darylf:
I'm using [graphql-codeg-enerator](https://github.com/dotansimha/graphql-code-generator) to create my objects, and it looks like even without the optional field, react-hook-form returns `nothing[]` for the fields object. I assume this is caused by the union type of `Array<T> | T`

I'm not sure if this would bloat the scope of any intended fix to the initial report, but it'd be cool if useFieldArray could handle a union type like this as well, by returning an array of the appropriate type.

---
**Generated code:**

    export type UpdateCreditsMutationVariables = Types.Exact<{
      bookId: Types.Scalars["ID"];
      input: Array<Types.CreditInput> | Types.CreditInput;
    }>;

---
**React-Hook-form:**

    const {
      control,
        formState: { errors, isSubmitting },
        handleSubmit,
        register,
      } = useForm<UpdateCreditsMutationVariables>();

    const { fields } = useFieldArray({
        control, 
        name: "input", 
      });

Lemaro86:
Try to use "@hookform/resolvers": "1.3.7", and react-hook-form@6.15.8
And for name use type like this: 
`control,
  name: `inputs.${nestIndex}.name` as 'name'
`

But you should update your type or interface for inputs array, like this:
`
interface InnerType { creditInput: string, debt: number  } 
`
and parent interface:
`
name: Array<InnerType>
`

And if you trust docs add in the input cast:
`... as const`

But if I using last version of RHF I have so much problem with resolver and yup validation. I'm not recommend to use last.

bluebill1049:
> But if I using last version of RHF I have so much problem with resolver and yup validation. I'm not recommend to use last.

[...]

PR #6826, opened 2021-10-16
✨ Add TypeScript support for optional array fields

LinusU:
Fixes https://github.com/react-hook-form/react-hook-form/discussions/5301

This adds TypeScript support for form interfaces that has the entire array as an optional/nullable field. Example:

```ts
import { useFieldArray, useForm } from 'react-hook-form'

interface FormData {
  name: string
  things?: Array<{ name: string }>
}

export default function App () {
  const { control, register } = useForm<FormData>({ defaultValues: { name: 'test' } })
  const { fields, append } = useFieldArray({ control, name: 'things' })

  return (
    <div className="App">
      <h1>useFieldArray demo</h1>
      <input {...register('name')} />
      <br />
      <button onClick={() => append({ name: '' })}>Add</button>
      {fields.map((field, index) => (
        <div key={field.id}>
          <input
            {...register(`things.${index}.name`)}
            defaultValue={field.name}
            placeholder={`Thing ${index + 1}`}
          />
        </div>
      ))}
    </div>
  )
}
```

Before this change the `field` parameter of the `map` function would receive `field` as `never` instead of `{ name: string }` because the `things` array was optional. This change fixes that.

This is my first contribution to this library, so I'm not familiar with the setup. If there is anything else you want me to do, like adding tests, I'd be happy to fix it if you point me in the right direction!

### DIFF
--- a/src/types/fieldArray.ts
+++ b/src/types/fieldArray.ts
@@ -24,9 +24,10 @@ export type FieldArrayWithId<
 export type FieldArray<
   TFieldValues extends FieldValues = FieldValues,
   TFieldArrayName extends FieldArrayPath<TFieldValues> = FieldArrayPath<TFieldValues>,
-> = FieldArrayPathValue<TFieldValues, TFieldArrayName> extends ReadonlyArray<
-  infer U
->
+> = FieldArrayPathValue<TFieldValues, TFieldArrayName> extends
+  | ReadonlyArray<infer U>
+  | null
+  | undefined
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks a lot for the contribution! 👍  I will follow up with a test case for this usage as well.

bluebill1049:
thanks this PR fix this issue as well: #6505 one stone two birds 🥳

PR #6843, opened 2021-10-19
change subject from class to function base to improve readability and reduce size

kotarella1110:


bluebill1049:
Thanks for the PR @kotarella1110 I love the idea and saving bytes! Could we start with matching the exact to start with before we start trimming the function? In fact, I feel my test cases are pretty weak. So let's convert all the `class` to `function` first?

kotarella1110:
@bluebill1049 OK! Change all classes to functions and fix them to be exactly the same as the existing implementation 👍🏻

bluebill1049 (approved):
Awesome stuff! - package size!

PR #6891, opened 2021-10-26
7.19.0 Feature/reset field

bluebill1049:
## `resetField`

RFC: https://github.com/react-hook-form/react-hook-form/discussions/6884

- [x] unit/integration tests
- [x] documentation update
- [x] unit test case for `setValue` with StrictMode
- [x] release beta version

doc: https://react-hook-form-website-git-7190-bluebill1049.vercel.app/api/useform/resetField/

```tsx
const { resetField } = useForm();

resetField('test')

resetField('test', {
  keepError: true,
  keepDirty: true,
  keepTouched: true,
  defaultValue: 'test1',
})
```

- improve `useController` by integrating with `useWatch`
- package size-reduction
- internally removed `skipEarlySubscription`

### DIFF
--- a/app/src/setValueStrictMode.tsx
+++ b/app/src/setValueStrictMode.tsx
@@ -1,0 +1,52 @@
+import React, { useEffect } from 'react';
+import { useForm, Controller } from 'react-hook-form';
+
+const SetValueAsyncStrictMode = () => {
+  const state = React.useRef(new Set());
+  const [, update] = React.useState({});
+  const { register, setValue, control } = useForm<{
+    firstName: string;
+  }>();
+
+  useEffect(() => {
+    setTimeout(() => {
+      setValue('firstName', 'A');
+    }, 10);
+
+    setTimeout(() => {
+      setValue('firstName', 'B', { shouldDirty: true });
+    }, 20);
+
+    setTimeout(() => {
+      setValue('firstName', 'C', { shouldTouch: true });
+    }, 30);
+
+    setTimeout(() => {
+      setValue('firstName', 'D', { shouldValidate: true });
+    }, 40);
+  }, [register, setValue]);
+
+  return (
+    <React.StrictMode>
+      <form>
+        <Controller
+          defaultValue={'test'}
+          control={control}
+          render={({ field }) => {
+            state.current.add(field.value);
+            return <input id={'input'} {...field} />;
+          }}
+          name={'firstName'}
+        />
+
+        <button id="submit" type={'button'} onClick={() => update({})}>
+          Submit
+        </button>
+
+        <p id="result">{JSON.stringify([...state.current])}</p>
+      </form>
+    </React.StrictMode>
+  );
+};
+
+export default SetValueAsyncStrictMode;
--- a/src/__tests__/useForm/reset.test.tsx
+++ b/src/__tests__/useForm/reset.test.tsx
@@ -535,7 +535,7 @@ describe('reset', () => {
         defaultValues: { test: 'test', test1: 'test1' },
       });
       const resetData = () => {
-        reset({}, { keepDefaultValues: true });
+        reset(undefined, { keepDefaultValues: true });
       };
 
       return (
--- a/src/__tests__/utils/createSubject.test.ts
+++ b/src/__tests__/utils/createSubject.test.ts
@@ -13,6 +13,8 @@ describe('createSubject', () => {
       next,
     });
 
+    expect(subject.observers.length).toBe(2);
+
     subject.next(2);
 
     expect(next).toBeCalledTimes(2);
@@ -21,19 +23,46 @@ describe('createSubject', () => {
 
   it('should unsubscribe observers', () => {
     const subject = createSubject();
+    const next1 = jest.fn();
+    const next2 = jest.fn();
+
+    const subscription = subject.subscribe({
+      next: next1,
+    });
+
+    subject.subscribe({
+      next: next2,
+    });
+
+    expect(subject.observers.length).toBe(2);
+
+    subscription.unsubscribe();
+
+    expect(subject.observers.length).toBe(1);
+
+    subject.next(2);
+
+    expect(next1).not.toBeCalled();
+    expect(next2).toBeCalledWith(2);
+  });
+
+  it('should unsubscribe all observers', () => {
+    const subject = createSubject();
     const next = jest.fn();
 
-    const tearDown = subject.subscribe({
+    subject.subscribe({
       next,
     });
 
-    const tearDown1 = subject.subscribe({
+    subject.subscribe({
       next,
     });
 
-    tearDown.unsubscribe();
+    expect(subject.observers.length).toBe(2);
 
-    tearDown1.unsubscribe();
+    subject.unsubscribe();
+
+    expect(subject.observers.length).toBe(0);
 
     subject.next(2);
     subject.next(2);
--- a/src/logic/generateWatchOutput.ts
+++ b/src/logic/generateWatchOutput.ts
@@ -1,6 +1,6 @@
 import { FieldValues, InternalFieldName, Names } from '../types';
-import convertToArrayPayload from '../utils/convertToArrayPayload';
 import get from '../utils/get';
+import isString from '../utils/isString';
 
 export function generateWatchOutput(
   names: string | string[] | undefined,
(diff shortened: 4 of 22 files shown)


app/src/setValueStrictMode.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const SetValueAsyncStrictMode = () => {
+  const state = React.useRef(new Set());
+  const [, update] = React.useState({});
+  const { register, setValue, control } = useForm<{
+    firstName: string;
+  }>();
+
+  useEffect(() => {
+    setTimeout(() => {
```
barrymay, line 12: Why is this necessary - was this moved from something else?

src/__tests__/utils/createSubject.test.ts
```diff
@@ -13,27 +13,56 @@ describe('createSubject', () => {
...
 
+    expect(subject.observers.length).toBe(2);
+
     subject.next(2);
 
     expect(next).toBeCalledTimes(2);
     expect(next).toBeCalledWith(2);
   });
 
   it('should unsubscribe observers', () => {
```
barrymay, line 24: @kotarella1110 @bluebill1049 this is likely an obvious q, but I assume we're not using rxjs here to keep package size low?

src/logic/generateWatchOutput.ts
```diff
@@ -0,0 +1,28 @@
+import { FieldValues, InternalFieldName, Names } from '../types';
+import get from '../utils/get';
+import isString from '../utils/isString';
+
+export function generateWatchOutput(
+  names: string | string[] | undefined,
+  _names: Names,
```
barrymay, line 7: is this meant to be an "internal" export or potentially used by end-users (for debugging I guess?). If it's for external usage, I might suggest a rename of names or _names to be a bit clearer?

src/useSubscribe.ts
```diff
@@ -1,56 +1,46 @@
...
-      _unsubscribe.current = props.subject.subscribe({
-        next: props.callback,
-      });
-    }
-  };
+const updateSubscriptionProps = <T>({ _subscription, props }: Payload<T>) => {
+  if (props.disabled) {
+    tearDown(_subscription);
+  } else if (!_subscription.current) {
+    _subscription.current = props.subject.subscribe({
```
barrymay, line 29: this is a rxjs/observable code-smell from my experience, so take It for what it is: can you keep the subject management (subscribing) to the creator of the subject as opposed to passing it down in props?  I usually aim to keep my subjects private/local to their creating context  so they can't get passed around and unnecessarily abused.

app/src/setValueStrictMode.tsx
bluebill1049: just for async testing with React strict mode. https://github.com/react-hook-form/react-hook-form/issues/6902

src/__tests__/utils/createSubject.test.ts
bluebill1049: correct 👍

src/logic/generateWatchOutput.ts
bluebill1049: it's only for internal tracking.

src/useFormState.ts
```diff
@@ -28,14 +33,16 @@ function useFormState<TFieldValues extends FieldValues = FieldValues>(
   useSubscribe({
     disabled,
     callback: (formState) =>
-      shouldSubscribeByName(_name.current, formState.name) &&
+      shouldSubscribeByName(
+        _name.current as InternalFieldName,
```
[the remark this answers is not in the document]
bluebill1049: I will try that, was trying to avoid `FieldPath`.

src/useSubscribe.ts
bluebill1049: could you provide an alternative? At the moment, this subscription is isolated at the per-hook instance, it can't be passed around as props, which is to avoid abuse of usage. It is only attached with each hook usage.

barrymay: I'm not sure what you mean by "it can't be passed around as props" since it's in the props object here, but I'm sure I'm missing something.  

Rx is a bit of a complex subject, but when I'm doing an rx implementation, it's better to either expose an **observable** you want to subscribe to publicly, or callbacks that can go back and manage the next() call on the subject.

Here's a fairly recent good best practice article on the topic (these are the generally big ones) - https://dev.to/nyagarcia/rxjs-best-practices-bhb

all that being said... since this is an internal custom implementation of rxjs and not the full rxjs platform itself, we can do it our own way, but if you're following the model, there's some "rules of the road" that can be helpful.

barrymay: I'll put it on my list of todo's to take a deeper look at this and see if it would be helpful to expose a (lightweight) observable interface

barrymay (dismissed):
Nothing blocking from what I can tell

bluebill1049:
Thanks, @barrymay for the review 👍

bluebill1049:
`Version 7.19.0-next.0` published

src/useSubscribe.ts
bluebill1049: Thanks, @barrymay at the moment react hook form is focused on the form, hence why we don't really expose any subscription-based hook, observable and related props. What I mean by props is to refer to something you can subscribe to, in fact, most of the use cases are purely just listener/subscriber which then followed with react state update. 

The only expectation made was to watch call back, in case you didn't wear:

```tsx
React.useEffect(() => {
  const unsubscribe = watch((data) => {
    console.log(data)
  });
  
  return () => unsubscribe();
}, [watch])
```

Let me know your thoughts and more than happy to discuss this further as well. If you have Twitter account, we can connect there and chat: https://twitter.com/bluebill1049

bluebill1049:
> I tried it today (localy bundled) and it works like a charm. Unlike `setValue` it is stable even in useEffect with async setTimeout (`setValue` is probably bound to scope that can change more often).

Thie async `setValue` is resolved as well which is due to react strict mode (double render).

jorisre (dismissed):
LGTM! 👏🏻  for the work

src/useSubscribe.ts
barrymay: Thanks Bill, this is probably just fine. I'll dig into how this works and let you know if I think it can improve in a useful way. Since our usage of rx patterns is light at the moment, I'm not very concerned.  

Happy to connect on Twitter as well regardless 😊 - I sent you a note over there.

bluebill1049:
thanks @jorisre for the review.

src/__tests__/utils/createSubject.test.ts
kotarella1110: yes 👍🏻

src/logic/generateWatchOutput.ts
```diff
@@ -0,0 +1,28 @@
+import { FieldValues, InternalFieldName, Names } from '../types';
+import get from '../utils/get';
+import isString from '../utils/isString';
+
+export function generateWatchOutput(
```
kotarella1110, line 5: 👍🏻

src/useController.ts
```diff
@@ -28,46 +28,37 @@ export function useController<
 ): UseControllerReturn<TFieldValues, TResult, TName> {
   const methods = useFormContext<TFieldValues>();
   const { name, control = methods.control, shouldUnregister } = props;
-  const [value, setInputStateValue] = React.useState(
-    get(
+  const value = useWatch({
```
kotarella1110, line 31: Nice refactoring!

src/__tests__/useForm/reset.test.tsx
```diff
@@ -535,7 +535,7 @@ describe('reset', () => {
         defaultValues: { test: 'test', test1: 'test1' },
       });
       const resetData = () => {
-        reset({}, { keepDefaultValues: true });
+        reset(undefined, { keepDefaultValues: true });
```
kotarella1110, line 538: Does this change make sense?
Is there any change in the behavior when an empty object is passed?

kotarella1110 (dismissed):
LGTM 🎉

src/__tests__/useForm/reset.test.tsx
bluebill1049: This is correct usage.

when the user decided to keep `defaultValues`, then we will not update `defaultValues`. If the user supplied the values which is the first argument, it should either be `undefined` or newly updated values. In this case, when supplied with `{}` which lead to one of the input being `undefined`.

```
reset(undefined, { keepDefaultValues: true }) // reset the entire form with defaultValues and don't update defaultValues

reset({test :'bill'}, { keepDefaultValues: true }) // rest the entire with values and don't update defaultValues
```

PR #6934, opened 2021-11-01
Fix 6933

bluebill1049:


src/types/utils.ts
```diff
@@ -187,8 +187,10 @@ export type FieldPathWithValue<
   TFieldValues extends FieldValues,
   TResult = unknown,
   FieldPaths extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
-> = {
-  [key in FieldPaths]: FieldPathValue<TFieldValues, key> extends TResult
-    ? key
-    : never;
-}[FieldPaths];
+> = TResult extends any
```
barrymay, line 190: in which scenario is this condition not truthy?  I can't fine one

bluebill1049: when expected type been set: 
```tsx
type ExpectedType = { test: string };

const Generic = <FormValues extends FieldValues>({
  name,
  control,
}: {
  name: FieldPathWithValue<FormValues, ExpectedType>;
  control: Control<FormValues>;
}) => {
  const {
    field: { value, ...fieldProps },
  } = useController<FormValues, ExpectedType>({
    name,
    control,
    defaultValue: { test: 'value' },
  });

  return <input type="text" value={value.test} {...fieldProps} />;
};
```

barrymay: hm.. but if you do
```ts
type x = { test: string } extends any ? true : false;
```

or 
```ts
type y = unknown extends any ? true : false;
```

and you hover over x and y, they're both true, so I don't get it...  Do you have a test that both sides of the conditional are doing what you expect?

If you want to test if an object is set vs leaving the default, what about 
```ts
TResult extends Record<string, any>
```
That would handle the expected test.

I'm not saying I know an answer to how to fix this issue, but I'm just not sure it's doing what you think it does.

bluebill1049: oh that's right...

bluebill1049: I did a test with the existing type and the error didn't show up for some reason.

```tsx
type FormValues = {
  firstName: string;
  lastName?: string;
};

type Props = UseControllerProps<FormValues>;

type Name = Props['name'];

const test1: Name = 'firstName'
const test2: Name = 'lastName'

console.log(test1);
console.log(test2);
```

bluebill1049: in addition, when i supplied with the type the error goes away as well:

```
<Controller<FormValues>
  name="lastName"
  defaultValue=""
  control={control}
  render={({ field: { value, onChange } }) => {
    return <input value={value} onChange={onChange} />;
  }}
/>
```

barrymay: So you're saying it's working as expected on the current released version?

src/types/utils.ts
```diff
@@ -188,7 +188,9 @@ export type FieldPathWithValue<
   TResult = unknown,
   FieldPaths extends FieldPath<TFieldValues> = FieldPath<TFieldValues>,
 > = {
-  [key in FieldPaths]: FieldPathValue<TFieldValues, key> extends TResult
+  [key in FieldPaths]: FieldPathValue<TFieldValues, key> extends
+    | TResult
+    | undefined
```
bluebill1049, line 193: this fixes the issue @barrymay

barrymay (dismissed):

kotarella1110 (changes requested):
umm...
This issue caused not only when the field value is optional, but also when it is Union type....
The reason is that `TResult` is inferred from the`defaultValue` type.

```tsx
<Controller
  name="lastName" // Since `TResult` is inferred to be a string, a type error will caused if the field value type is other than `string`.
  defaultValue="" // UseControllerProps's `TResult` is inferred to `string`
  control={control}
  render={({ field: { value, onChange, onBlur } }) => {
    return <input value={value} onChange={onChange} />;
  }}
/>
```

```ts
type FormValues = {
  firstName: string;
  lastName?: string;
};

type A = FieldPathWithValue<FormValues, string>; // "firstName"
```

Therefore, if `defaultValue` is not passed, or if Generics is passed, it will work correctly.

```tsx
// `defaultValue` is not passed
<Controller
  name="lastName" // no type error
  control={control}
  render={({ field: { value, onChange, onBlur } }) => {
    return <input value={value} onChange={onChange} />;
  }}
/>
```

```tsx
// Generics is passed
<Controller<FormValues>
  name="lastName" // no type error
  defaultValue=""
  control={control}
  render={({ field: { value, onChange, onBlur } }) => {
    return <input value={value} onChange={onChange} />;
  }}
/>
```

We need to think of ways to support the Union type.

barrymay:
@kotarella1110 you're absolutely right.  I was focused on the smaller typing fix, but there's a significantly larger issue here...

It's not really just a union problem either - it's any type that's not a string...

Check out this:
https://codesandbox.io/s/infallible-jepsen-dmdgo?file=/src/App.tsx:143-207

I took the original 6933 issue, changed the type of lastName from a string to a number, and the typing fails.  If I revert to 7.17.* it's fine.  Furthermore, only if both are strings does it pass, and defaultValues don't seem to work unless they're strings.  @bluebill1049 maybe we need to revisit the reason for the controller typing change that came in to 7.18?

bluebill1049:
Ok I will revert it today. That's sad but don't want to impact more applications.

PR #6944, opened 2021-11-02
upgrade to ts 4.5.2

bluebill1049:
- update to 4.5.0-dev
- build pass
- revert b83fb02


bluebill1049:
looks like 4.5 is no longer working with ✨ enable sharing form sections between forms in typescript (#6807)

src/__tests__/type.test.tsx
```diff
@@ -27,21 +27,15 @@ test('should not throw type error with path name', () => {
...
     foo: string;
     bar: 'test1' | 'test2';
     test2: boolean;
-  } & FormFieldsSmall;
+  };
 
   function App() {
     const form = useForm<FormFieldsBig>();
-    const test: Omit<UseFormReturn<FormFieldsSmall>, 'watch'> = form;
+    const test: Omit<UseFormReturn<FormFieldsBig>, 'watch'> = form;
```
bluebill1049, line 38: result in type error

bluebill1049:
Made some update let me know your thoughts @barrymay @kotarella1110.

barrymay:
Excellent work Bill well done!  It's great to see that simplify clarifying the passed down types fixed the build and the tests!

src/__tests__/type.test.tsx
barrymay: In this one I would really revert the code and (maybe) comment out the test.  The original code was really what @SimonSelg (understandably) wanted, whereas changing FormFieldsSmall to FormFieldsBig no longer lets the test agree with tits own description.

bluebill1049: yea, we will have to revert it because its not doing what it desired anymore.

barrymay (dismissed):

barrymay:
Acknowledging our need to discuss the internal 'as' changes (which we'll do!) and type clarifications (and the need to investigate tsd to support testing the types), I agree this is good to go for now.   (to be clear, I wouldn't deploy this until we can run fine on 4.5 final, but it's good otherwise)

bluebill1049:
yea @barrymay, I am keeping this feature branch running until 4.5 is released.

PR #6983, opened 2021-11-05
feature: useWatch/useFormState exact prop

### LINKED ISSUE #6977: issue: useController() trigger a render when a similar field is updated
### Describe the bug

When using `useController()` hook, a render is trigger on all fields that starts with "name" prop.
By example, if you have fields name :
 - foo
 - bar
 - fooo

touching field "foo" or "fooo" will trigger a render on both of them.

### Version Number

7.18.1

### Codesandbox/Expo snack

https://codesandbox.io/s/55p3h

### Steps to reproduce

1. Go on the sandbox
2. Open the console
3. Touch first field (named "foo")
4. Notice that both "foo" and "fooo" fields were re rendered.

Optionally, try the same thing with the other fields and notice they're working properly.

### Expected behaviour

`useController` should not trigger renders for other fields with similar name.

Firefox, Chrome

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
close #6977

```tsx
useWatch({
 name: 'test',
 exact: true
})

useFormState({
 name: 'test',
 exact: true
})
```


src/__tests__/useFormState.test.tsx
```diff
@@ -599,4 +599,30 @@ describe('useFormState', () => {
...
+      );
+    };
+
+    render(<App />);
+
+    fireEvent.focus(screen.getByRole('textbox'));
+
+    fireEvent.blur(screen.getByRole('textbox'));
+
+    expect(screen.queryByText('touched')).toBeNull();
```
barrymay, line 626: can you add something so touched becomes visible in the test?

bluebill1049: will do 👍

src/logic/isWatched.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export default (
+  name: InternalFieldName,
+  _names: Names,
+  isBlurEvent?: boolean,
+) =>
+  !isBlurEvent &&
+  (_names.watchAll ||
+    _names.watch.has(name) ||
+    _names.watch.has((name.match(/\w+/) || [])[0]));
```
barrymay, line 11: How is this line tested?

bluebill1049: mostly through integration tests, but yea worth adding a unit test on this one.

bluebill1049: good one, actually found a bug with it, will use `startsWith` instead.

barrymay: If you want more complex, I'm happy to help build the Regex, I have a lot of practice there 😉

bluebill1049: take over, please! 👍 I will submit my breaking test.

barrymay: Cool - I can finish it can you describe but exactly the problem is you're fixing? If you have any other tests that cover the situation, even if they break, that's helpful.  I'll sort it out tomorrow

src/logic/isWatched.ts
```diff
@@ -0,0 +1,13 @@
...
+export default (
+  name: InternalFieldName,
+  _names: Names,
+  isBlurEvent?: boolean,
+) =>
+  !isBlurEvent &&
+  (_names.watchAll ||
+    _names.watch.has(name) ||
+    [..._names.watch].some((watchName) =>
+      /^\.\w+/.test(name.slice(watchName.length)),
```
barrymay, line 12: The reduction is good but we're missing the match of the starting text which is important.  I found a way to bring the existing text into the regex so it matches my meaning so I'll do that, but I can't sign this off. (I'll add the change in a bit)

Alternatively we can add the startsWith back but both likely add the same number of bytes back in.  (Nothing egregious but I think managing having concise code here is a balancing act - fully support the ... and .some move tho!)

barrymay: PR to fix is in 😊

bluebill1049:
all review comments are addressed 👍

barrymay (approved):

PR #6988, opened 2021-11-06
close #6987 avoid shallow clone for object (immutable) field

### LINKED ISSUE #6987: issue: Version 7.19.0 Breaks react-hook-form with react-draft-wysiwyg
### Version Number

7.19.0

### Codesandbox/Expo snack

https://codesandbox.io/s/amazing-glade-19cmo-forked-bimw3

### Steps to reproduce

Click on any of the buttons or on the text area

### Expected behaviour

I've been storing the Draftjs EditorState in field.value and updating it with field.onChange and it worked perfectly as seen [here](https://codesandbox.io/s/amazing-glade-19cmo) (version 7.18.1). The 7.19.0 release it broke this and i'm not sure if this was intented. 

Firefox, Chrome, Safari, Edge

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
ref: https://codesandbox.io/s/amazing-glade-19cmo-forked-zvi50?file=/src/App.tsx

### DIFF
--- a/src/useWatch.ts
+++ b/src/useWatch.ts
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { generateWatchOutput } from './logic/generateWatchOutput';
 import shouldSubscribeByName from './logic/shouldSubscribeByName';
 import isObject from './utils/isObject';
+import isString from './utils/isString';
 import isUndefined from './utils/isUndefined';
 import {
   Control,
@@ -17,6 +18,7 @@ import {
 } from './types';
 import { useFormContext } from './useFormContext';
 import { useSubscribe } from './useSubscribe';
+import { get } from './utils';
 
 export function useWatch<
   TFieldValues extends FieldValues = FieldValues,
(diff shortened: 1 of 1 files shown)


src/useWatch.ts
```diff
@@ -72,7 +74,11 @@ export function useWatch<TFieldValues>(props?: UseWatchProps<TFieldValues>) {
         );
 
         updateValue(
-          isObject(fieldValues)
+          isObject(fieldValues) &&
+            !(
+              isString(_name.current) &&
+              get(control._fields, _name.current as InternalFieldName, {})._f
```
barrymay, line 80: what exactly does InternalFieldName mean?  Is it just any string, or does it have a greater meaning (as in, is it a property that would be  same as keyof FieldValues or something else?  

There are multiple ways to avoid this "as" and make it work (not to mention it isn't needed for the 'get').    If it actually means something and it's not just anyt string, we can strength the testing too (and make TypeScript work well for us)

bluebill1049: 💯 agree. The reason with such type is to avoid internal field path which result in exceed recursive, if u have some time, please take a look at it see if there is better solution for this.

barrymay: Very fair - I've started down the path of helping on this - see https://github.com/react-hook-form/react-hook-form/pull/6992

PR #6992, opened 2021-11-08
Type Cleanup Initial Step - Remove Readonly types (for now)

barrymay:
This PR removes all ReadonlyArray usage (moved to Array) and removes the readonly keyword.

- Start initial steps through removing the 'as' keyword, which was getting caught on the readonl\y nature of the types
- While readonly definitely has a good meaning in TS, I feel we need to have clear meaning and better names to all our types we're applying before using it.
- I want to start being more rigorous about avoiding the 'as' keyword except when absolutely required.  I generally find that it can be somewhat helpful when dealing with a 3rd party library (particularly one with poor types), but when dealing with something that's our own codebase, it's can be an overused hack and should have a comment (next to the code) explaining why we needed it (or else we'll forget)

If we have use cases that NEED a readonly type, let's discuss it, but we have no tests that require it.

### DIFF
--- a/src/types/utils.ts
+++ b/src/types/utils.ts
@@ -68,7 +68,7 @@ export type DeepMap<T, TValue> = IsAny<T> extends true
   ? { [K in keyof T]: DeepMap<NonUndefined<T[K]>, TValue> }
   : TValue;
 
-/** Is object greater than 1 level deep? */
+/** Is object greater than 1 level deep (not counting NestedValues)? */
 export type IsFlatObject<T extends object> = Extract<
   Exclude<T[keyof T], NestedValue | Date | FileList>,
   Array<unknown> | object
@@ -92,7 +92,7 @@ type PathImpl<K extends string | number, V> = V extends Primitive
   ? `${K}`
   : `${K}` | `${K}.${Path<V>}`;
 
-/** Union of all possible object paths for given type, including Arrays (that's what ArrayKey is for)  */
+/** All dot-notation paths for a given type, including Array indexing */
 export type Path<T> = T extends Array<infer V>
   ? IsTuple<T> extends true
     ? {
@@ -103,10 +103,10 @@ export type Path<T> = T extends Array<infer V>
       [K in keyof T]-?: PathImpl<K & string, T[K]>;
     }[keyof T];
 
-/** Path based on RHF FieldValues  */
+/** Path<T> for RHF FieldValues  */
 export type FieldPath<TFieldValues extends FieldValues> = Path<TFieldValues>;
 
-/** Similar to PathImpl, but only allow string literal paths to array sub-elements */
+/** Same concept as PathImpl, but only allow string literal paths to array sub-elements */
 type ArrayPathImpl<K extends string | number, V> = V extends Primitive
   ? never
   : V extends Array<infer U>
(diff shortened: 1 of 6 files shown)


bluebill1049:
Awesome @barrymay This is in a good direction, one of the reasons we had `readonly` is to avoid users mutating the array fields.

bluebill1049 (dismissed):
LGTM

barrymay:
Thanks Bill - let me keep iterating on this one a bit to see if I can keep the concept of the immutable inputs while avoiding the typing problems were seeing

src/types/utils.ts
```diff
@@ -114,20 +129,26 @@ export type ArrayPath<T> = T extends Array<infer V>
...
+      ? // If the property after the dot is a valid child path, recurse into it
+        Rest extends Path<T[K]>
+        ? PathValue<T[K], Rest>
         : never
-      : K extends `${ArrayKey}`
+      : // if the property after the dot is valid Array key (a number) and the child is an array, recurse with the array type in mind
+      K extends `${ArrayKey}`
       ? T extends Array<infer V>
-        ? PathValue<V, R & Path<V>>
+        ? PathValue<V, Rest & Path<V>>
```
barrymay, line 144: @bluebill1049 @kotarella1110 Can you help me understand if passing Rest here is necessary; it doesn't seem to be.   If you got PathValue from another site, I'd be happy to check that out.

bluebill1049: hey @barrymay sorry miss the message, I will check it out. in addition to the `readonly`, I think there is another reason: https://github.com/react-hook-form/react-hook-form/pull/7016

bluebill1049: I think it's safe to remove that.

dborysov:
@barrymay, could you maybe give an example of issues which you're trying to solve by this PR?

To me, it's not really obvious since ReadonlyArray is less specific than Array (I mean that ReadonlyArray contains just a sub-set of methods from Array). And because of that, you can path both Array and ReadonlyArray to the places, where ReadonlyArray is required (but not the other way around).

IMO, it should be the other way around: it's better to use readonly arrays everywhere until there is a strong reason to do otherwise

barrymay:
Hi @dborysov thank you for the question, and I agree with you that the readonly-only arrays and types do make sense in most cases.  

My main purpose in doing this investigation is that more and more I've been seeing more code issues "fixed" with "as" casting and I'm trying to work out ways to reduce it.  The problem with casting this way is that it's an override to the type system and I see these changes are having knock-on effects to other code (more for run-time JS breaks that are typical for users who aren't using TS at all)

So the readonly change was honestly to loosen things a bit so that you didn't need to force a readonly type if it didn't come from one, yet the types were good enough to agree and avoid the casting override need.

The flip side of your argument to me is that if we do that, the end user dev may need to start using readonly arrays (or 'as const') when they interface with RHF, which may require a lot of end-user code change that we generally want to avoid.  (If this was a self-contained app and not a lib I'd agree with going all-readonly a lot more) I'm not sure there's a perfect answer here.

Again I appreciate your thoughts, and any further would be helpful.

dborysov:
@barrymay, making arrays non-readonly would solve some cases, but would also introduce problems in different places.

E.g.

```ts
const fieldsToWatch = ['fieldA', 'fieldB'] as const; // Type of the variable is `readonly ['fieldA', 'fieldB']`

const [fieldA, fieldB] = useWatch<Schema, typeof fieldsToWatch>({name: fieldsToWatch});
```

This example would work properly only if in the definition of `useWatch` we change `TFieldNames extends FieldPath<TFieldValues>[]` to `TFieldNames extends readonly FieldPath<TFieldValues>[]` (which was done in #7016)

Would it maybe make sense to fix `readonly array`-related issues one-by-one?

barrymay:
@dborysov  It's a fair request although fraught with more questions than answers.  The merge of your PR broke the branch I was on, and I know my merge would just undo what you did, so I might start from scratch here.  

Trying to make everyone happy (including the complier) is a particularly challenging proposition in this scenario, as the types have immensely deep impacts.

(@bluebill1049 FYI)

PR #7016, opened 2021-11-10
chore(useWatch): making name a readonly array

dborysov:
see the motivation in #5845


bluebill1049:
thanks for the PR.

bluebill1049:
hey @dborysov this introduced a regression, which i will revert and add test coverage for this. let me know if you have a better solution for this.

dborysov:
Changing line 40 was my mistake, thanks for correcting it!

PR #7017, opened 2021-11-10
Adds types for no supplied props argument

sfrieson:
## Issue 
The `useWatch` hook's documentation has an object with all optional parameters on the `props`, and the code sets a default value for the same, but the type checking does not allow it to be used without a `props` argument even though the function implementation marks the parameter as optional.

image

## Solution

This MR adds a function overload signature without the `props` parameter to avoid this error.


bluebill1049:
thanks for the PR.

PR #7034, opened 2021-11-11
chore: Modify package.json to include es6 imports

karanaditya993:
#7033 

Enable ES modules for Node.js


bluebill1049:
thanks @karanaditya993 for the contribution 👍

PR #7041, opened 2021-11-13
Improve tests around isWatched

barrymay:
@bluebill1049 I wanted to do this with pure regex but I feel this is the clearest way to match the start of the string without getting caught up by regex special char issues.  Feel free to merge this into your branch and commit if you're good w/ it


bluebill1049:
LGTM! thanks @barrymay  I will try to reduce the bytes in my PR.

barrymay:
Cool - at present I was looking for code correctness vs byte correction, but if you can beat the minimizer's work, good luck ;)

PR #7042, opened 2021-11-13
Feature/exact PR startsWith fix

barrymay:
Put back startsWith check (fastest perf wise vs escaped regex) with new test to explain why it's needed


bluebill1049:
Awesome! Make sense, thanks @barrymay

PR #7046, opened 2021-11-13
improve dx on control object when hooks are optional inside compoennt

bluebill1049:
ref: https://codesandbox.io/s/xenodochial-clarke-kxk1z?file=/src/App.js

related: #4033

## rational

right now control object is a hard lock when used with a component, this could be a problem when using a component in which hooks are optional, this will prevent this sort of error and allow developers to integrate with hooks easier.

- [ ] RFC

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -22,3 +22,31 @@ export const INPUT_VALIDATION_RULES = {
   required: 'required',
   validate: 'validate',
 };
+
+const noop = () => {};
+
+const subscribeNoop = {
+  next: noop,
+  subscribe: noop,
+};
+
+export const CONTROL_DEFAULT = {
+  register: noop,
+  _getWatch: noop,
+  _removeUnmounted: noop,
+  _updateFieldArray: noop,
+  _getFieldArray: noop,
+  _subjects: {
+    watch: subscribeNoop,
+    state: subscribeNoop,
+    array: subscribeNoop,
+  },
+  _proxyFormState: {},
+  _formValues: {},
+  _stateFlags: {},
+  _names: {
+    watch: new Set(),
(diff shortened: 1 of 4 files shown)


src/constants.ts
```diff
@@ -22,3 +22,31 @@ export const INPUT_VALIDATION_RULES = {
...
 };
+
+const noop = () => {};
+
+const subscribeNoop = {
+  next: noop,
+  subscribe: noop,
+};
+
+export const CONTROL_DEFAULT = {
```
barrymay, line 33: this only adds my concerns to making control optional.

If you do this, and suddenly you decide to start using the defined control object in some convenient way to useWatch or other hooks, we've now caused everyone who assumed it was "magically" interconnected to lose functionality. 

Do you have an example of where this a significant nuisance to have control passed to the hook?

bluebill1049: validate point. The benefit I can see is that you can embed `useController` optional rather than have a hard dependency without `control` object. eg for components library, now you can include this hook without worrying about having run time error. happy to pause this feature as those concerns are valid.

barrymay: If I'm being honest, I see a lot more downside than upside integration/potential bug wise, and I don't feel passing the control down is a pain point for users.

bluebill1049: 👍 sounds valid to me. closing.

AmitNeuhaus: @bluebill1049 
So just to understand, what is the current way to conditionally attach controller?
if i for example want a field to be visible but to not be registered(when currently "disabled" rule isnt attached to useController yet) , what should i do?
The only way is to create similar component without useController hook and conditionally switch rendering both?

PR #7072, opened 2021-11-18
Update setValue.test.tsx

marcelolima-hotmart:


mestir (approved):

src/__tests__/useForm/setValue.test.tsx
```diff
@@ -25,6 +25,24 @@ describe('setValue', () => {
     result.current.register('test1');
     result.current.setValue('test1', 'data');
   });
+  
+  it('should not setValue for unmounted state with shouldUnregister', () => {
```
barrymay, line 29: Maybe this is a draft, but I don't see where the use of shouldUnregister comes into play?  Is the test description accurate?

PR #7150, opened 2021-11-25
fix: Supporting Number Object comparison in deepEqual

swapnilrohom:


src/utils/deepEqual.ts
```diff
@@ -8,6 +8,10 @@ export default function deepEqual(object1: any, object2: any) {
     return object1 === object2;
   }
 
+  if (!isNaN(object1) || !isNaN(object2)) {
+    return object1.toString() === object2.toString();
```
barrymay, line 12: hi @bluebill1049 - that's creating a looser equals comparison - values of 10 and "10" will match.  Did you intend for that?

barrymay, line 12: something bothers me about this - how does anything that's non null get past this code?  I get that we're avoiding the new Number equality case, but this is going to convert any set of nested objects by string?

I know you want to avoid other dependencies, but have you looked at something like https://www.npmjs.com/package/fast-deep-equal to do this.

bluebill1049: thanks @barrymay i have convert them to number instead: https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/deepEqual.ts#L12

bluebill1049: > I get that we're avoiding the new Number equality case, but this is going to convert any set of nested objects by string?

That's a valid concern!

bluebill1049: I am going to revert this for now.

barrymay: I wish there was an easier way to compare those objects cleanly but I can't find one.  There are so few times I have ever used the Number objects throughout my career, I'd love to know a use case where you need it.

PR #7153, opened 2021-11-26
unify register onChange with useController retrieve value

bluebill1049:
credit to @LucasLemanowicz @getTobiasNielsen


src/logic/createFormControl.ts
```diff
@@ -609,7 +610,9 @@ export function createFormControl<
     if (field) {
       let error;
       let isValid;
-      const fieldValue = target.type ? getFieldValue(field._f) : target.value;
+      const fieldValue = target.type
+        ? getFieldValue(field._f)
+        : getEventValue(event);
```
LucasLemanowicz, line 615: Thanks for starting work on this so quickly!

Is this enough of a change to fix the issue we were chatting about? My concern is specifically with the Mantine RadioGroup use case where the onChange handler's type is `(value: string) => void` instead of `(e: Event) => Promise<void | boolean>`. If the "event" in this function is just a string, we will still get an error:

```typescript
const handleChange: ChangeHandler = async (event) => {
  const target = event.target; // undefined
  let name = target.name; // "Cannot read properties of undefined (reading 'name')"
  const field: Field = get(_fields, name); // won't get here, but even if it did, undefined

  if (field) { // undefined, so false and then the code changes are a NOOP
    // ...
```

bluebill1049: no, not really. However, it gives a workaround for people who are using javascript

```
const { onChange } = register('whatever')
onChange('anyThing')
```

In react hook form, we are supporting both uncontrolled and controlled components.

## uncontrolled

the input value is live inside the input itself which is `getFieldValue`.

## controlled

input value is updated via state update which passing through via `onChange` => `getEventValue`

The different controlled components can send different data set, some radio components will send `string` or `event` alike objects, which is the reason why we have some parsing logic inside the `getEventValue`. Should we attach more parse logic inside, I would avoid it. Personally, i always lean towards web standard input event.

LucasLemanowicz: I think where we would also need to make a change is here to create a synthetic event if it doesn't exist: https://github.com/react-hook-form/react-hook-form/blob/764ce28/src/logic/createFormControl.ts#L897

Something along the lines of:
```typescript
// ...
onChange: (event: any) => {
  return event.target
    ? handleChange(event)
    : handleChange({ target: { name, value: event } });
},
// ...
```

bluebill1049: I don't think we need such, what's the use case and reason for such?

LucasLemanowicz: I think I misunderstood what this PR is about. I was under the impression that it would make `register` and the controller have the same `onChange` type definition which would let me connect my form components using `register` method instead of using a controller.

Ex:
```typescript
<RadioGroup
  label="Option 1 or 2?"
  {...register("options", { required: true })}
>
  <Radio value="1">Option 1</Radio>
  <Radio value="2">Option 2</Radio>
</RadioGroup>
```
rather than
```typescript
<Controller
  name="options"
  control={control}
  rules={{ required: true }}
  render={({ field }) => (
    <RadioGroup {...field} label="Option 1 or 2?">
      <Radio value="1">Option 1</Radio>
      <Radio value="2">Option 2</Radio>
    </RadioGroup>
  )}
/>
```

bluebill1049: for an external controlled component that doesn't forward or expose the input `ref`, we will have to use `controller` or `useController`. even we make `onChange` working, there will be issue with set value and reset the form.

bluebill1049: However, with your example above, this PR will work if you only want to retrieve input value.

https://github.com/react-hook-form/react-hook-form/pull/7153/files#diff-5a85c1643c71f45e050ce584d7c0db2bdf2e56f8a7dae86d74fedf01be2d2cabR615

will return that string value.

LucasLemanowicz: Understood, thanks for explaining!

cwagner22: > However, with your example above, this PR will work if you only want to retrieve input value.
> 
> https://github.com/react-hook-form/react-hook-form/pull/7153/files#diff-5a85c1643c71f45e050ce584d7c0db2bdf2e56f8a7dae86d74fedf01be2d2cabR615
> 
> will return that string value.
> 
> basically this `handleChange('string')` will resolve inside `getEventValue`

@bluebill1049 Unless I'm missing something, it won't work, because as @LucasLemanowicz said, `event.target` will be undefined so `handleChange('string')` will throw an error.
I'm trying to figure out how to make RHF works with my custom inputs that don't return the event in onChange. I wanted to avoid having to use controllers for each of them.

Issue #6121, opened 2021-07-29
[Bug]: Cannot assign to read only property '0' of object '[object Array]'


hugodelnegro:
### Version Number

v7

### Codesandbox

https://codesandbox.io/s/react-hook-form-get-started-forked-w7zev?file=/src/index.js

### Steps to reproduce

Submit form and check console logs

### Expected behaviour

I would expect to run this on my system as in codesandbox.

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell
Unlike in Codesandbox I am unable to change checkbox fields getting:  "Cannot assign to read only property '0' of object '[object Array]'".
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bluebill1049:
<img width="824" alt="Screen Shot 2021-07-29 at 9 22 43 pm" src="https://user-images.githubusercontent.com/10513364/127483673-c7d84861-d8e4-4750-871b-6705fb3b36a7.png">

can't reproduce

hugodelnegro:
I can only run locally and I see it on compilation

bluebill1049:
- could be your dependency if can't be reproduce in codesandbox
- for complex checkbox like you got above, build a controlled component and host your state inside and let controller collect value and validate

ZipBrandon:
I ran across this same issue when I lazy initialized the form after having provided `defaultValues`. 


Given the shape of a `reviewSite` being `{ label: string, active: boolean }` and the checkbox manipulates the active field.

```
  const methods = useForm({
    defaultValues: {
      reviewSites: [],
    },
  });
```

This useEffect with setValue creates an issue when the checkbox field is later manipulated
```
  useEffect(() => {
    if (loadingComplete) {
      methods.setValue(`reviewSites`, loadedData); 
    }
  }, [loadingComplete]);
```
However this useEffect works with the reset.
```
  useEffect(() => {
    if (loadingComplete) {
      methods.reset({reviewSites: loadedData}); 
    }
  }, [loadingComplete]);
```

Andreychikov-Vasiliy:
Have same problem when migrating from 7.12.2 --> 7.13.0-next.0

ZipBrandon:
[...]

PR #7155, opened 2021-11-26
fix(Controller): update fieldValues with copy object when Controller …

Holl9:
The problem is in src/utils/set.ts，when the form data is stored in the immer store，the object: FieldValues is not expandable，

```ts
object[key] = newValue;
```
is wrong, the original object should not be manipulated directly. I'll submit a PR later. Please check and merge it after passing.😄

fix(Controller): update fieldValues with copy object when Controller component rerender，Related [[Bug]: Cannot assign to read only property '0' of object '[object Array]'](https://github.com/react-hook-form/react-hook-form/issues/6121#issuecomment-979698278)

### DIFF
--- a/src/utils/set.ts
+++ b/src/utils/set.ts
@@ -1,5 +1,6 @@
 import { FieldValues } from '../types';
 
+import cloneObject from './cloneObject';
 import isKey from './isKey';
 import isObject from './isObject';
 import stringToPath from './stringToPath';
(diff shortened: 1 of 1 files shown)


bluebill1049:
Thanks for the PR, could you please supply a valid use case that can be included in a test case as well? Ideally not include external dependency as well (such as immer).

PR #7159, opened 2021-11-26
improve deep equal

bluebill1049:


swapnilrohom:
@bluebill1049, while improving deep-equal, please make sure that `isObject` should return false for `Number` Objects. As the current version of the library is not capable of rendering `Number` objects. If you add a check for excluding `Number` objects inside `isObject`, it will start rendering `Number` Objects.

swapnilrohom:
I will try to submit PR for deep Equal

bluebill1049:
thanks, @swapnilrohom we will not support `Number` in this library, as this creates unnecessary complexity to the deep equal. please use the number instead. thanks for the PR.

PR #7200, opened 2021-12-02
Take Chinese document to the repo home page

ThankYouDishwasher:


bluebill1049:
thanks for the PR, you may want to create a dedicated v7 version of the readme (due to the syntax changes). eg: https://github.com/react-hook-form/react-hook-form/blob/master/docs/README.V7.ja-JP.md

Issue #7201, opened 2021-12-02
Use browser native validation messages


enisdenjo:
**Is your feature request related to a problem? Please describe.**

Registering an input with a `required: true` option on a `shouldUseNativeValidation: true` form shows an empty validation tooltip when invalid.

**Describe the solution you'd like**

Registering an input with a `required: true` option on a `shouldUseNativeValidation: true` form shows the browser native message when invalid.

**Describe alternatives you've considered**

You can just put the `required` prop directly on the input (`<input {...register('name')} required />`) and get the browser message, but is this the intended usage that `react-hook-form` would advocate? If so, I'd naturally expect the `required` option to get spread to the input field when using the form with the `shouldUseNativeValidation: true` option.

**Additional context**

It's simply annoying to always have to supply the exact same message on all inputs that are required. Especially annoying when you have internationalization. For example sake, this is how all required inputs would have to look like with [`react-intl`](https://github.com/formatjs/formatjs/tree/main/packages/react-intl):
```diff
+ import { FormattedMessage } from 'react-intl';

...

+ <FormattedMessage id="FIELD_REQUIRED">
+   {([message]) => (
      <input
        {...register('name', {
+         required: message,
        })}
      />
+   )}
+ </FormattedMessage>
```

Whereas, the browser would translate the appropriate message for you depending on the user's browser language settings.

_* I guess the expected behavior should happen for all other browser native validation properties, like `pattern` and `min/max`_

bluebill1049:
similar disucssion here: https://github.com/react-hook-form/react-hook-form/discussions/7193

bluebill1049:
> Whereas, the browser would translate the appropriate message for you depending on the user's browser language settings.

It doesn't translate messages for you, only those default ones.

bluebill1049:
move to the project board.

enisdenjo:
> It doesn't translate messages for you, only those default ones.

Yes, that's what I meant. It translates the built-in default messages to the browsers language.

Anyways, thanks for the quick response and the PR, much appreciated! 😄

PR #7207, opened 2021-12-03
disable JS validation (shouldUseNativeValidation)

bluebill1049:
related: #7201

```tsx
const { register } = useForm()


<input {...register('name', { required: true })} />
```

to 

```tsx
<input name="name" required /> // both client and server render
```


jorisre (dismissed):
LGTM! :)

src/logic/createFormControl.ts
```diff
@@ -890,75 +888,68 @@ export function createFormControl<
         )
       : updateValidAndValue(name, true);
 
-    return isWindowUndefined
-      ? ({ name: name as InternalFieldName } as UseFormRegisterReturn)
```
kotarella1110, line 894: Does this mean that We can use `register` in ReactNative as well?

bluebill1049: If you set to use browser in build validation: `shouldUseNativeValidation`

bluebill1049: > Does this mean that We can use register in ReactNative as well?

You can use register with React Native, if you don't need to programmatically set field value or reset form.

kotarella1110: > You can use register with React Native, if you don't need to programmatically set field value or reset form.

Am I correct in understanding that this change will make that possible?

bluebill1049: no, this is only for web-native validation API.

src/types/validator.ts
```diff
@@ -28,8 +28,8 @@ export type RegisterOptions<
   required: Message | ValidationRule<boolean>;
   min: ValidationRule<number | string>;
   max: ValidationRule<number | string>;
-  maxLength: ValidationRule<number | string>;
-  minLength: ValidationRule<number | string>;
+  maxLength: ValidationRule<number>;
+  minLength: ValidationRule<number>;
```
kotarella1110, line 32: 👍🏻 

```ts
React.InputHTMLAttributes<HTMLInputElement>.maxLength?: number | undefined
React.InputHTMLAttributes<HTMLInputElement>.minLength?: number | undefined
```

kotarella1110 (approved):
LGTM 🎉

src/types/validator.ts
bluebill1049: Better!

bluebill1049: actually it support object as well, so i will leave as it is, feel free to send a PR if you reckon this can be improved.

```tsx
export type ValidationRule<
  TValidationValue extends ValidationValue = ValidationValue,
> = TValidationValue | ValidationValueMessage<TValidationValue>;
```

src/logic/createFormControl.ts
kotarella1110: What I want to know is why it used to return only `name` when the `window` object was `undefined`? 👀

Issue #7251, opened 2021-12-10
Can't bundle CRA with v7.21.1


ddolcimascolo:
### Version Number

7.21.1

### Codesandbox/Expo snack

https://github.com/ddolcimascolo/react-hook-form-issue-7251

### Steps to reproduce

```
npx create-react-app react-hook-form-bundling-issue
cd react-hook-form-bundling-issue
npm i --save -E react-hook-form
```

Then use the lib in its simplest form

```diff
import logo from './logo.svg';
import './App.css';
+
+import { useForm } from 'react-hook-form'

function App() {
+  const { register } = useForm();
+
  return (
    <div className="App">
      <header className="App-header">
        <img src={logo} className="App-logo" alt="logo" />
        <p>
          Edit <code>src/App.js</code> and save to reload.
        </p>
        <a
          className="App-link"
          href="https://reactjs.org"
          target="_blank"
          rel="noopener noreferrer"
        >
          Learn React
        </a>
      </header>
+      <div>
+        <form>
+          <input {...register('myField')} />
+        </form>
+      </div>
    </div>
  );
}

export default App;
```

Build the app

```
npm run build
```

### Expected behaviour

Build succeeds

### Relevant log output

```shell
node_modules/react-hook-form/dist/index.esm.mjs
Can't import the named export 'createContext' from non EcmaScript module (only default export is available)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bluebill1049:
provide more detail and a GitHub repo if you have plz: https://github.com/react-hook-form/react-hook-form/issues/2848

ddolcimascolo:
On my way :)

ddolcimascolo:
Reproduction steps + repo added

bluebill1049:
thanks.

bluebill1049:
publishing 7.21.2 thanks for the reporting

evocateur:
@ddolcimascolo If you could verify #7261 works in your project, that would be very appreciated.

ddolcimascolo:
Ah sorry I didn't post back 😅

This was actually handled without me being involved! Our dependency bot (renovate) did pick up v7.21.2 and updated the pull request, our pipeline went green and got merged :)

So yes, I can confirm v7.21.2 works as expected

Thx guys, keep up the great work

David

bluebill1049:
would you like to confirm @ddolcimascolo 
yarn add react-hook-form@7.21.3-beta.0

[...]

PR #7244, opened 2021-12-08
build(esm): Use explicit .mjs extension for ESM

evocateur:
This ensures that Node always treats the ESM output as ESM, despite the package.json (implicit) type of commonjs. Bundlers that expect ESM, such as Next v12, no longer explode when encountering this module.

Refs #7095
Refs #7088
Refs https://github.com/react-hook-form/resolvers/issues/271
Refs https://github.com/vercel/next.js/issues/30750

PR #7095, opened 2021-11-19
revert module config

bluebill1049:
close #7088

Issue #7088, opened 2021-11-19
issue: react hook form breaks blitz.js/next.js build since 7.20


JohannesKlauss:
### Version Number

7.20.1

### Codesandbox/Expo snack

https://github.com/JohannesKlauss/blitzjs-rhf-build-error

### Steps to reproduce

1. Check out the repo mentioned above
2. Run `yarn install`
3. Run `yarn run build`

### Expected behaviour

The build completes without error.

I guess this is related to this commit: https://github.com/react-hook-form/react-hook-form/commit/461c7db172f34e28c2439f64b23dfd173191a58d

Not quite sure though. I can confirm that is worked without a problem for version `7.19.5`. Let me know if you need any help!

### Relevant log output

```shell
(node:6594) UnhandledPromiseRejectionWarning: Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/johannesklauss/Documents/Github/_experimentation/blitzjs-rhf-build-error/node_modules/react-hook-form/dist/index.cjs.js
require() of ES modules is not supported.
require() of /Users/johannesklauss/Documents/Github/_experimentation/blitzjs-rhf-build-error/node_modules/react-hook-form/dist/index.cjs.js from /Users/johannesklauss/Documents/Github/_experimentation/blitzjs-rhf-build-error/node_modules/@hookform/resolvers/zod/dist/zod.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.cjs.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/johannesklauss/Documents/Github/_experimentation/blitzjs-rhf-build-error/node_modules/react-hook-form/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1080:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
    at require (internal/modules/cjs/helpers.js:88:18)
    at Object.<anonymous> (/Users/johannesklauss/Documents/Github/_experimentation/blitzjs-rhf-build-error/node_modules/@hookform/resolvers/zod/dist/zod.js:1:7)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
[...]

PR #7261, opened 2021-12-10
build(esm): Use explicit .mjs extension for ESM, with CJS interop

evocateur:
Building off of #7244, hopefully addressing CRA issues (#7251) and https://github.com/react-hook-form/resolvers/issues/301.

I believe the root cause for the CRA issue is that RHF uses the namespace import form (`import * as React from 'react'`) when it should be using a default import (`import React from 'react'`). React itself is not ESM, it's just a default export that Babel et al make-believe is a series of named exports, thus the error message folks are encountering in non-ESM tooling.

I was surprised RHF isn't using `esModuleInterop: true` in its tsconfig, and that has been addressed here. The test mock changes (one unused, one needing `__esModule: true` flags) were triggered by this tsconfig change.

Refs #7095
Refs #7088
Refs https://github.com/react-hook-form/resolvers/issues/271
Refs https://github.com/vercel/next.js/issues/30750

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,11 +3,11 @@
   "description": "Performant, flexible and extensible forms library for React Hooks",
   "version": "7.21.2",
   "main": "dist/index.cjs.js",
-  "module": "dist/index.esm.js",
+  "module": "dist/index.esm.mjs",
   "umd:main": "dist/index.umd.js",
   "unpkg": "dist/index.umd.js",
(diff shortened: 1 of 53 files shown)


bluebill1049:
thank you so much @evocateur 👍

bluebill1049:
going to release a beta on this.

PR #7262, opened 2021-12-11
beta: build(esm): Use explicit .mjs extension for ESM, with CJS interop (#7…

bluebill1049:
…261)

* build(esm): Use explicit .mjs extension for ESM

This ensures that Node always treats the ESM output as ESM, despite the package.json (implicit) type of commonjs. Bundlers that expect ESM, such as Next v12, no longer explode when encountering this module.

Refs #7244
Refs #7095
Refs #7088
Refs https://github.com/react-hook-form/resolvers/issues/271
Refs https://github.com/vercel/next.js/issues/30750

* fix(esm-interop): Import non-ESM React as default


bluebill1049:
- [ ] test CRA
- [ ] test nextjs

PR #7305, opened 2021-12-16
add CustomEvent check on getEventValue function

vinioo:
## Description
We use some inputs built with Web Component and they are all in Shadow DOM.
These inputs use some CustomEvents to propagate the input events outside the Shadow DOM, it's exactly the same way as Ionic Framework does.

When using these kind of input is necessary to wrap it with the Controller component, otherwise it won't be possible to catch the Custom Events, so we did this way: 

```
      <Controller
        control={control}
        name="password"
        render={({ field: { onChange, onBlur, value } }) => (
          <Input
            id="password"
            data-testid="password"
            onBdsChange={onChange}
            onBdsOnBlur={onBlur}
            value={value}
            required
          ></Input>
        )}
      />
```
When testing it won't work. To simulate the typing on a shadow dom input, we need to emit the custom event related to the typing, in my case, the event is the onBdsChange, e.g: 

``` 
export const bdsChangeEvent = (value: string) => {
  return new CustomEvent('bdsChange', { detail: { value } });
};
```
usage with react-testing-library:
```
    const input = (await screen.findByTestId('counter-input')) as HTMLBdsInputElement;
    fireEvent(input, bdsChangeEvent('1500'));
```
It works only when the input isn't wrapped on the Controller component.
This solution is also used by Ionic to test their inputs, as we can see on [this repository](https://github.com/ionic-team/ionic-react-test-utils)

The problem is: in CustomEvents we tend to use the [detail property](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent) to pass custom properties and the value from the input itself. This way, react hook form wasn't recognizing the event.detail.value property and wasn't possible to test the inputs and the form.

A quick fix to this would be on every input add this: 
```
      <Controller
        control={control}
        name="password"
        render={({ field: { onChange, onBlur, value } }) => (
          <Input
            id="password"
            data-testid="password"
            **onBdsChange={(ev) => { onChange(ev.detail.value) }}**
            onBdsOnBlur={onBlur}
            value={value}
            required
          ></Input>
        )}
      />
```
But it would be very verbose and this situation can be fixed with a few lines on the library. That's what I did.

I've added a check on the getEventValue file, that verifies if the event received is CustomEvent, and if it is, it will return the event.detail.value property, which in most of cases will be the value of custom inputs.

### DIFF
--- a/src/logic/getEventValue.ts
+++ b/src/logic/getEventValue.ts
@@ -1,11 +1,21 @@
 import isCheckBoxInput from '../utils/isCheckBoxInput';
+import isCustomEvent from '../utils/isCustomEvent';
 import isObject from '../utils/isObject';
 
 type Event = { target: any };
 
-export default (event: unknown) =>
-  isObject(event) && (event as Event).target
-    ? isCheckBoxInput((event as Event).target)
-      ? (event as Event).target.checked
-      : (event as Event).target.value
-    : event;
+export default (event: unknown) => {
+  if (isObject(event)) {
+    if ((event as Event).target) {
+      return isCheckBoxInput((event as Event).target)
+        ? (event as Event).target.checked
+        : (event as Event).target.value;
+    }
+
+    if ((event as CustomEvent).detail && isCustomEvent(event)) {
+      return (event as CustomEvent).detail.value;
+    }
+  }
+
+  return event;
+};
--- a/src/utils/isCustomEvent.ts
+++ b/src/utils/isCustomEvent.ts
@@ -1,0 +1,1 @@
+export default (event: unknown) => event instanceof CustomEvent;
--- a/src/__tests__/logic/getEventValue.test.ts
+++ b/src/__tests__/logic/getEventValue.test.ts
@@ -11,11 +11,14 @@ test('getEventValue should return correct value', () => {
       target: { checked: true, type: 'checkbox', value: 'test' },
     }),
   ).toEqual(true);
-  expect(getEventValue({ target: { value: 'test' }, type: 'test' })).toEqual(
-    'test',
-  );
   expect(getEventValue({ data: 'test' })).toEqual({ data: 'test' });
   expect(getEventValue('test')).toEqual('test');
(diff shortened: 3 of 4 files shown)


arthursoas (approved):

bluebill1049:
Thanks for the PR, please include the custom check at the App level.

vinioo:
> Thanks for the PR, please include the custom check at the App level.

What do you mean by "include at the App level"? On the /app folder?

src/logic/getEventValue.ts
```diff
@@ -1,11 +1,21 @@
...
-    : event;
+export default (event: unknown) => {
+  if (isObject(event)) {
+    if ((event as Event).target) {
+      return isCheckBoxInput((event as Event).target)
+        ? (event as Event).target.checked
+        : (event as Event).target.value;
+    }
+
+    if ((event as CustomEvent).detail && isCustomEvent(event)) {
```
long-lazuli, line 15: you should be able to do:
`if (isCustomEvent(event) && event.detail) {`

src/utils/isCustomEvent.ts
```diff
@@ -0,0 +1 @@
+export default (event: unknown) => event instanceof CustomEvent;
```
long-lazuli, line 1: For clarity, let's type the return of the function as `(event: unknown): event is CustomEvent => event instanceof CustomEvent`

long-lazuli:
What's up with this PR ?
@vinioo , could you rebase this to the base branch, and maybe address the two little things I suggested ?

src/utils/isCustomEvent.ts
long-lazuli:
```suggestion
export default (event: unknown): event is CustomEvent => event instanceof CustomEvent;
```

src/logic/getEventValue.ts
long-lazuli:
```suggestion
    if (isCustomEvent(event) && event.detail) {
```

Issue #7290, opened 2021-12-15
issue: FieldPath slows down TypeScript considerably


felixschorer:
### Version Number

7.20.1

### Codesandbox/Expo snack

https://codesandbox.io/s/charming-browser-9ujqf?file=/src/App.tsx

### Steps to reproduce

1. Open `App.tsx` in the provided repo link.
2. Change the value of the `name` prop on any of the `<Controller />` to an invalid path.
3. Wait for TS to recognize the mistake.
4. Make further changes.

Typescript will come to a crawl:
- Simple changes will take seconds to process.
- IntelliSense becomes unusable.

### Expected behaviour

Changes shouldn't take multiple seconds to process on complex form data.

The issue seems to be the definition of the `FieldPath` type and similar types like `ArrayPath`.
These types eagerly enumerate all paths of an object type.
The resulting union type becomes excessively big for nested objects containing complex objects like `Date`s.

This could be fixed by lazily evaluating the paths. See [`ts-toolbelt`](https://github.com/millsp/ts-toolbelt)'s [`AutoPath`](https://millsp.github.io/ts-toolbelt/modules/function_autopath.html).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bluebill1049:
what's under the hood: https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html and this is a design limitation.

bluebill1049:
> This could be fixed by lazily evaluating the paths. See ts-toolbelt's AutoPath.

thanks for the suggestion with `AutoPath` as well, I will evaluate it see if this can be potential solution.

bluebill1049:
cc @barrymay @kotarella1110

bluebill1049:
That `AutoPath` really is just `get` type, it doesn't replace the functionality with `FieldPath` to produce autocomplete/hint for fields name when register with inputs. I don't think this will resolve the underlining limitation. Correct me if I am wrong here.

felixschorer:
We're currently using `AutoPath` like this.
https://codesandbox.io/s/vigilant-flower-oydoe?file=/src/App.tsx

Basically, `name` is just constrained to `string`.
If `name` isn't a valid path into the object, then every type using that path evaluates to never.

felixschorer:
> That `AutoPath` really is just `get` type, it doesn't replace the functionality with `FieldPath` to produce autocomplete/hint for fields name when register with inputs.

[...]

PR #7311, opened 2021-12-17
🐷  Implement lazy versions of Path and ArrayPath

felixschorer:
Implements lazy versions of `Path` and `ArrayPath` as suggested in #7290

The project does **NOT** make use of these types yet. They just exist for now. Integrating them into the API is subject of another PR.

- [x] Extract path related types into their own submodule (`types/utils.ts` -> `types/path/common.ts` / `types/path/eager.ts`)
- [x] Implement primitive types to manipulate paths
- [x] Implement lazy alternatives to eager types (`Path` -> `LazyPath`, `ArrayPath` -> `LazyArrayPath`)
- [x] Type tests
  - [x] Introduce [`tsd`](https://github.com/SamVerschueren/tsd) for type testing
  - [x] Integrate [`tsd`](https://github.com/SamVerschueren/tsd) into the build pipeline
  - [x] Test that the types behave as expected on unions
  - [x] Test that the types behave as expected on optional/nullable fields

 ## `LazyPath` and `LazyArrayPath`
Both types validate paths (e.g. `'foo.bar.42.baz'`) against a type and return a union of suggestions which may include the given path if it is valid. They're supposed to be used on function arguments.

**before**
```ts
function foo<T, P extends Path<T>>(value: T, path: P) {}
```

**after**
```ts
function foo<T, P extends string>(value: T, path: LazyPath<T, P>) {}
```
By using those types on function arguments, the path will be validated against `T`. I.e. TS will report an error when the path does not exist within `T`. Also, only the immediate successor/child paths will be suggested to the user when typing. Previously `Path` would report all possible paths, resulting in bad DX for large types.

## Typescript below version 4.5
 - The new types are implemented tail recursively which allows TS to do some optimizations on them. I've tested the types under TS 4.3 and TS 4.4 to make sure that these optimizations aren't needed for the types to work properly. There is also no noticeable difference in performance.

 - Index signatures using template literal types aren't supported in TS 4.3. The test cases explicitly testing that feature fail under TS 4.3.

## IntelliSense
WebStorm and VS Code will show the path suggestions in a dropdown while typing.

Should a path be rejected, neither (WS or VSC) will show the expanded union of suggested paths when hovering over the error.
The error message will some times include *"Did you mean ...?"*, but not always.

**WebStorm**

**VS Code**

  
## Import changes
Since a few types have been moved to different files, there may be breaking changes for users which imported these types by their full path.

```ts
// ❌ breaks
import { Path } from 'react-hook-form/dist/types/utils'

// ✔ fine
import { Path } from 'react-hook-form'
```

This could be solved with a re-export in `types/utils.ts`.
```ts
export type {
  ArrayPath,
  FieldArrayPath,
  FieldArrayPathValue,
  FieldPath,
  FieldPathValue,
  FieldPathValues,
  Path,
  PathValue,
} from './path/eager'
```


barrymay:
Hi @felixschorer this is great stuff, thank you! Actually the more comments the merrier on these types, so please from the top of each add a jsdoc explanation (I would like to do them for all of them).  

Separately question: did you see this in another codebase? They look different from the ones we saw in ts-toolbelt..

bluebill1049:
Honestly, this is really impressive! ✨

> I am a fairly new to react-hook-form. Is there anything in particular that I should type test?

💯 Would be great to have some test coverage for those types.  

> Seeing that there aren't any comments outside of test cases, how should I document those new types?
I'd be happy to write docstrings in case something isn't clear enough.

Yes, please add comments. That would be really helpful.

src/types/path/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './eager';
+export * from './lazy';
```
bluebill1049, line 2: nice separation for the path look up.

bluebill1049: Question: when will we need an eager path look up? since lazy path looks up probably is going to be the better option. my initial thought was to replace eager with lazy.

felixschorer: 1. Backward comparability. Those types are currently part of the public API, so we can't remove them.
2. The lazy types can only be used for function arguments. Anywhere else we'll still have to use the eager types. E.g. default type parameters. 
```ts
declare function example<
  T, 
  P extends PathString = Path<T>
>(obj: T, path: LazyPath<T, P>>): void
```

felixschorer:
> Separately question: did you see this in another codebase? They look different to me from the ones we saw in ts-toolbelt..

I wrote those types from scratch.
1. I didn't want to rip any code from ts-toolbelt
2. The requirements are slightly different
3. It's easier for me to write it from scratch than trying to understand the implementation of AutoPath
  - I'm very familiar with TS, this is fun for me
  - The concept of validating a type instead of constraining it is not new to me
  - Knowing that and having used AutoPath before, it is immediately obvious how it was implemented
4. TypeScript supports tail call optimization since 4.5

Are you implying that I stole that code from somewhere?

src/types/path/index.ts
bluebill1049: > Backward compatability. Those types are currently part of the public API, so we can't remove them.

sounds good to me, I did have some concern over people using `FieldPath` as well.

bluebill1049:
> Are you implying that I stole that code from somewhere?

@felixschorer I am pretty sure barry didn't mean that. I think he was just surprised with all the types that you have written so far.

felixschorer:
> > Are you implying that I stole that code from somewhere?
> 
> @felixschorer I am pretty sure barry didn't mean that. I think he was just surprised with all the types that you have written so far.

No offense taken. I am not a native speaker, so I may have interpreted it differently than how it was meant. It's also the middle of the night 😅

bluebill1049:
Indeed I am appreciating (so does the community)  the contribution and work that you have been putting this. It's going to be a huge improvement!

barrymay:
Are you implying that I stole that code from somewhere?
- Not at all!  Just wanted to know, since it is indeed quite a bit of new code, so if it was used elsewhere it would actually give a bit of comfort that it was already "battle tested".  
- Since these are new type defs it would be helpful to see some behavior test cases with them

felixschorer:
@bluebill1049 Would it be possible to add a type assertion library as a dev dependency? Testing for type equality is sadly not that trivial due to a lot of edge cases. I'd rather use a library for that instead of rolling my own.

[`conditional-type-checks`](https://www.npmjs.com/package/conditional-type-checks) worked very well for me so far. I am open for suggestions though.

Also, which version of `yarn` are you using? My installed version tries to migrate the `yarn.lock` file to a newer version everytime I run it.

barrymay:
@felixschorer have you checked out https://github.com/SamVerschueren/tsd?  I've heard quite a number of refs to it and I was hoping to consider that route (it was also mentioned by a Microsoft contact we heard from a few days ago)

It's somewhat of an extra test step but I free feel well worth considering

barrymay:
I'd also be interested in checking out https://www.npmjs.com/package/conditional-type-checks as well and see how it compares

felixschorer:
@barrymay I have tsd on my radar. I actually mentioned it in my comment but then removed it again in the edit. 

I haven't used it yet, but from what I've seen you need to run a script to execute the type tests instead of getting compile time errors. It also seems to be awkward that one has to create a value first. 

Honestly, I didn't want to mess with build pipeline 😅

barrymay:
Ha - testing ts types is awkward to start with, so I totally get it!  I'd be thrilled if you took Tsd for a spin and see what you thought, it's hard to overlook it, just haven't personally had time to integrate yet.  

Considering the (awesome) new types your introducing here, I think it's a perfect time to try it.

bluebill1049:
thanks, @barrymay's input. I just quickly checked both they are both actively maintained, and syntax wise both are quite similar too. Consider we don't have a type assertion library at the moment. I am keen to take either onboard. Another great add-on we always talked about but never implemented. Really exciting stuff.

src/types/path/lazy.ts
```diff
@@ -0,0 +1,244 @@
...
+  TPathString extends PathString,
+> = CompletePathString<T, TPathString, ReadonlyArray<any>>;
+
+/**
+ * See {@link LazyArrayPath}
+ */
+export type LazyFieldArrayPath<
+  TFieldValues extends FieldValues,
+  TPathString extends PathString,
+> = LazyArrayPath<TFieldValues, TPathString>;
```
bluebill1049, line 244: thanks for adding those comments 👍

bluebill1049:
`yarn version: 1.22.15`

felixschorer:
I've experimented with `tsd` a bit, however, it doesn't seem to be the right tool for the job.
Before running `tsd` one has to first generate the typings, i.e. `dist/index.d.ts`. Otherwise `tsd` will fail. 
That's also mentioned in this long standing issue: https://github.com/SamVerschueren/tsd/issues/32

I did some more digging and found an alternative to `conditional-type-checks` which is maintained by the same organization as `ts-node`: https://github.com/TypeStrong/ts-expect

Using a type assertion library also means that type test failures are part of the regular `jest` test report instead of some separate report.

src/__tests__/types/__fixtures__/expectTypesEqual.ts
```diff
@@ -0,0 +1,7 @@
+import { expectType, TypeEqual } from 'ts-expect';
+
+export function expectTypesEqual<Expected, Actual>(
```
barrymay, line 3: I assume this Is strictly an abstraction so we can change testing libraries more easily, as opposed to depending on expectType directly.  (I support this, just wanted to confirm)

felixschorer: Typing `expectType<TypeEqual<Expected, Actual>>(true)` gets old quickly. That was my main motivation. But I suppose that solves that issue as well. 😂

barrymay:
Thanks once again @felixschorer - that's a very interesting point on tsd.  I'm perfectly fine with ts-expect, it's certainly simple enough (could probably just take those limited expect types and make them local tbh, since they're fairly straightforward to understand).  

(I noticed that ts-toolbelt that it was largely doing the same checks at its core)

felixschorer:
@barrymay I too was thinking about just copying that `TypeEqual` type. Do you then also write tests to ensure that it does is correct. Who brings the mailman his mail? 😂

barrymay:
As I said, I'm perfectly fine leaving the reference to ts-expect in there.  We should test (in a deployed beta probably) that it doesn't make it a dependency library via npm/yarn but I don't expect that it would.

felixschorer:
@barrymay I think I've found the original source for that type. `ts-toolbelt` uses the same type for its [`Equals` type](https://github.com/millsp/ts-toolbelt/blob/master/sources/Any/Equals.ts).

`ts-toolbelt` -> StackOverflow -> https://github.com/Microsoft/TypeScript/issues/27024#issuecomment-421529650

It seems to [fail for intersection types](https://www.typescriptlang.org/play?#code/C4TwDgpgBAogjgVwIYBsDOAeAGgGigTQD4oBeAKCkqgAoMAVQ6gSlOLqggA9gIA7AEzRQsUAPxQAjFABcUAEwsuPAWgpVaDZqyjslfQQTGSZ8luOAAnBNFkAzVGggBuMmQD0bqPfQQyoSDoQaMCksIgOGADeUJyyUgC+UABkUNEgsnJQ8XjRsZJ46fJZhEA) 😞

barrymay:
Interesting - can you expand the playground with  both types and the pass/fail example just to prove it? (I don't see it clearly by looking at the two codebases)

As an aside, I'm more than happy to discuss this off the PR thread so it doesn't get out of hand, my twitter account is @Barry_A_May

felixschorer:
@barrymay [there you go](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAbwIYGcUFNYBo4EkUCiAHkgMYwC+cAZlBCHAESkQB2AJsDMG0gDYBaGAE8w6AaQAW6UgGsUjAFCKEiuOrgixcAEroUAVz7wAvHkIlyAHgRwiALjgBGOFQBkiOMMcAmVzlsHZxxvOD8KAD41DVQMWCs9Q2MIgAoYKAN0AEpFCmVQSFhEdCIxcgAVUXQcSrECAEcDfipaeiYYFAESspglFWj1LXRdfSNTOFr0Bqa+GztHF3dPUPCA+eCvX1cojThumRhJhNHktIzs3KA)

I've just created an account. I am @felixschorer on Twitter.

barrymay:
never mind - [I got it](https://www.typescriptlang.org/play?#code/C4TwDgpgBAogjgVwIYBsDOAeAGgGigTQD4oBeAKCkqgAoMAVQ6gSlOLqggA9gIA7AEzRQsUAPxQAjFABcUAEwsuPAWgpVaDZqyjslfQQTGSZ8luOAAnBNFkAzVGggBuMmQD0bqPfQQyoSDoQaMCksIgOGADeUJyyUgC+UABkUNEgsnJQ8XjRsZJ46fJZhK4AlgC2YAD2FiGRAJJoMJxIAMbAibYWVeVQAEStVQKlwKVDqAC0-hATrQAWEK0A1mh9rh5QltZ+4NB0QcCZJFCNzW3AUTFxWcmpUIWZ2Xd5EgUZxUA)

Now I'm wondering if the conditional-type-checks Exactness has it's own failings, since that check seems simpler/more straightforward.  I feel it's a covariance/contravariance issue.

barrymay:
Review so far - Lazy types look great so far, and testing working beautifully with tsd!

bluebill1049:
> Thanks once again @felixschorer - that's a very interesting point on tsd. I'm perfectly fine with ts-expect, it's certainly simple enough (could probably just take those limited expect types and make them local tbh, since they're fairly straightforward to understand - I don't care one way or another here, it's up to @bluebill1049).

I don't mind. As long as it does the job.

src/types/path/lazy.ts
```diff
@@ -0,0 +1,251 @@
...
+ * SuggestPath<{foo: {bar: string}}, 'foo', string> = 'foo.bar'
+ * SuggestPath<{foo: {bar: string}}, 'foo.ba', string>
+ *   = 'foo' | 'foo.bar'
+ * SuggestPath<{foo: {bar: string}}, 'foo.bar', string>
+ *   = 'foo' | 'foo.bar'
+ * SuggestPath<{foo: {bar: {baz: string}}}, 'foo.bar', string>
+ *   = 'foo' | 'foo.bar.baz'
+ * ```
+ */
+export type SuggestPath<
```
bluebill1049, line 170: 🥳

bluebill1049:
those unit tests area really great, it actually help review this awesome PR.

src/types/path/lazy.ts
```diff
@@ -0,0 +1,342 @@
...
+
+export type ContainsTuple<T> = (
+  T extends ReadonlyArray<any> ? (_: T['length']) => void : never
+) extends (_: infer I) => void
+  ? number extends I
+    ? false
+    : true
+  : false;
+
+export type ContainsTupleKey<T> = keyof {
```
barrymay, line 41: Can you explain this one?

felixschorer: That was still a work in progress. Sorry for the confusion. I've cleaned it up and added test cases. Please have another look.

src/__tests__/types/path/common.test-d.ts
```diff
@@ -0,0 +1,793 @@
...
+    expectType<['foo', 'bar', '0', 'baz']>(actual);
+  }
+
+  /** it should split a PathString which does not contain a "." */ {
+    const actual = _ as SplitPathString<'foo'>;
+    expectType<['foo']>(actual);
+  }
+
+  /** it should split a PathString containing only a "." */ {
+    const actual = _ as SplitPathString<'.'>;
```
bluebill1049, line 112: Hmm, maybe this should be `never`? I don't think anything should start with '.'.

src/__tests__/types/path/common.test-d.ts
```diff
@@ -0,0 +1,793 @@
...
+    expectType<['', '']>(actual);
+  }
+
+  /** it should be implemented tail recursively */ {
+    const actual = _ as SplitPathString<HundredPathString<'foo'>>;
+    expectType<HundredTuple<'foo'>>(actual);
+  }
+
+  /** it should work on unions */ {
+    const actual = _ as SplitPathString<'foo.bar' | 'bar.foo'>;
```
bluebill1049, line 122: oh nice! union works as well

src/__tests__/types/path/common.test-d.ts
```diff
@@ -0,0 +1,793 @@
...
+        value: EvaluatePath<T, SplitPathString<PS>>,
+      ) =>
+        value;
+
+    const setter = makeSetter<{ foo: string }>();
+
+    const actual = setter('foo', 'bar');
+    expectType<string>(actual);
+  }
+}
```
bluebill1049, line 793: at least 100 plus tests for all the types! impressive ✨ wish we had this earlier.

src/__tests__/types/path/common.test-d.ts
felixschorer: They all work on unions 😅

src/__tests__/types/path/common.test-d.ts
felixschorer: Hm, good point. I'll have to check how `get` behaves. My thought was that you could have an object like this `{'': {'': 42}}`.

src/__tests__/types/path/common.test-d.ts
felixschorer: It should be closer to 200 I think. It was 182 last time I've checked.

src/__tests__/types/path/common.test-d.ts
felixschorer: You were correct. I fixed it. It now removes blank keys from the tuple.

Instead of `never` it returns the empty tuple though. Semantically, this works better. I.e. it's an empty path, not a non-existing path.

I've also altered `Keys` to not return blank keys. And also `AutoCompletePath` to not accept the empty string as a valid path string.

bluebill1049: That's awesome!

barrymay:
This is truly an amazing piece of TS work along with great tsd testing and superb jsdoc comments.  I need a couple days to finalize my review but this change makes RHF so much stronger.

bluebill1049:
Thanks, @barrymay sorry to take a bit longer to review this. If you are happy with I can merge it into the branch and publish a beta first.

felixschorer:
@bluebill1049 @barrymay This PR doesn't change the function definitions yet. It just add the types. They are currently unused, so releasing a beta is a bit pointless 😅

felixschorer:
To keep the individual PRs manageable (this one is already too large imo) my plan would be to split the work like this:

1. Add lazy types, but don't use them yet (this PR)
2. Refactor eager types to be up to standard with the lazy types and ensure compatability between them
3. Integrate the lazy types into the function signatures
4. (Fix `PathValue` being used for function arguments, i.e. setters, as is distributive on the path when it should intersect the types)

bluebill1049:
sounds good to me @felixschorer I will merge this one in first, I will keep reading the code and probably asking you questions. at this can unblock the rest of the feature.

PR #5945, opened 2021-07-14
improve useWatch and watch types

kotarella1110:
## Overview

- fixed using [Variadic Tuple Types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#variadic-tuple-types) to correctly infer `useWatch` return value type
- refactor: removed unnecessary type definitions in `watch` and `getValues`

## Issue

close #5928

## CodeSandbox

https://codesandbox.io/s/5928-t17rg?file=/src/App.tsx

PR #7353, opened 2021-12-23
🎳 fix: change type order to make typescript hints for watch available

pawk3k:
The problem is in `src/types/form.ts` after last update when hardcoded 10 [pr](https://github.com/react-hook-form/react-hook-form/pull/5945/files) values where changed when you provide array of values to watch like watch(['name','lastname']) typescript is not smart enough to give hints
Here is the sandbox with reordered types 
https://codesandbox.io/s/react-hook-form-get-started-ts-forked-bx0x1?file=/src/index.tsx
Here is a codesandbox like it works right now
https://codesandbox.io/s/zen-chatterjee-j46b6?file=/src/App.tsx


felixschorer:
Did the line ending of the API Extractor report change?

pawk3k:
> Did the line endings of the API Extractor report change?
Yeah , gh actions told me that public api changed so i had to make api extractor report as was suggested.

felixschorer:
Yeah, that's correct. However, it seems you have committed CRLF line endings.

You can setup Git to convert CRLF to LF on Windows automatically. See https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

Alternatively, most editors allow you to change the line endings for a file.

pawk3k:
> Yeah, that's correct. However, it seems you have committed CRLF line endings.
> 
> You can setup Git to convert CRLF to LF on Windows automatically. See https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings
> 
> Alternatively, most editors allow you to change the line endings for a file.

Thanks 🔥  , today i learned that this issue might exist 👍

felixschorer:
Do I understand correctly that that's only an issue with `useWatch`?

The issue is that Typescript does not infer the type of the `name` property to be a tuple by default.
That shouldn't be the case with the bound `watch` though, i.e.  the `watch` property returned by `useForm`.

You can, however, force Typescript to infer it as a tuple using `as const` like so:
```ts
const [platform, title, amount] = useWatch({
  control: form.control,
  name: ["platform", "title", "amount"] as const,
});
```

It seems odd to me that reordering the overloads of the bound `watch` would fix an issue with `useWatch`. Can you please confirm that this fixes the issue with `useWatch`? Your code sandbox only demonstrates this works for the bound `watch`.

felixschorer:
Did you change anything just now on https://codesandbox.io/s/zen-chatterjee-j46b6?file=/src/App.tsx ?
I can't reproduce it anymore. Even without the `as const` it seems to work for me.

pawk3k:
I believe i didn't change anything with that codesandbox , one more thing it doesn't fixes `useWatch` and i was only reordering types with bounded `watch` i will take a look on `useWatch` as well but i'm not typescript expert :smile: . By the way thanks for your engagement.

felixschorer:
Which version of Typescript are you using?

pawk3k:
https://codesandbox.io/s/react-hook-form-get-started-ts-forked-bx0x1?file=/src/index.tsx in this sandbox and on work 4.5.4

felixschorer:
As I said, I can't reproduce the issue in the sandbox anymore. Could you please create a test case in `src/__tests__/type.test.tsx` that would raise a compiler error of you hadn't reordered the overloads?

pawk3k:
i will try but the issue i'm trying to show is that how i expect bounded  `watch` should behave:
https://codesandbox.io/s/react-hook-form-get-started-ts-forked-bx0x1?file=/src/index.tsx works as expected
image
Here what i'm getting in 
https://codesandbox.io/s/zen-chatterjee-j46b6?file=/src/App.tsx
image with the regular order of types

felixschorer:
Oh, I see. There has been a misunderstanding. I thought you were getting a compiler error / too broad types. No need to write a regression test in that case.

pawk3k:
Sorry for unclear codesanboxes 😅  , anyway i'm glad we are on the same page now 👍

felixschorer:
Does the IntelliSense "hints" still work for the other overload? I.e. `watch('foobar')`?

pawk3k:
Yeah, i've tested it. Intelisense works for both overloads 
image
Here is a proof

bluebill1049:
thanks for the PR @pawk3k awesome fix, wondering why is changing order fix this issue?

bluebill1049 (approved):
LGTM.

PR #7367, opened 2021-12-25
💅 include prettier ignore to protect api report

bluebill1049:


felixschorer:
@bluebill1049 I guess reports should never be touched by any formatter? It might make sense to future proof this by excluding the entire directory.

bluebill1049:
yep correct @felixschorer

PR #7368, opened 2021-12-25
Refactor Path, ArrayPath, and PathValue

felixschorer:
- [ ] Determine differences between eager and lazy path types.
  - [x] `Path`
    - [x] Run tests for `Keys` against `Path`
    - [x] Write additional tests
  - [ ] `ArrayPath`
    - [ ] Run tests for `Keys` against `ArrayPath`
    - [ ] Write additional tests
  - [ ] `PathValue`
    - [ ] Run tests for `EvaluatePath` and `EvaluateKey` against `PathValue`
- [ ] Refactor types
  - [ ] Refactor `Path`
    - [ ] Use `Keys` to ensure consistency with lazy types
    - [ ] Refactor type to be tail-recursive, i.e. breadth-first traversal
  - [ ] Refactor `ArrayPath`
    - [ ] Use `Keys` to ensure consistency with lazy types
    - [ ] Refactor type to be tail-recursive, i.e. breadth-first traversal
  - [ ] Refactor `PathValue` to use `EvaluatePath`
- [ ] Test compatability
  - [ ] Write tests to ensure that `Path` is assignable to `LazyPath`
  - [ ] Write tests to ensure that `ArrayPath` is assignable to `LazyArrayPath`

## Current `Path` type inconsistencies
- Paths following a string index signature, aren't expanded anymore.
  **This is technically not an error and doesn't improve the IntelliSense.**
  ```
  Input:    Record<string, { bar: string }>
  Expected: string | `${string}.bar`
  Output:   string
  ```
- Paths containing a number index signature are omitted.
  ```
  Input:    Record<number, { bar: string }>
  Expected: `${number}` | `${number}.bar`
  Output:   never
  ```
- Non-overlapping paths are included. Also applies to tuples, arrays, and objects with numeric keys.
  ```
  Input:    { foo: string } | { bar: string }
  Expected: never
  Output:   'foo' | 'bar'
  ```
- Paths which are not supported by `get` are included. (1)
  ```
  Input:    { 'foo.bar': string }
  Expected: never
  Output:   'foo.bar'
  ```
- Paths which are not supported by `get` are included. (2)
  ```
  Input:    { '': string }
  Expected: never
  Output:   ''
  ```
- Primitive types are expanded at the top level.
  ```
  Input:    string 
  Expected: never
  Output:   keyof string
  ```

### DIFF
--- a/src/__tests__/types/__fixtures__/traversable.ts
+++ b/src/__tests__/types/__fixtures__/traversable.ts
@@ -13,7 +13,4 @@ export type NullableInfiniteType<T> = NullableBase<InfiniteType<T>, T>;
 
 export type Depth3Type<T> = Base<Base<Base<never, T>, T>, T>;
 
-export type NullableDepth3Type<T> = NullableBase<
-  NullableBase<NullableBase<never, T>, T>,
-  T
->;
+export type NullableDepth2Type<T> = NullableBase<NullableBase<never, T>, T>;
--- a/src/__tests__/types/__fixtures__/traversable.ts
+++ b/src/__tests__/types/__fixtures__/traversable.ts
@@ -5,15 +5,15 @@ interface Base<T, V> {
   value: V;
 }
 
+type NullableBase<T, V> = null | undefined | Partial<Base<T, V>>;
+
 export type InfiniteType<T> = Base<InfiniteType<T>, T>;
 
-export type NullableInfiniteType<T> =
-  | null
-  | undefined
-  | Partial<Base<NullableInfiniteType<T>, T>>;
+export type NullableInfiniteType<T> = NullableBase<InfiniteType<T>, T>;
 
 export type Depth3Type<T> = Base<Base<Base<never, T>, T>, T>;
 
-export interface Nested<T> {
-  nested: T;
-}
+export type NullableDepth3Type<T> = NullableBase<
+  NullableBase<NullableBase<never, T>, T>,
+  T
+>;
--- a/src/__tests__/types/__fixtures__/traversable.ts
+++ b/src/__tests__/types/__fixtures__/traversable.ts
@@ -13,4 +13,4 @@ export type NullableInfiniteType<T> = NullableBase<InfiniteType<T>, T>;
 
 export type Depth3Type<T> = Base<Base<Base<never, T>, T>, T>;
 
-export type NullableDepth2Type<T> = NullableBase<NullableBase<never, T>, T>;
+export type NullableDepth1Type<T> = NullableBase<never, T>;
--- a/src/__tests__/types/path/common.test-d.ts
+++ b/src/__tests__/types/path/common.test-d.ts
@@ -28,7 +28,6 @@ import {
   HundredPathString,
   HundredTuple,
   InfiniteType,
-  Nested,
   NullableInfiniteType,
 } from '../__fixtures__';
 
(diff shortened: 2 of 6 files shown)


felixschorer:
I've tried 3 different approaches to fixing the issues with the `Path` type:

1. Same approach as the current implementation. 
  Depth-first traversal using mapped types, replacing `keyof T` with `Keys<T>` and `T[K]` with `EvaluateKey<T, K>`. 
  Didn't work. Typescript complained about the type being excessively deep.
2. Depth-first traversal, but using conditional types instead of mapped types. (9d5b10e)
  Didn't work either. Some intermediate unions were getting too large to represent.
3. Breadth-first traversal. (d184af6)
  Type is very slow to evaluate due to all the intermediate types.
  For large objects TS complains about the type being excessively deep.

While approach 3 works in most cases, I am really not confident in using it.
I doubt I can make it much faster or fix the excessive depth issue.

bluebill1049:
I would vote to go with option 3 for now.

felixschorer:
Maybe I should elaborate why these issues need to be fixed in the first place. 

- `LazyPath` is a lot stricter than the current `Path` implementation. `Path` can include paths which aren't valid. See https://github.com/react-hook-form/react-hook-form/pull/7368#issue-1088576455
- To ensure backward compatibility `Path` needs to be assignable to `LazyPath`. This is currently not the case.

The only options are to marginally improve approach 3, or to make a breaking change by removing `Path` entirely.

bluebill1049:
what would be a migration path if we make a breaking change with `path` entirely.

bluebill1049:
> The only options are to marginally improve approach 3

I can include some references and examples in the changelog as well if that's the path we will take.

felixschorer:
> what would be a migration path if we make a breaking change with `path` entirely.

## Functions
- `LazyPath` would replace all usages of `Path` in function signatures.
- Any component / function wrapping `register`, `Controller`, etc. would also have to switch to `LazyPath`.
- `Path` is used as a default type argument in a lot of the exported function signatures.
  There are no possible migrations for that. However, you would have to go out of your way to depend on that.
  The only way to depend on that is to partially annotate the type arguments. 
  E.g. `useWatch<FooBar>({ control, name })`.

```diff
- interface SomeComponentProps<T extends FieldValues, P extends Path<T>> {
+ interface SomeComponentProps<T extends FieldValues, P extends string> { 
    control: Control<T>
-   name: P
+   name: LazyPath<T, P>
  }

- function SomeComponent<T extends FieldValues, P extends Path<T>>({ 
+ function SomeComponent<T extends FieldValues, P extends string>({
    control, 
    name,
  }: SomeComponentProps<T, P>) {
    const { field } = useController({ control, name })
    // ...
  }
```

## Other Usages
`LazyPath` can only be used for function arguments. https://github.com/react-hook-form/react-hook-form/discussions/7354 could be used to cover all other use cases.

```diff
  interface TextFieldDescriptor<T extends FieldValues> {
    type: "text"
    label: string
-   path: Path<T>
+   path: PathWithMeta<T, string>
  }

  interface NumberFieldDescriptor<T extends FieldValues> {
    type: "number"
    label: string
-   path: Path<T>
+   path: PathWithMeta<T, number>
  }

  type FormDescriptor<T extends FieldValues> = Array<
    TextFieldDescriptor<T> | NumberFieldDescriptor<T>
  >

  interface FooBar {
    foo: string
    bar: number
  }

  const fooBarForm: FormDescriptor<FooBar> = [
-   { type: 'text', label: 'Foo', path: 'foo' },
-   { type: 'number', label: 'Bar', path: 'bar' },
+   { type: 'text', label: 'Foo', path: of('foo') },
+   { type: 'number', label: 'Bar', path: of('bar') },
  ]
```

bluebill1049:
Let's prepare a breaking change if compatible is difficult. so you don't over-stress on this problem.

PR #7388, opened 2021-12-29
Force Git to keep CRLF line endings for API Extractor report

felixschorer:
API Extractor produces CRLF on all operating systems.
This PR forces Git to keep CRLF line endings.


bluebill1049 (approved):
👍

PR #7397, opened 2021-12-30
Implement SetKey and SetPath

felixschorer:
# Changes
- Rename `EvaluateKey` to `GetKey`
- Rename `EvaluatePath` to `GetPath`
- Add `SetKey` as contravariant equivalent of `GetKey`
- Add `SetPath` as contravariant equivalent of `GetPath`

# Examples

```ts
// ✔ correct
function get<T, P extends PathTuple>(obj: T, path: P): GetPath<T, P>

// ❌ wrong
function get<T, P extends PathTuple>(obj: T, path: P): SetPath<T, P>
                                                       ^^^^^^^^^^^^^
```

```ts
// ✔ correct
function set<T, P extends PathTuple>(obj: T, path: P, value: SetPath<T, P>): void

// ❌ wrong
function set<T, P extends PathTuple>(obj: T, path: P, value: GetPath<T, P>): void
                                                             ^^^^^^^^^^^^^
```

### DIFF
--- a/src/__tests__/types/path/common.test-d.ts
+++ b/src/__tests__/types/path/common.test-d.ts
@@ -17,6 +17,7 @@ import {
   NumericKeys,
   ObjectKeys,
   PathString,
+  SetKey,
   SplitPathString,
   ToKey,
   TupleKeys,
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):
LGTM ✌🏻

PR #7407, opened 2022-01-01
Script for running API Extractor in Github Actions

felixschorer:
Adds a script for running API Extractor in Github Actions

1. Finds the API Extractor config for the current working directory (just like the vanilla `api-extractor run` command)
2. Checks that the API report is up-to-date with the API
3. Runs the API Extractor again to check that there are no errors or warnings. This is done in a separate step to keep the output clean. If the API report is not up-to-date it should be immediately visible.
4. Checks that the line endings of the API report matches the API Extractor configuration to detect if someone has altered the line endings during commit.

### DIFF
--- a/.github/workflows/api-extrator.yml
+++ b/.github/workflows/api-extrator.yml
@@ -14,7 +14,7 @@ jobs:
         uses: bahmutov/npm-install@v1
 
       - name: Build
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):
LGTM! thanks

PR #7409, opened 2022-01-01
fix append util

michaltarasiuk:
there is no point in trying to convert a value to array since data is of a predetermined type of array


bluebill1049:
thank you @MichalTarasiuk that's correct!

PR #7410, opened 2022-01-01
fix compact util type

michaltarasiuk:
Hello,
I fixed type in compact util.

I also changed the logic/createFormControl.ts and utils/get.ts, because the previous type of compact util influenced the returned result (functions return the same as before, I did not interfere with it).

I believe that the function itself should decide when it returns "any". Then the subsequent type change is easier if possible, because we know what causes the return type.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -484,7 +484,7 @@ export function createFormControl<
     return generateWatchOutput(names, _names, fieldValues, isGlobal);
   };
 
-  const _getFieldArray = (name: InternalFieldName) =>
+  const _getFieldArray = (name: InternalFieldName): any[] =>
     compact(
       get(
         _stateFlags.mount ? _formValues : _defaultValues,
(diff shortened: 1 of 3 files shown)


src/logic/createFormControl.ts
```diff
@@ -484,7 +484,7 @@ export function createFormControl<
     return generateWatchOutput(names, _names, fieldValues, isGlobal);
   };
 
-  const _getFieldArray = (name: InternalFieldName) =>
+  const _getFieldArray = (name: InternalFieldName): any[] =>
```
bluebill1049, line 487: would it be possible to convert to `unknown` instead?

bluebill1049 (approved):
LGTM.

bluebill1049:
thank you @MichalTarasiuk ❤️

src/logic/createFormControl.ts
MichalTarasiuk: No, this function must return "any". The create Form Control function requires this. Function _getFieldArray before my changes also inference "any" type.

PR #7405, opened 2022-01-01
fix type infer with UseFormHandleSubmit

bluebill1049:
fix `UseFormHandleSubmit` with correct form values infer

PR #7411, opened 2022-01-02
rewerite object get utility

adrianos10:
Hi, I changed the implementation of the objet get utility.
Firstly we can define `T extends Record<string | number | symbol, any> | undefined` which checks the typing of obj param on the TS level without the need to call the `isObject` utility. Secondly I noticed that when using the reduce implementation we cannot break out of the loop when some property in the nested path like `a,b,c,d,e,f` is undefined or null. There is no sense to loop through the other properties when for example `b` is undefined. When using the for...of loop we can exit early. Generally I think this is now more straightforward to read :)

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -262,7 +262,7 @@ export type FormStateSubjectRef<TFieldValues> = Subject<Partial<FormState<TField
 }>;
 
 // @public (undocumented)
-export const get: <T extends Record<string | number | symbol, any> | undefined, U = undefined>(obj: T, path: string, defaultValue?: U | undefined) => any;
+export const get: <T extends Record<string, any>, U = undefined>(obj: T | undefined, path: string, defaultValue?: U | undefined) => any;
 
 // @public (undocumented)
 export type GetIsDirty = <TName extends InternalFieldName, TData>(name?: TName, data?: TData) => boolean;
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -262,7 +262,7 @@ export type FormStateSubjectRef<TFieldValues> = Subject<Partial<FormState<TField
 }>;
 
 // @public (undocumented)
-export const get: <T>(obj: T, path: string, defaultValue?: unknown) => any;
+export const get: <T extends Record<string | number | symbol, any> | undefined, U = undefined>(obj: T, path: string, defaultValue?: U | undefined) => any;
 
 // @public (undocumented)
 export type GetIsDirty = <TName extends InternalFieldName, TData>(name?: TName, data?: TData) => boolean;
--- a/src/utils/get.ts
+++ b/src/utils/get.ts
@@ -3,20 +3,49 @@ import isNullOrUndefined from './isNullOrUndefined';
 import isObject from './isObject';
 import isUndefined from './isUndefined';
 
-export default <T>(obj: T, path: string, defaultValue?: unknown): any => {
-  if (isObject(obj) && path) {
-    const result = compact(path.split(/[,[\].]+?/)).reduce(
-      (result, key) =>
-        isNullOrUndefined(result) ? result : result[key as keyof {}],
-      obj,
-    );
+export default <
+  T extends Record<string | number | symbol, any> | undefined,
+  U = undefined,
+>(
+  obj: T,
+  path: string,
+  defaultValue?: U,
+) => {
+  if (obj && path) {
+    // handle case like 'betty.test.test1[0].test1': 'test'
+    if (!isUndefined(obj[path])) {
+      return obj[path];
+    }
 
-    return isUndefined(result) || result === obj
-      ? isUndefined(obj[path as keyof T])
-        ? defaultValue
-        : obj[path as keyof T]
-      : result;
+    const pathKeys = compact(path.split(/[,[\].]+?/));
+    const pathKeysLastIndex = pathKeys.length - 1;
+
+    let result: any = obj;
+
+    for (const [index, key] of pathKeys.entries()) {
+      const currentValue = result[key];
+
+      if (isNullOrUndefined(currentValue)) {
+        result = isUndefined(currentValue)
+          ? defaultValue ?? currentValue
+          : currentValue;
+        break;
+      }
+
+      if (
+        index < pathKeysLastIndex &&
+        !isObject(currentValue) &&
+        !Array.isArray(currentValue)
+      ) {
+        result = defaultValue ?? undefined;
+        break;
+      }
+
+      result = currentValue;
+    }
+
+    return result;
   }
 
-  return undefined;
+  return defaultValue ?? undefined;
 };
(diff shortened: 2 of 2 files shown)


src/utils/get.ts
```diff
@@ -3,20 +3,49 @@ import isNullOrUndefined from './isNullOrUndefined';
...
-      obj,
-    );
-
-    return isUndefined(result) || result === obj
-      ? isUndefined(obj[path as keyof T])
-        ? defaultValue
-        : obj[path as keyof T]
-      : result;
+export default <
+  T extends Record<string | number | symbol, any> | undefined,
```
felixschorer, line 7: `Record<string | number | symbol, any>` can be shortened to `Record<string, any>`. `obj` will only be indexed by `string`.

I would probably move the `undefined` to the argument, i.e. `obj: T | undefined`. In my opinion keeping generics free of `null` and `undefined` makes it easier to read. E.g.

```ts
declare function isNotNull<T>(value: T | null): value is T
```
vs
```ts
declare function isNotNull<T>(value: T): value is Exclude<T, null>
```
Not that it matters here. Just personal preference. I am not a maintainer, so feel free to ignore any of this. 😅

src/utils/get.ts
```diff
@@ -3,20 +3,49 @@ import isNullOrUndefined from './isNullOrUndefined';
...
+>(
+  obj: T,
+  path: string,
+  defaultValue?: U,
+) => {
+  if (obj && path) {
+    // handle case like 'betty.test.test1[0].test1': 'test'
+    if (!isUndefined(obj[path])) {
+      return obj[path];
+    }
```
felixschorer, line 18: Moving this check to the start of the function changes the behavior of the function.
```ts
get({ 'foo.bar': 1, foo: { bar: 2 } }, 'foo.bar')
```
Would be `1` after the change, bust was `2` before.

@bluebill1049 This should probably be removed in v8. Properties like this aren't officially supported afaik.

src/utils/get.ts
```diff
@@ -3,20 +3,49 @@ import isNullOrUndefined from './isNullOrUndefined';
...
-      : result;
+export default <
+  T extends Record<string | number | symbol, any> | undefined,
+  U = undefined,
+>(
+  obj: T,
+  path: string,
+  defaultValue?: U,
+) => {
+  if (obj && path) {
```
felixschorer, line 14: I am a fan of early returns. I.e.
```ts
if (!obj || !path) {
  return defaultValue
}
```
Feel free to ignore this.

src/utils/get.ts
```diff
@@ -3,20 +3,49 @@ import isNullOrUndefined from './isNullOrUndefined';
...
+
+    for (const [index, key] of pathKeys.entries()) {
+      const currentValue = result[key];
+
+      if (isNullOrUndefined(currentValue)) {
+        result = isUndefined(currentValue)
+          ? defaultValue ?? currentValue
+          : currentValue;
+        break;
+      }
```
felixschorer, line 33: This can be simplified to 
```ts
if (currentValue === null) {
  return null
} else if (currentValue === undefined) {
  return defaultValue
}
```

src/utils/get.ts
```diff
@@ -3,20 +3,49 @@ import isNullOrUndefined from './isNullOrUndefined';
...
+
+      if (
+        index < pathKeysLastIndex &&
+        !isObject(currentValue) &&
+        !Array.isArray(currentValue)
+      ) {
+        result = defaultValue ?? undefined;
+        break;
+      }
```
felixschorer, line 42: A more robust solution solution would be to check that the key exists before accessing it rather than introspecting the value at that key.

```ts
if (key in result) {
  result = result[key]
} else {
  return defaultValue
}
```

felixschorer:
@bluebill1049 The API Extractor script doesn't recognize that the line endlings have changes due to the setting in `.gitattributes`. That causes the CI to convert the line endings to CRLF on checkout. 😅 

It should work once https://github.com/react-hook-form/react-hook-form/pull/7405 is merged.

src/utils/get.ts
felixschorer: I think `get` is part of the public API. Constraining `T` to be a `Record` would be a breaking change 🤔

reports/api-extractor.md
```diff
@@ -1,711 +1,711 @@
...
+
+// Warning: (ae-forgotten-export) The symbol "Subject" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export type FormStateSubjectRef<TFieldValues> = Subject<Partial<FormState<TFieldValues>> & {
+    name?: InternalFieldName;
+}>;
+
+// @public (undocumented)
+export const get: <T extends Record<string | number | symbol, any> | undefined, U = undefined>(obj: T, path: string, defaultValue?: U | undefined) => any;
```
felixschorer, line 265: Constraining `get` to only accept `Record`s is a breaking change.

adrianos10:
@felixschorer Thanks do the review :) I will apply changes when I get home.

bluebill1049:
Thank you very much @adrianos10 I am a big fan of type improvement 👍 Could I ask for a favor here? let's only improve the type for now instead change the implementation details.

adrianos10:
@bluebill1049 I've added the separate typing fix PR in #7417.

src/utils/get.ts
adrianos10: @felixschorer @bluebill1049  I moved this under the loop so it can behave as some kind of fallback before returning the undefined value. For me it would be strange to call a property in object like this, but maybe I am not aware of some specific usage inside the lib. I've added this just to cover the test case.

src/utils/get.ts
adrianos10: I am not sure how to apply this solution. With this `if` I wanted to cover the case when there is more keys to be checked but one of the former keys has a primitive value so there is no need to check further and just return `undefined` or the provided` defaultValue`.

bluebill1049:
thanks a lot @adrianos10 would be able to merge react-18 into your PR? This changes the get API signature and let's ship it in V8.

src/utils/get.ts
felixschorer: `key in result` checks whether the key exists. If it doesn't then you've hit a primitive value. You would do the check at the start of the loop body.

adrianos10:
@bluebill1049 Maybe it's worth looking again because one comment from @felixschorer gave me some thinking and I have simplified it even more :)

bluebill1049:
I thank you for putting a lot of effort and thinking into this (by all means, not trying to stop you from contributing to this project) however since get is a public API and I prefer not to touch or change it at this stage.  however, i would like to accept this part:

```tsx
if (!obj || !path) {
  return defaultValue;
}
```

which has none impact on the code result, if you are happy to update. hope for your understanding.

adrianos10:
@bluebill1049 Sure, no problem, totally understand your point of view. I didn't want to leave some code that was commented without any fixes ;). I've added one PR to master #7421 and one to the v8 base branch #7422. We don't need to check `obj` with `isObject` utility because in v8 there will be check on TS level.

PR #7414, opened 2022-01-02
fix isPrimitive utility

michaltarasiuk:
Hello,
I fixed utils/isPrimitive.ts. In my opinion this is a simpler way to check if a value is primitive.


felixschorer:
In my opinion the current implementation is a lot easier to understand than the proposed change. Dumb code is better than clever code.

Issue #7279, opened 2021-12-14
Allow customisation of object cloning when the form is reset


gerrod:
**Is your feature request related to a problem? Please describe.**
Resetting the form does a "simple" clone of objects, which does not allow for cloning objects (or class instances). As an example, if a Luxon `DateTime` object is passed as a property the form's values, after the form has been `reset()`, the `DateTime` instance is no longer valid.

Refer to full discussion here: https://github.com/react-hook-form/react-hook-form/issues/6592

**Describe the solution you'd like**
There are a few alternatives:

* When the form is created, allow the `clone` method to be overwritten:

```js
const form = useForm({
    defaultValues: { ... },
    clone: (original) => ({ ...original }),
});
```

* Allow an array of "replicators" to be passed to the form, which will be used internally by its clone method. Each replicator could be keyed by a function which returns true/false that determines if that function should be used, and then the method to do the replication itself. e.g. again using Luxon's `DateTime` as an example, we might provide something like:

```js
const form = useForm({
    defaultValues: { ... },
    replicators: [
        [obj => obj instanceof DateTime, (obj) => DateTime.fromObject(obj.toObject())]
    ]
});
```

Then, internally, when the form was cloning the default values, it could try the replicators first, and use its default implementation if there wasn't a suitable one found. Simplistically, something like:

```js
const cloneValues = (defaultValues) => {
    return Object.keys(defaultValues)
        .reduce((clone, key) => {
            const value = defaultValues[key];
            const replicator = replicators.find(value);
            return {
                ...clone,
                [key]: replicator ? replicator(value) : value,
            };
        }, {}};
};
```

**Describe alternatives you've considered**
With the current implementation, all values for the form need to be reduced to a primitive in order for them to be cloneable. This is ok for some scenarios - e.g. the Luxon `DateTime` can be changed to millis-since-epoch, but this is less than ideal since it means a lot of conversion each time a value is going in and out of the form.

bluebill1049:
Thanks for the detailed feature request, putting it into the backlog, and waiting for the upvote. personally, i like

```tsx
const form = useForm({
    defaultValues: { ... },
    clone: (original) => ({ ...original }),
});
```

bluebill1049:
another purposal

```tsx
const form = useForm({
  utitlites: {
    clone: yourOwnCloneMethod,
    equal: yourOwnDeepEqualMethod,
  }
})
```

SafaAlfulaij:
[...]

PR #7416, opened 2022-01-02
fix deepEqual utility

michaltarasiuk:
Hello
I rewrote  the deepEqual utility.


bluebill1049:
thanks very much for the PR, for deep equal, I think we will keep it as it is. In fact, we are considering exposing the utility, so users can provide their own: https://github.com/react-hook-form/react-hook-form/issues/7279

PR #7424, opened 2022-01-03
fix prepend utility

michaltarasiuk:


src/utils/prepend.ts
```diff
@@ -1,5 +1,5 @@
 import convertToArrayPayload from './convertToArrayPayload';
 
 export default function prepend<T>(data: T[], value: T | T[]): T[] {
```
bluebill1049, line 3: I think it's save to remove the value: T and replace with value: T[]

src/utils/prepend.ts
```diff
@@ -1,5 +1,5 @@
...
-  return [...convertToArrayPayload(value), ...convertToArrayPayload(data)];
+  return [...convertToArrayPayload(value), ...data];
```
bluebill1049, line 4: could you double-check if `convertToArrayPayload` is required for the first argument as well?

PR #7427, opened 2022-01-03
rewrite objectHasFunction utility

michaltarasiuk:
Hello 👋
I think that functional programing easier to read. What do you think about this.


bluebill1049:
thanks for the PR, `for of` is faster and with an early exit.

PR #7429, opened 2022-01-03
Add AccessPattern type to constrain property accesses by a Get and Set type

felixschorer:
Adds the `AccessPattern` type to constrain a path or key by a `Get` and `Set` type.
The `Get` type is covariant and the `Set` type is contravariant. 
See https://dmitripavlutin.com/typescript-covariance-contravariance/

The idea is that accessing the property `foo` of the type `{ foo?: number } | { foo: number }` yields the type `(number | undefined) | number`, or just `number | undefined`. 
However, when trying to set the value of the `foo` property, the type would have to match both `number | undefined` and `number`, i.e. `(number | undefined) & number` or just `number`. 

`AccessPattern` allows these constraints to be modelled. In this case the access pattern of `foo` would be `AccessPattern<number | undefined, number>`.

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -298,10 +298,10 @@ export type KeepStateOptions = Partial<{
 }>;
 
 // Warning: (ae-forgotten-export) The symbol "AutoCompletePath" needs to be exported by the entry point index.d.ts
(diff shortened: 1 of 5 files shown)


bluebill1049 (approved):
Amazing! `AccessPattern` 🔥

PR #7438, opened 2022-01-05
Reorganize & rename path types

felixschorer:
- Moved types from `path/eager.ts` to `path/index.ts` to make room for branded types. Will be removed in the following PRs.
- Extracted types from  `path/common.ts` to `path/internal`
- Extracted types from `path/lazy.ts` to `path/internal`
- Renamed
  - `GetKey` to `KeyGetValue`
  - `SetKey` to `KeySetValue`
  - `GetPath` to `PathGetValue`
  - `SetPath` to `PathSetValue`


felixschorer:
For some reason API Extractor includes unexported types in the report...

felixschorer:
`import * as Internal from './internal';` caused API Extractor to include those types in the report.

src/__tests__/types/path/internal/autoCompletePath.test-d.ts
```diff
@@ -0,0 +1,386 @@
...
+import { _, InfiniteType, NullableInfiniteType } from '../../__fixtures__';
+
+/** {@link SuggestParentPath} */ {
+  /** it should evaluate to the parent path */ {
+    const actual = _ as SuggestParentPath<['foo', 'bar', 'baz']>;
+    expectType<'foo.bar'>(actual);
+  }
+
+  /** it should evaluate to never if there is no parent path */ {
+    const actual = _ as SuggestParentPath<['foo']>;
```
bluebill1049, line 20: nice

felixschorer: That test has always been there. This PR is just moving and renaming things. No code added or removed. 😅

bluebill1049: yea, last time merged with a proper look. this time going to take my time with it. huge effort!

src/__tests__/types/path/internal/autoCompletePath.test-d.ts
```diff
@@ -0,0 +1,386 @@
...
+      'foo.value',
+      AccessPattern<number>
+    >;
+    expectType<'foo'>(actual);
+  }
+
+  /** it should not accept trailing dots */ {
+    const actual = _ as AutoCompletePath<InfiniteType<string>, 'foo.'>;
+    expectType<'foo' | 'foo.foo' | 'foo.bar' | 'foo.baz' | 'foo.value'>(actual);
+  }
```
bluebill1049, line 186: What would be this use case?

src/__tests__/types/path/internal/autoCompletePath.test-d.ts
```diff
@@ -0,0 +1,386 @@
...
+
+  /** it should not accept repeated dots */ {
+    const actual = _ as AutoCompletePath<InfiniteType<string>, 'foo..foo'>;
+    expectType<
+      | 'foo'
+      | 'foo.foo'
+      | 'foo.foo.foo'
+      | 'foo.foo.bar'
+      | 'foo.foo.baz'
+      | 'foo.foo.value'
```
bluebill1049, line 201: does that mean 'foo..too' will return the above suggestion?

src/__tests__/types/path/internal/autoCompletePath.test-d.ts
```diff
@@ -0,0 +1,386 @@
...
+    expectType<'bar.0'>(actual);
+  }
+
+  /** TS should be able to infer the generic */ {
+    const fn = <P extends PathString>(
+      path: AutoCompletePath<InfiniteType<string>, P>,
+    ) => path;
+
+    const actual = fn('foo.bar');
+    expectType<'foo' | 'foo.bar' | 'foo.bar.0'>(actual);
```
bluebill1049, line 306: how did this infer to `foo.bar.0`?

src/__tests__/types/path/internal/autoCompletePath.test-d.ts
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  /** it should suggest/accept only overlapping paths */ {
+    const actual = _ as AutoCompletePath<
+      { foo: { bar: string } } | { foo: { baz: string } },
+      'foo'
+    >;
+    expectType<'foo'>(actual);
+  }
+}
```
bluebill1049, line 386: Awesome type tests! super comprehensive.

src/__tests__/types/path/internal/hasPath.test-d.ts
```diff
@@ -0,0 +1,150 @@
...
+  ValidPathPrefix,
+} from '../../../../types/path/internal/hasPath';
+import {
+  _,
+  HundredTuple,
+  InfiniteType,
+  NullableInfiniteType,
+} from '../../__fixtures__';
+
+/** {@link HasKey} */ {
```
bluebill1049, line 15: `hasKey` only works for the root level right?

src/__tests__/types/path/internal/hasPath.test-d.ts
```diff
@@ -0,0 +1,150 @@
...
+    const actual = _ as ValidPathPrefix<InfiniteType<string>, []>;
+    expectType<[]>(actual);
+  }
+
+  /** it should be implemented tail recursively */ {
+    const actual = _ as ValidPathPrefix<
+      InfiniteType<string>,
+      HundredTuple<'foo'>
+    >;
+    expectType<HundredTuple<'foo'>>(actual);
```
bluebill1049, line 80: 🧠

src/__tests__/types/path/internal/autoCompletePath.test-d.ts
felixschorer: Yes, it works exactly like `compact(path.split('.'))`. So `'foo.foo'` and `'....foo....foo....'` will have the same suggestions.

bluebill1049: cool

src/__tests__/types/path/internal/autoCompletePath.test-d.ts
felixschorer: For the path `'foo.bar'`
- `'foo.bar'` will be included as it is a valid path. Not including it would show an error on the argument.
- `'foo'` will be included as a suggestion as it is the parent path.
- `'foo.bar'` points to a one element tuple. So `'foo.bar.0'` will be included as a suggestion as it is a child path (the only one in this case)

Keep in mind that this test is testing the return value of the function, which is `AutoCompletePath`. The generic will be inferred to just `'foo.bar'`.

src/__tests__/types/path/internal/keys.test-d.ts
```diff
@@ -0,0 +1,259 @@
...
+
+/** {@link ObjectKeys} */ {
+  /** it should return the keys of an object */ {
+    const actual = _ as ObjectKeys<{ foo: string; bar: number }>;
+    expectType<'foo' | 'bar'>(actual);
+  }
+
+  /** it should return the overlapping keys of a union of objects */ {
+    const actual = _ as ObjectKeys<
+      { foo: string; bar: number } | { bar: number; baz: string }
```
bluebill1049, line 109: Interestingly, ObjectKeys give me the impression it will return all `keys`, for union is actually return the interaction types.

src/__tests__/types/path/internal/hasPath.test-d.ts
felixschorer: Correct there is always a type that works on a single `Key` and one that works on a tuple of `Key`s, i.e. a `PathTuple`. `HasKey` vs `HasPath` in this instance.

src/__tests__/types/path/internal/pathGetValue.test-d.ts
```diff
@@ -0,0 +1,283 @@
...
+    expectType<number | InfiniteType<number>>(actual);
+  }
+
+  /** it should add undefined if one of the paths doesn't exist */ {
+    const actual = _ as PathGetValue<
+      InfiniteType<number>,
+      ['foo', 'value'] | ['foo', 'foobar']
+    >;
+    expectType<number | undefined>(actual);
+  }
```
bluebill1049, line 210: I hope no one else does that. lol

src/__tests__/types/path/internal/keys.test-d.ts
felixschorer: Yes, I believe `keyof` works the same. `'foo'` can't be used to index `{ bar: number; baz: string }` and `'baz'` can't be used to index `{ foo: string; bar: number }`.

src/__tests__/types/path/internal/pathSetValue.test-d.ts
```diff
@@ -0,0 +1,305 @@
...
+/** {@link KeySetValue} */ {
+  /** it should traverse an object */ {
+    const actual = _ as KeySetValue<{ foo: number; bar: string }, 'foo'>;
+    expectType<number>(actual);
+  }
+
+  /** it should traverse an index signature */ {
+    const actual = _ as KeySetValue<Record<string, number>, string>;
+    expectType<number>(actual);
+  }
```
bluebill1049, line 20: what's the difference between the `pathGetValue` type?

felixschorer: You need to differentiate between `KeyGetValue`/`KeySetValue` and `PathGetValue`/`PathSetValue`. As with `HasKey` and `HasPath`, the former operates on individual `Key`s and the latter on tuples of `Key`s, i.e. `PathTuple`s.

The difference between `*GetValue` and `*SetValue` is how they handle unions and non-existing keys/paths.
 - If a key/path doesn't exist `*GetValue` will add `undefined` to the type and `*SetValue` will evaluate to `never`. think about it this way, you can always get the value of a non-existing key/path (i.e. `undefined`), but you shouldn't be able to set it.
 - If `T` is a union, then `*GetValue` will distribute over that type. I.e. it gets value for each individual `T` and then joins them to a union in the end.
 - If the key/path is a union, then `*GetValue` will distribute over that key/path. I.e. it gets value for each individual key/path and then joins them to a union in the end.
 - If `T` is a union, then `*SetValue` will **not** distribute over that type. I.e. it gets value for each individual `T` and then intersects them in the end. To set the `'foo'` key of `{ foo: string } | { foo: number }`, you have to provide a value which is both a `string` and a `number`.
 - If the key/path is a union, then `*SetValue` will **not** distribute over that key/path. I.e. it gets value for each individual key/path and then intersects them in the end. To set the `'foo' | 'bar'` key of `{ foo: string, bar: number }`, you have to provide a value which is both a `string` and a `number`.

See the `AccessPattern` type. That's the same topic essentially. https://github.com/react-hook-form/react-hook-form/pull/7429

src/types/path/internal/autoCompletePath.ts
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Type to check the current path against the constraint type.
+ * Returns the path if it is valid and matches the constraint type.
+ * @typeParam T  - type which is indexed by the path
+ * @typeParam PS - the current path into the type as a string
+ * @typeParam PT - the current path into the type as a tuple
+ * @typeParam C  - constraint
+ * @internal
+ */
```
bluebill1049, line 136: can we include an example in the doc as well?

src/types/path/internal/autoCompletePath.ts
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Type to implement {@link AutoCompletePath} without having to compute the
+ * key list more than once.
+ * @typeParam T  - type which is indexed by the path
+ * @typeParam PS - the current path into the type as a string
+ * @typeParam PT - the current path into the type as a tuple
+ * @typeParam C  - constraint
+ * @internal
+ */
```
bluebill1049, line 158: If possible this one as well with example. 🙏

src/types/path/internal/autoCompletePath.ts
felixschorer: Sure, it's just an implementation detail of `AutoCompletePath` though. Think of it as a `AutoCompletePathImplHelper`. Not a very nice name though.

src/types/path/internal/keys.ts
```diff
@@ -0,0 +1,157 @@
...
+export type Keys<
+  T,
+  C extends AccessPattern = AccessPattern,
+> = IsAny<T> extends true
+  ? Key
+  : IsNever<T> extends true
+  ? Key
+  : IsNever<NonNullable<T>> extends true
+  ? never
+  : CheckKeyConstraint<T, KeysImpl<NonNullable<T>>, C>;
```
bluebill1049, line 157: I was wondering what's `Key` lol it's `string`

bluebill1049 (approved):
Huge huge effort, it probably going to take me some time to absorb all the types and when i start using them. Amazing!

src/types/path/internal/autoCompletePath.ts
bluebill1049: kk

src/__tests__/types/path/internal/pathSetValue.test-d.ts
bluebill1049: > The difference between *GetValue and *SetValue is how they handle unions and non-existing keys/paths.

Got it! thanks

src/types/path/internal/keys.ts
felixschorer: There are 3 aliases like this. In my opinion they improve the readability.
```ts
type Key = string
type ArrayKey = number
type PathString = string
```

bluebill1049: yea, i agree.

PR #7443, opened 2022-01-05
fix swap utility

michaltarasiuk:
Hello,
I rewrote swap utility. In my opinion, this method is clearer.


felixschorer:
It could be made even simpler by using a temporary variable.

JamesWong1999:
Thanks for the improvemnt. If it's working right now, perhaps we can skip the change?

PR #7452, opened 2022-01-06
Add user facing path types

felixschorer:
- Adds `Branded`, `Lazy`, and `Either` type namespaces.
  - Each contain a `FieldPath`, `TypedFieldPath`, `FieldArrayPath`, and `TypedFieldArrayPath` type
  - `Lazy`: These types only offer autocompletion for paths at function call sites.
  - `Branded`: Branded paths should only be obtainable by proving to the compiler that the paths are valid, i.e. by calling a function which accepts lazy paths and returns branded paths. These types can be used in all contexts (unlike `Lazy`).
  - `Either`: These types accept branded and lazy paths. `Either` can be used wherever `Lazy` can be used and should be preferred over `Lazy`.
- Adds `FieldPathValue`, `FieldPathValues`, and `FieldPathSetValue` types for getting the value type of lazy and branded paths.
- Various smaller fixes and improvements to type tests.

The types from `Branded` and the types `FieldPathValue`, `FieldPathValues`, and `FieldPathSetValue` will become the default export of the `types/path` module once the old types are removed in the next PR.

This is the final PR before the storm (i.e. replacing all the types in the public API).

@bluebill1049

### DIFF
--- a/src/__tests__/types/path/value.test-d.ts
+++ b/src/__tests__/types/path/value.test-d.ts
@@ -52,10 +52,18 @@ import { _ } from '../__fixtures__';
 }
 
 /** {@link FieldPathValues} */ {
-  /** it should resolve all paths */ {
+  /** it should resolve string paths */ {
     const actual = _ as FieldPathValues<
       { foo: string; bar: number },
       ['foo', 'bar']
+    >;
+    expectType<[string, number]>(actual);
+  }
+
+  /** it should resolve branded paths */ {
+    const actual = _ as FieldPathValues<
+      {},
+      [Branded.TypedFieldPath<{}, string>, Branded.TypedFieldPath<{}, number>]
     >;
     expectType<[string, number]>(actual);
   }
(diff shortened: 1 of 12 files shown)


felixschorer:
Maybe the `Lazy` namespace shouldn't be exported?
The types in `Either` have the same capabilities as the types in `Lazy`, but also accept the types from `Branded`. 
There is no point in ever using `Lazy`.

Instead the `Either` namespace could be exported as `Lazy`. `Either` is a common name in fp libraries.

Thoughts?

src/__tests__/types/path/value.test-d.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  /** it should resolve branded paths */ {
+    const actual = _ as FieldPathValues<
+      {},
+      [Branded.TypedFieldPath<{}, string>, Branded.TypedFieldPath<{}, number>]
+    >;
+    expectType<[string, number]>(actual);
+  }
+}
```
bluebill1049, line 70: It's awesome to see the usages.

bluebill1049 (approved):
LGTM. bring in the strom.

PR #7467, opened 2022-01-07
fix cloneObject utility

michaltarasiuk:
Hello,
I rewrote cloneObject utility. 
In this pull request, the file responsible for testing the object tool has also been changed. I changed the assertions from toEqual or toBe.

My justification:
 - toEqual answers the question if you have the same
 - toBe answers the question of whether you are the same (In JavaScript, objects are a reference type.)

### DIFF
--- a/src/__tests__/utils/cloneObject.test.ts
+++ b/src/__tests__/utils/cloneObject.test.ts
@@ -92,11 +92,6 @@ describe('clone', () => {
       other: 'string',
     };
 
-    expect(cloneObject(test)).toEqual({
-      data: {
-        testFunction,
-      },
-      other: 'string',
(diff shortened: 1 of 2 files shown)


MichalTarasiuk:
Hi, can I have a review please? 😃

MichalTarasiuk:
Checking if a value is a function I would transfer to an optional callback as additional condition

bluebill1049:
thanks for the PR. at this stage, we have no plan to change this utility.

bluebill1049:
thanks for understanding, more references are here: https://github.com/react-hook-form/react-hook-form/issues/7279 we going to expose those utilities eventually.

PR #7475, opened 2022-01-07
unstable `getFieldState` API for retrieve field level state

bluebill1049:
New API: `getFieldState`

## Context

Today we have this annoying issue with primitive input (single value) vs nested value input (object/array), it's impossible to determine the type at the build time and hence we introduced `NestedValue`, however, it isn't perfect and does cause other issues.

## Proposal

This API will be the solution for nested/primitive field state issues with TS, get it ready for v8's TS improvement. The end goal is deprecated [`NestedValue`](https://react-hook-form.com/ts/#NestedValue) type and in the flavor of using this method for the nested field. We are planning to introduce to v7 and reduce the migration headache when eventually move into v8. 

## Example

You can retrieve the relevant single field state.

```tsx
const { _getFieldState, formState } = useForm(); // the _ will be removed eventually when it rolls out

// formState subscription is not notified
const { isDirty, isTouched, invalid, error } = _getFieldState('nestedValueField', formState);

// when formState subscription is notified
const { isDirty, isTouched, invalid, error } = _getFieldState('nestedValueField');
```

## Todos

- [x] unit tests
- [ ] release next version (beta)
- [ ] documentation update
- [x] API poll: https://twitter.com/HookForm/status/1482928929061347329

### DIFF
--- a/src/__tests__/types/form.test-d.ts
+++ b/src/__tests__/types/form.test-d.ts
@@ -1,5 +1,6 @@
 import { expectType } from 'tsd';
 
+import { FieldError } from '../../types';
 import { useForm } from '../../useForm';
 
 /** {@link UseFormHandleSubmit} */ {
@@ -23,5 +24,23 @@ import { useForm } from '../../useForm';
     }>();
 
     handleSubmit((data) => expectType<{ test: string; test1: number }>(data));
+  }
+}
+
+/** {@link _UseFormGetFieldState} */ {
+  /** it should return associated field state */ {
+    /* eslint-disable react-hooks/rules-of-hooks */
+    const { _getFieldState } = useForm({
+      defaultValues: {
+        test: '',
+      },
+    });
+
+    expectType<{
+      invalid: boolean;
+      isDirty: boolean;
+      isTouched: boolean;
+      error: FieldError;
+    }>(_getFieldState('test'));
   }
 }
--- a/src/__tests__/types/form.test-d.ts
+++ b/src/__tests__/types/form.test-d.ts
@@ -43,4 +43,20 @@ import { useForm } from '../../useForm';
       error: FieldError;
     }>(_getFieldState('test'));
   }
+
+  /** it should return associated field state when formState is supplied */ {
+    /* eslint-disable react-hooks/rules-of-hooks */
+    const { _getFieldState, formState } = useForm({
+      defaultValues: {
+        test: '',
+      },
+    });
+
+    expectType<{
+      invalid: boolean;
+      isDirty: boolean;
+      isTouched: boolean;
+      error: FieldError;
+    }>(_getFieldState('test', formState));
+  }
 }
--- a/src/__tests__/useForm/getFieldState.test.tsx
+++ b/src/__tests__/useForm/getFieldState.test.tsx
@@ -591,4 +591,44 @@ describe('getFieldState', () => {
       });
     });
   });
+
+  describe('when field is not found', () => {
+    it('should return field state', async () => {
+      const App = () => {
+        const { control, _getFieldState, formState } = useForm<FormValues>({
+          defaultValues: {
+            nested: {
+              first: '',
+              last: '',
+            },
+          },
+        });
+
+        // @ts-expect-error expected to show type error for field name
+        const { isDirty } = _getFieldState(formState, 'nestedMissing');
+
+        // @ts-expect-error expected to show type error for field name
+        const { isTouched } = _getFieldState('nestedMissing');
+
+        return (
+          <form>
+            <NestedInput control={control} />
+            <p>{isDirty ? 'dirty' : 'notDirty'}</p>
+            <p>{isTouched ? 'touched' : 'notTouched'}</p>
+          </form>
+        );
+      };
+
+      render(<App />);
+
+      await act(async () => {
+        fireEvent.change(screen.getAllByRole('textbox')[0], {
+          target: { value: ' test' },
+        });
+      });
+
+      screen.getByText('notDirty');
+      screen.getByText('notTouched');
+    });
+  });
 });
--- a/src/__tests__/useForm/getFieldState.test.tsx
+++ b/src/__tests__/useForm/getFieldState.test.tsx
@@ -5,334 +5,590 @@ import { Control } from '../../types';
 import { useController } from '../../useController';
 import { useForm } from '../../useForm';
 
+type FormValues = {
+  nested: {
+    first: string;
+    last: string;
+  };
+};
+
+const NestedInput = ({ control }: { control: Control<FormValues> }) => {
+  const { field } = useController({
+    control,
+    name: 'nested',
+    rules: {
+      validate: (data) => {
+        return data.first && data.last ? true : 'This is required';
+      },
+    },
+  });
+
+  return (
+    <fieldset>
+      <input
+        value={field.value.first}
+        onChange={(e) => {
+          field.onChange({
+            ...field.value,
+            first: e.target.value,
+          });
+        }}
+        onBlur={field.onBlur}
+      />
+      <input
+        value={field.value.last}
+        onChange={(e) => {
+          field.onChange({
+            ...field.value,
+            last: e.target.value,
+          });
+        }}
+        onBlur={field.onBlur}
+      />
+    </fieldset>
+  );
+};
+
 describe('getFieldState', () => {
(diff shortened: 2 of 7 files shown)


src/__tests__/useForm/getFieldState.test.tsx
```diff
@@ -0,0 +1,634 @@
...
+            register,
+            _getFieldState,
+            formState: { errors },
+          } = useForm({
+            defaultValues: {
+              test: '',
+            },
+          });
+
+          errors;
```
barrymay, line 103: is this for debugging?  maybe it's worth validating the value in the test if possible?

src/__tests__/useForm/getFieldState.test.tsx
```diff
@@ -0,0 +1,634 @@
...
+          );
+        };
+
+        render(<App />);
+
+        await act(async () => {
+          fireEvent.click(screen.getByRole('button'));
+        });
+
+        screen.getByText('error');
```
barrymay, line 122: while this will work (this is not a blocker), a suggestion I have for testing things like this, with a specific output, is to generate a random number (this is what faker.js was very handy at ;) ) and test that the getByText matches the random number which you'll output in line 112.


(you can apply this to the other tests too).  the main benefit is that if you put the words error or dirty somewhere else in the example, it won't accidentally catch those.

src/__tests__/types/form.test-d.ts
```diff
@@ -25,3 +26,21 @@ import { useForm } from '../../useForm';
...
+/** {@link _UseFormGetFieldState} */ {
+  /** it should return associated field state */ {
+    /* eslint-disable react-hooks/rules-of-hooks */
+    const { _getFieldState } = useForm({
+      defaultValues: {
+        test: '',
+      },
+    });
+
+    expectType<{
```
barrymay, line 39: nice using tsd! 😃 

want to check both modes of the function?

src/logic/createFormControl.ts
```diff
@@ -793,6 +794,21 @@ export function createFormControl<
       : fieldNames.map((name) => get(values, name as InternalFieldName));
   };
 
+  const _getFieldState: _UseFormGetFieldState<TFieldValues> = (
+    formState,
+    name?,
+  ) => {
+    const isFieldName = isString(formState);
```
barrymay, line 801: this is my only significant issue w/ this PR - why not swap the arguments and if formState is passed, it's the nested version?  Even tho this is internal, I'm not a fan of formState having different types because of the overload (you should put TS types on your arguments)

barrymay (commented):
overall I think I get it but I'm stumped on the actual behavior of the _getFieldState function - wondering if there's a way we can keep the arguments swapped the other way, as formState is the part that's optional?

Generally speaking in code that I write, my function arguments and outputs are always typed, this makes sure the other code using the function agrees w/ it.

src/logic/createFormControl.ts
bluebill1049: I agree! will do the swap.

src/__tests__/useForm/getFieldState.test.tsx
bluebill1049: to avoid lint error.

src/__tests__/types/form.test-d.ts
bluebill1049: will do 👍

bluebill1049:
@barrymay please re-review.

src/__tests__/useForm/getFieldState.test.tsx
```diff
@@ -0,0 +1,634 @@
...
+          defaultValues: {
+            nested: {
+              first: '',
+              last: '',
+            },
+          },
+        });
+
+        // @ts-expect-error expected to show type error for field name
+        const { isDirty } = _getFieldState(formState, 'nestedMissing');
```
barrymay, line 608: this line now needs the field swapped, correct?

barrymay (commented):
I have one note on the tests, but feel free to merge once that's fixed ;)

bluebill1049:
thanks @barrymay for the review.

bluebill1049:
FYI @kotarella1110 @jorisre this new method will be included, this method is try to resolve nested value input. which potentially to replace `NestedValue`

src/types/form.ts
```diff
@@ -180,6 +180,18 @@ export type UseFormGetValues<TFieldValues extends FieldValues> = {
...
+  TFieldName extends FieldPath<TFieldValues>,
+>(
+  name: TFieldName,
+  formState?: FormState<TFieldValues>,
+) => {
+  invalid: boolean;
+  isDirty: boolean;
+  isTouched: boolean;
+  error: FieldError;
+};
```
kotarella1110, line 193: I think there is an error in the type of the return value.
The type of a property other than invalid may return undefined in some cases.

https://github.com/react-hook-form/react-hook-form/blob/ad6abef617a8cc25fcda38c93cf64dfb01b0902c/src/logic/createFormControl.ts#L812-L815

Therefore, the following type definition is correct.

```suggestion
export type _UseFormGetFieldState<TFieldValues extends FieldValues> = <
  TFieldName extends FieldPath<TFieldValues>,
>(
  name: TFieldName,
  formState?: FormState<TFieldValues>,
) => {
  invalid: boolean;
  isDirty: boolean | undefined;
  isTouched: boolean | undefined;
  error: FieldError | undefined;
};
```

kotarella1110:
@bluebill1049 Is there any change in the `formState.errors` type by deprecating `NestedValue`?

src/types/form.ts
bluebill1049: valid point, let me fix it with !!

bluebill1049:
> @bluebill1049 Is there any change in the `formState.errors` type by deprecating `NestedValue`?

not within v7, @felixschorer has a really awesome idea on the [errors](https://www.typescriptlang.org/play#code/C4TwDgpgBAshBOBzCAeAggGigIQHxQF4oBvKAWACgpqoBtAaSgEsA7KAawhAHsAzKNFAA+HLnxwBdAFxRGEAB7AILACYBnUT35pKNPVAD8sqAqWqNnLTl37bR2mgYSs2JxJOLl6utwBGAKwgAY2AsP0CQiRtbGMNYBGR0JxcnXGjY-RlHencRVxz02KynQpoZOU9zTXFsUrscEqoM6hkWCAA3BEoAX0pKViV4XgBDIOgAUXh4bngYNUQSdIBbCDU1YeQDGTVgeFZEHr6KUEgoSen4NQB5AODgFAAVfCJidIZmNktxB4ktqAenB4zN5wndyE16nAkKhzjM5ogsLDLjcIvcATlcGkIZkzlM4fNDhQjipggAbYbwaBBbgsHYmPGXGRI663EIoYi8bjcGTEXwU7a7fZYFZrDYQAV7FgLbrdLGUBAXNQAOk53AMSr58AA3PKGcrVeqRetkJQgA) type:

kotarella1110:
@bluebill1049 @felixschorer amazing! thx 💚

barrymay:
FYI - I made a small adjustment that [playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAshBOBzCAeAggGigIQHxQF4oBvKAWACgpqoBtAaSgEsA7KAawhAHsAzKNFAA+HLnxwBdAFxRGEAB7AILACYBnUT35pKNPVAD8sqAqWqNnLTl37bR2mgYSs2JxJOLl6utwBGAKwgAY2AsP0CQiRtbGMNYBGR0JxcnXGjY-RlHencRVxz02KynQpoZOU9zTXFsUrscEqoM6hkWCAA3BEoAX0pKViV4XgBDIOgAUXh4bngYNUQSdIBbCDU1YeQDGTVgeFZEHr6KUEgoSen4NQB5AODgFAAVfCJidIZmNktxB4ktqAenB4zN5wndyE16nAkKhzjM5ogsLDLjcIvcATlcGkIZkzlM4fNDhQjipggAbYbwaBBbgsHYmPGXGRI663EIoYi8bjcGTEXwU7a7fYAbigKzWGwgAr2LEQ3W6WMoCAuagAdJzuAYVXz4ELFQzVerNWL1sgjpQSUFyZSoNTacB6cqAExM-Uou7s9U87VS-ZyhUUJUzNSOtVczXa3UB-Uhw0q40SyNAA) @felixschorer to show it works w/ and without the message prop. 

Awesome stuff regardless!

PR #7501, opened 2022-01-12
🧪 improve test cases with checkbox default value as array

### LINKED ISSUE #7500: support register with uncontrolled  dynamic checkboxes
### Version Number

7.22.5

### Codesandbox/Expo snack

https://codesandbox.io/s/quiet-morning-678cv?file=/src/App.tsx

### Steps to reproduce

1. Go to [the CSB](https://codesandbox.io/s/quiet-morning-678cv?file=/src/App.tsx) (or this [alternate example](https://codesandbox.io/s/hungry-night-kn5nt?file=/src/App.tsx)).
2. Compare the `defaultValues` to the rendered checkboxes, and see checkbox "A" checked by default even though the `defaultValues` does not specify it should be (nor does the input have `defaultChecked`).

### Expected behaviour

Given a checkbox input with a `defaultValues` that is an array, `register()` should only check the checkbox by default if the array includes the input's value.

The issue appears to come from this line: https://github.com/react-hook-form/react-hook-form/blob/cf2c633280bf3f752912263f0e10d07c937ea010/src/logic/createFormControl.ts#L539

I will open a PR shortly.

Chrome

### Relevant log output

_No response_

### Code of Conduct

(edited to clarify `defaultValues` vs `defaultValue`)

spacenate:
Fixes #7500

When calling `register()` on checkbox input groups, `setFieldValue()` usually gets called multiple times, such that the last time it is called, it finds `fieldReference.refs.length` to be greater than 1, and it loops over all checkbox refs and sets their `checked` attribute correctly per these lines:

https://github.com/react-hook-form/react-hook-form/blob/cf2c633280bf3f752912263f0e10d07c937ea010/src/logic/createFormControl.ts#L530-L537
 
In certain edge cases, `setFieldValue()` only gets called once, which means `refs.length` will not be greater than 1. When that happens, this alternate logic is used:

https://github.com/react-hook-form/react-hook-form/blob/cf2c633280bf3f752912263f0e10d07c937ea010/src/logic/createFormControl.ts#L538-L539

Given the context of checkbox inputs, I suggest it is more appropriate to compare values in the same way as when there are multiple checkbox inputs in a group.


spacenate:
I see checkbox inputs with no `value` attribute are returned as a value of `true` by getCheckboxValue.ts. 

To support these checkbox inputs with no `value` attribute, I think `fieldValue === true` should also be considered a match. I added this logic in the "fieldReference.refs.length === 1" scenario block, rather than in `_isCheckboxValuePresent()`, to maintain the existing behavior for multi-checkbox groups (tho I think it would be fine either way IMO).

bluebill1049 (approved):

PR #7520, opened 2022-01-13
V8: remove keyName prop from useFieldArray

bluebill1049:


standup75:
Why do you want to remove this feature? It's important for to have a distinct id from the database. So that:
- When an item arrives in our API without id we know that we need to create one.
- When an item arrives in our API with an id we know how to update it.

We are right now using the `keyname` prop and I wonder if everything will break when we update to v8

pfcodes:
+1

Please don't remove this feature without providing an escape hatch to fact that useFieldArray's `id` field overwrites the field's existing `id` value.

PR #7529, opened 2022-01-14
Add createPath and joinPath

felixschorer:
- Rename `Either` namespace to `Auto`
- Move type tests to `__typetests__`
- Add `createPath` and `joinPath` functions

@bluebill1049


bluebill1049 (approved):
Awesome!

PR #7535, opened 2022-01-14
add getMountedFieldPaths

stutrek:
This adds an API to get the field paths that are currently in the DOM.

## Simple Example

```js
const { register, getMountedFieldPaths } = useForm();

useLayoutEffect(() => 
    const paths = getMountedFieldPaths();
    console.log(paths)
    // ['test', 'long.path.to.field' ]
}, [])

return <div>
    <input {...register('test')} />
    <input {...register('long.path.to.field')} />
</div>
```

## Use Case

I'm making a number of multi-page forms, and I need to ensure that the user has filled out each page before moving on to the next. Using `trigger` without any arguments will check fields that haven't been shown to the user, and always return false. This method allows us to trigger validation only on fields that the user can currently act upon.

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -582,6 +582,7 @@ export type UseFormReturn<TFieldValues extends FieldValues = FieldValues, TConte
     control: Control<TFieldValues, TContext>;
     register: UseFormRegister<TFieldValues>;
     setFocus: UseFormSetFocus<TFieldValues>;
+    getMountedFieldPaths: () => FieldPath<TFieldValues>[];
 };
 
 // @public (undocumented)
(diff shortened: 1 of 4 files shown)


bluebill1049:
Thanks for sending a PR, normally we would like to go through a feature request process first and see if the community does require such a feature before going into RFC and its implementation.

stutrek:
Cool, how do I do that? I know that this will help at least one user, and it's someone willing to do the work. 

I have no specific opinions on the best way to expose this. It could be part of formState, it could be updated by register, it may even be an option passed into trigger that causes it to only checks mounted inputs.

stutrek:
https://github.com/react-hook-form/react-hook-form/issues/7561

stutrek:
I created a new feature request, in the mean time is there a way that I can get this out of the API in a supported way, even if I need to do something hacky?

PR #7541, opened 2022-01-15
Replace path type

felixschorer:
- Removes the eager path types
- Use `Auto.FieldPath` instead of eager types
- Adds `of` for brandings paths
- Adds `join` to join branded paths

|         |                                                                                                                                                                                             | Branded (default)                                                                                | Auto                                                                                             | (Lazy)                                                                       |
|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------|
|         |                                                                                                                                                                                             | Branded types can be used in all contexts, but require branding using `of()`.                    | Auto types can only be used on function arguments. Branded and Lazy types can be assigned to it. | Lazy types can only be used on function arguments. Always prefer Auto types. |
| untyped | Untyped types only prove the existence of a path, but do **not** constrain the type of the value at that path.  Most useful when the path should be treated as a generic. See **Example 1** | `FieldPath`, `Branded.FieldPath`, `FieldArrayPath`, `Branded.FieldArrayPath`                     | `Auto.FieldPath`, `Auto.FieldArrayPath`                                                          | `Lazy.FieldPath`, `Lazy.FieldArrayPath`                                      |
| typed   | Typed types prove the existence of a path **and** constrain the type of the value at that path.  Useful if you for example want to create a text input. See **Example 2**.                  | `TypedFieldPath`, `Branded.TypedFieldPath`, `TypedFieldArrayPath`, `Branded.TypedFieldArrayPath` | `Auto.TypedFieldPath`, `Auto.TypedFieldArrayPath`                                                | `Lazy.TypedFieldPath`, `Lazy.TypedFieldArrayPath`                            |

The difference between the array and the regular variants is that the array variants expect the path to point to a array of a certain type.

**Example 1**
```ts
function get<T, P extends FieldPath<T>>(obj: T, path: P): FieldPathValue<T, P> {
  // ...
}
```

**Example 2**
```tsx
function Input<T>(props: { control: Control<T>, name: TypedFieldPath<T, string> }) {
  const { value, onChange } = useController(props)
  return <input type="text" value={value} onChange={onChange} />
}
```


src/__tests__/useController.test.tsx
```diff
@@ -594,12 +594,12 @@ describe('useController', () => {
...
     let data: unknown;
 
-    function Input<T>({
+    function Input<T, N extends PathString>({
       control,
       name,
     }: {
       control: Control<T>;
-      name: FieldPath<T>;
+      name: Auto.FieldPath<T, N>;
```
bluebill1049, line 602: This would be the breaking change for existing app which using FieldPath?
all FieldPath will need to change to the above?

src/__tests__/useFieldArray/append.test.tsx
```diff
@@ -366,33 +367,33 @@ describe('append', () => {
       test: { name: { deep: string } }[];
     };
 
-    function Input({
+    function Input<N extends PathString>({
       name,
       control,
     }: {
-      name: FieldPath<FormValues>;
+      name: Auto.TypedFieldPath<FormValues, N, string>;
```
bluebill1049, line 374: `Auto.TypedFieldPath`
`Auto.TypedFieldArrayPath`
`Auto.FieldArrayPath`
`Auto.FieldPath`
`FieldPath`

what's the correct usage for each one of them?

src/logic/validateField.ts
```diff
@@ -194,7 +194,7 @@ export default async <T extends NativeFieldValue>(
 
   if (validate) {
     if (isFunction(validate)) {
-      const result = await validate(inputValue);
+      const result = await validate(inputValue as never);
```
bluebill1049, line 197: What's the reason behind this `as never`?

bluebill1049 (approved):
Awesome stuff, another huge step! Just some general questions, especially around those different `Auto` vs `FieldPath`, so we can clearly documented this later on.

src/__tests__/controller.test.tsx
```diff
@@ -18,7 +18,7 @@ function Input<TFieldValues>({
   onChange,
   onBlur,
   placeholder,
-}: Pick<ControllerRenderProps<TFieldValues>, 'onChange' | 'onBlur'> & {
+}: Pick<ControllerRenderProps<TFieldValues, never>, 'onChange' | 'onBlur'> & {
```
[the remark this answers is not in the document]
felixschorer: `ControllerRenderProps` requires a `PathString` type parameter now to determine the field value. However, here we're just interested in the `onChange` and `onBlur` methods (i.e. the surrounding `Pick` type). So we can simply pass `never` for the path as we don't care about the value.

src/__tests__/logic/validateField.test.tsx
```diff
@@ -1227,8 +1227,8 @@ describe('validateField', () => {
             value: 'This is a long text input',
             required: true,
             validate: {
-              test: (value) => value.toString().length < 3,
-              test1: (value) => value.toString().length > 10,
+              test: (value) => (value as string).toString().length < 3,
+              test1: (value) => (value as string).toString().length > 10,
```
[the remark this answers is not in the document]
felixschorer: The `Field` type is not generic and only used internally. Before the change `value` had type `any`, now it's `unknown`.

src/__tests__/useController.test.tsx
felixschorer: You can still use `FieldPath<T>`, but then have to use `of` and possibly a type annotation, e.g. `<Input<FormData> control={control} name={of(name)} />`.

Using `Auto.FieldPath` only requires a change to the types.

src/__tests__/useFieldArray/append.test.tsx
felixschorer: |  |  | Branded (default) | Auto | (Lazy) |
|---|---|---|---|---|
|  |  | Branded types can be used in all contexts, but require branding using `of()`. | Auto types can only be used on function arguments. Branded and Lazy types can be assigned to it. | Lazy types can only be used on function arguments. Always prefer Auto types. |
| **untyped** | Untyped types only prove the existence of a path, but do **not** constrain the type of the value at that path.  Most useful when the path should be treated as a generic. See **Example 1** | `FieldPath`, `Branded.FieldPath`, `FieldArrayPath`, `Branded.FieldArrayPath` | `Auto.FieldPath`, `Auto.FieldArrayPath` | `Lazy.FieldPath`, `Lazy.FieldArrayPath` |
| **typed** | Typed types prove the existence of a path **and** constrain the type of the value at that path.  Useful if you for example want to create a text input. See **Example 2**. | `TypedFieldPath`, `Branded.TypedFieldPath`, `TypedFieldArrayPath`, `Branded.TypedFieldArrayPath` | `Auto.TypedFieldPath`, `Auto.TypedFieldArrayPath` | `Lazy.TypedFieldPath`, `Lazy.TypedFieldArrayPath` |

The difference between the array and the regular variants is that the array variants expect the path to point to a array of a certain type.

**Example 1**
```ts
function get<T, P extends FieldPath<T>>(obj: T, path: P): FieldPathValue<T, P> {
  // ...
}
```

**Example 2**
```tsx
function Input<T>(props: { control: Control<T>, name: TypedFieldPath<T, string> }) {
  const { value, onChange } = useController(props)
  return <input type="text" value={value} onChange={onChange} />
}
```

src/logic/validateField.ts
felixschorer: That was obsolete, good catch.

src/__tests__/useFieldArray/append.test.tsx
bluebill1049: awesome table 👍

tehwalris:
We have a project that uses `react-hook-form` with a large object. The old non-lazy version of `FieldPath` returned ~600 paths and took 10s to evaluate. To make matters worse, TS seemed to evaluate this type on every unrelated code change, so any error check or auto-completion anywhere in our codebase would take at least 10s.

Switching to `react-hook-form@8.0.0-alpha.0` with Felix's lazy `FieldPath` made a **huge** difference for our project. TS error checking and auto-completions are now fast again (~0.5s), just like they would be without using `react-hook-form` at all. Also, the upgrade didn't seem to introduce any new issues.

Thanks a lot for these awesome PRs!

felixschorer:
Seems like I wasn't the only one throwing large objects into `useForm` 😅

bluebill1049:
Going to release `alpha.1` this weekend with some more updates and feature brought over from v7.

bluebill1049:
`react-hook-form@8.0.0-alpha.1`

Issue #4353, opened 2021-03-05
Feature request - Add fieldArrayContext


wdfinch:
**Is your feature request related to a problem? Please describe.**
Yes

**Describe the solution you'd like**
Currently, multiple `useFieldArray` hooks with the same name and control are not connected. This means that methods called on field array A will **not** update the `fields` property of field array B with the same name and control. There are also other issues that are caused by trying to do this. 

Therefore, I propose adding something similar to `useFormContext` and `FormProvider`. I view the API as being very similar where we simply pass the output of the `useFieldArray` to a `FieldArrayProvider`. From there we can use `useFieldArrayContext` to access the properties of the field array. An operation like append would then update all the `fields` properties across multiple components. 

**Describe alternatives you've considered**
Right now use callbacks and custom context to pass around field array methods and properties from the top level.

joefraley:
whatever happened to this? it seems like a super useful feature

sscotth:
Just came across this today myself.

bluebill1049:
Just to be clear the form context API will be applied to the field array, it's not a subscription on each input.

PR #7544, opened 2022-01-15
🪢 feature: useFormContext support field array

bluebill1049:
Field Array context API

```tsx
type FormValues = {
  test: { name: string }[];
};

const Test = () => {
  const context = useFormContext<FormValues>();

  return (
    <>
      {context?.fieldArrays?.test.map((field) => (
        <input key={field.id} />
      ))}
    </>
  );
};

const App = () => {
  const methods = useForm();
  const testFieldArray = useFieldArray({
    control: methods.control,
    name: 'test',
  });

  return (
    <FormProvider
      {...methods}
      fieldArrays={{
        test: testFieldArray,
      }}
    >
      <form>
        <Test />
      </form>
    </FormProvider>
  );
};
```

- [x] unit tests
- [x] API poll https://twitter.com/HookForm/status/1485838122558849026
- [ ] documentation


Moshyfawn:
Hey, @bluebill1049, do you mind if I ask questions here about this feature? I'm not completely sure about what it solves

bluebill1049:
@Moshyfawn I don't mind :) I am trying to lay a good foundation for the breaking change in v8 by removing `setValue` to update an entire field array. 

> will no longer update the entire field array (focus on inputs level), use `replace`, update from useFiedlArray instead.

ref: https://github.com/react-hook-form/react-hook-form/discussions/7433

currently `setValue` does support updating the entire field array which semantically it's not the right thing to do, I would like to encourage developers to use useFieldArray methods instead. For users which need to update the field array at the root level,  using `useFormContext` can be a valid option without the `props` drilling.

Moshyfawn:
From what I've read in [the feature request that presumably sparked this PR](https://github.com/react-hook-form/react-hook-form/issues/4353), the initial issue the OP was talking about was that you can't update the same field array from different components using the `useFieldArray` hook with the same `name` and `control` set. I remember some similar issue I've ran into when first adapting RHF.

I've created [a simple CSB](https://codesandbox.io/s/modest-lovelace-16vle?file=/src/App.tsx) to test the use-case and turns out that the value of the field with the name set as the `name` for both `useFieldArray`s in different components does indeed update the "final" value and you can see the result of a field array operation _(replace in [this CSB](https://codesandbox.io/s/modest-lovelace-16vle?file=/src/App.tsx) case)_ `onSubmit`. Which leads me to believe that it's not that `useFieldArray`s aren't connected, but that they don't subscribe to the value change, per say.

Now, I haven't delved into the code of the `useFieldArray` hook that much, so, excuse my ignorance if I just don't know something about the hook yet, but from [the CSB](https://codesandbox.io/s/modest-lovelace-16vle?file=/src/App.tsx) behavior I've experienced, it seems like to "solve" the issue what needs to be done is a form values subscription rather than a React Context to pass down the fields. The Context means I'll need to call all my `useFieldArray` hooks at the parent level which doesn't seem right; feels like an unnecessary complexity level. Calling `useFieldArray` at any component tree level with the same name of a field I want to access the methods / fields of sounds right; just like I'd call `useWatch` at different components level to watch for the same field value.

> removing `setValue` to update an entire field array

I'm not sure I understand how feeding field arrays into the Context keeps devs from using `setValue` to update the entire field array. Can you elaborate on this one, please?

bluebill1049:
> The Context means I'll need to call all my useFieldArray hooks at the parent level which doesn't seem right; feels like an unnecessary complexity level

Yes, I agree. It also depends on user cases some use cases could be cleaner to avoid the prop drilling.

> I'm not sure I understand how feeding field arrays into the Context keeps devs from using setValue to update the entire field array. Can you elaborate on this one, please?

```tsx
useFieldArray({
  name: 'test'
})
setValue('test', [{ data: 'test' }]) // this usage 🤮
```

Moshyfawn:
> you would like to update field array from different level of components

But what makes adding the Context layer the solution to update field array from different components as opposed to having multiple `useFieldArray` with the same `name` that would subscribe to the form value changes?
> not allow multiple field arrays with the same name.

What's the reason for wanting to deviate from multiple `useFieldArray` hooks with the same `name`, tho? Referring again to [my CSB](https://codesandbox.io/s/modest-lovelace-16vle?file=/src/App.tsx), I can have a "Clear All" button calling `replace([])` in a different place in a form from the field array inputs list by having `useFieldArray` with the same `name` which does update the actual form value in the current latest RHF but just doesn't update the actual components _(UI)_ 'cuz there's no form value subscription in place. In simple words, adding a subscription to "sync" `useFieldArray`s with the same `name` in different places feels much cleaner

bluebill1049:
> But what makes adding the Context layer the solution to update field array from different components as opposed to having multiple useFieldArray with the same name that would subscribe to the form value changes?

That's a valid point. It's not officially supported, may have edge cases which not working. I think still prefer to have a single instance to host the field array

> I can have a "Clear All" button calling replace([]) in a different place in a form from the field array inputs list by having useFieldArray with the same name which does update the actual form value in the current latest RHF but just doesn't update the actual components (UI) 'cuz there's no form value subscription in place. 

Yea, we don't support that at the moment for inputs update, but if more feature requests are related to that we can look into that. It does make things a bit more complicated. In fact, at the moment we are recommended to keep each name unique.

I think we want to avoid multiple instances with the same name for useFieldArray for now, as it's not really covered by the library or having any test coverage.

bluebill1049:
> Currently, multiple useFieldArray hooks with the same name and control are not connected. This means that methods called on field array A will not update the fields property of field array B with the same name and control.

I see what are you coming from now @Moshyfawn I don't think this feature will resolve the above.

bluebill1049:
@Moshyfawn here is a surprise feature: https://codesandbox.io/s/cocky-franklin-sn0vn?file=/src/src/useController.ts

src/useForm.ts
```diff
@@ -19,7 +19,7 @@ import { useSubscribe } from './useSubscribe';
  * @remarks
  * [API](https://react-hook-form.com/api/useform) • [Demo](https://codesandbox.io/s/react-hook-form-get-started-ts-5ksmm) • [Video](https://www.youtube.com/watch?v=RkXv4AXXC_4)
  *
- * @param props - form configuration and validation parameters.
+ * @param props - form options and validation parameters.
```
bluebill1049, line 22: cc @Moshyfawn

bluebill1049:
Thanks for the feedback @Moshyfawn 

>  Using multiple useFieldArray linked to the same fields

we don't support or documented such usage, I still think useFieldArray should have a single source of truth or hook.

```tsx
useFieldArray({ name: 'test' }) 
useFieldArray({ name: 'test' }) // ❌
useFieldArray({ name: 'test' }) // ❌
```

The intended usage for this `fieldArray` context is just so a **single** field array can be actioned in other component siblings or deep nested. I shouldn't relate this PR to that feature request on the board as it doesn't solve that issue. 

```tsx
type FormValues = {
  test: { name: string }[];
};

const Test = () => {
  const context = useFormContext<FormValues>();

  return (
    <>
      {context?.fieldArrays?.test.map((field) => (
        <input key={field.id} />
      ))}
    </>
  );
};

const FieldArrayAction = () => {
  const context = useFormContext<FormValues>();
  
  // doing such action in other component easier than prop drilling
  return <button onClick={() => context.fieldArray?.test?.append()}>test</button>
}

const App = () => {
  const methods = useForm();
  const testFieldArray = useFieldArray({
    control: methods.control,
    name: 'test',
  });

  return (
    <FormProvider
      {...methods}
      fieldArrays={{
        test: testFieldArray,
      }}
    >
      <form>
        <Test />
      </form>

       <FieldArrayAction />
    </FormProvider>
  );
};
```

Moshyfawn:
> it doesn't solve that issue

Ah, that's where I got confused :P It makes much more sense now.

> so a single field array can be actioned in other component siblings or deep nested

What I'd suggest is to, maybe, create a helper hook that would accept the field array name / names and pull the associated field array(s) from the Context. What do you think about that?

bluebill1049:
> What I'd suggest is to, maybe, create a helper hook that would accept the field array name / names and pull the associated field array(s) from the Context. What do you think about that?

Do you have an example? Curious to know why context is not a good option.

Moshyfawn:
No-no, not to get rid of the Context, but to have a convenient hook to pull the field arrays by the name
```tsx
// Somewhere at the form top level
const guestsFieldArray = useFieldArray({
  control: methods.control,
  name: 'test',
});

return (
  <FormProvider
    fieldArrays={{
      guestsFieldArray
    }}
  >
    <Form />
  </FormProvider>
)

// down the form tree
// a different hook name, maybe
const { fields } = useGetFieldArray("guestFieldArray")

return (
  <ul>
    {fields.map(field => (
      -- do whatever with your field array --
    ))}
  </ul>
)

// instead of
const context = useFormContext()

return (
  <ul>
    {context?.fieldArrays.guestFieldArray?.map(field => (
      -- do whatever with your field array --
    ))}
  </ul>
)
```

bluebill1049:
good suggestion @Moshyfawn I did consider that, the reason I didn't go for that path is 

- run time
- and more code footprint

the current approach leverage 100% on context, there is no extra code required other than just TS definition.

Screen Shot 2022-02-12 at 11 30 53 am

Moshyfawn:
Makes sense!

Moshyfawn:
Ah, looking at the functionality and the feature requests around it, I still have doubts about whether RHF should include a custom array Context. Creating a custom Context at the app level to pass around your field arrays feels right at times.
Not saying against the feature, just some thoughts

bluebill1049:
> Creating a custom Context at the app level to pass around your field arrays feels right at times.

Thanks @Moshyfawn for the feedback. What's stopping us from providing it other than asking users to have more context? I think it's just an option that you can supply field array into "our" context, it's optional.

Moshyfawn:
I see the convenience to to have a native to RHF array Context. Basically, less boilerplate + the "recommended" pattern of managing field arrays. That's what I mean, I'm not against it.
I guess, it makes more sense, now that I've listed "the recommended pattern" benefit of it..
Alright, never mind, I guess, I need some coffee before looking at the code xD

bluebill1049:
No rush at all on getting this feature out, I am happy to sit on it and think about it for a while. enjoy your coffee ☕

bluebill1049:
any second thought on this @Moshyfawn?

ZuBB:
> const { fields } = useGetFieldArray("guestFieldArray")

I just wanted to say that my team is in desperate need of this ^ and we will go with some ugly workarounds to overcome that 😞

bluebill1049:
@ZuBB any thoughts on the proposed API above?

ZuBB:
Hi @bluebill1049! 

I am sorry for the slow response.

My teammate @iMykhailychenko and I continue working on a task where we use heavily `useFieldArray()`, and we can confirm (again) that the suggested improvement would make our code a lot simpler.

I am stuck in my own tasks+work, but I will do my best and provide the requested feedback by the end of the weekend. Or maybe I will jump in

bluebill1049:
thanks for the feedback, i will come back to this PR soon as well. hopefully get a pre release so you guys can try it out first.

ZuBB:
Hi @bluebill1049!

I read the whole issue and tried to understand everything happening here. So basically, we have two options here.

* `const { fields } = useGetFieldArray("guestFieldArray")`
* `useFormContext()?.fieldArrays.guestFieldArray?.map(field => ...`

In 1st case, other array stuff (`append`, `remove`, etc) lives at the same level. eg `const { fields, append } = useGetFieldArray("myArray1")`

In the second case, `fields` points to the array itself.

```
useFormContext()?.fieldArrays.guestFieldArray?.map(field => ...
```

But to access other helper methods, we go one level deeper.
```
  return <button onClick={() => context.fieldArray?.test?.append()}>test</button>
```
Which is quite confusing. Because at the same level, we have both props of native JavaScript's `Array` and other methods produced by `useFieldArray()`.

Considering all that is above, I for sure like 1st option more. Also, its usage is shorter and looks more convenient.

bluebill1049:
```
const { fields } = useGetFieldArray("guestFieldArray")

fields.map
````

vs

```tsx
// no optional china operator required
const { fieldArrays: { guestFieldArray } } = useFormContext() 

guestFieldArray.map
```

honestly, I don't see much different here instead to have another hook.

ZuBB:
what about `append()`?

bluebill1049:
> what about `append()`?

What's the difference?

```tsx
const { fields, append } = useGetFieldArray("guestFieldArray")

fields.map
append({})
````

vs

```tsx
// - no optional china operator required
// - type safety will check field array type name
const { fieldArrays: { guestFieldArray } } = useFormContext() 

guestFieldArray.fields.map
guestFieldArray.append({})
```

ZuBB:
Ok, I got where I was wrong (use different messages from 2 ppl)

Yeah, in that case, the difference is neglectable. maybe extending `useFormContext` is even better because we extending the existing API, but not adding a new one

angelocala94:
Do you have any updates on this? Currently I'm using a very bad workaround

bluebill1049:
> @bluebill1049 Is this the place to follow for a pre-release notification for this? I need to know when we should start migrating away from the `setValue('fieldArray', [...])` calls.

Will update once it's going to pre release

Medsestrun:
Hi, any chance to get this merged?

bluebill1049:
I will come back to this after Form component.

SafaAlFulaijLumofy:
Hi all.
This will be really useful and remove a lot of prop drilling!
Thanks for the hard work.

I have a question, how to handle nested field arrays?

```js

const App = () => {
  const methods = useForm();
  const testFieldArray = useFieldArray({
    control: methods.control,
    name: 'test',
  });

  // how to define field arrays inside test? For example if we have a form for products, and each product have availableColors.
  // `products.0.availableColors.0`
 // We can't loop hooks here

  return (
    <FormProvider
      {...methods}
      fieldArrays={{
        test: testFieldArray,
      }}
    >
      <form>
        <Test />
      </form>
    </FormProvider>
  );
};```

**Comment by bluebill1049** [comment]
@Moshyfawn what do you think of pushing this out next minor update?

**Comment by Moshyfawn** [comment]
> @Moshyfawn what do you think of pushing this out next minor update?

OMG, sorry, it completely slipped my mind! 😱

I'm open to releasing it in the next minor update, but I do have concerns about this API potentially misinterpreted spawning issues related to incorrect usage discussed before in this thread and you addressed in https://github.com/react-hook-form/react-hook-form/pull/7544#issuecomment-1036104979.

**Comment by Moshyfawn** [comment]
> @SafaAlFulaijLumofy: ..how to handle nested field arrays?

🤔 That's an interesting one.

@bluebill1049, I'm gonna throw a very raw idea at you just to see if it makes sense. Not proposing a change per say; just curious if it could resolve this question.

What if it was a separate `FieldArray` context instead of `FormProvider` taking it in as a single object?

**Comment by bluebill1049** [comment]
> What if it was a separate FieldArray context instead of FormProvider taking it in as a single object?

would like to share some API example?

**Comment by Moshyfawn** [comment]
Mmm, on second thought..

So, if I have a `fields` field array that contains a nested field array, something like `fields.${index}.nestedFields`, would you say, that there's no way to put the `nestedField` field array **in the context** because you'd have to map over the `fields` field array first?

**Comment by bluebill1049** [comment]
> So, if I have a fields field array that contains a nested field array, something like fields.${index}.nestedFields, would you say, that there's no way to put the nestedField field array in the context because you'd have to map over the fields field array first?

Need to test, but if you have the correct execution order with your `useFielddArray` then it should be fine.

**Comment by Moshyfawn** [comment]
> Need to test, but if you have the correct execution order with your `useFielddArray` then it should be fine.

I mean, I'd have to call 
```jsx
useFieldArray({ name: `fields.${index}.nestedFields` })
```
in a nested component, which means I've already put my `form` inside the `FormProvider`, so I can't update my context from a nested component with the `nestedFields` field array

bluebill1049:
yes, valid one. wouldn't be able to support that

Moshyfawn:
> yes, valid one. wouldn't be able to support that.

K, fair! Just making sure I understand the API!

Moshyfawn:
Another question would be: my field array is called `object.fields`, so the form values would look something like
```js
const defaultValues = {
  object: {
    fields: [{ name: "John Doe" }]
  }
}
```

and the `useFieldArray` call would look like
```js
const objectFieldsArray = useFieldArray({
  name: "object.fields"
})
```

Does `fieldArrays` support special characters in this case?
```jsx
<FormProvider
  fieldArrays={{
    'object.fields': objectFieldsArray
  }}
>
  ...
</FormProvider>
```

or would it be a custom object property name
```jsx
<FormProvider
  fieldArrays={{
    objectFieldsArray: objectFieldsArray
  }}
>
  ...
</FormProvider>
```

bluebill1049:
would be the following: 

```tsx
<FormProvider
  fieldArrays={{
    'object.fields': objectFieldsArray
  }}
>
  ...
</FormProvider>
```

due to type check.

Moshyfawn:
Sorry for a ton of questions! This PR doesn't have a generated codesandbox and I can't put together a custom one at this time

Moshyfawn:
> would be the following:
> 
> ```tsx
> <FormProvider
>   fieldArrays={{
>     'object.fields': objectFieldsArray
>   }}
> >
>   ...
> </FormProvider>
> ```
> 
> due to type check.

I see! As long as it handles nested field array names, I'm content 😋

bluebill1049:
> Sorry for a ton of questions! This PR doesn't have a generated codesandbox and I can't put together a custom one at this time

You can finish this PR if you want to as well 🤣

delyan-peev:
> >
> >I mean, I'd have to call
> >
> > ```js
> > useFieldArray({ name: `fields.${index}.nestedFields` })
> > ```
> >
> > in a **nested component**, which means I've already put my `form` inside the `FormProvider`, so I can't update my context from a nested component with the `nestedFields` field array
>
> yes, valid one. wouldn't be able to support that.



So, how should we approach avoiding prop-drilling for nested arrays. We use RHF pretty extensively in our app and we have several levels of field array nesting that we currently update with `setValue`, since it is the most convenient way we've found to do so. To give an example, we have a form state similar to:

```js
{
  steps: [{
    ...,
    elements: [{
      ...,
      settings: [{
        ...,
        values: [...]
      }]
    }, ...]
  }, ...]
}
```

What we currently do is to have a custom hook for updating elements that looks sth like this:

```ts
export const useFormElements = (stepIndex: number) => {
  const formKey = `steps.${stepIndex}.elements` as const;
  const { setValue } = useFormContext<DesignPopupFormData>();
  const elements = useWatch<DesignPopupFormData, `steps.${number}.elements`>({ name: formKey });
  
  return {
    elements,
    addElement: useCallback((type: ElementType) => {
      setValue(
        formKey,
        [...elements, newElement(getNextOrder(), type)],
        { shouldDirty: true }
      );
    }, [formKey, elements]),
    duplicateElement: useCallback((element: FormElement) => {
      setValue(
        formKey,
        [...elements, { ...duplicateElement(element), order: getNextOrder() }],
        { shouldDirty: true }
      );
    }, [formKey, elements]),
    toggleVisibility: useCallback((index: number) => {
      setValue(
        formKey,
        elements.map((el, i) => (i === index ? { ...el, isVisible: !el.isVisible } : el)),
        { shouldDirty: true }
      );
    }, [formKey, elements]),
    deleteElement: useCallback((index: number) => {
      const newElements = elements.filter((_, i) => i !== index);
      setValue(
        formKey,
        adjustElementsOrder(newElements),
        { shouldDirty: true }
      );
    }, [formKey, elements]),
    setElements: useCallback((elements: FormElement[]) => {
      setValue(formKey, elements, { shouldDirty: true });
    }, [formKey, elements])
  };
}
```

This hook is called in the component of each `Step` and we use its functions to update the elements within it. This process is repeated for every single level of nesting (think the `Element` component to edit its `Settings` and `Setting` component to update its values).

We were aware that this usage was advised against in the docs, but we could not really find a more convenient way to achieve this with `useFieldArray` due to the already discussed subscription issue. Having said all that, what would be the advised way to make use of the `useFieldArray` and the context once this PR is merged?

P.S.
FWIW, I double `Moshyfawn`'s opinion on this:

>The Context means I'll need to call all my useFieldArray hooks at the parent level which doesn't seem right; feels like an unnecessary complexity level. Calling useFieldArray at any component tree level with the same name of a field I want to access the methods / fields of sounds right; just like I'd call useWatch at different components level to watch for the same field value.

I just feel it makes the API across the library more consistent. Plus, calling numerous `useFieldArray`s at the parent level just looks dubious to me :(

trevorpfiz:
anything planned that allows me to call `useFieldArray` in multiple components where I need it and it will all use the same form state?

bjmvercelli:
> Is this change no longer planned? This seems like it would be a really useful feature.

100%.
That would be awesome

dilipthakre:
It will be cool to have separate decoupled logic for field array or do we have any other work around for this.

PR #7636, opened 2022-01-24
fix: getFieldState - error might be undefined

michalbundyra:
Fix to the feature introduced in 7.25.0

when the form is not validated yet or field is valid the `error` field is `undefined`

todo:
- [x] test to cover the fix

### DIFF
--- a/src/__tests__/useForm/getFieldState.test.tsx
+++ b/src/__tests__/useForm/getFieldState.test.tsx
@@ -204,7 +204,11 @@ describe('getFieldState', () => {
           return (
             <form>
               <input {...register('test')} />
-              <p>{getFieldState('test')?.error === undefined ? 'error undefined' : ''}</p>
+              <p>
+                {getFieldState('test').error === undefined
+                  ? 'error undefined'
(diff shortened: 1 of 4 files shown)


src/__tests__/useForm/getFieldState.test.tsx
```diff
@@ -605,16 +719,17 @@ describe('getFieldState', () => {
         });
 
         // @ts-expect-error expected to show type error for field name
-        const { isDirty } = getFieldState(formState, 'nestedMissing');
+        const { isDirty } = getFieldState('nestedMissing', formState);
```
michalbundyra, line 722: it is separate bug in a test - order of the params was wrong here - happy to create separate PR with it, if you prefer

bluebill1049: you are right, I think my last commit forgot to push in the original PR... :( I was pointed out during the review process that undefined value as well. my bad

bluebill1049 (approved):
Awesome PR! thanks for the contribution.

PR #7656, opened 2022-01-25
fix #7662 issue with reset file input

### LINKED ISSUE #7652: issue: resetField method doesn't reset an uncontrolled file input
### Version Number

7.25.0

### Codesandbox/Expo snack

https://codesandbox.io/s/sweet-haibt-rrc4o?file=/src/App.tsx

### Steps to reproduce

1. Go to [this CSB](https://codesandbox.io/s/sweet-haibt-rrc4o?file=/src/App.tsx)
2. Click on XML File input
3. Choose any file and upload
4. Click Reset File Input button
5. See xmlFile form field reset to undefiend but the actual file input still has it's value

### Expected behaviour

An uncontrolled file input is been reset along with it's form field value to it's provided default value

Firefox

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
ref: https://codesandbox.io/s/busy-tesla-flqd7?file=/src/App.tsx&resolutionWidth=677&resolutionHeight=675

fix #7652

### DIFF
--- a/src/__tests__/useForm/resetField.test.tsx
+++ b/src/__tests__/useForm/resetField.test.tsx
@@ -197,6 +197,50 @@ describe('resetField', () => {
     });
   });
 
+  it('should reset input file to empty string only', () => {
+    const getValuesFn = jest.fn();
+
+    const App = () => {
+      const { register, resetField, getValues } = useForm({
+        defaultValues: {
+          test: '',
+        },
+      });
+
+      return (
+        <form>
+          <input type={'file'} {...register('test')} />
+          <button
+            type={'button'}
+            onClick={() => {
+              resetField('test', { defaultValue: null });
+            }}
+          >
+            reset
+          </button>
+          <button
+            type={'button'}
+            onClick={() => {
+              getValuesFn(getValues());
+            }}
+          >
+            getValues
+          </button>
+        </form>
+      );
+    };
+
+    render(<App />);
+
+    fireEvent.click(screen.getByRole('button', { name: 'getValues' }));
+
+    expect(getValuesFn).toBeCalledWith({ test: '' });
+
+    fireEvent.click(screen.getByRole('button', { name: 'reset' }));
+
+    expect(getValuesFn).toBeCalledWith({ test: '' });
+  });
+
   describe('when provided with options', () => {
     it('should update input value and its defaultValue', () => {
       const App = () => {
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -555,7 +555,7 @@ export function createFormControl<
             );
           }
         } else if (isFileInput(fieldReference.ref)) {
-          fieldReference.ref.value = isString(fieldValue) ? fieldValue : '';
+          fieldReference.ref.value = '';
         } else {
           fieldReference.ref.value = fieldValue;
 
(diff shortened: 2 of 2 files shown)


Moshyfawn:
I might've got smth wrong _(so, ignore me if that's the case)_, but wouldn't it be appropriate for the file input's default value be `null` instead of an empty string? The actual value that you gonna get for the input is FileList, which would be of type object, right?

bluebill1049:
you can only programmatically set a file input to an empty string. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file let me know if that's not the case @Moshyfawn 👍

Moshyfawn:
I know about the security measures for file inputs to not allow for setting their default values, but I'm thinking more along the lines of..

Like, if I were to create a validation schema and infer it's types, it would be something like
```js
const schema = z.object({
  file: z.instanceof(FileList).nullable()
})

type FormValues = z.infer<typeof schema>
```
Which will allow for my `defaultValues` be
```js
defaultValues: DefaultValues<FormValues> = {
  file: null
}
```
Having a_(n empty)_ string for the file input's default value would mean
```js
const schema = z.object({
  file: z.instanceof(FileList).or(z.string())
})

type FormValues = z.infer<typeof schema>
```
Which is fair enough, tho, feels.. weird, I guess, as the type of FileList is `object` as it is for null. But setting the file input's default value as string would mean different types.

bluebill1049:
You can still use `null`, it is just when you manipulate the file input, it will be set to empty string.

Moshyfawn:
Huh, I'm still a bit confused, but looking at the [ref CSB](https://codesandbox.io/s/busy-tesla-flqd7?file=/src/App.tsx&resolutionWidth=677&resolutionHeight=675) you've provided, everything works as I'd expect, so I'm happy and will stop bothering you about it now xD

bluebill1049:
not bother at all.

```tsx
setValue('fileInputDate', null) // for the native input it will be set to empty string 
setValue('fileInputDate', undefined) // for the native input it will be set to empty string 
setValue('fileInputDate', '') // for the native input it will be set to empty string 
```
or 

perhaps we can let users set whatever and they can deal with errors coming from the browser perhaps?

bluebill1049:
> perhaps we can let users set whatever and they can deal with errors coming from the browser perhaps?

actually i will do above instead.

bluebill1049:
I remember the reason why now check out this repo: https://codesandbox.io/s/busy-tesla-flqd7?file=/src/src/logic/createFormControl.ts we do store input value, so when input unmounts and remount set input with file list will cause an error, whish is the reason why we do the file input type check and stop programmatically update the value to others type except empty string.

Moshyfawn:
Ah, I see what you mean now, thx!
It's interesting that the input's value is always gonna be an empty string whenever you try to set it to like `null` or `undefined`. In case of the file input, it's gonna thrown an exception as well..
I'm just curious how it's gonna play out as the controlled input with the value set to the empty string

Moshyfawn:
Lol, it's the second time today I'm commenting right after you've commented a similar thing xD

bluebill1049:
lol, for controlled input you are basically hosting your input value inside a local state, but for file input, you have the same security restriction.

Moshyfawn:
Sounds good!
I'm always double checking what working with controlled inputs, what's gonna happened when I do `{...field}` into the controlled input

Issue #7683, opened 2022-01-28
Preact - isBlurEvent is always false since it emits a "focusout" event


vattenapa:
**Is your feature request related to a problem? Please describe.**

react-hook-form explicitly checks the event object type equivalency to "blur". Can we add "focusout" as well? This solves compatibility issues when using preact.

[Reproduction](https://codesandbox.io/s/wizardly-waterfall-8koh5)
1. Click the "First Name" field
2. Focus another element
3. Note in the console that the event type is "focusout", instead of "blur" (vanilla React)

This leads to the field never being registered as touched in RHF, causing field validation to never trigger.

**Describe the solution you'd like**
react-hook-form to handle both "focusout" and "blur" as acceptable event types when setting the `isBlurEvent` boolean.

**Describe alternatives you've considered**
Creating and emitting a new event in the event handler, but I'm not sure it will work as expected since it's not an event emitted by the user.

PR #7684, opened 2022-01-28
Feature/#7683

Moshyfawn:
Ref: #7683
In case you wanna support [the `focusout` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusout_event)
As RHF doesn't officially support anything non-React, IDK if we wanna have tests with Preact.
Please, let me know, what you think.

_P.S. my 5 cents, RHF should support native Web event types as it's the Web API_

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -655,7 +655,7 @@ export function createFormControl<
       const fieldValue = target.type
         ? getFieldValue(field._f)
         : getEventValue(event);
-      const isBlurEvent = 
+      const isBlurEvent =
         event.type === EVENTS.BLUR || event.type === EVENTS.FOCUS_OUT;
       const shouldSkipValidation =
         (!hasValidation(field._f) &&
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -653,7 +653,9 @@ export function createFormControl<
       const fieldValue = target.type
         ? getFieldValue(field._f)
         : getEventValue(event);
-      const isBlurEvent = [EVENTS.BLUR, EVENTS.FOCUS_OUT].some(eventType => event.type === eventType);
+      const isBlurEvent = [EVENTS.BLUR, EVENTS.FOCUS_OUT].some(
+        (eventType) => event.type === eventType,
+      );
       const shouldSkipValidation =
         (!hasValidation(field._f) &&
           !_options.resolver &&
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -655,9 +655,7 @@ export function createFormControl<
       const fieldValue = target.type
         ? getFieldValue(field._f)
         : getEventValue(event);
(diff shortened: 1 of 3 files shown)


src/logic/createFormControl.ts
```diff
@@ -655,7 +655,9 @@ export function createFormControl<
       const fieldValue = target.type
         ? getFieldValue(field._f)
         : getEventValue(event);
-      const isBlurEvent = event.type === EVENTS.BLUR;
+      const isBlurEvent = [EVENTS.BLUR, EVENTS.FOCUS_OUT].some(
+        (eventType) => event.type === eventType,
+      );
```
bluebill1049, line 660:
```suggestion
      const isBlurEvent = [EVENTS.BLUR, EVENTS.FOCUS_OUT].includes(event.type);
```

Save some bytes :)

bluebill1049 (commented):
Thanks for the PR @Moshyfawn 🙏

src/logic/createFormControl.ts
```diff
@@ -655,7 +655,7 @@ export function createFormControl<
       const fieldValue = target.type
         ? getFieldValue(field._f)
         : getEventValue(event);
-      const isBlurEvent = event.type === EVENTS.BLUR;
+      const isBlurEvent = [EVENTS.BLUR, EVENTS.FOCUS_OUT].includes(event.type);
```
bluebill1049, line 658:
```suggestion
      const isBlurEvent = event.type === EVENTS.BLUR || event.type === EVENTS.FOCUS_OUT;
```

bluebill1049: actually go with cheapest 👍

bluebill1049 (approved):
Merge and getting it published.

Moshyfawn:
Fair enough!
Sry, it was night time for me :P

I wasn't sure about inducing any tests as, technically, in React there shouldn't be a `focusout` even, right?
```js
fireEvent.focusOut(screen.getByRole('textbox'));
```

bluebill1049:
I think React still recognise focus event from the Dom.

Moshyfawn:
I was referring [to this in React's codebase](React repo](https://github.com/facebook/react/blob/9198a5cec0936a21a5ba194a22fcbac03eba5d1d/packages/react-dom/src/shared/ReactDOMUnknownPropertyHook.js#L36)
```js
console.error(
  'React uses onFocus and onBlur instead of onFocusIn and onFocusOut. ' +
    'All React events are normalized to bubble, so onFocusIn and onFocusOut ' +
    'are not needed/supported by React.',
);
```

bluebill1049:
I see, probably need to update that unit test then 👍

Moshyfawn:
What do u wanna update?

bluebill1049:
I think the test case just need to covert that we accept both events.

Moshyfawn:
Idk, maybe it's good as a standalone test, since the `focusout` event is not a part of React. I don't really know what's the best strategy to test this case, tbh

bluebill1049:
i will come back to it 

```tsx
const { onChange } = register()

onChange(pass a fake focus event)
```

PR #7690, opened 2022-01-29
TSDoc

bluebill1049:
## TSDoc

Improve DX with TSDoc for each public API.

- Description
- API Link
- Demo Link
- Video Tutorial Link
- Code example

### Screenshot:

Screen Shot 2022-01-29 at 11 28 44 pm


Moshyfawn:
Oh, hell, yeah! I was going to suggest this as well!
Also, might be a good idea to include a ref link to the docs section of the method TSDoc is describing, like, say, MUI does
image

bluebill1049:
> Oh, hell, yeah! I was going to suggest this as well! Also, might be a good idea to include a ref link to the docs section of the method TSDoc is describing, like, say, MUI does image

👍

bluebill1049:
Screen Shot 2022-01-29 at 9 45 02 pm
@Moshyfawn what do you think?

Moshyfawn:
What do you think about showing the code example below the method's description and the links to demos and docs after? I just think in what order I would need such information.

Additionally, what about using the MD for links like
```md
/**
 * API:
 * - [register API](https://react-hook-form.com/api/useform/register)
 * Demos:
 * - [register](https://codesandbox.io/s/react-hook-form-js-wbybv)
 */
 to reduce the clutter?
```

bluebill1049:
Screen Shot 2022-01-29 at 9 57 50 pm

I think example will gets put under desc automatically

bluebill1049:
Screen Shot 2022-01-29 at 10 03 58 pm

This is the cleanest.

Moshyfawn:
Something like that should work in theory..
```ts
/**
 * Register field desc
 * More words maybe
 *
 * ### Example
 * ```jsx
 * const fieldName = register("fieldName")
 *
 * return <input {...fieldName} />
 * ```
 *
 * Demos:
 *
 * - [register](https://codesandbox.io/s/react-hook-form-js-wbybv)
 *
 * API:
 *
 * - [register API](https://react-hook-form.com/api/useform/register)
 */
```

Moshyfawn:
Ah, I like the links you've got going! Sry, didn't see your replay before posting the prev example. Tho, you can add syntax highlighting to your example code like in the example

bluebill1049:
@Moshyfawn would you like to have a review on this PR?

src/controller.tsx
```diff
@@ -1,6 +1,48 @@
 import { ControllerProps, FieldPath, FieldValues } from './types';
 import { useController } from './useController';
 
+/**
+ * Component based on `useController` hook to work with controlled component.
+ *
+ * @remarks
+ * [API](https://react-hook-form.com/api/usecontroller/controller) • [Demo](https://codesandbox.io/s/react-hook-form-v6-controller-ts-jwyzw) • [Video](https://www.youtube.com/watch?v=N2UNk_UCVyA)
```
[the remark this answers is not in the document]
bluebill1049: It's actually v7 :( the URL link is fixed with v6. I will update the dep. Thanks 👍

src/useFormContext.tsx
```diff
@@ -5,11 +5,66 @@ import { FieldValues, FormProviderProps, UseFormReturn } from './types';
...
+ * }
+ * ```
+ */
 export const useFormContext = <
   TFieldValues extends FieldValues,
 >(): UseFormReturn<TFieldValues> =>
   React.useContext(HookFormContext) as unknown as UseFormReturn<TFieldValues>;
 
+/**
+ * Form provider to host all useForm methods.
```
Moshyfawn, line 44: What do you think about "A provider component that propagates the `useForm` methods to all children components via React Context.". Maybe not exactly like that, but something with "Context" in there to emphasize the use of React Context and that the form methods are accessible in the children components via `useFormContext`.

src/useFormContext.tsx
```diff
@@ -5,11 +5,66 @@ import { FieldValues, FormProviderProps, UseFormReturn } from './types';
...
+ * @param props -  all useFrom methods
+ *
+ * @example
+ * ```tsx
+ * function App() {
+ *   const methods = useForm();
+ *   const onSubmit = data => console.log(data);
+ *
+ *   return (
+ *     <FormProvider {...methods} >
```
Moshyfawn, line 58: Same as the `FormProvider` question earlier: what do you think about including the `useFormContext` call in the example to show the form methods retrieval along the use of `FormProvider`? Not to focus on the hook, but just to hint on how to retive the Context values. Maybe a mention of `useFormContext` in the description with a link on [`useFormContext`](https://react-hook-form.com/api/useformcontext)?

src/useFormContext.tsx
bluebill1049: Agree.

src/useFormContext.tsx
bluebill1049: 👍

src/types/controller.ts
```diff
@@ -45,6 +45,26 @@ export type UseControllerProps<
   control?: Control<TFieldValues>;
 };
 
+/**
```
Moshyfawn, line 48: I'm not sure you need to describe the types, tbh. Everything I've seen, it's methods (functions) with their arguments that are described with TSDoc

src/useController.ts
```diff
@@ -18,6 +18,44 @@ import { useFormContext } from './useFormContext';
...
 
+/**
+ * Custom hook to work with controlled component, this function provide you with both form and field level state. Re-render is isolated at the hook level.
+ *
+ * @remarks
+ * [API](https://react-hook-form.com/api/usecontroller) • [Demo](https://codesandbox.io/s/usecontroller-0o8px)
+ *
+ * @param props - the path name to the form field value, and validation rules.
+ *
+ * @returns provide field handler functions, field and form state. {@link UseControllerReturn}
```
Moshyfawn, line 29: Maybe omit the "provide" word, as you're using `@returns` tag? Like, instead of starting a new sentence, you can continue 
> `@returns` field properties, it's event handlers and the form state

src/useController.ts
```diff
@@ -18,6 +18,44 @@ import { useFormContext } from './useFormContext';
...
+ *   return (
+ *     <div>
+ *       <input {...field} placeholder={props.name} />
+ *       <p>{fieldState.isTouched && "Touched"}</p>
+ *       <p>{formState.isSubmitted ? "submitted" : ""}</p>
+ *     </div>
+ *   );
+ * }
+ *
+ *  export default function App() {
```
Moshyfawn, line 44: Does the dev need the whole "working" example? What do you think about including just a bit with the most relevant pieces of the example, i.e. removing the App component? Lengthy tool-tips can also lead to devs just skipping mist of the info, 'cuz they need to scroll a lot.

src/useForm.ts
```diff
@@ -13,6 +13,35 @@ import {
 } from './types';
 import { useSubscribe } from './useSubscribe';
 
+/**
+ * useForm is a custom hook to mange the entire form by invoke the hook it returns functions.
```
Moshyfawn, line 17: Idk if you need to specify that `useForm` returns functions as it's basically a given. Maybe doesn't even need a description as it's a self-explanatory core hook that contains all the form management methods.

src/useForm.ts
```diff
@@ -13,6 +13,35 @@ import {
...
+ *
+ * @remarks
+ * [API](https://react-hook-form.com/api/useform) • [Demo](https://codesandbox.io/s/react-hook-form-get-started-ts-5ksmm) • [Video](https://www.youtube.com/watch?v=RkXv4AXXC_4)
+ *
+ * @param props - useForm props
```
Moshyfawn, line 22: Maybe something a bit more descriptive like.. form configuration and validation parameters?

src/useFormState.ts
```diff
@@ -12,6 +12,36 @@ import {
 import { useFormContext } from './useFormContext';
 import { useSubscribe } from './useSubscribe';
 
+/**
+ * This custom hook allows you to subscribe to each form state, and isolate the re-render at the custom hook level. It has its scope in terms of form state subscription, so it would not affect other useFormState and useForm. Using this hook can reduce the re-render impact on large and complex form application.
+ *
+ * @remarks
+ * [API](https://react-hook-form.com/api/useformstate) • [Demo](https://codesandbox.io/s/useformstate-75xly)
+ *
+ * @param props - include options on specify fields to subscribe. {@link UseFormStateReturn}
```
Moshyfawn, line 21: It looks like you're describing the `name` param of the `useFormState` but appending the description to the `props` param, which is a bit misleading _(I actually didn't know about the param, lol. Looks super usefull)_.
How do you feel about providing a nested object description for some of the args instead? Like
```ts
  /**
  * @param props.name optionally specify the form states to subscribe to
  */
```

src/useWatch.ts
```diff
@@ -27,6 +48,26 @@ export function useWatch<
   disabled?: boolean;
   exact?: boolean;
 }): UnpackNestedValue<DeepPartialSkipArrayKey<TFieldValues>>;
+/**
+ * Behaves similarly to the watch API, however, this will isolate re-rendering at the component level and potentially result in better performance for your application.
```
Moshyfawn, line 52: I feel like a reference to some other method's API in a quick description snippet isn't very useful as it would make me remember / look up the other API to understand what's this one's about. Maybe we can think of a short descriptive sentence or two that wouldn't duplicate the `watch` API description, but will have the same meaning?..

src/useWatch.ts
```diff
@@ -47,10 +111,39 @@ export function useWatch<
...
+ *
+ * @remarks
+ *
+ * [API](https://react-hook-form.com/api/usewatch) • [Demo](https://codesandbox.io/s/react-hook-form-v7-ts-usewatch-h9i5e)
+ *
+ * @example
+ * ```tsx
+ * // when the component is wrapped with FormProvider
```
Moshyfawn, line 123: This description might confuse someone unfamiliar / new to RHF as to why `useWatch` when the component is wrapped with FormProvider. I understand that ppl should read the docs first. But maybe we can help too by saying smth like
> can skip passing down the `control` if the form is wrapped with the FormProvider

Smth that states a bit clearer that the the `control` is not being passed in this example, 'cuz someone might not know there can be a `control` passed at all :P

src/useFieldArray.ts
```diff
@@ -30,13 +30,50 @@ import {
 import { useFormContext } from './useFormContext';
 import { useSubscribe } from './useSubscribe';
 
-export const useFieldArray = <
+/**
+ * Custom hook for managing Field Arrays (dynamic inputs). The motivation is to provide better user experience and form performance.
```
Moshyfawn, line 34: What do you think about instead of providing a motivation for the hook, describing what it does, like:
> A custom hook that exposes convenient methods to perform operations with a list of dynamic inputs that need to be appended, updated, removed etc.

src/types/controller.ts
```diff
@@ -45,6 +45,26 @@ export type UseControllerProps<
...
+ * These props will provide the control for the field.
+ *
+ * @returns all the event handler, and relevant field and form state.
+ *
+ * @example
+ * ```tsx
+ * const { field, fieldState, formState } = useController();
+ *
+ * console.log(formState.isDirty);
+ * console.log(fieldState.error);
```
Moshyfawn, line 58: Maybe `console.log` isn't the best representation of the controller use?
What do you think about enchaining the form to include these properties instead?
```js
const { field, fieldState } = useController()

return (
  <div>
    <input 
      onChange={field.onChange}
      onBlur={field.onBlur}
      name={field.name}
      ref={field.ref} // optional for focus management
   />
   {!!fieldState.error && <p>{fieldState.error.message}</p>}
  </div>
)
```
`formState` might not be needed here as it's about the controlled input case. Maybe it's better to properly describe it with it's use-case example at the `useFormState`

src/useFormState.ts
bluebill1049: `@param props.name` is invalid for TSDoc, because the type definition should do that.

bluebill1049:
Thanks, @Moshyfawn all feedback above has been addressed.

src/useFormState.ts
Moshyfawn: Defining the types in description makes my TS say
> JSDoc types may be moved to TypeScript types.ts(80004) 
```ts
/**
 * @param {{ a: string, b: number }} props the **foo**, not the **bar**
*/
```
image

But accessing the nested values to describe the arg works fine
```ts
/**
 * @param props.foo the **foo**, not the **bar**
 * @param props.bar the **bar**, definitely not the **bar**
 * @returns props
 */
```
image

bluebill1049: Screen Shot 2022-02-02 at 9 33 17 pm

> tsdoc-param-tag-with-invalid-name

Moshyfawn: Fair enough!

Moshyfawn (approved):

src/useFormState.ts
bluebill1049: Think twice on this during the shower, I think it makes sense to remove all those @param, as its server has very little value rather than just noise. type definition should be the real helper here on the developer experience. Thanks for the detailed review 👍

bluebill1049:
Thank you very much @Moshyfawn for the review! 🙏

src/useFormState.ts
Moshyfawn: Ah. the place where everything is debugable and solvable - the shower xD
You might be right. Tho, I quite often describe in words the nested args properties in my projects by default, it's definitely a clutter the more properties you have

bluebill1049: lol agree with you, sorry i have deleted my message.

PR #7697, opened 2022-01-31
save byte on parseFloat

bluebill1049:


src/logic/validateField.ts
```diff
@@ -118,8 +118,7 @@ export default async <T extends NativeFieldValue>(
 
     if (!isNaN(inputValue as number)) {
```
klarstrup, line 119: could save an internal cast in isNaN by moving the below cast to above this clause

bluebill1049: would you like to send a PR?

PR #7698, opened 2022-01-31
🚚 fix reset test route typo

Moshyfawn:
I was trying to manually check out the reset functionality and couldn't see the reset route going from the list of routes via the link.


bluebill1049 (approved):
Thank you!

PR #7704, opened 2022-02-01
fix: don't expose internal state in handleSubmit

### LINKED ISSUE #7703: issue: [handleSubmit] exposes internal state references for nested values and errors
### Version Number

`>= 7.13.0`

### Codesandbox/Expo snack

https://codesandbox.io/s/priceless-cookies-kynvc

### Steps to reproduce

**Trigger read-only error**
1. Go to https://codesandbox.io/s/priceless-cookies-kynvc
2. Click on submit button
3. Change value in the field
4. The application raises an error, because we tried to set the value on a read-only object.

**Change internal form state**
1. Go to https://codesandbox.io/s/priceless-cookies-kynvc
2. Click on submit button
3. See the console output - it prints the input value
4. Click on submit button again
5. See the log output. It now prints the changes to the data object of the previous `handleSubmit` call

### Expected behaviour

Changes on nested values of the data argument in `handleSubmit` should not change the internal state of `control._formValues`

I think this is the cause of this already reported issue: #7155

Let me explain, what causes this error in our workflow:

We are using `@reduxjs/toolkit`, with default middlewares.

When the form is submitted, we dispatch an action with the data. The payload of this action is frozen somewhere along the way. Which results in a frozen object in the internal `_formValues` state.

While our specific problem is caused by a third party library this is also reproducable without it (see sandbox), though I guess it's not that likely to happen.

Firefox, Chrome

### Relevant log output

```shell
-- Chrome:
index.esm.mjs:475 Uncaught (in promise) TypeError: Cannot assign to read only property 'email' of object '#<Object>'
  at set (index.esm.mjs:475:1)
    at _callee7$ (index.esm.mjs:2152:1)
    at tryCatch (runtime.js:63:1)
    at Generator.invoke [as _invoke] (runtime.js:294:1)
    at Generator.next (runtime.js:119:1)
    at asyncGeneratorStep (index.esm.mjs:12:1)
    at _next (index.esm.mjs:14:1)
    at index.esm.mjs:14:1
    at new Promise (<anonymous>)
    at index.esm.mjs:14:1

-- Firefox:
 Uncaught (in promise) TypeError: "email" is read-only
    set index.esm.mjs:475
    _callee7$ index.esm.mjs:2152
    tryCatch runtime.js:63
    invoke runtime.js:294
    defineIteratorMethods runtime.js:119
    asyncGeneratorStep index.esm.mjs:12
    _next index.esm.mjs:14
    _asyncToGenerator index.esm.mjs:14
    _asyncToGenerator index.esm.mjs:14
    onChange index.esm.mjs:2229
    React 14
    unstable_runWithPriority scheduler.development.js:470
    React 4
```

**The same problem exists for errors in `onInvalid`, but the `cloneObject` method can't currently handle the circular references within react elements.** 

### Code of Conduct

tiii:
Fixes #7703.

Create a deep clone of `formValues` before passing them on as the `data` argument to `handleSubmit`. Otherwise nested objects directly reference the internal `_formValues` state.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1010,9 +1010,7 @@ export function createFormControl<
         e.persist && e.persist();
       }
       let hasNoPromiseError = true;
-      let fieldValues: any = _options.shouldUnregister
-        ? cloneObject(_formValues)
-        : { ..._formValues };
+      let fieldValues: any = cloneObject(_formValues);
 
       _subjects.state.next({
         isSubmitting: true,
--- a/src/__tests__/useForm/handleSubmit.test.tsx
+++ b/src/__tests__/useForm/handleSubmit.test.tsx
@@ -87,6 +87,38 @@ describe('handleSubmit', () => {
     });
   });
 
+  it('should not provide reference to _formValues as data', async () => {
+    const { result } = renderHook(() =>
+      useForm<{ test: string; deep: { values: string } }>({
+        mode: VALIDATION_MODE.onSubmit,
+        defaultValues: {
+          test: 'data',
+          deep: {
+            values: '5',
+          },
+        },
+      }),
+    );
+
+    await act(async () => {
+      await result.current.handleSubmit((data: any) => {
+        data.deep.values = '12';
+      })({
+        preventDefault: () => {},
+        persist: () => {},
+      } as React.SyntheticEvent);
+    });
+
+    await act(async () => {
+      await result.current.handleSubmit((data: any) => {
+        expect(data.deep).toEqual({ values: '5' });
+      })({
+        preventDefault: () => {},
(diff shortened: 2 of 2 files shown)


src/logic/createFormControl.ts
```diff
@@ -1035,9 +1035,9 @@ export function createFormControl<
             errors: {} as FieldErrors<TFieldValues>,
             isSubmitting: true,
           });
-          await onValid(fieldValues, e);
+          await onValid(cloneObject(fieldValues), e);
```
tiii, line 1038: Maybe we should just line 1013 to `let fieldValues: any = cloneObject(_formValues)`?

bluebill1049:
thanks for the PR.

PR #7718, opened 2022-02-02
Don't expose internal errors reference in handleSubmit

### LINKED ISSUE #7717: issue: [handleSubmit] exposes internal errors references in `onInvalid` callback
### Version Number

>= 7.13.0

### Codesandbox/Expo snack

https://codesandbox.io/s/hardcore-firefly-598rs

### Steps to reproduce

1. Go to https://codesandbox.io/s/hardcore-firefly-598rs
2. Click on submit button
2. Click on set custom error
4. See read-only error

### Expected behaviour

Freezing errors in `onInvalid` callback should not freeze internal errors object.

Firefox

### Relevant log output

```
-- Firefox:
 Uncaught (in promise) TypeError: "email" is read-only
    set index.esm.mjs:475
```

### Code of Conduct

tiii:
Fixes #7717

`_formState.errors` are cloned before passing them on to `onInvalid` handler.

This requires a change in the `cloneObject` method:

On the web context we have lots of references on a `HTMLElement` which would already slow the method down. On top of that, a `HTMLInputElement` has a reference to the form, which itself references it's inputs again, causing an infinite loop.

Because of that we now ignore `HTMLElement`s on the web and accordingly `Component`s when using `react-native` and instead pass those on without cloning.

### DIFF
--- a/src/__tests__/useForm/handleSubmit.test.tsx
+++ b/src/__tests__/useForm/handleSubmit.test.tsx
@@ -395,6 +395,33 @@ describe('handleSubmit', () => {
     });
   });
 
+  it('should not provide internal errors reference to onInvalid callback', async () => {
+    const { result } = renderHook(() =>
+      useForm<{
+        test: string;
+      }>(),
+    );
+    result.current.register('test', { required: true });
+
+    await act(async () => {
+      await result.current.handleSubmit(
+        () => {},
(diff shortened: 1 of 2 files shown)


bluebill1049:
thank you again for your contribution.

bluebill1049 (approved):

PR #7755, opened 2022-02-05
Fix IntelliSense

felixschorer:
- Reorders overloads to fix IntelliSense for
  - `useWatch`
    - make `useWatch({ name: ['foo'] })` the default overload
    - deprecate the `useWatch({ name: 'foo' })` overload
    - fix typing of the `defaultValue` prop when `name` is a tuple.
  - TODO
- Adds `completionsFactory` fixture to get IntelliSense completions for a TS snippet. Imports within the snippet are relative to the passed directory.
- Adds tests to test IntelliSense completions. The tests are regular Jest test cases.


src/__tests__/__fixtures__/completions.ts
```diff
@@ -0,0 +1,86 @@
...
+      completionArgs.push({
+        position: strings[i].length + completionArgs[i - 1].position,
+        options: options[i],
+      });
+    }
+  }
+  return completionArgs;
+}
+
+export function completionsFactory(dirname: string) {
```
bluebill1049, line 34: Interesting tests.

src/__tests__/__fixtures__/completions.ts
```diff
@@ -0,0 +1,86 @@
...
+    },
+    getCurrentDirectory: () => dirname,
+    getCompilationSettings: () => ({}),
+    getDefaultLibFileName: (options) => ts.getDefaultLibFilePath(options),
+    fileExists: ts.sys.fileExists,
+    readFile: ts.sys.readFile,
+    readDirectory: ts.sys.readDirectory,
+    directoryExists: ts.sys.directoryExists,
+    getDirectories: ts.sys.getDirectories,
+  };
```
bluebill1049, line 61: Is there any good reference for those methods?

bluebill1049:
looks like the test failed as well :(

src/__tests__/__fixtures__/completions.ts
felixschorer: Not that I know of. I took that straight from https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#incremental-build-support-using-the-language-services

src/__tests__/__fixtures__/completions.ts
felixschorer: I am making use of template tag functions to mark the positions where I want to get IntelliSense completions. Imagine `${undefined}` to be the position of your cursor. 

Instead of `undefined` you can also pass an object with options for that TS language service method (`getCompletionsAtPosition`).

felixschorer:
Sadly, for `useWatch` only the first overload signature will get IntelliSense completions. That's why that test fails. If `name` is a tuple there won't be an completions since that's not the first overload. If you move that overload to the top, the other test will fail.

bluebill1049:
I see @felixschorer worst case we will just have to treat it as limtaiton then :(

src/__tests__/__fixtures__/completions.ts
felixschorer: Added documentation for the `completionsFactory` function.

src/__tests__/__fixtures__/completions.ts
felixschorer: Added a reference to that wiki article in the code.

bluebill1049:
oops looks like I mess up the build, will fix it later. are we deprecating useWatch(string) only for IntelliSense?

felixschorer:
@bluebill1049 That's correct. It's your decision though.

bluebill1049:
thanks, @felixschorer let me think about it.

src/useWatch.ts
```diff
@@ -33,50 +33,56 @@ import { useSubscribe } from './useSubscribe';
...
  * })
  * ```
  */
-export function useWatch<TFieldValues extends FieldValues>(props: {
-  defaultValue?: DeepPartialSkipArrayKey<TFieldValues>;
+export function useWatch<
+  TFieldValues extends FieldValues,
+  TFieldNames extends readonly PathString[],
+>(props: {
+  name: readonly [...Auto.FieldPaths<TFieldValues, TFieldNames>];
```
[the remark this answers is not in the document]
bluebill1049: i can fix that in a separate branch 👍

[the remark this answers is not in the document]
bluebill1049: hmmm i am not too sure on this, some tests works but not others.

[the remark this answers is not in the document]
bluebill1049: I will rever this PR first, sorry to merge ahead early on...

PR #7756, opened 2022-02-06
V8: remove internal native reset from reset API

bluebill1049:
FYI @Moshyfawn The `reset` behavior which we discussed before.


Moshyfawn:
Thx for the heads up!

PR #7757, opened 2022-02-06
V8: include readonly for handleSubmit payload

bluebill1049:
FYI: @tiii


bluebill1049:
will fix the test at v8 root

PR #7759, opened 2022-02-06
close #7758 useController should subscribe to exact field name

### LINKED ISSUE #7758: issue: Nested controlled input doesn't re-render when using setValue
### Version Number

7.26.0

### Codesandbox/Expo snack

https://codesandbox.io/s/react-hook-form-object-controlled-not-updated-qx4e2?file=/src/index.js

### Steps to reproduce

1. Go to the codesandbox link
2. Click on 'Update' 
3. The "Object controlled field" nested value will not re-render in the UI (highlighted in red) although the watched value of that same field shows the proper value is being set 
4. Click "Update object controlled" and the nested value will re-render properly when using `setValue` on the nested field directly

### Expected behaviour

Calling `setValue` should re-render the nested controlled field correctly.
I'm not sure if this is expected behavior or not but it seems counter-intuitive as updating a nested controlled field works if that field is not an object.  Note that I also added nested controlled value inside an array and it updates properly as well.

Firefox, Chrome, Safari, Edge

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:


Moshyfawn (approved):

bluebill1049:
thanks for the review @Moshyfawn 🙏

Issue #7661, opened 2022-01-25
issue: handle global error in forms


alikazemkhanloo:
**Is your feature request related to a problem? Please describe.**
when I get response after submitting a form, I get two types of errors: a) field related errors, which can be handled by `setError` b)  some global errors, like user not found. there is no place for this kind of errors in `react-hook-form`

**Describe the solution you'd like**
there can be a field starting with dangle (_) to used as global error for the form.

**Describe alternatives you've considered**
Now I use a custom field named `_error`, but typescript keeps complaining about it, as there is no `_error` field in my inputs.
another workaround was to use a local state.

**Additional context**
in `redux-form` we could use `_error` field for global errors. and it's handled nicely providing a prop named `error`.

bluebill1049:
thanks, pinging to the project board.

alikazemkhanloo:
I'm willing to contribute to this specific issue, but the question is how we should handle it. My solution is to add _error as a reserved field for the global error.

Moshyfawn:
> I'm willing to contribute to this specific issue, but the question is how we should handle it. My solution is to add _error as a reserved field for the global error.

Any feature request goes through community up-voting the feature to see if the feature is in demand before it goes into development

julio3266:
@Moshyfawn  I understand that there are limitations, but since I have a separate form on another useForm screen, and it generates multiple arrays, would it be possible to generate an error for each different entity?

bluebill1049:
Idea: 

cc @Moshyfawn 

related: https://github.com/react-hook-form/react-hook-form/issues/7278

```tsx
handleSubmit((data) => {
  const result = await request();

  return {
    values: data, // will return by handleSubmit
    errors: {
      dataMismatch: {
        type: 'server',
        message: 'whatever',
      }
    } // will update errors object and to be persisted and only `clearErrors` and `handleSubmit` method can remove those error
  }
})
```

Moshyfawn:
Hm.. what about the fact that it's pretty common to do
```js
handleSubmit(data => updateEntry(data).then(res => success(res)).catch(err => fail(err)))
```

Meaning, if you use `.then().catch()`, the API can return an entry `object`, which is gonna be `return`ed

bluebill1049:
thanks @Moshyfawn correct above solution is not valid then.

bluebill1049:
Actually we don't return anything at the moment, only catch part is valid.

Moshyfawn:
[...]

PR #7792, opened 2022-02-10
feat: add validation error prop in FieldError

PupoSDC:
This follows up a discussion [here](https://github.com/react-hook-form/react-hook-form/discussions/3808#discussioncomment-261851).

When dealing with validation libraries like `yup` and `zod` it is convenient to have the possibility to be able to access the original error thrown by the validation library outside the context of react hook form. 

This tiny change would allow a standardised param where we can retrieve the original error from the validation library we are using.

This is the only change that is needed in this package, and it's backwards compatible. We can then add the original error in the resolver packages, for example [here for Zod](https://github.com/react-hook-form/resolvers/blob/master/zod/src/zod.ts#L29).

### DIFF
--- a/src/types/errors.ts
+++ b/src/types/errors.ts
@@ -10,12 +10,12 @@ export type MultipleFieldErrors = {
   [key: string]: ValidateResult;
 };
 
-export type FieldError = {
+export type FieldError<T = unknown> = {
   type: LiteralUnion<keyof RegisterOptions, string>;
   ref?: Ref;
   types?: MultipleFieldErrors;
   message?: Message;
-  validationError?: unknown;
+  validationError?: T;
 };
 
 export type ErrorOption = {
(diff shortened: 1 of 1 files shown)


bluebill1049:
Thanks for the PR, please upvote this feature request:  https://github.com/react-hook-form/react-hook-form/issues/7661 we can then figure out something at the global level for both build-in and schema validation.

PupoSDC:
@bluebill1049 Thanks for the heads up. I went over that ticket, but it seems to be adressing a different problem. I created a different issue: https://github.com/react-hook-form/react-hook-form/issues/7797

(FWIW, i did check the suggestions board, but I could only see an empty board, so i thought that flow had been abandoned. Github was a bit goofy yesterday).

PupoSDC:
Closing as the idea was not accepted by the maintainers: https://github.com/react-hook-form/react-hook-form/issues/7797

Issue #7730, opened 2022-02-03
Yup validation transforms date string to datetime


chrisbag:
### Version Number

V7

### Codesandbox/Expo snack

https://codesandbox.io/s/react-hook-form-v7-yup-and-dates-p80hu

### Steps to reproduce

Hello,
First of all thanks for this wonderful library and the new V7.
I have been experiencing an unexpected behavior when using react-hook-form with yup for date validation. This also happens when using v6 I believe.

1. Go to sandbox
2. Choose a date
3. Click submit and check console
4. Date field value in watch is a string of type yyyy-mm-dd
5. However data from on submit is of type Date (basically a timestamp)

### Expected behaviour

I would expect form data not not to be transformed by yup resolver, meaning formState = data passed in on submit. I assume what's happening is that Yup parse the date string into a Date Object and that is the element returned

### Relevant log output

```
submited data =  Thu Feb 03 2022 00:00:00 GMT+0100 (heure normale d’Europe centrale)
form Data =  2022-02-03 
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bluebill1049:
native input doesn't return `date` type instead of `string`.

Moshyfawn:
Huh, if you, say,[ use zod with it's RHF resolver](https://codesandbox.io/s/react-hook-form-v7-yup-and-dates-forked-nhd1h?file=/src/index.js), you'd get `Expected date, received string` on submit validation. But in the OP's case, yup silently transforms the date string into date and successfully submits it. I'd think, yup should also report an error instead of transforming the value according to the set type

bluebill1049:
oh, I miss the part of the data transform in the `resolver`, we just return what's returned by the schema validation library. we don't need any parsing/transform on the data.

Moshyfawn:
So, the question is then": "does Yup parse the value and return it parsed?"

bluebill1049:
> So, the question is then": "does Yup parse the value and return it parsed?"

I think it doesn't then from the example showed above.

bluebill1049:
I just did a quick reading, i think in `yup`, you will need to use `.transform(parseDateString)` API

Moshyfawn:
If you return a non-date type in `.transform()`, Yup will yell the "Incorrect type" error instead. See [this CSB](https://codesandbox.io/s/react-hook-form-v7-yup-and-dates-forked-ljr6z?file=/src/index.js)

[...]

PR #7795, opened 2022-02-10
Fix output types of transformed values onSubmit

christopherklint97:
## Summary

This fixes a typing/validation issue when using a resolver that transforms the form input types into different output types.

Example using zod:
```typescript

import { z } from "zod";
import { zodResolver } from "@hookform/resolvers/zod";

const countSchema = z.object({
  count: z.string().transform(val => Number(val)),
});

type CountForm = z.input<typeof countSchema>; // input type is inferred => string
type CountSchema = z.infer<typeof countSchema>; // output type is inferred => number

  const { register, handleSubmit } = useForm<CountForm>({resolver: zodResolver(countSchema)});
  const onSubmit = (data: CountSchema) => console.log(data);

  return (
    <div>
      ...
      // Error: Type '{ count: string; }' is not assignable to type '{ count: number; }'.
        <input {...register("count")} placeholder="Count" />
        <input type="submit" />
      </form>
    </div>
  );
}
```
Error was also reproduced here: https://codesandbox.io/s/ecstatic-pine-brxt1?file=/src/App.tsx:0-892

### Version

7.26.1

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -550,7 +550,7 @@ export type UseFormGetValues<TFieldValues extends FieldValues> = {
 };
 
 // @public
-export type UseFormHandleSubmit<TFieldValues extends FieldValues> = <_>(onValid: SubmitHandler<TFieldValues>, onInvalid?: SubmitErrorHandler<TFieldValues>) => (e?: React_2.BaseSyntheticEvent) => Promise<void>;
+export type UseFormHandleSubmit<TFieldValues extends FieldValues> = <TSubmitFieldValues = TFieldValues>(onValid: SubmitHandler<TSubmitFieldValues>, onInvalid?: SubmitErrorHandler<TFieldValues>) => (e?: React_2.BaseSyntheticEvent) => Promise<void>;
 
 // @public (undocumented)
 export type UseFormProps<TFieldValues extends FieldValues = FieldValues, TContext extends object = object> = Partial<{
--- a/src/__tests__/useForm/handleSubmit.test.tsx
+++ b/src/__tests__/useForm/handleSubmit.test.tsx
@@ -349,6 +349,34 @@ describe('handleSubmit', () => {
       });
       expect(callback.mock.calls[0][0]).toEqual({ test: 'test' });
     });
+
+    it('should transform input values and return correct output types', async () => {
+      const resolver = async () => {
+        return {
+          values: { count: 0 },
+          errors: {},
+        };
+      };
+
+      const { result } = renderHook(() =>
+        useForm<{ count: string }>({
+          mode: VALIDATION_MODE.onSubmit,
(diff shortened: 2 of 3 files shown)


src/__tests__/useForm/handleSubmit.test.tsx
```diff
@@ -349,6 +349,34 @@ describe('handleSubmit', () => {
       });
       expect(callback.mock.calls[0][0]).toEqual({ test: 'test' });
     });
+
+    it('should transform input values and return correct output types', async () => {
```
bluebill1049, line 353: Thanks for the PR. I don't think this test does cover the type fix as it passes today's codebase.

reports/api-extractor.md
```diff
@@ -550,7 +550,7 @@ export type UseFormGetValues<TFieldValues extends FieldValues> = {
 };
 
 // @public
-export type UseFormHandleSubmit<TFieldValues extends FieldValues> = <_>(onValid: SubmitHandler<TFieldValues>, onInvalid?: SubmitErrorHandler<TFieldValues>) => (e?: React_2.BaseSyntheticEvent) => Promise<void>;
```
bluebill1049, line 553: `<_>` was left there to avoid breaking change.

src/__tests__/useForm/handleSubmit.test.tsx
christopherklint97: Thank you for your comments! You were right about that, I change the test so it should be failing now.

reports/api-extractor.md
christopherklint97: Ahh ok, then I wonder if this is the intentional behavior of react-hook-forms, that the output values of the resolver should be the `TFieldValues` and not the input values if they are transformed? 
For example:
```typescript
import { z } from "zod";
import { zodResolver } from "@hookform/resolvers/zod";

const countSchema = z.object({
  count: z.string().transform(val => Number(val)),
});

type CountSchema = z.infer<typeof countSchema>; // output type is inferred => number

  const { register, handleSubmit } = useForm<CountSchema>({resolver: zodResolver(countSchema)});
  const onSubmit = (data: CountSchema) => console.log(data);
  ...
```

Been trying to think of other ways to transform the typing but come up short 🤷🏾

Moshyfawn:
Does #7730 help?

christopherklint97:
> Does #7730 help?

Yes it does help definitely, thanks! I think I misunderstood the problem and thought that the resolvers should notify react-hook-form somehow about the types of the return values.

I guess it is more of an issue with zod's paradigm then, that values can be transformed to new types after being parsed.

Really unsure though how to type the `useForm` and `data` object from `handleSubmit` since the input values are the only things being considered from either the `defaultValues` or the generic type in `useForm<T>` 🤔

PR #7805, opened 2022-02-11
errors: include received and expected (build-in validation)

### LINKED ISSUE #3950: Include information about failing validation constrain in formState.errors.
I have a helper function that takes the `formState.errors[fieldName]` as parameter and uses it's `type` field to constructs an error message. For the failing minLength/maxLength validations I'd like to indicate the desired length in the error message.
````javascript
function formErrorMessages(fieldError) {
  if (fieldError.message) return fieldError.message;

  switch (fieldError.type) {
    case ERROR_TYPES.required:
      return 'Field is requried';
    case ERROR_TYPES.minLength:
      return 'Value is too short';   // <-- I want `Value must be at least {x} characters long`.
    default:
      return 'Invalid';
  }
}
````

`formState.errors[fieldName]` to include optional payload field, similar to how angular [does](https://angular.io/api/forms/Validators#minLength-usage-notes) it:
````javascript
{
  message,
  type: "minLength",
  ref,
  minLength: {  // <-- this key to match the type, as there might be multiple errors.
    required,
    actual
  }
}
````

Pass the required length as error message in register: `register({ minLength: { value: 2, message: 2 } })`. But it's it's a duplication, and `2` is not really a message. Could also provide the whole error message here, but I still need to repead the `2` inside the error message, and my helper function is useless in this case.
Or I could pass it as optional 2nd parameter for my `formErrorMessages` function, but again, its a duplication, that would require extracting the `2` into variable to make sure validation and error message always match when in future someone only updates the validation rule and not the message.

**Additional context**
-

### PR DESCRIPTION
**Pull request by bluebill1049**
```diff
export type FieldError = {
  type: LiteralUnion<keyof RegisterOptions, string>;
  ref?: Ref;
  types?: MultipleFieldErrors;
  message?: Message;
+ received?: unknown;
+ expected?: unknown;
};
```

close [#3950](https://github.com/react-hook-form/react-hook-form/issues/3950)

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -189,6 +189,8 @@ export type FieldError = {
     ref?: Ref;
     types?: MultipleFieldErrors;
     message?: Message;
+    received?: unknown;
+    expected?: unknown;
 };
 
 // @public (undocumented)
--- a/src/__tests__/logic/__snapshots__/validateField.test.tsx.snap
+++ b/src/__tests__/logic/__snapshots__/validateField.test.tsx.snap
@@ -3,6 +3,7 @@
 exports[`validateField should handle pattern with g flag 1`] = `
 Object {
   "test": Object {
+    "expected": 10,
     "message": "",
     "received": "a",
     "ref": Object {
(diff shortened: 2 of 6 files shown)


bradlocking:
This would have been a lovely addition. Was there a reason why it was closed?

PR #7836, opened 2022-02-16
fix(typo): remove duplicated 'the' from messages.default's comment

t-i-0414:
This PR is for fix json's comment. so it doesn't affect app's function.


bluebill1049:
thanks.

PR #7859, opened 2022-02-19
🐞 fix #7849 issue with unset on multiple dimension array

### LINKED ISSUE #7849: issue: value of double depth arrays lost when one entry is unregistered
### Version Number

7.27.0

### Codesandbox/Expo snack

https://codesandbox.io/s/funny-edison-sqdvk9

### Steps to reproduce

1. Enter something in name 1, observe that the dumped value changes
2. Check the checkbox. This enables another field.
3. Enter something in name 2, observe that the dumped value changes
4. Uncheck the checkbox. This unregisters the new field.
5. Observe that the dumped value is empty, even though the first field still contains a value.

### Expected behaviour

After unregistering `items.1.0.name`, the value for `items.0.0.name` should still be in the watched value when watching `items`.

When I use a single-depth array this works fine (so `items.0.name` instead of `items.0.0.name`).

Chrome

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
fix #7849 on unset array of array issue

### DIFF
--- a/src/__tests__/utils/unset.test.ts
+++ b/src/__tests__/utils/unset.test.ts
@@ -1,4 +1,3 @@
-// @ts-nocheck
 import unset from '../../utils/unset';
 
 describe('unset', () => {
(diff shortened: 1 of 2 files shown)


jorrit:
Thanks! You're the best!

PR #7901, opened 2022-02-23
enable deps with string syntax

bluebill1049:


Moshyfawn:
Are we moving ahead with `useWatch` array only field names? If so, does it make sense to allow a string for `deps`?

bluebill1049:
oh yes @Moshyfawn for `watch` and `useWatch`, this is for `trigger`. I get your concern for inconsistency here tho.

Moshyfawn (approved):

bluebill1049:
thanks for the review @Moshyfawn 👍

PR #7923, opened 2022-02-26
include issue auto check and closer

bluebill1049:


Moshyfawn:
That's great! One remark is that ppl sometimes prefer to use stackblitz.com instead of codesanbox.io due to its performance. The latter is a bit more resource heavy and might affect how hard it's to create a good repro

Moshyfawn (approved):
Might piss some ppl off, but it's for our own mental well-being ;)

bluebill1049:
thanks for the review! i will merge it in and test.

PR #7943, opened 2022-03-01
fix: compare html elements against correct global

zenoamaro:
When checking whether a value is an instance of HTMLElement,
the reference class must be owned by the same window which
owns the value, or the comparison will fail.


zenoamaro:
This change fixes Radio and Checkbox handling when react-form-hook is running in a Web Extension's content script. 

In such a case, `window` is not the same as the element's `window`, and so `element` is not instanceof `HTMLElement`, which means the element ref is discarded during `register`.

Here is an example of the issue in action:

Screen Shot 2022-03-01 at 13 01 30

PR #7956, opened 2022-03-03
fix: useFieldArray example crash

mfalthaw:
Summary of fixes
1. Null-safe item key
2. Remove unused `useWatch` import
3. Remove unused `watch` destruct

Crash happens when swapping/moving with initial values
Screen Shot 2022-03-03 at 19 30 28

Original sandbox: https://codesandbox.io/s/react-hook-form-usefieldarray-ssugn?file=/src/index.js
Forked sandbox: https://codesandbox.io/s/react-hook-form-usefieldarray-forked-31u8q8
diff: https://www.diffchecker.com/mEwLVofv

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -35,7 +35,7 @@ import { useSubscribe } from './useSubscribe';
  * A custom hook that exposes convenient methods to perform operations with a list of dynamic inputs that need to be appended, updated, removed etc.
  *
  * @remarks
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks @mfalthaw i have updated to the latest version of that csb, could you see if the issue is resolved itself? `id` should always presented this may be a bug related to old version.

mfalthaw:
> thanks @mfalthaw i have updated to the latest version of that csb, could you see if the issue is resolved itself? `id` should always presented this may be a bug related to old version.

I can no longer reproduce the crash. Thanks @bluebill1049!

bluebill1049:
thanks @mfalthaw 🙏

PR #7968, opened 2022-03-04
Prevent error on setFocus for unfocusable inputs

danielweinmann:
If we call `setFocus` on a field that doesn't have a regular input associated with it, it will now fail silently.

That's useful in case we have a generic focus management function that tries to focus fields regardless of their input type.

I'm using react-hook-form under the hood on [Remix Forms](https://remix-forms.seasoned.cc/), and I came across this issue when writing a custom HTML field with it.

### DIFF
--- a/app/src/setFocus.tsx
+++ b/app/src/setFocus.tsx
@@ -7,6 +7,8 @@ type FormValues = {
   selectInputContent: string;
   focusTextarea: string;
   selectTextareaContent: string;
+  focusUnfocusable: number;
+  selectUnfocusable: number;
 };
 
 const defaultValues: DefaultValues<FormValues> = {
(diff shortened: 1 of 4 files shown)


bluebill1049:
Thanks, @danielweinmann  for the PR. The current implementation is intentional to throw an error when input ref is not focusable. 

> I came across this issue when writing a custom HTML field with it.

Could you provide some more details on the above?

danielweinmann:
I just realized I can use a `try..catch` inside Remix Forms and we're good to go :smile: :man_facepalming: 

With that in mind, this becomes a non-issue for me. If you think this will be beneficial for react-hook-form, feel free to merge the PR. Otherwise, feel free to close it.

Thank you so much for the amazing library! It's being a real pleasure to build on top of it! :pray: :muscle:

bluebill1049:
Thanks for your understanding @danielweinmann it's a great PR (solid one including good testing coverage). Let's keep it open for a while, if we encounter a more similar demand or request then we merge it. 👍

PR #8023, opened 2022-03-15
💚 ci(issues): disable closing issues without repro

Moshyfawn:
I've experimented a bit with filtering out the GH labels and such, but couldn't get it to work properly.

Currently, the action causes more confusion then helping us to manage "bad" issues.

I'll spend some time on writing, maybe, a custom action instead of relying on the pre-made actions in the future. Don't have the capacity to support it, I guess :P


bluebill1049:
thanks, @Moshyfawn 🙏 no worries on this.

bluebill1049 (approved):

PR #8064, opened 2022-03-21
Structured clone

edjonesdev:


bluebill1049:
I would love to use `StructuredClone` 🔥 however, the browser support is pretty weak so far. maybe it's something worth considering for the next major.

bluebill1049:
hmmmm, this got me thinking now... we can use structureClone if detect browser support!

ed-jones:
Whoops! I didn't actually mean to create this PR 😂. I've just been playing around with trying to get luxon to work with react-hook-form. I believe the error I am having is due to the shallow clone function not copying the prototype methods on the luxon `Interval` object. It would be awesome to see support for this in a future version.

> hmmmm, this got me thinking now... we can use structureClone if detect browser support! cc @Moshyfawn

I think for use cases like mine this would mean that forms would break on older browsers. A polyfill solution might be the better way to do it for my use case.

bluebill1049:
> I think for use cases like mine this would mean that forms would break on older browsers. A polyfill solution might be the better way to do it for my use case.

you can supply a polyfill at your app-level 👍  hook form can just detect if such API is supported. Luxon and moment the custom object is pretty hard to deal with I would avoid if you can and supply with string instead. This will be problematic with deep equal as well.

ed-jones:
Ok that makes sense. Thanks for the advice! I'll look into treating luxon types as strings in react-hook-form instead.

PR #8127, opened 2022-04-02
Update package.json exports rule to allow importing specific file

smeng9:
When building our code using react-hook-form as an external library, some are shipped using esm and some are using cjs, which causes useFormContext to return null. See https://github.com/react-hook-form/react-hook-form/discussions/3894

In one of the replies https://github.com/react-hook-form/react-hook-form/discussions/3894#discussioncomment-2135181 it mentioned the cause is due to the mixed usage of esm and cjs.

Currently the import esm or cjs module is automatically determined and the system may not be very smart. It always causes some problem.

This pull request changes the package.json to allow directly selecting of module type during import by forcing a filename.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     ".": {
       "import": "./dist/index.esm.mjs",
       "require": "./dist/index.cjs.js"
-    }
+    },
+    "./dist/": "./dist/"
   },
   "scripts": {
     "clean": "rimraf dist",


smeng9:
E.g. allow import of react-hook-form/dist/index.cjs.js possible

bluebill1049:
Thanks for the detailed explanation. so this is caused by the following usage (CJS import ESM)?

`const HookForm = require("react-hook-form/dist/index.esm.mjs");`

smeng9:
Yes, we will have to manually do CJS import ESM to make sure there is only one instance of the package in the compiled code.

bluebill1049:
sorry to pull you into this conversation @evocateur wondering if it's safe (none breaking) to introduce this change into the package.json? (any potential side effects) 🙏  many thanks

evocateur:
@smeng9 The linked roadmap ticket does not seem to indicate this error is being experienced there? Where are you actually experiencing this error? It smells like a Webpack problem, not react-hook-form. If you're using ESM, you have to use `import`, or a bundler that understands it correctly.

@bluebill1049 I am -1 on this change, as it exposes additional public API surface for no perceptible gain. Or, rather, it's "fixing" the issue incorrectly, and if we're truly experiencing this "dual package hazard", we should fix it correctly, instead of hacking it. In my experience, we're almost guaranteed to regret a change like this one in the future, as consumers come to rely on the hacky workaround and become dependent on implementation details.

evocateur (changes requested):
If we're truly hitting the "dual package hazard" with standard Webpack usage, we should fix the problem correctly, not expose implementation details as "public" API.

smeng9:
I will close this PR for now as explained in #8154 as there may be better way to fix the problem mentioned here https://nodejs.org/api/packages.html#approach-1-use-an-es-module-wrapper

bluebill1049:
thanks for the feedback and help @evocateur. @smeng9 I don't think it's an issue for the library to resolve. I will reply and close the other issue as well.

PR #8155, opened 2022-04-09
Make some changes on tests

romain-trotard:
Hello :)

In this PR, I make some changes in tests. 

**Let's explain what I did:**
- Remove the use of `act` where it's not needed. By default `render` and `fireEvent` actions are wrapped with `act`. If there are some asynchronous changes, it's better to use `findBy` , `waitFor` or `waitForElementToBeRemoved` after triggering an action (`click`, `change`, ...)
- I replaced the `await waitFor(() => screen.getBy*)` pattern by `await screen.findBy*`
- I removed the `getBy` / `findBy` which were alone in the code to `expect` that the selected element is visible. Yep it does the same thing but I find it harder to understand that the line is here to check that the element is present  (because of no expect). And there was also some `await waitFor(() => screen.queryByText('Some text.'));`  that will always passed that I changed with `expect(await screen.findByText('Some text')).toBeVisible()`.
- I replaced some `toBeNull` with `not.toBeInTheDocument` because I find `not.toBeInTheDocument` more meaningful.
- There was some `click` and `focus` were directly done on the element. I use instead `fireEvent` which wraps with `act` in order to simulate at best real user experience
- I removed snapshots which hides a lot what we expect. It's easy to accept some changes in the snapshots that are not good.
- Upgrade dependencies: `ts-jest`, `@testing-library/jest-dom` and `@testing-library/react-native` (there are new breaking versions for `@testing-library/react` but they dropped the support for React 17 so it will be needed to upgrade the devDepencies `react` and `react-dom`, maybe best to do it in another PR)

Notes:
Each commit can be read separately.
Most of changes that I have made are things that _Kent C. Dodds_ has recommended in his articles.

**What could be done next?**
- It can be nice to use `@testing-library/user-event` that will give some better API for interaction than `fireEvent`. For example when typing in a field, it will be easier to code it and really simulate that we type each letter one after the other.

```javascript
fireEvent.change(input, { target: { value: 'New value'} });
```

would become 

```javascript
userEvent.type(input, 'New value');
```

- Upgrade to latest breaking version `@testing-library/react`

If you have some questions about my changes or there are some changes that you don't want let me know :)


src/__tests__/useFieldArray.test.tsx
```diff
@@ -2090,7 +2242,7 @@ describe('useFieldArray', () => {
       }[];
     };
 
-    const watchedValue: unknown[] = [];
+    let watchedValue: object = {};
```
bluebill1049, line 2245: maybe worth to keep them as an array of objects for now? There is still value to knowing each render what's been rendered as watch output. (I think a snapshot would be fine if it's getting too big for the output.)

src/__tests__/useFieldArray.test.tsx
```diff
@@ -2222,7 +2389,7 @@ describe('useFieldArray', () => {
       }[];
     };
 
-    const watchedValues: FormValues[] = [];
+    let watchedValues: object = {};
```
bluebill1049, line 2392: similar as above.

src/__tests__/useFieldArray/update.test.tsx
```diff
@@ -260,7 +278,7 @@ describe('update', () => {
         control,
       });
 
-      fieldArrayValues.push(fields);
+      fieldArrayValues = fields;
```
bluebill1049, line 281: maybe there is a better way to do this, but the original idea was to capture each render `fields` similar to `watch` API feedback above.

bluebill1049:
> I removed snapshots which hides a lot what we expect. It's easy to accept some changes in the snapshots that are not good.

This PR is awesome thanks @romain-trotard so much improvement. I am still reviewing and I notice you have removed all the snapshots, which I agree is better to be specific. However, there are some tests which are specifically snapshot those render outputs such as `watch` API on each render. shall we keep them as they do serve a purpose, I generally do find good catches with those snapshots. eg: https://github.com/react-hook-form/react-hook-form/pull/8155/files#diff-c5bbecb45cfec14a514551118b256a2ae527cbff87f39a8973506c0f422b8ed2L293

src/__tests__/useFieldArray.test.tsx
romain-trotard: I answer only to this comment but it's the same for other ones.

Yep I wasn't sure if it will be sufficient to check only a single value like I did. And I understand that you prefer to check values of all render.
For now I'm gonna do something simple, but in the future we could imagine a custom matcher which would take an array of a custom type, for example:

```typescript
type ExpectedValue = {
  value: unknown;
  renderNumber: number;
}
```

And will be used:

```javascript
expect(watchedValues).toHaveValuesDuringRender([
      {
        value: {
          test: [
            {
              value: 'data',
            },
          ],
          renderNumber: 2,
        },
      },
      {
        value: {
          test: [
            {
              value: 'data',
            },
            { value: 'test' },
            { value: 'test1' },
          ],
        },
        renderNumber: 2,
      },
    ]);
```

Or maybe like I did after each actions which checks more finely the values with the number of render.

What do you think about it? If you would like have a glimpse, I can try to make an example in a future PR.

For now I will maybe use some snapshot like before in function of the output like you said ;)

romain-trotard:
@bluebill1049 I have made the changes about checking all renders values in the commit https://github.com/react-hook-form/react-hook-form/pull/8155/commits/095809a6498815eafc036ba10f51c5f7b6fa8b9a  :)

bluebill1049 (approved):
huge uplift on the tests, thank you so much! small suggestion maybe keeps PR relative smaller, but keep them coming!

PR #8167, opened 2022-04-11
Support field array parents with numbers in name

jgullstr:
**Describe the bug**
Field arrays with names containing numbers (for example "address1") currently break down when using "shouldUnregister: true" on the form level.

**To Reproduce**
[This sandbox](https://codesandbox.io/s/react-hook-form-usefieldarray-forked-855mle?file=/src/index.js) demonstrates the bug:

1. Append a row and alter its value.
2. Delete the first row.
3. The remaining field is reset to its initial value (empty string in this case).

See it working as intended by removing the "1" from the arrayName constant at line 9.

https://codesandbox.io/s/react-hook-form-usefieldarray-forked-855mle?file=/src/index.js

**Expected behavior**
The remaining fields should have unaltered values.

**Solution**
The expression used in "getNodeParentName" splits the field name at any digit instead of checking for an all-digit fragment.

I have modified the expression used by getNodeParentName to look for explicit numerical indices, either of the form "fieldName[123]" or "fieldName.123"

### DIFF
--- a/src/__tests__/logic/getNodeParentName.test.ts
+++ b/src/__tests__/logic/getNodeParentName.test.ts
@@ -3,11 +3,14 @@ import getFieldArrayParentName from '../../logic/getNodeParentName';
 describe('getFieldArrayParentName', () => {
   it('should return parent name when name is field array', () => {
     expect(getFieldArrayParentName('test.0')).toBe('test');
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks, @jgullstr for the issue report and PR 🙏 I have adjusted the regex to be a bit simpler and so it's faster.

PR #8181, opened 2022-04-13
getNodeParentName missing case(s)

jgullstr:
In reference to [this pull request](https://github.com/react-hook-form/react-hook-form/pull/8167):

The merged expression does not consider object keys that start with a digit "test.1st". I have altered the expression to make sure that the key is a positive amount of digits that either (1) ends with a period or (2) is at end of field name.

_Field names using square brackets, ie "test[0]", were removed from tests and are not supported at all now. Have they been deprecated?_

Another small optimization: Using Set.has in isNameInFieldArray has better performance than spreading the set into an array.


bluebill1049:
> Field names using square brackets, ie "test[0]", were removed from tests and are not supported at all now. Have they been deprecated?

yes, no longer supported.

bluebill1049 (approved):

PR #8203, opened 2022-04-17
V8: Feature/batch state update

bluebill1049:
```tsx
useEffect({
  // will update when errors updated
}, [errors])

useEffect({
  // will update when dirty fields updated
}, [dirtyFields])
```

Will works for individual form state as deps


Thirsty39 (commented):
.

bluebill1049:
@Thirsty39 anything you would like to report?

PR #8219, opened 2022-04-19
WIP: V8 make `of` mostly optional

felixschorer:


felixschorer:
Making `of` optional in generic contexts leads to weird edge cases. I think it's best to always require a call to `of` in those contexts.

PR #8267, opened 2022-04-28
Set defaultValue in state when initialized in useController

romain-trotard:
Hello :)

**Explanation**
Currently, there is a behavior difference:
- if we initialize a field directly in `useForm` with the prop `defaultValues` 
- if we initialize a controlled field in `useController` with the prop `defaultValue`

We can see the difference with the `isDirty` value.
_Initialized with `useForm`_ : https://codesandbox.io/s/rhf-initialized-with-useform-s7cl3y?file=/src/App.js
_Initialized with `useController`_: https://codesandbox.io/s/rhf-initialized-with-useform-forked-ch10pd?file=/src/App.js

In the case of the intialization with `useForm`, if we touch the input the form (field) stays pristine. Unlike when initializing with `useController` which becomes `dirty`.

**Research**
I did some research about this difference, but did not find anything in the documentation, issues, discussions and tests. I don't know if it's something wanted.
Therefore, I propose this PR that unifies the behavior.

**What I did?**
I just initialized the field with `useController` id it has not been initialized with `useForm`. If the user change the `defaultValue` after the first render, I do nothing.
The use of `useRef` is only here to satisfy the eslint `react-hooks/exhaustive-deps` rule.

If you have some questions about my changes let me know :)

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -295,6 +295,86 @@ describe('useController', () => {
     );
   });
 
+  it('should not be dirty when initializing field with controller defaultValue', async () => {
+    const App = () => {
+      const {
+        control,
+        formState: { isDirty },
+      } = useForm();
+
+      return (
+        <div>
+          <p>Form dirty: {isDirty.toString()}</p>
+          <Controller
+            render={({ field }) => {
+              return <input {...field} />;
+            }}
+            control={control}
+            name="test"
+            defaultValue="luo"
+          />
+        </div>
+      );
+    };
+
+    render(<App />);
+
+    const input = screen.getByRole('textbox');
+
+    expect(input).toHaveValue('luo');
+    expect(screen.getByText('Form dirty: false')).toBeVisible();
+
+    fireEvent.change(input, { target: { value: 'Bill' } });
+
+    expect(screen.getByText('Form dirty: true')).toBeVisible();
+
+    fireEvent.change(input, { target: { value: 'luo' } });
+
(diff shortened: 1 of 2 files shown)


bluebill1049:
Thanks, @romain-trotard for another awesome PR.

> You need to either set defaultValue at the field-level or useForm's defaultValues. undefined is not a valid value.

for such use cases, we are deliberay to avoid mixing with top-level and field-level default values.

bluebill1049:
In fact, we can keep the first test case as it's a valid one.

bluebill1049:
nvm, that's not valid as well, for checking dirty state, we required to profited defaultValues.

romain-trotard:
Thanks @bluebill1049 for the information. I close the PR.
Is it okay for you if I make a PR in the doc to add a word about `dirtyFields` and `isDirty`, which explains that to make them work default values should be set with `defaultValues` ?

bluebill1049:
Screen Shot 2022-05-03 at 10 01 33 pm

no worries, I believe it's already there. Let me know if there is anything we should/can improve on it.

romain-trotard:
> Screen Shot 2022-05-03 at 10 01 33 pm
> 
> no worries, I believe it's already there. Let me know if there is anything we should/can improve on it.

Oh my bad. Id didn't see this information. It's nice :)

bpinto:
Sorry but I have one question, it says " provide all inputs' defaultValues at the userForm", does that mean I cannot use `defaultValue` per input?

I was assuming the following was true in this scenario as well:

> You need to either set defaultValue at the field-level or useForm's defaultValues. undefined is not a valid value.

PR #8287, opened 2022-05-01
Allow unused spread siblings

Moshyfawn:
Suppress eslint warning in case
```
const { **name**, ...formState } = formStateData
```

It's a very common case to spread only relevant props and omit the unused ones by destructuring


bluebill1049:
👍 yep, that's great we do omit quite a bit here

PR #8293, opened 2022-05-02
Add examples from RFC 7389 and 7354

felixschorer:
Add example usage of `of` and `join` from RFC 7389 and 7354, respectively.


bluebill1049:
sweet! thanks @felixschorer

Moshyfawn:
Ah, super helpful!

I might try making use of generated CSB to provide a live example of V8 as well!

bluebill1049:
That would be amazing!

Issue #8256, opened 2022-04-27
issue: Uncaught TypeError: Cannot read properties of undefined (reading 'substring')


dwome:
### Version Number

7.30.0

### Codesandbox/Expo snack

not available

### Steps to reproduce

Since the 7.30.0 release we are expecting issues when using the React Hook Form Controller.
When rendering the component the value and the options might be undefined.

This worked in 7.29.0 like a charm so we are downgrading for now.

Any ideas how to resolve this issue?

```
const {
    control,
    formState: { errors },
  } = useFormContext();
  return (
    <Controller
      control={control}
      name={props.name}
      render={({ field }) => (
        <UncontrolledSelect
          {...props}
          isClearable={true}
          inputRef={field.ref}
          options={options}
          onBlur={field.onBlur}
          value={field.value}
          error={errors[props.name]?.message}
          onChange={(v, action) => {
            field.onChange(v);
            props?.onChange && props?.onChange(v, action);
          }}
        />
      )}
    />
  );
```

### Expected behaviour

No crashing and handling undefined values like in 7.29.0

### Relevant log output

```shell
Uncaught TypeError: Cannot read properties of undefined (reading 'substring')
    at getNodeParentName (index.esm.mjs?682b:21:1)
    at isNameInFieldArray (index.esm.mjs?682b:23:1)
    at useController (index.esm.mjs?682b:353:1)
    at Controller (index.esm.mjs?682b:465:1)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bluebill1049:
https://github.com/react-hook-form/react-hook-form/issues/2848

PR #8294, opened 2022-05-03
fix issue #8256

JerryYuanJ:
Hi @bluebill1049 , firstly I want to apologize that there's some problem for me to share the code snippet to show you the issue I've met, but when I looked at the board I found that issue https://github.com/react-hook-form/react-hook-form/issues/8256 is exactly what I encountered. 

I checked the according PR and found the cause of the issue. Just in terms of javascript, this update has its potential issue:

old:
```js
export default (names: Set<InternalFieldName>, name: InternalFieldName) =>
  [...names].some((current) => getNodeParentName(name) === current);
```
new:
```js
export default (names: Set<InternalFieldName>, name: InternalFieldName) =>
  names.has(getNodeParentName(name));
```

When the `names` is empty and `name` is `undefined`, the old way worked as expected because there's no iteration if the array is empty, so the function `getNodeParentName(name)` will never be called. 
But for the new one, the `getNodeParentName(name)` will always be called, and if it's undefined, it will break the whole thing.

So I added the condition to check if the name is valid, if yes, then we can do next thing, if not, just check if its origin value is in the set.

Please review this PR, let me know if this is correct way to do it, thanks!

### DIFF
--- a/src/logic/isNameInFieldArray.ts
+++ b/src/logic/isNameInFieldArray.ts
@@ -3,4 +3,4 @@ import { InternalFieldName } from '../types';
 import getNodeParentName from './getNodeParentName';
 
 export default (names: Set<InternalFieldName>, name: InternalFieldName) =>
-  names.has(getNodeParentName(name));
+  names.has(name ? getNodeParentName(name) : name);


bluebill1049:
thanks for the PR, curious to know when this would happen before we introduce the fix.

JerryYuanJ:
@bluebill1049 I'm using it together with mui `ChecoboxGroup` component, I didn't give its options item a filedName so it will cause this issue. I'll try to create a repo to reproduce it but it might take some time.

JerryYuanJ:
@bluebill1049 Hi, thanks for reply so quickly. I identified the root cause of it. In my project, the form's filed are dynamic and could be undefined/null, and when that happens, it will cause this error and break the whole page.

I created a very simple repo to reproduce it, please take a look: [https://stackblitz.com/edit/react-ts-y2nsop?file=App.tsx](https://stackblitz.com/edit/react-ts-y2nsop?file=App.tsx)

For my point of view, I don't think this error is that fatal to break the app, we can just leave the field empty. What do you think ?

bluebill1049:
>You need to either set defaultValue at the field-level or useForm's defaultValues. undefined is not a valid value.

thanks @JerryYuanJ you are missing `defaultValue` or `defaultvalues` at useForm

JerryYuanJ:
@bluebill1049 yes, I'm aware of that, but the use case of mine is that the form's field is dynamic and sometimes we don't know the defaultValue of current condition. So there might be some cases we encountered this issue. Can't we just have a simple check for `getNodeParentName(name)` function call? I mean it shouldn't be that fatal to break the app.

Moshyfawn:
> @bluebill1049 yes, I'm aware of that, but the use case of mine is that the form's field is dynamic and sometimes we don't know the defaultValue of current condition. So there might be some cases we encountered this issue. Can't we just have a simple check for `getNodeParentName(name)` function call? I mean it shouldn't be that fatal to break the app.

Why do you need to render the input when you don't have data to know what's the name of the input?

JerryYuanJ:
@Moshyfawn We implement a tiny JSON driven form, the component is generated based on the form json. That's why we might lose the fieldName for it.
But anyway, as I explained, the new `names.has(getNodeParentName(name));` has its potential risk right ? It's a runtime error and would stop the app.

bluebill1049:
thanks @JerryYuanJ however for incorrect usage, I actually prefer to have an early runtime error than break your app with consumers. In your use case above is not something that this library support as it states quite clear that we don't support `undefined`.

JerryYuanJ:
Hmm...I'm not convinced. As a user, I think that "", undefined, null should both have the same result because they're meaning the same that we don't have the name for this field. And again like I mentioned before, if we can't guarantee the value of name, we should check before we use it, but anyway, it's your call. Thanks for reply :)

PR #8301, opened 2022-05-04
fix: clone object which have functions

airman5573:
Above all, I don't think I understand the context of this library properly. **If there is any problem that my code will cause, please let me know**.

I think you decided not to just clone an object that contains a function in order not to create a new File object(maybe..?). But I think it's an error that when a user puts a function in defaultValues, the object itself containing the function is not copied.

So I changed cloneObject like below.

```javascript
let copy: any;
const isArray = Array.isArray(data);

if (data instanceof Date) {
  copy = new Date(data);
} else if (data instanceof Set) {
  copy = new Set(data);
} else if (globalThis.Blob && data instanceof Blob) { // check it is Blob
  copy = data;
} else if (isArray || isObject(data)) {
  copy = isArray ? [] : {};
  for (const key in data) { // dont check data[key] is function or not
    copy[key] = cloneObject(data[key]); // process clone
  }
} else {
  return data;
}

return copy;
```

This code prevent a bug(maybe..) below.

```javascript
function DefaultValues() {
  const defaultValues = {
    testFunction: () => { console.log('this is function'); },
    test: 'test',
    checkbox: ['1', '2'],
    test1: {
      firstName: 'firstName',
      lastName: ['lastName0', 'lastName1'],
      deep: {
        nest: 'nest',
      },
    },
  }
  const { register, handleSubmit, getValues } = useForm<{
    testFunction: () => void;
    test: string;
    test1: {
      firstName: string;
      lastName: string[];
      deep: {
        nest: string;
      };
    };
    checkbox: NestedValue<string[]>;
  }>({
    defaultValues,
  });
  const [show, setShow] = React.useState(true);

  const onSubmit = (data: any) => {
    defaultValues.test = 'another-test-string'; // this change the result of `getValues()`;
    console.log(getValues());
  }

  return (
    <>
      {show ? (
          <input {...register('test')} />
          <input {...register('test1.firstName')} />
          <input {...register('test1.deep.nest')} />
          <input {...register('test1.deep.nest')} />
          <input {...register('test1.lastName.0')} />
          <input {...register('test1.lastName.1')} />
          <input type="checkbox" value={'1'} {...register('checkbox')} />
          <input type="checkbox" value={'2'} {...register('checkbox')} />
          <button type="submit">SUBMIT</button>
        </form>
      ) : null}
      <button type={'button'} id={'toggle'} onClick={() => setShow(!show)}>
        toggle
      </button>
    </>
  );
}

export default DefaultValues;
```

If we skip clone an object which has a function, the result of `getValues().test` will be `another-test-string`.

### DIFF
--- a/src/__tests__/useForm/setValue.test.tsx
+++ b/src/__tests__/useForm/setValue.test.tsx
@@ -74,7 +74,6 @@ describe('setValue', () => {
       0: file,
       1: file,
       length: 2,
-      item: () => file,
     } as any as FileList;
 
     act(() => result.current.setValue('test', fileList));
--- a/src/__tests__/utils/cloneObject.test.ts
+++ b/src/__tests__/utils/cloneObject.test.ts
@@ -82,19 +82,30 @@ describe('clone', () => {
     expect(copy.items).toEqual([2]);
   });
 
-  it('should skip clone if a node contains function', () => {
+  it('should skip clone if a node is instance of Blob', () => {
     function testFunction() {}
 
-    const test = {
-      data: {
+    const data = {
+      test: {
         testFunction,
+        test: 'inner-string',
+        deep: {
+          test: 'deep-string',
+        },
       },
       other: 'string',
     };
 
-    expect(cloneObject(test)).toEqual({
-      data: {
+    const copy = cloneObject(data);
+    data.test.deep.test = 'changed-deep-string';
+
+    expect(copy).toEqual({
+      test: {
         testFunction,
+        test: 'inner-string',
+        deep: {
+          test: 'deep-string',
+        },
       },
       other: 'string',
     });
--- a/src/__tests__/utils/cloneObject.test.ts
+++ b/src/__tests__/utils/cloneObject.test.ts
@@ -101,7 +101,6 @@ describe('clone', () => {
 
     expect(copy).toEqual({
       test: {
-        testFunction,
         test: 'inner-string',
         deep: {
           test: 'deep-string',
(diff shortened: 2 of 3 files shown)


bluebill1049:
thanks for the PR. Please avoid the clone function as well, this is related to complex objects (which contain functions) such as draftJS or Moment.

airman5573:
> thanks for the PR. Please avoid the clone function as well, this is related to complex objects (which contain functions) such as draftJS or Moment.

Thanks for checking !

How to avoid clone function ? 

I think there are two way to avoid clone function.  

1. throw error
the structuredClone function [throw an error](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#things_that_dont_work_with_structured_clone) when clone an object which contain functions.

2. clone other property except function.
```
let copy: any;
const isArray = Array.isArray(data);

if (data instanceof Date) {
  copy = new Date(data);
} else if (data instanceof Set) {
  copy = new Set(data);
} else if (isArray || isObject(data)) {
  copy = isArray ? [] : {};
  for (const key in data) {
    if (!isFunction(data[key])) {
      copy[key] = cloneObject(data[key]);
    }
  }
} else {
  return data;
}
```

Am I thinking right ? check please!

bluebill1049:
That's awesome feedback above, yes we want to move into `structuredClone` eventually, for current I think your proposal 2 works well and fixes the bug.

bluebill1049:
Thanks for the update, I am leaving blob and file list coverage out for now as this is not the original issue.

bluebill1049:
Sorry to mess around with the commit, I am going to merge with this PR and have a play with the blob and filelist, ideally we can skip those checks.

bluebill1049:
thanks very much for the issue report and help on the PR! ❤️

PR #8338, opened 2022-05-10
try to save byte by move string to variable

bluebill1049:


src/utils/isWeb.ts
```diff
@@ -1,3 +1,5 @@
-export default typeof window !== 'undefined' &&
-  typeof window.HTMLElement !== 'undefined' &&
-  typeof document !== 'undefined';
+const UNDEFINED = 'undefined';
```
NMinhNguyen, line 1: Would you save more bytes by using `let`? 😆

```suggestion
let UNDEFINED = 'undefined';
```

NMinhNguyen, line 1: Would you save more bytes by using `let`? 😆

```suggestion
let UNDEFINED = 'undefined';
```

bluebill1049: I have been thinking about it change all const to let

`1:5  error  'UNDEFINED' is never reassigned. Use 'const' instead  prefer-const`

bluebill1049:
Failed :)

PR #8355, opened 2022-05-13
fix: compare html elements against correct global

zenoamaro:
When checking whether a value is an instance of HTMLElement,
the reference class must be owned by the same window which
owns the value, or the comparison will fail.

### DIFF
--- a/app/src/app.tsx
+++ b/app/src/app.tsx
@@ -4,6 +4,7 @@ import AutoUnregister from './autoUnregister';
 import Basic from './basic';
 import Watch from './watch';
 import BasicSchemaValidation from './basicSchemaValidation';
+import CrossFrameForm from './crossFrameForm';
 import SetError from './setError';
 import SetFocus from './setFocus';
 import SetValue from './setValue';
@@ -51,6 +52,7 @@ const App: React.FC = () => {
           path="/re-validate-mode/:mode/:reValidateMode"
           element={<ReValidateMode />}
         />
+        <Route path="/crossFrameForm" element={<CrossFrameForm />} />
         <Route path="/manual-register-form" element={<ManualRegisterForm />} />
         <Route path="/watch" element={<Watch />} />
         <Route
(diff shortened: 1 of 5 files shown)


zenoamaro:
This change fixes Radio and Checkbox handling when react-form-hook is running in a Web Extension's content script. 

In such a case, `window` is not the same as the element's `window`, and so `element` is not **instanceof** `window.HTMLElement`, which leads to the element ref being discarded during `register`, and its value being `null`.

Here is an example of the issue in action:

Screen Shot 2022-03-01 at 13 01 30

zenoamaro:
Thanks for the swift answer. 

It is not straight-forward to provide a test-case that matches our issue, as it comes from the special environment of a Chrome WebExtension. However the new solution for the `isHTMLElement` check should in general be the most technically correct way to compare instances owned by `window`.

What would you recommend?

bluebill1049:
can you reproduce the issue in the browser? if so maybe worth to add a cypress test.

bluebill1049:
any thought on the above? I am keen to merge this fix in. 👍

zenoamaro:
@bluebill1049 Thanks! I have experimented with some reproduction cases so far, but haven't find anything relevant/realistic enough yet. I will try once more, and let you know how it goes.

zenoamaro:
@bluebill1049 this test should be reasonably representative. It renders a basic form inside an iframe, while keeping component and callback outside of it. There's a unit test for `isHTMLElement`, and an integration test. Feedback on test names / content / style / routes is welcome.

Results before the fix:
Screen Shot 2022-05-31 at 18 29 34
Screen Shot 2022-05-31 at 16 33 50

Results after the fix:
Screen Shot 2022-05-31 at 18 30 25
Screen Shot 2022-05-31 at 16 34 35

bluebill1049:
awesome! thanks, I will add a few commits to see if we can reduce the package size impact.
Screen Shot 2022-06-01 at 8 47 48 am

bluebill1049:
Screen Shot 2022-06-01 at 9 23 28 am

better 👍

bluebill1049 (approved):

bluebill1049:
Thank you very much for your contribution. ❤️

PR #8368, opened 2022-05-17
fix(reset): set isSubmitted to formState value if keepIsSubmitted is true

blabute:
https://github.com/react-hook-form/react-hook-form/pull/8237 introduced a bug where `isSubmitted` would be set to the value of `keepIsSubmitted` (not the `isSubmitted` value that was in form state).

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1186,7 +1186,9 @@ export function createFormControl<
               keepStateOptions.keepDefaultValues &&
               !deepEqual(formValues, _defaultValues)
             ),
-      isSubmitted: !!keepStateOptions.keepIsSubmitted,
+      isSubmitted: keepStateOptions.keepIsSubmitted
+        ? _formState.isSubmitted
+        : false,
       dirtyFields:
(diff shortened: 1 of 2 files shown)


src/logic/createFormControl.ts
```diff
@@ -1186,7 +1186,9 @@ export function createFormControl<
               keepStateOptions.keepDefaultValues &&
               !deepEqual(formValues, _defaultValues)
             ),
-      isSubmitted: !!keepStateOptions.keepIsSubmitted,
+      isSubmitted: keepStateOptions.keepIsSubmitted
+        ? _formState.isSubmitted
+        : false,
```
bluebill1049, line 1191:
```suggestion
      isSubmitted: keepStateOptions.keepIsSubmitted && _formState.isSubmitted,
```

save some bytes?

bluebill1049: nvm on the above.

bluebill1049:
thanks for the fixing! ❤️

blabute:
> thanks for the fixing! ❤️

Thank you for the great package!

PR #8414, opened 2022-05-28
feat: make UseFormRegisterReturn generic over TFieldName

omjadas:
Improve the type for `UseFormRegisterReturn.name`

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -574,14 +574,14 @@ export type UseFormProps<TFieldValues extends FieldValues = FieldValues, TContex
 }>;
 
 // @public
-export type UseFormRegister<TFieldValues extends FieldValues> = <TFieldName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>>(name: TFieldName, options?: RegisterOptions<TFieldValues, TFieldName>) => UseFormRegisterReturn;
+export type UseFormRegister<TFieldValues extends FieldValues> = <TFieldName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>>(name: TFieldName, options?: RegisterOptions<TFieldValues, TFieldName>) => UseFormRegisterReturn<TFieldName>;
 
 // @public (undocumented)
-export type UseFormRegisterReturn = {
+export type UseFormRegisterReturn<TFieldName extends InternalFieldName = InternalFieldName> = {
     onChange: ChangeHandler;
     onBlur: ChangeHandler;
     ref: RefCallBack;
-    name: InternalFieldName;
+    name: TFieldName;
     min?: string | number;
     max?: string | number;
     maxLength?: number;
@@ -741,7 +741,7 @@ export type WatchObserver<TFieldValues> = (value: UnpackNestedValue<DeepPartial<
 
 // Warnings were encountered during analysis:
 //
-// src/types/form.ts:410:3 - (ae-forgotten-export) The symbol "Subscription" needs to be exported by the entry point index.d.ts
+// src/types/form.ts:412:3 - (ae-forgotten-export) The symbol "Subscription" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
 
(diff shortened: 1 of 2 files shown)


Moshyfawn:
What's the use-case for this feature?

omjadas:
> What's the use-case for this feature?

In my case, I have a custom component that acts as a wrapper for an `<input />` and I want to pass the name of the field and a corresponding control to that component so that I can use `useWatch` inside of it. Something like the following simplified example.

```tsx
import { Control, FieldPath, useForm, useWatch } from "../../../../react-hook-form";

interface MyInputProps<T extends string, U> extends React.HTMLProps<HTMLInputElement> {
  name: T,
  control: Control<U, any>,
}

const MyInput = <T extends FieldPath<U>, U extends Record<string, any>>(
  {
    name,
    control,
    ...props
  }: MyInputProps<T, U>,
): JSX.Element => {
  const value = useWatch({ name, control });

  // use value

  return (
    <input
      name={name}
      {...props}
    />
  );
};

interface Form {
  email: string,
}

const MyForm: React.FC = () => {
  const {
    register,
    control,
  } = useForm<Form>();

  return (
    <form>
      <MyInput
        control={control}
        {...register("email")}
      />
    </form>
  );
};
```

bluebill1049:
Thanks for the PR, this is actually intended to keep the name as `string` without impact ts error with the `name` type.

omjadas:
> Thanks for the PR, this is actually intended to keep the name as `string` without impact ts error with the `name` type.

Sorry, I'm not sure if I understand your comment. Are you saying that you would prefer to keep the name as a string instead of making it generic?

bluebill1049:
Yea, as this would lead to a breaking change to the existing user's controlled component when their child component accepts name as `string` instead of `FieldPath`

omjadas:
I don't think that should be an issue, the actual type of name will end up being a string literal type (in the case of my example `"email"`), which is a subtype of `string`, so can still be passed to a component that accepts name as `string`. The reason I used `FieldPath` in my example was to allow the name and control to be passed to `useWatch`.

omjadas:
As an example, the following also works

```tsx

interface MyInputProps extends React.HTMLProps<HTMLInputElement> {
  name: string,
}

const MyInput: React.FC<MyInputProps> = (
  {
    name,
    ...props
  },
): JSX.Element => {
  return (
    <input
      name={name}
      {...props}
    />
  );
};

interface Form {
  email: string,
}

const MyForm: React.FC = () => {
  const {
    register,
  } = useForm<Form>();

  return (
    <form>
      <MyInput
        {...register("email")}
      />
    </form>
  );
};
```

Moshyfawn:
Hm.. I see your point.

There's an ongoing development of the next major v8 RHF is happening. The goal is to improve on the existing type-safety of the package.

On of the improvements is to bring [the branded strings](#7354) into RHF, which would allow for a better DX around building reusable inputs.

Do you thing this is something can help you with your use-case?

omjadas:
> Hm.. I see your point.
> 
> There's an ongoing development of the next major v8 RHF happening. The goal is to improve on the existing type-safety of the package.
> 
> On of the improvements is to bring [the branded strings](https://github.com/react-hook-form/react-hook-form/discussions/7354) into RHF, which would allow for a better DX around building reusable inputs.
> 
> Do you thing this is something can help you with your use-case?

I think if the type of `UseFormRegisterReturn.name` was changed to `PathWithMeta` and `useWatch` was updated to support it, then it would work.

bluebill1049 (dismissed):

bluebill1049:
@omjadas could you please resolve the conflict?

kotarella1110 (approved):
> I don't think that should be an issue, the actual type of name will end up being a string literal type (in the case of my example `"email"`), which is a subtype of `string`, so can still be passed to a component that accepts name as `string`.

This is correct 👍🏻 
As far as I can think, this change will have no impact on users.

LGTM 🎉

bluebill1049:
thanks for the confirmation and double check @kotarella1110 🙏

PR #8484, opened 2022-06-10
Funnel along appropriate constraints for (upcoming) TypeScript 4.8

DanielRosenwasser:
Hi there! 👋

TypeScript recently added some stricter checking around

* unconstrained generics being incompatible with emptyish object types.
* narrowing on generic values

The changes are only in nightly versions of TypeScript right now, but we've been exploring how open-source projects have been impacted [here](https://github.com/microsoft/TypeScript/issues/49461). 

These are some of the changes necessary for react-hook-form to build cleanly, though there are two left that require manual intervention from a maintainer.

```
src/logic/getFieldValueAs.ts:14:10 - error TS2533: Object is possibly 'null' or 'undefined'.

14       : +value
            ~~~~~

src/logic/validateField.ts:121:53 - error TS2533: Object is possibly 'null' or 'undefined'.

121         (ref as HTMLInputElement).valueAsNumber || +inputValue;
                                                        ~~~~~~~~~~


Found 2 errors in 2 files.

Errors  Files
     1  src/logic/getFieldValueAs.ts:14
     1  src/logic/validateField.ts:121
```

These seem like legitimate errors, and would have been caught if `value` and `inputValue` were not declared as generics. In fact, I don't think they need to be declared as generics.

### DIFF
--- a/app/src/crossFrameForm.tsx
+++ b/app/src/crossFrameForm.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import {useForm} from 'react-hook-form';
+import { useForm } from 'react-hook-form';
 
 const FRAME_CONTENT = `
   <style>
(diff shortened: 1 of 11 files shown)


bluebill1049:
Thanks very much for the PR, would you mind to re-run the API extractor?

DanielRosenwasser:
Done!

bluebill1049 (approved):
Thanks very much for the contribution and help ❤️

PR #8528, opened 2022-06-18
deprecate on `NestedValue` and `UnpackNestedValue`

bluebill1049:


tetchel:
So, what can we use instead? The deprecation notice says "consult docs", but doesn't say what the new/alternative API is?

image

bluebill1049:
try to simply remove them.

tetchel:
thanks. it looks like `FieldPathValue` on its own has the same effect.

PR #8552, opened 2022-06-23
Improve inference of browser environment

### LINKED ISSUE #8551: Rely on presence of window.HTMLDivElement instead of window.HTMLElement to infer browser environment
Custom elements extend HTMElement (`class MyCustomElement extends HTMLElement`), so when rendering them server-side one needs to shim the presence of HTMLElement. In such an environment react-hook-form will therefore return a false positive, because it infers a browser environment like this in [isWeb.ts](https://github.com/react-hook-form/react-hook-form/blob/111c2b651c876086e264279895425d0556598913/src/utils/isWeb.ts#L2):
```typescript
export default typeof window !== 'undefined' &&
  typeof window.HTMLElement !== 'undefined' &&
  typeof document !== 'undefined';
```

Instead of relying on presence of window.HTMLElement, I suggest relying on the presence of window.HTMLDivElement, which you do not need to shim when rendering custom elements.

None

**Additional context**
None

wolthers:
Fixes https://github.com/react-hook-form/react-hook-form/issues/8551


bluebill1049:
thanks for the PR, will pend on merging until more users request this related feature request.

PR #8102, opened 2022-03-27
V8: feature: useFieldArray rules props

bluebill1049:
close: #6879 

```tsx
useFieldArray({
  name: 'test',
  rules: {
    required: true,
    minLength: 2,
    maxLength: 10,
    validate: (fieldArrayValues) => {
      if (fieldArrayValues) {
        return 'valdiate Error'
      }
    }
  }
})

errors?.test?.root?.message
```

ref: https://codesandbox.io/s/react-hook-form-usefieldarray-forked-iyejbp?file=/src/src/logic/createFormControl.ts

- [x] functional tests
- [x] units & automation tests
- [x] type tests

cc @Moshyfawn

PR #8562, opened 2022-06-24
useFieldArray `rules` props (#8102)

bluebill1049:
close: #6879 

```tsx
useFieldArray({
  name: 'test',
  rules: {
    required: true,
    minLength: 2,
    maxLength: 10,
    validate: (fieldArrayValues) => {
      if (fieldArrayValues[2].title === 'test') {
        return 'valdiate Error'
      }
    }
  }
})

errors?.test?.root?.message
```

ref: https://codesandbox.io/s/react-hook-form-usefieldarray-forked-iyejbp?file=/src/src/logic/createFormControl.ts

- [x] functional tests
- [x] units & automation tests
- [x] type tests
- [x] API poll: https://github.com/react-hook-form/react-hook-form/discussions/8625


src/logic/validateField.ts
```diff
@@ -86,11 +87,13 @@ export default async <T extends NativeFieldValue>(
   };
 
   if (
-    required &&
-    ((!isRadioOrCheckbox && (isEmpty || isNullOrUndefined(inputValue))) ||
-      (isBoolean(inputValue) && !inputValue) ||
-      (isCheckBox && !getCheckboxValue(refs).isValid) ||
-      (isRadio && !getRadioValue(refs).isValid))
+    isFieldArray
+      ? Array.isArray(inputValue) && !inputValue.length
```
bluebill1049, line 91: Should also check for `undefined` value which means the field array is not presented at all.

mmahalwy:
@bluebill1049 thanks for creating this! Just so I understand, the validate cannot target specific fields in the array? So if we are virtualizing 1000s of inputs, we cant target a field that's not mounted with this other than erroring on the root.

bluebill1049:
I don't think so @mmahalwy

Pulverizer2020:
I found a small bug: 

If you have your form on "onTouched" mode and have multiple validations for useFieldArray, it will properly validate and change the error each time you clear an error, but the error message will not clear when all validation checks are passed.

Ex:

```
useFieldArray({
  name: 'test',
  rules: {
    validate: {
        isFirstNamePresent: (v) => {
          if (v[0].firstName) {
            return "Please add first name";
          }
          return true;
        },
        isLastNamePresent: (v) => {
          if (v[0].lastName) {
            return "Please add last name";
          }
          return true;
        },
      }
  }
})
```

If both fields are validated, the error message in `formState.errors?.test?.root?.message` is not removed.

miller-productions:
@bluebill1049

Thanks for implementing this, Bill 🙏🙂

bluebill1049:
most welcome @miller-productions 🙏

zned45:
This does not seem to be working. There is not error messages in the root field, instead ```formState.errors.test``` is equal to an empty array.

PR #8566, opened 2022-06-24
Ensure field is focused when selected

elstgav:
As [noted by MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/select#notes), calling `element.select()` doesn’t necessarily focus the element. As such, they recommend calling `element.focus()` too.

I believe the expectation when calling `setFocus(name, { shouldSelect: true })` is that the field would be both selected _and_ focused.


bluebill1049:
thanks for the PR.

PR #8568, opened 2022-06-25
add react v16 and 17 job to unit test workflow

kotarella1110:
Since RHF continues to support React ^16.8.0 and 17, it is better to test with them.
Cypress should be tested as well, but should be considered due to its long execution time.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -3,7 +3,37 @@ name: Unit Test / Lint
 on: [pull_request]
 
 jobs:
-  build:
+  react-16-17:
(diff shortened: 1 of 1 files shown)


bluebill1049:
awesome to start seeing PR from @kotarella1110 🕺

bluebill1049:
- Server-side rendering is going to have a slightly different output.
- Rendering count maybe a bit different compare < 18 

We may need to update the tests to make them tolerate for those minor adjustment due to react major update.

kotarella1110:
@bluebill1049 I will update the tests 👍

PR #8571, opened 2022-06-25
use const instead of props to win some bytes

cdrikd:
Use new const created instead of props.


Upinwv:
I’m sorry buddy

--

bluebill1049:
YES YES every bytes count: thank you @cdrikd please send more if you find any other. I do sacrifice readability over package size, so don't hesitate to change code and we have pretty good test coverage as well.

image

bluebill1049 (approved):

PR #8608, opened 2022-06-30
Close #8596

### LINKED ISSUE #8596: issue: Destructuring methods on the provider are causing unnecessary re-renders
### Version Number

7.33.0

### Codesandbox/Expo snack

https://codesandbox.io/s/resolver-context-forked-1q82fl?file=/src/App.tsx

### Steps to reproduce

1. Go to ['...'](https://codesandbox.io/s/resolver-context-forked-1q82fl?file=/src/App.tsx)
2. Inspect AppWithProvider and AppWithCustomProvider
3. Look that second field is re-rendering unnecessary 

### Expected behaviour

second field should not be rerendering.
We should not be spreading methods, because it creates a new instance of the object, and causes the rerenders

_No response_

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
Close #8596

### DIFF
--- a/src/useFormContext.tsx
+++ b/src/useFormContext.tsx
@@ -71,10 +71,8 @@ export const useFormContext = <
  */
 export const FormProvider = <TFieldValues extends FieldValues, TContext = any>(
   props: FormProviderProps<TFieldValues, TContext>,
-) => {
-  return (
-    <HookFormContext.Provider value={props as unknown as UseFormReturn}>
-      {props.children}
-    </HookFormContext.Provider>
-  );
-};
+) => (
+  <HookFormContext.Provider value={props as unknown as UseFormReturn}>
+    {props.children}
+  </HookFormContext.Provider>
+);
--- a/src/useFormContext.tsx
+++ b/src/useFormContext.tsx
@@ -72,10 +72,9 @@ export const useFormContext = <
 export const FormProvider = <TFieldValues extends FieldValues, TContext = any>(
   props: FormProviderProps<TFieldValues, TContext>,
 ) => {
-  const { children, ...data } = props;
   return (
-    <HookFormContext.Provider value={data as unknown as UseFormReturn}>
-      {children}
+    <HookFormContext.Provider value={props as unknown as UseFormReturn}>
+      {props.children}
     </HookFormContext.Provider>
   );
 };


src/useFormContext.tsx
```diff
@@ -71,11 +71,8 @@ export const useFormContext = <
...
-) => {
-  const { children, ...data } = props;
-  return (
-    <HookFormContext.Provider value={data as unknown as UseFormReturn}>
-      {children}
-    </HookFormContext.Provider>
-  );
-};
+) => (
+  <HookFormContext.Provider value={props as unknown as UseFormReturn}>
```
alexandrzavalii, line 75: would it be okay to pass children to the context?

bluebill1049: In theory, this should be fine as `children's` props are a reference for the actual children's component. however I am not in a hurry to have this fix to be merged, your current work around it's suitable to resolve the issue that you have reported.

Moshyfawn: The way you pass `props` exposes the Context's `children` to.. `children`. Which is not correct.
Like, you're able to do _(pseudo-code)_
```js
const formContext = useContext(FormContext)

formContext.children
```

As we're simply returning the `Context` `value` in the `useFormContext` hook
```ts
export const useFormContext = <
>(): UseFormReturn<TFieldValues> =>
  React.useContext(HookFormContext) as unknown as UseFormReturn<TFieldValues>;
```
the user can actually access `children` this way, even though the TS will scream due the type assertion `as UseFormReturn<TFieldValues>`

Moshyfawn: I can't say for sure rn _('cuz I'm going through a Pfizer shot lol)_, but is assigning the props one by one to the proper `Context` `value` to avoid passing `children` into it is the only solution without spreading the rest?
```diff
- const { children, ...data } = props
+ const { children, watch, getValues, getFieldState, /* ... all the other form props one by one ... */ } = props

return (
  <HookFormContext.Provider value={{ watch, getValues, getFieldState, /* ... all the other form props one by one ... */ }}>
      {children}
    </HookFormContext.Provider>
)
```

bluebill1049: thanks @Moshyfawn I am yet convinced even such change is required, so not sure we will see until more and more users report this is actually a real issue impacting app performance, I will start to look into alternatives.

alexandrzavalii: any re-render of the component that uses FormProvider would trigger re-renders for every single component that uses hook from `react-hook-form`.
I think its a big concern especially considering the fact that the first word describing the library is Performant

> Performant, flexible and extensible forms with easy-to-use validation.

bluebill1049:
For anyone else who encounters this or a related issue, first please check your application structure when a single re-render costs a lot on your application performance. If yet to identify a root cause or eliminated edge case, you can consider to the use the following and build your context.

```tsx
export const FormProvider = <TFieldValues extends FieldValues, TContext = any>({
  register,
  setFocus,
  formState,
  handleSubmit,
  getValues,
  setValue,
  control,
  clearErrors,
  children,
  setError,
  trigger,
  watch,
  getFieldState,
  resetField,
  reset,
  unregister,
}: FormProviderProps<TFieldValues, TContext>) => (
  <HookFormContext.Provider
    value={
      {
        register,
        setFocus,
        formState,
        handleSubmit,
        getValues,
        setValue,
        clearErrors,
        control,
        setError,
        trigger,
        watch,
        resetField,
        getFieldState,
        reset,
        unregister,
      } as unknown as UseFormReturn
    }
  >
    {children}
  </HookFormContext.Provider>
);
```

alexandrzavalii:
@bluebill1049 
To add to the solution of using custom provider, you would also need to always pass your own control to the hooks and `Controller`

```
  const control = useYourCustomController();
  const val = useWatch({name: 'field', control})
  return (
      <Controller control={control}/>
  )

PR #8678, opened 2022-07-14
Native form reset isn't executed if first fieldReference isn't a HTMLElement

alexviar:
If the first field evaluated in `isHTMLElement(fieldReference)` returns `false` then loop is skipped and native reset is not executed.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1134,7 +1134,7 @@ export function createFormControl<
                 : field._f.ref;
 
               try {
-                if(isHTMLElement(fieldReference)){
+                if (isHTMLElement(fieldReference)) {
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks for the PR. however, this is expected behavior, if we detect there are nonnative input presented in the form, we will have to flush the state update instead of using the native reset API.

alexviar:
> thanks for the PR. however, this is expected behavior, if we detect there are nonnative input presented in the form, we will have to flush the state update instead of using the native reset API.

I don't quite understand what you mean, the current code also doesn't seem to do what you say because if the first field that is evaluated is a native input, the native reset is executed and the loop breaks regardless of whether or not there are non-native inputs in the form

bluebill1049 (approved):

PR #8726, opened 2022-07-25
Memoize using `React.memo`

sanjaiyan-dev:
Memoizing some components using `React.memo` which might improve performance.

_Sorry if I made any mistakes :(_


minorgod:
Looks like your build would pass if you follow the instructions here:
https://github.com/react-hook-form/react-hook-form/runs/7493599184?check_suite_focus=true
and fix the code formatting issues reported here:
https://github.com/react-hook-form/react-hook-form/runs/7493599180?check_suite_focus=true

PR #8737, opened 2022-07-26
Fix Typo : excuteBuildInValidation -> excuteBuiltInValidation

ANTARES-KOR:
In my opinion, `executeBuiltInValidation` should be more reasonable name for this function.

And the docs says "Built-in validators", too.

스크린샷 2022-07-26 오후 3 23 50

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -161,7 +161,7 @@ export function createFormControl<
     if (_proxyFormState.isValid) {
       isValid = _options.resolver
         ? isEmptyObject((await _executeSchema()).errors)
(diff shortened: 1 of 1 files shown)


Moshyfawn:
Fair enough, but it's a breaking change, so, if we go with it, it's gonna be in the next major release, or even later, because we have a lot of changes in v8

Moshyfawn:
Thanks, @ANTARES-KOR!
@bluebill1049 I don't see why not

bluebill1049:
thanks for the review @Moshyfawn 🙏

bluebill1049 (approved):

PR #8772, opened 2022-07-30
fix Date, FileList, File and Blob FieldErrors mapping

### LINKED ISSUE #8749: [FieldError] - unexpected type error for Date property which is not nullable
### Version Number

7.33.1

### Codesandbox/Expo snack

https://codesandbox.io/s/recursing-jang-9l8kqm?file=/src/App.tsx

### Steps to reproduce

1. Go to https://codesandbox.io/s/recursing-jang-9l8kqm?file=/src/App.tsx
2. Check the `App.tsx` and check the type error for the `mandatoryDateError`

### Expected behaviour

After upgrading to version 7.33.1 I got TS-errors because my `DateInput` expect a `FieldError` regardless if the Date is nullable or not. This is not an urgent issue but I thought it might be a edge case which has been discovered. Or perhaps I just don't understand why this behaviour makes sense. 

Thanks for your work and the awesome library!

_No response_

### Relevant log output

_No response_

### Code of Conduct

vemoo:
It seems #8754 didn't fully fix #8749.
I've added a test and completed `DeepRequired` to fix it.

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -131,7 +131,7 @@ export type DeepPartialSkipArrayKey<T> = T extends Date | FileList_2 | File_2 | 
 };
 
 // @public (undocumented)
-export type DeepRequired<T> = {
+export type DeepRequired<T> = T extends Date | FileList | File | Blob ? T : {
     [K in keyof T]-?: NonNullable<DeepRequired<T[K]>>;
 };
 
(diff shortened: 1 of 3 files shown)


bluebill1049:
thanks for the fix 👍

bluebill1049 (approved):

mneira10:
Hi! Could you release this fix please? Currently getting some errors when trying to pass `errors` from the `formState` to a component as props in typescript

PR #8804, opened 2022-08-06
fix(path): keys of `Date | FileList | File` shouldn't be add to the `PathImpl`

ronny1020:
keys of `Date | FileList | File` shouldn't be add to the `PathImpl`

If using `Path<{foo: {bar: File}}>`, the pahs of File keys shouldn't be include.

### DIFF
--- a/src/types/path/eager.ts
+++ b/src/types/path/eager.ts
@@ -1,5 +1,5 @@
 import { FieldValues } from '../fields';
-import { Primitive } from '../utils';
+import { BrowserNativeObject, Primitive } from '../utils';
 
 import { ArrayKey, IsTuple, TupleKeys } from './common';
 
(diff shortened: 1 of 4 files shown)


src/types/path/eager.ts
```diff
@@ -7,7 +7,11 @@ import { ArrayKey, IsTuple, TupleKeys } from './common';
  * Helper type for recursively constructing paths through a type.
  * See {@link Path}
  */
-type PathImpl<K extends string | number, V> = V extends Primitive
+type PathImpl<K extends string | number, V> = V extends
+  | Primitive
+  | Date
+  | FileList
+  | File
```
bluebill1049, line 14: Mabe those can be converted into `BrowserNativeObject`?

bluebill1049 (dismissed):
Thanks for the PR 👍

bluebill1049 (approved):

kotarella1110 (approved):
Thanks. LGTM!

bluebill1049:
thanks @kotarella1110's reivew .

PR #8829, opened 2022-08-10
move isSubmitting state update after valid form

bluebill1049:


bryantobing12:
why is this reverted?

PR #8837, opened 2022-08-12
feature: field path by value generic implementation

carvalheiro:
I came to this solution after a discussion on the discord support channel. related to: https://github.com/react-hook-form/react-hook-form/discussions/8557

The purpose of the `FieldPathByValue` is to return a union with all available paths that match the passed value. This is really helpful to have strongly type custom components using the `ControllerApi`. As e.g.:

```jsx
import {
  FieldPathByValue,
  FieldValues,
  Control,
  useController,
  useForm
} from "react-hook-form";

// this component uses a date value
function SomeCustomFormComponent<
    TPath extends FieldPathByValue<TFieldValues, Date>
>({
    control,
    name
}: {
    control: Control<TFieldValues>,
    name: TPath
}) {
    const { field } = useController({
      control,
      name,
    });

    // rest of components rules
}

function ExampleOfUsage () {
    const {control} = useForm<{
        foo: Date;
        baz: string;
        bar: {
            baz: Date;
        }
    }>()

    return (
      <>
        <SomeCustomFormComponent control={control} name="foo" /> {/* throw no error */}
        <SomeCustomFormComponent control={control} name="baz" /> {/*  throw an error since baz is string */}
        <SomeCustomFormComponent control={control} name="bar.baz" /> {/* throw no error */}
      </>
    );
}
```

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -218,6 +218,11 @@ export type FieldNamesMarkedBoolean<TFieldValues extends FieldValues> = DeepMap<
 export type FieldPath<TFieldValues extends FieldValues> = Path<TFieldValues>;
 
 // @public
+export type FieldPathByValue<TFieldValues extends FieldValues, TValue> = {
+    [Key in FieldPath<TFieldValues>]: FieldPathValue<TFieldValues, Key> extends TValue ? Key : never;
+}[FieldPath<TFieldValues>];
+
+// @public
 export type FieldPathValue<TFieldValues extends FieldValues, TFieldPath extends FieldPath<TFieldValues>> = PathValue<TFieldValues, TFieldPath>;
 
 // @public
--- a/src/types/path/eager.ts
+++ b/src/types/path/eager.ts
@@ -140,3 +140,25 @@ export type FieldPathValues<
     TPath[K] & FieldPath<TFieldValues>
   >;
 };
+
+/**
+ * Type which eagerly collects all paths through a fieldType that matches a give type
+ * @typeParam TFieldValues - field values which are indexed by the paths
+ * @typeParam TValue       - the value you want to match into each type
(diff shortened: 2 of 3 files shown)


bluebill1049:
thanks for the PR, we will get back to you soon.

bluebill1049:
I wonder if this could apply to `useFieldArray` as well?

carvalheiro:
> I wonder if this could apply to `useFieldArray` as well?
Doesn't seems to me  that it's possible since we only can map through unions for apply it's rules. I guess We could think in a different strategy that make it works with that other map.

bluebill1049:
thanks for the response @carvalheiro a separate PR would be great.

Moshyfawn (dismissed):
Looks good to me!
I can see how we might get bug reports about field array fields "not working", since we support single fields with this PR. Let's see what we can do

bluebill1049:
yea @Moshyfawn that's why i am suggesting if we can solve field array in a similar type as well.

kotarella1110 (commented):
@carvalheiro Please run `yarn api-extractor`

bluebill1049 (dismissed):

carvalheiro:
> @carvalheiro Please run `yarn api-extractor`

@kotarella1110 Done! 
image

kotarella1110:
@carvalheiro Thanks!
Please commit & push output file (`reports/api-extractor.md`).

carvalheiro:
> @carvalheiro Thanks! Please commit & push output file (`reports/api-extractor.md`).

Oh sorry! done 🚀

kotarella1110 (approved):
@carvalheiro 
awesome PR 🎉
thx!

bluebill1049:
thanks, everyone! Will release a beta release for this one soon.

PR #8840, opened 2022-08-12
Fix: nested fields example

HarmonyEarth:
Fixed imports and exports on the nested fields example, along with other minor errors.


bluebill1049:
thanks 👍

PR #8917, opened 2022-08-25
fix: DeepMap infers types correctly when using union types

### LINKED ISSUE #8815: issue: DeepMap not working correctly with Union Types
### Version Number

7.4.1

### Codesandbox/Expo snack

https://codesandbox.io/s/laughing-pascal-ev9v9x?file=/src/App.tsx

### Steps to reproduce

In this example I'm using `zod` for schema validation.

1. See given Codesandbox
2. Check `useEffect` and `showErrors` for type errors

----

Given the following schema:

```ts
const Gelato = z.literal("gelato");
const Sorbetto = z.literal("sorbetto");

const GelatoSchema = z.object({
  type: Gelato,
  creamFlavor: z.string().min(1)
});

const SorbettoSchema = z.object({
  type: Sorbetto,
  fruitFlavor: z.string().min(1)
});

const FormSchema = z.object({
  name: z.string().min(1),
  favoriteIceCream: z.union([GelatoSchema, SorbettoSchema])
});

type FormSchemaType = z.infer<typeof FormSchema>;
```

and using `useForm` like the following:

```ts
const {
    register,
    handleSubmit,
    control,
    formState: { touchedFields, errors }
  } = useForm<FormSchemaType>({
    defaultValues: {
      name: "",
      favoriteIceCream: {
        type: Gelato.value,
        creamFlavor: ""
      }
    },
    resolver: zodResolver(FormSchema)
  });

  const formState = useWatch({ control });
```

will throw a type error message when trying to access `creamFlavor` or `fruitFlavor` from `touchedFields` or `errors`:

```ts
Property 'creamFlavor' does not exist on type 'DeepMap<{ type: "gelato"; creamFlavor: string; } | { type: "sorbetto"; fruitFlavor: string; }, true>'.
```

### Expected behaviour

I was expecting both `touchedFields` and `errors` have the same type as `formValues`, but only the common field between the two types (in this case, the `type: 'gelato' | 'sorbetto'`) is available 

Firefox, Chrome, Safari, Edge

### Relevant log output

_No response_

### Code of Conduct

tatianainama:
Fixes #8815

This fixes DeepMap not applying to union types using a combination of `UnionToIntersection` and avoiding distributivity on conditional types (`[T] extends [...]`) 

I am not entirely sure this is the correct approach, please let me know what you think about this solution.

I'm still figuring out a fix for `FieldErrorsImpl` but wanted to check with you this one first

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -116,9 +116,11 @@ export type CustomElement<TFieldValues extends FieldValues> = {
     focus?: Noop;
 };
 
+// Warning: (ae-forgotten-export) The symbol "UnionToIntersection" needs to be exported by the entry point index.d.ts
+//
 // @public (undocumented)
-export type DeepMap<T, TValue> = IsAny<T> extends true ? any : T extends BrowserNativeObject | NestedValue ? TValue : T extends object ? {
-    [K in keyof T]: DeepMap<NonUndefined<T[K]>, TValue>;
+export type DeepMap<T, TValue> = IsAny<T> extends true ? any : [T] extends [BrowserNativeObject | NestedValue] ? TValue : [T] extends [object] ? {
+    [K in keyof UnionToIntersection<T>]: DeepMap<NonUndefined<UnionToIntersection<T>[K]>, TValue>;
 } : TValue;
 
 // @public (undocumented)
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -213,7 +213,12 @@ export function createFormControl<
           args.argA,
           args.argB,
         );
-        shouldSetValues && set(_formState.touchedFields, name, touchedFields);
+        shouldSetValues &&
+          set(
+            _formState.touchedFields as unknown as TFieldValues,
+            name,
+            touchedFields,
+          );
       }
 
       if (_proxyFormState.dirtyFields) {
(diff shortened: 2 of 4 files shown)


reports/api-extractor.md
```diff
@@ -116,9 +116,11 @@ export type CustomElement<TFieldValues extends FieldValues> = {
     focus?: Noop;
 };
 
+// Warning: (ae-forgotten-export) The symbol "UnionToIntersection" needs to be exported by the entry point index.d.ts
```
tatianainama, line 119: I didn't know what to do with this warning 🙏

src/logic/createFormControl.ts
```diff
@@ -298,15 +303,19 @@ export function createFormControl<
 
       isCurrentFieldPristine
         ? unset(_formState.dirtyFields, name)
-        : set(_formState.dirtyFields as TFieldValues, name, true);
+        : set(_formState.dirtyFields as unknown as TFieldValues, name, true);
```
tatianainama, line 306: There was a cast before this, I'm not sure why, but since this change I had to cast to `unknown` first, which already gives me a bad feeling

bluebill1049:
thanks for looking into this 🙏 @tatianainama

reports/api-extractor.md
bluebill1049: we are not export common types, hence this warning. It's safe to ingore

Moshyfawn (approved):

Moshyfawn (changes requested):
https://github.com/react-hook-form/react-hook-form/issues/8815#issuecomment-1227602302

PR #8974, opened 2022-09-03
🛵 close #8969 improve type for `useFieldArray` rules validate prop

### LINKED ISSUE #8968: Add options to useFieldArray functions
I recently came upon using `useFieldArray` to update my array of objects across the application instead of using setValue, since my arrays are quite large and instead of entirely replacing them with `setValue`, I preferred to use `update` or `replace` from `useFieldArray` instead. The problem that I am facing is this, my submit button is linked to the forms `dirty` state, so when I use the `update` method to make specific changes to my array, by form becomes dirty, when the user physically made no change and the button becomes enabled when it shouldn't.  

Similar to the `setValue` function, I believe there should be an options object available in the `update` and `replace` functions of `useFieldArray` where we can set booleans such as `shouldDirty` to false, so that when required, we can update the form values without dirtying it.  

```
const { update } = useFieldArray({ name, control })

update(index, updatedObject, { shouldDirty: false })
```

The only alternate that I can think of is using the `setValue` function or checking for dirty state before using `update` function

### LINKED ISSUE #8969: issue: FieldArray rules.validate type is overly broad
### Version Number

7.34.2

### Codesandbox/Expo snack

https://codesandbox.io/s/rhf-fieldarray-validate-sznb1h

### Steps to reproduce

I'm trying to use the new `rules.validate` from 7.34 (🙌 released just when I needed it!).  However, I find that the type of `values` provided to the validate function are very broad, and seem to not only include types from the provided fieldArray name, but other fields as well.  

### Expected behaviour

I expected that the type of the `value` for the validate function would be the same as what I provided to `FormValues`.

_No response_

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
close #8969

Screen Shot 2022-09-03 at 2 58 19 pm

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -528,7 +528,9 @@ export type UseFieldArrayProps<TFieldValues extends FieldValues = FieldValues, T
     name: TFieldArrayName;
     keyName?: TKeyName;
     control?: Control<TFieldValues>;
-    rules?: Pick<RegisterOptions<TFieldValues>, 'maxLength' | 'minLength' | 'validate' | 'required'>;
+    rules?: {
+        validate?: Validate<FieldArray<TFieldValues, TFieldArrayName>[]> | Record<string, Validate<FieldArray<TFieldValues, TFieldArrayName>>>;
+    } & Pick<RegisterOptions<TFieldValues>, 'maxLength' | 'minLength' | 'required'>;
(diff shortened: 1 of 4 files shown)


Moshyfawn:
Fair, but also a bit messy :P

PR #8984, opened 2022-09-05
Forward custom props from Controller to render

NetoBraghetto:
I have this situation where I need to wrap inputs inside a Controller depending on any parent prop Eg:

```javascript
export function Foo({error}) {
  {/* ... */}
  const WrappedComponent = ({ field }) => {
    return (
      <TextField
        error={error}
        {/* ... */}
      />
    );
  };

  return (
    <Controller
      name="firstName"
      control={control}
      render={WrappedComponent}
    />
  );
}
```

And creating components inside components without memoization leads to the nested component and all its children being recreated during each re-render. A property to pass down custom properties could prevent this behavior

something like this:

```javascript
const WrappedComponent = ({ field, customProps }) => {
  return (
    <TextField
      error={customProps.error}
      {/* ... */}
    />
  );
};

export function Foo({error}) {
  ...
  return (
    <Controller
      name="firstName"
      control={control}
      render={WrappedComponent}
      customProps={{ error }}
    />
  );
}
```

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -85,10 +85,11 @@ export type ControllerFieldState = {
 
 // @public
 export type ControllerProps<TFieldValues extends FieldValues = FieldValues, TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>> = {
-    render: ({ field, fieldState, formState, }: {
+    render: ({ field, fieldState, formState, customProps, }: {
         field: ControllerRenderProps<TFieldValues, TName>;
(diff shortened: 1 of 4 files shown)


Moshyfawn:
I don't think `customProps` API will be merged, TBH
@bluebill1049

bluebill1049:
thanks for the PR, however, we are not looking into expanding the API at the moment. you can build a customer controller component with useController and merge the props at your app/consumer level.

PR #8966, opened 2022-09-01
🍨 feature: include `defaultValues` inside `formState`

bluebill1049:
## Description

Allow user to read defaultValues within `useForm` and `useFormState` hook

- feature-request: https://github.com/react-hook-form/react-hook-form/issues/7289
- Twitter poll: https://twitter.com/HookForm/status/1564869849095933952

## Example

CSB: https://codesandbox.io/s/react-hook-form-v7-ts-template-forked-qhqnfc?file=/src/App.tsx

```tsx
const { formState: { defaultValues } } = useForm()
const { defaultValues } = useFormState()
```

## Todos

- [x] code implementation
- [x] unit tests
- [x] doc update

PR #9051, opened 2022-09-17
🚂 feature: `reset` to support callback syntax

bluebill1049:
## Context

There is a lot of use case, in which we observe users are reset with partial form values and required to invoke `getValues` at the same time.

```tsx
reset({
  ...getValues(),
  partialData: 'onlyChangeThis'
})
```

## Prospal API

This following API eliminates the need to forward the getValues method and also type safe with the callback payload. This also align with React's core `useState/setState` callback syntax.

```tsx
reset((formValues) => {
  return {
    ...formValues,
    partialData: 'onlyChangeThis'
  }
})
```

## Todos

- [x] code implementation
- [x] unit tests
- [x] doc update

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -31,8 +31,8 @@ export type BatchFieldArrayUpdate = <T extends Function, TFieldValues extends Fi
     argB: unknown;
 }>, shouldSetValue?: boolean, shouldUpdateFieldsAndErrors?: boolean) => void;
 
-// Warning: (ae-forgotten-export) The symbol "FileList_2" needs to be exported by the entry point index.d.ts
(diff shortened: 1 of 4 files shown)


Evalon:
This is cool. Is it possible to provide default values there along with formValues? My concern is that most of the time we want to reset to current defaultValues (they can be dynamic) and preserve some data that user already provided to us. With getValues we get user data and set is as defaultValues in current implementation.

bluebill1049:
@Evalon you will be able to access `defaultVlaues` from `formState` soon: https://github.com/react-hook-form/react-hook-form/pull/8966

Evalon:
Okay, thank you for your work!

PR #9079, opened 2022-09-22
📽 feature: `validate` function to include formValues

bluebill1049:
## Benefits

- include `formValues` as part of validating function payload
- this also improves the capability by exporting validation function external without extra function dependency

## Example

CSB: https://codesandbox.io/s/pedantic-babycat-cj3q8r?file=/src/App.tsx

```tsx
register('test', {
  validate: (value: string, formValues: FormValues) => {
    return formValues.data == value;
  }
})
```

## Full Example

```tsx
type FormValues = {
  number1: number;
  number2: number;
};

function validateNumber(_: number, formValues: FormValues) {
  return formValus.number1 + formValues.number2 === 3;
}

  const {
    register,
    handleSubmit,
    formState: { isValid }
  } = useForm({
    mode: "onChange",
    defaultValues: {
      number1: 0,
      number2: 0
    }
  });

  return (
    <div>
      <form onSubmit={handleSubmit((data) => console.log(data))}>
        <input
          type="number"
          {...register("number1", {
            validate: validateNumber,
            valueAsNumber: true
          })}
          placeholder="Number1"
        />
        <input
          type="number"
          {...register("number2", {
            validate: validateNumber,
            valueAsNumber: true
          })}
          placeholder="Number2"
        />
        <p>{isValid ? "yes" : "no"}</p>
        <input type="submit" />
      </form>
    </div>
  );
}
```


timkindberg:
may be good to pass `formState` as well while you are at it. I have a use case where I have a nested react-json-schema-form that I want to validate but normal validation happens too often... I only want this nested form to validate when the user submits the form (aka formState.isValidating). So I am checking that like this:

```
  const { field } = useController({
    name,
    rules: {
      async validate() {
        // Validate runs a bit too often for this nested form I have
        if (!formState.isValidating) return
        
        // If parent form is validating then go ahead and validate nested form as well
        const isValid = await ref.current?.isValid?.({ focusFirstError: false })
        if (!isValid) return `Invalid ${_name} values`
        return true
      }
    }
  })
```

bluebill1049:
> react-json-schema-form

doesn't work with resolver, its only for build-in validation, also formState is trigger such as touched field state and dirty, state update is on the next render

timkindberg:
> doesn't work with resolver, its only for build-in validation, also formState is trigger such as touched field state and dirty, state update is on the next render

I don't fully understand... but let's ignore my use case for now.

First may I please confirm how the `validate` function works? `validate` seems to run more often than just when the form is submitting. What things "trigger" a particular field's validate function to run?

If you answer that it runs more often than submit, I think it would be nice to control validate to run or not run in the following scenarios:
- on submit
- on field change
- on watch field change
- etc, not sure all the things that trigger

Thoughts??

bluebill1049:
> If you answer that it runs more often than submit, I think it would be nice to control validate to run or not run in the following scenarios:

That's a good use case to use `trigger` as you control when to fire the validation.

timkindberg:
> That's a good use case to use trigger as you control when to fire the validation.

hmm ok but can you control when to NOT fire? Validate function seems to fire whenever it wants to. I do t want it to fire sometimes when it does and I can’t stop it. My use case is not simple though so maybe it’s not common.

Moshyfawn:
> > doesn't work with resolver, its only for build-in validation, also formState is trigger such as touched field state and dirty, state update is on the next render
> 
> I don't fully understand... but let's ignore my use case for now.
> 
> First may I please confirm how the `validate` function works? `validate` seems to run more often than just when the form is submitting. What things "trigger" a particular field's validate function to run?
> 
> If you answer that it runs more often than submit, I think it would be nice to control validate to run or not run in the following scenarios:
> 
>     * on submit
> 
>     * on field change
> 
>     * on watch field change
> 
>     * etc, not sure all the things that trigger
> 
> 
> Thoughts??

The frequency of the `validate` callback running corresponds to the `mode` passed to `useForm`:
```js
useForm({
  mode: "onSubmit"
})
``` 

Check out [this codesandbox](https://codesandbox.io/s/charming-pateu-9lgsry?file=/src/App.tsx) for a live example

timkindberg:
> The frequency of the validate callback running corresponds to the mode passed to useForm:

that’s not what I’ve experienced. We have it set to onSubmit but I see that this fields gets triggered before submitting, for example when I remove an unrelated array field item.

PR #9080, opened 2022-09-22
🖨 close #9058 avoid clone object contains prototype methods

### LINKED ISSUE #9058: useForm defaultValues -> cloneObject breaks Luxon DateTime's toJSON implementation.
### Version Number

7.35.0

### Codesandbox/Expo snack

https://codesandbox.io/s/intelligent-currying-r9hs5i?file=/src/App.tsx

### Steps to reproduce

1. Go to https://codesandbox.io/s/intelligent-currying-r9hs5i?file=/src/App.tsx
2. Enter a firstname and last name in the form and click submit
3. Look at the console output of codesandbox

As you can see in this code sandbox, when you create a model with some fields on it where one of those fields is a Luxon DateTime, it serializes properly until you bind it to a react hook form via default Values.

Once bound to a form via useForm, submitting the form has turned the DateTime into a plain old javascript object with no methods on it, there for it's toJSON implementation has been stripped off and now when you stringify it, it does not convert to an ISO String as expected.

This causes are date times to post to our API's as complex objects instead of simple ISO Strings, and none of the methods are on the DateTime anymore in the submitted form, so we have to manually recreate the DateTime's from the year/month etc data and then serialize them ourselves which is quite tedious.

I have download the source code and found the problem to be with "cloneObject"  which is located at https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/cloneObject.ts

Reading this function I am not sure what isFunction(data[key]) is supposed to be doing, but that looks like where functions are being dropped off.

### Expected behaviour

I would expect React Hook forms to not modify my object at all when I use defaultValues, I want the object to be left untouched, accept for the fields I actually register.

_No response_

### Relevant log output

```shell
--form data before serialization 
TestModel {firstName: null, lastName: null, birthDate: DateTime, constructor: Object}
--expected serialization of a date time, becomes an iso string 
{"firstName":null,"lastName":null,"birthDate":"2000-04-12T00:00:00.000-04:00"} 
because RHF cloneObject breaks DateTime, it now serializes to an object 
{"firstName":"sdsd","lastName":"sdsd","birthDate":{"ts":955512000000,"_zone":{},"loc":{"locale":"en-US","numberingSystem":null,"outputCalendar":null,"intl":"en-US","weekdaysCache":{"format":{},"standalone":{}},"monthsCache":{"format":{},"standalone":{}},"meridiemCache":null,"eraCache":{},"specifiedLocale":null,"fastNumbersCached":null},"invalid":null,"weekData":null,"c":{"year":2000,"month":4,"day":12,"hour":0,"minute":0,"second":0,"millisecond":0},"o":-240,"isLuxonDateTime":true}} 
```
```


### Code of Conduct

bluebill1049:
thanks go to @kaisermann's feedback and proposal and this package 🙏 https://github.com/jonschlinkert/is-plain-object.

ref: https://codesandbox.io/s/throbbing-leaf-ojyutz?file=/src/App.tsx

- [x] update doc
- [x] update test
- [x] pre release 

related issue: #9058
original RFC: https://github.com/react-hook-form/react-hook-form/discussions/8030#discussioncomment-3707166


kaisermann (dismissed):
💯

bluebill1049:
https://github.com/react-hook-form/react-hook-form/releases/tag/v7.37.0-next.0

PR #9098, opened 2022-09-27
Add link for V7 folder

spencerstith:


bluebill1049:
thanks for the PR, the link was deliberated removed when user is landed on v7 doc page.

PR #9113, opened 2022-09-29
Fix: Add missing blur method to field ref

finn-leonidkaida:
Adding a missing `blur` method to field ref in `useController` hook.

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -135,6 +135,7 @@ export function useController<
               setCustomValidity: (message: string) =>
                 elm.setCustomValidity(message),
               reportValidity: () => elm.reportValidity(),
+              blur: () => elm.blur(),
             };
           }
         },


Moshyfawn:
Thank you for the contribution. The `ref` content is created in accordance with the web standard. I doubt this PR is going to be merged

leonidkaida:
@Moshyfawn thank you for the response! We are using react-hook-form in React Native and are relying on the presence of `.blur()` method which is also a “standard” for an input in RN. Do you know if there is a workaround for this problem?

bluebill1049:
thanks for the PR, what usage that you try to cover in react native?

leonidkaida:
@bluebill1049 we were checking if there are any fields in a form unfilled and if there were no such fields we blurred the last field which removed the focus and dismissed the keyboard at the same time. Closing this PR because we found a workaround.

PR #9132, opened 2022-10-03
🤯 use input reference instead of custom object

bluebill1049:
ref https://codesandbox.io/s/sleepy-chatterjee-y3dj1h?file=/src/App.js


Moshyfawn:
Yea, I think it's a good idea since `Controller` is the primary way of dealing with RN inputs, where they need a lot of `ref` passing

bluebill1049:
yea 👍 and save lots of bytes for other features.

gps035:
I'm also interested in that intermediate approach. If it makes sense, I would be happy to open a PR to add an `elm` key so that users can access other methods on the input.

timkindberg:
Not sure if it's relevant but I noticed that the reason I cannot use valueAsNumber with a controlled input is because a lot of the validation for valueAsNumber also checks the `ref.value` directly instead getting the value via other means. So I have controlled inputs that have a value in the UI but rhf sees them as empty because the ref object doesn't have a `value` prop. I was excited to see this thinking maybe I could upgrade and fix the issue but i see you reverted it.

PR #9152, opened 2022-10-05
Improve naming of options forEach local

simenbrekken-visma:
Was reading the source code and got a bit confused reading the local variable name.


bluebill1049 (approved):
LGTM

PR #9170, opened 2022-10-08
🚔 check field array value type before updating

bluebill1049:
- check `fieldValues` type before updating the field array.


vanyadymousky:
@bluebill1049 Hey, why didn't you update any tests for this case?

PR #9174, opened 2022-10-08
🧦 fix accessibility issue on examples with input missing id

BogdanCerovac:

PR #9177, opened 2022-10-08
for attribute needs id on the target form element +

BogdanCerovac:
radio buttons should be grouped in fieldsets


bluebill1049:
could you apply these changes to this PR? https://github.com/react-hook-form/react-hook-form/pull/9174

PR #9299, opened 2022-10-30
fix valueAsNumber with whitespace input returning 0 instead of NaN

johnpm-12:
This change makes whitespace inputs convert to NaN instead of 0, when using valueAsNumber.

Currently, an empty input converts to NaN (as expected) but any whitespace converts to the number 0, which is confusing at best and can cause serious problems at worse. No user would expect a spacebar entered in a form to result in the number 0 sent to an API or elsewhere.

Also, 2 test are added for empty string and whitespace, when using valueAsNumber.

This issue can currently be replicated here: https://codesandbox.io/s/react-hook-form-js-forked-8dynim which is a fork of the form on the landing page.

Screenshot with nothing entered and hitting submit:
image

Screenshot with whitespace entered and hitting submit:
image

### DIFF
--- a/src/__tests__/logic/getFieldValueAs.test.ts
+++ b/src/__tests__/logic/getFieldValueAs.test.ts
@@ -13,4 +13,28 @@ describe('getFieldValueAs', () => {
       }),
     ).toBeUndefined();
   });
+  it('should return NaN when value is empty string ""', () => {
+    expect(
+      getFieldValueAs('', {
+        ref: {
+          name: 'test',
+        },
+        name: 'test',
+        valueAsNumber: true,
+        valueAsDate: false,
+      }),
+    ).toBeNaN();
+  });
+  it('should return NaN when value is whitespace string " "', () => {
+    expect(
+      getFieldValueAs(' ', {
+        ref: {
+          name: 'test',
+        },
+        name: 'test',
+        valueAsNumber: true,
+        valueAsDate: false,
+      }),
+    ).toBeNaN();
+  });
 });
--- a/src/logic/getFieldValueAs.ts
+++ b/src/logic/getFieldValueAs.ts
@@ -9,7 +9,7 @@ export default <T extends NativeFieldValue>(
   isUndefined(value)
     ? value
     : valueAsNumber
-    ? value === ''
+    ? isString(value) && value.trim() === ''
       ? NaN
       : value
       ? +value


bluebill1049:
Thanks a lot for the PR! for addition value manipulation such as trimming please refer to using `setValueAs` as it's the flexible option for set value.

johnpm-12:
Yeah, we've been using setValueAs to get around this in the meantime.

But entering a space in a number input and it getting sent as the number 0 seems like a bug.
Especially when the code doesn't do that with empty input, where it correctly returns NaN.

johnpm-12:
To clarify, this doesn't trim the input to convert it. It uses the trim function to detect if the input is nothing but whitespace, so it can return NaN

bluebill1049:
> But entering a space in a number input and it getting sent as the number 0 seems like a bug.
Especially when the code doesn't do that with empty input, where it correctly returns NaN.

input type number should not allow empty space in the first place. If you are applying setValueAsNumber on text input, i think it should use `setValueAs` instead.

johnpm-12:
> > But entering a space in a number input and it getting sent as the number 0 seems like a bug.
> > Especially when the code doesn't do that with empty input, where it correctly returns NaN.
> 
> input type number should not allow empty space in the first place. If you are applying setValueAsNumber on text input, i think it should use `setValueAs` instead.

Fair enough. We avoid using type='number' even for number inputs due to inconsistent handling on different browsers and concerns like scrolling the page could accidentally increment or decrement the input without the user realizing it.

Is valueAsNumber only meant to be used for inputs with type='number'? If so, adding clarification to the docs would be helpful.

I see the docs have type='number' in the example next to valueAsNumber, but it also says: "Only applicable and support to text input" which is unclear

johnpm-12:
This article lists the issues with type='number': https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/
We use inputmode='numeric' or inputmode='decimal' instead and validate the input when user submits.

I know this is not directly relevant to the PR, but I thought I'd clarify our use case, and why I view it as a bug

bluebill1049:
> Is valueAsNumber only meant to be used for inputs with type='number'? If so, adding clarification to the docs would be helpful.

> I see the docs have type='number' in the example next to valueAsNumber, but it also says: "Only applicable and support to text input" which is unclear

thanks, will make this clear in the doc.

PR #9369, opened 2022-11-14
fix #9366 useWatch with shadow clone `Set` data

### LINKED ISSUE #9366: issue: instanceof Set as field value is not working
### Version Number

> 7.20.2

### Codesandbox/Expo snack

https://codesandbox.io/s/determined-chaplygin-s4e4fp?file=/src/App.tsx

### Steps to reproduce

1. Go to the sandbox link: https://codesandbox.io/s/determined-chaplygin-s4e4fp?file=/src/App.tsx
2. Click on **Click me to Set value**
4. See wrong value provided in Controller field.value: **Not a set** 

### Expected behaviour

Upto version 7.20.1, linked code works as expected by returning in Controller valid `Set(['ok'])` value as is preserve in fieldValues, rather then empty `{}` object as its in version post 7.20.2

I've observed strange behavior here:  https://github.com/react-hook-form/react-hook-form/blob/master/src/useWatch.ts#L194, as this useState returns this empty `{}` rather then proper value. 

Firefox, Chrome

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
ref: https://codesandbox.io/s/romantic-ritchie-e3zuhp?file=/src/App.tsx

### DIFF
--- a/src/__tests__/__snapshots__/useWatch.test.tsx.snap
+++ b/src/__tests__/__snapshots__/useWatch.test.tsx.snap
@@ -67,7 +67,7 @@ exports[`useWatch reset with useFieldArray should return current value with radi
   {
     "options": [
       {
-        "option": "no",
+        "option": "yes",
       },
       {
         "option": "yes",
@@ -77,7 +77,7 @@ exports[`useWatch reset with useFieldArray should return current value with radi
   {
     "options": [
       {
-        "option": "no",
+        "option": "yes",
       },
       {
         "option": "yes",
--- a/src/__tests__/useWatch.test.tsx
+++ b/src/__tests__/useWatch.test.tsx
@@ -1426,6 +1426,30 @@ describe('useWatch', () => {
       expect(await screen.findByText('no')).toBeVisible();
     });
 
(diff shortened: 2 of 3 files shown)


DarryQueen:
This caused a pretty wonky and undetected change in our code. We're calling `useWatch` on an object field `objectField` in a component. We've then got a `useEffect` hook that uses `objectFieldWatchedValue` as deps and calls `setValue` on a completely independent field `otherField`. This causes an infinite render cycle, where calling `setValue` on an unrelated field causes `objectField`'s value to be updated, and vice versa.

What's the guidance here? Truthfully, this feels like odd behavior.

michaelkay-wrisk:
Hi. This change also messed up our application as we use the results of useWatch as dependencies of a useMemo to decide whether to show/hide conditional questions. This causes an infinite render cycle for us.

DarryQueen:
It's been a few months with no attention on this issue, so I assume this is not the best forum. I've filed https://github.com/react-hook-form/react-hook-form/issues/10027 to track as a GH issue, so we can move this discussion over there.

PR #9448, opened 2022-11-26
fix: fieldArray changes only notify relevant subscribers

### LINKED ISSUE #9430: issue: Modifying useFieldArray re-renders all subscribers to useFormState
First of all, hello and thanks for this amazing library! 👋 

### Version Number

7.39.5

### Codesandbox/Expo snack

https://codesandbox.io/s/react-hook-form-bug-form-state-ysd0lv?file=/src/App.js

### Steps to reproduce

1. Go to [codesandbox](https://codesandbox.io/s/react-hook-form-bug-form-state-ysd0lv?file=/src/App.js)
2. On the UI, append and remove some items on the `Field Array`
3. Notice that `Input With Error` and `Controlled Input` render counter is the same as `Field Array`

### Expected behaviour

I expected that subscribing to specific errors by providing `name` would only re-render at the hook level when that specific field has an error:
```
// re-render only if `address` field has error
const { errors } = useFormState({ name: 'address' })
```

### Investigation
I investigated a bit and I think that it happens because:
- `append` and `remove` from `useFieldArray` calls `control._updateFieldArray`
- [_updateFieldArray](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L217-L222) publish to `observers` without providing a specific name.
- [useFormState](https://github.com/react-hook-form/react-hook-form/blob/master/src/useFormState.ts#L64-L82) subscription guard calls `shouldSubscribeByName`
- [shouldSubscribeByName](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/shouldSubscribeByName.ts#L8-L11) without `signalName` returns true
- Consequently every subscriber to `useFormState` updates.

### Extra notes

If it's not the expected behavior, I'm willing to work on this 🙋‍♂️! (probably on the weekend) 

### Code of Conduct

elton-okawa:
* closes #9430
* pass `name` parameter to only notify root and `useFormState` field subcribers

### DIFF
--- a/src/__tests__/useFieldArray.test.tsx
+++ b/src/__tests__/useFieldArray.test.tsx
@@ -19,6 +19,7 @@ import {
 import { useFieldArray } from '../useFieldArray';
 import { useForm } from '../useForm';
 import { FormProvider } from '../useFormContext';
+import { useFormState } from '../useFormState';
 
 let i = 0;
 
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
Awesome update! thanks for the PR and those unit tests 🙏 🚀

Issue #9287, opened 2022-10-28
issue: FieldErrors can't handle union type


joan-saum:
### Version Number

7.38.0

### Codesandbox/Expo snack

https://codesandbox.io/s/nostalgic-butterfly-j84vtt?file=/src/App.tsx

### Steps to reproduce

1. Create a form with a union type. The types used in the union should have a property not in common
2. When trying to access to errors.property, typescript is not happy

### Expected behaviour

The errors object should have the same shape as the FormValues, so in case of union type, we could use a type predicate to correctly infer the type

With
```
type Foo = {
      id: 1;
      foo: string;
    };
type Bar = {
      id: 2;
      bar: string;
    }
type Form = Foo | Bar;
```

We should be able to do:
```
function isFoo(foobar: Foo | Bar): foobar is Foo {
  return foobar.foo !== undefined;
}

if (isFoo(errors)) {
  errors.foo // OK
}
```

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bluebill1049:
This is a limitation which we have spent quite bit time for a solution but coudln't.

PR #9540, opened 2022-12-06
Keep track of traversed types to avoid self-referencing while constructing paths for a type

SimplyLinn:
[Link to TS Playground showing off what purpose this PR has](https://www.typescriptlang.org/play?ssl=11&ssc=1&pln=12&pc=1#code/C4TwDgpgBAShDGBXATgZwJYDcIBVzQF4oBvAKCijGXQFt1gsIB+ALigDtEaAjCZAbnKVqdBtgCCyZKw5deyANoBdQRSq16jHIjAAbZmwWcefFUID23AFYyyFCpasAFEW2PzV9xy9qTkbuT5lTwdrHxptPQhDd1NPAF9PRz9bIVCrP3CAkwEheOChZAQUDHN2GTgkNC18QUTSUlBIKABRAA8AQxoopw7gAAsoIgAxdAhdABNegYAeSpKayAA+QSbodq6ovw6QacGRscnt3b7+ueLq7DxlwQaAegAqJ+eX19e70keoXuQGDt0oBAOgBzPgAOmAqCgDw+jze8LeH0+T3IDygAAlxpBkFA1lAAGbmHFFKoYbC6EBQeBlVDAZCIeAMdjAyinKEDZDmRDAwYdXH4MGoqAAZQg0GIAAFdOh2ABrb6neKoj54vYASW6uhmAGlAW1gBB2BMobTqMyoAAfWQ5AA0UAAanacMgOtg0BAJtcIKglkMHXqDUbUEIreFNNgQ1AAEKcgDuqD4ADk+owAPLWBDAIRMKAAAwAJMRtfFc0I2PaA4bjVBna6+AnPfhgxQcwp2nSOoyZo6ay63Q2vT6lJWg1A2-qXV3a-2PYO7falko0lAcwWiyXl2w18Xc5a84Xi2DC3tu06+-XZ029wuNxQtweS3vt-Ej8QTz3pxfG5AoVab7nbjhIUvSgWN+nQeBBiBUFkApKlzF0fRGShf4ATANlcX6TluV5flICFCU1h+LoaygABaPDoDAiDBlQfouUmKBeCgGU6XMVBIEZD0CKIjoXRoXs63db9vXIyioWoyCoH6OsoH+IoOgmSleENXFz2E3FzDkzBzHQCYWPYfEZXoCAhRJBYygIiBOk1UzoTzXNS3sk9iEJcw2GIbg+LYU0ZWBeJ4l9IgAHI3OCvdQvMcwwS85BgqFRynJhUhrLAIlgEohVZhwM8hIHK8iHYCA3SC0jrMDas4EUsoKWOGYZXxPgHSWbMoDVVBIn0GYcF9cqq3Zek7PsHM7HsMaFF1GUax0fRtQgEBUG6xcyJkdVNR1KAADIoF85knQmpRcpnESfRCex8k6iA5oWpalzGtg1r0GZjmu+cjq-QcWrvEhlwoCaDKgWV5vMfEayUFaHtODUnt1bbduBfbtUOwTjs+s78iBkAQbBwCUXs0VxSlGV5T2JVoQ+VL0sy0ZximU5uppyZ7X+RBRL60dGYmZndFZn0-RPIQAFV2F6eBZXEVAQHYeAABEIHxDpEF0YBud5hnDi5lnvRalZ7jxtFMV0bFMsJYkLjJcZKWpdhTQZJkWXQgZ2SwrkeTkyjBXxsUSCJuUoGOUnlUafB-akHZHq1XV2ereG91iZA3pRj6m1Kito+bPcwzEIarRjcx4yTFNsHTKxM1aoq3TLf109gIEJhqkA6oaprBa+lcoEFkdqyzxg9zzgvkGTbOS7LsacwrvhN2r-V+qT4TB2XVt20nYBTzn-Kf0XLuoXHDsp3UjfvXnRdfvbifkFP+9113K1n1fAP6Y-A-Lx-a8llvewr53J8H3vsOTlmE-PKL9RJ-nfk5KA5Zt7rxARnJeE5OyryAajFOw4a67xXt1Z+J1j53WGhwYqk97r7nXH-F0ACzjIOTq-MBH8v4vkLA-QB7155XloQBPWDxgIh0klBEEfA4LUkQpmFCiFWRO0wthN2fI8S8MoLpdgGVgBaQ6OwOS-8hRrE9miQi+BiICRwGJWR4EpJ0QYvpZirFOQcUzNxeyujID6JgSdIxV45EyWwHJXQCklJMTFGojsx1NLaV0vpBqxkDRmXNugSy9jrKbH0AlRyQoXJuQ8rFHydI-LKDtF5AAXtkeQygAqlUitFWK4UrRlJih0PJ8V7KJSDpTX4mUmFnBys4wcfpz6lUMTXKq9d2C1X-jMIQzccT2lIG3HM7VLpLWgXSVmi8fpjXGpNNRl1rqLR6uDGQbToaRy2jtLJe0awHRYYfU6p8LozSuvNbZJ9iH7PWi9eaidPysM3lXUaayAaY2xjgXZbBnkwyOfDRGyMPmXN1mNDGwNQaAtxlwr2hNpR+zaWTZKzSlEh05m09WtNVZsxnhzDWRK+ZEHxULEWnZxaS2lnLBWSsVZa22ZzclOskUIgRLCNEapEwABl+UtDagAWScKmGAOBhTk2RNynlKU2hpRaXidlrK-SVCJBMGY4K5LsBALrbFrT-7XW6YEXIRq8SzNud1aBAyG51VUQa0q8doE4AUMFfQzIBjBTwTmBWugExV0WRAQQlqQ6bPuba-pdcHUjKdUsUq7R4A8wmBAUZFB-kIptEILNeqQDKCmWGpVVNVQiHDIQSMnBEKRkQEaeWMo7F7nhlW81kZuBRX0KoyMkseAIXbegYErFi3Kpxc0fuCZB5FwgCPRkfoZZ9GgFaUY+gBXoFpHuFdobFWjsysLUWdKpay3lorZW5KCVM1ZUFIQOA1U82JRVKEAAKAAlEMX0LhzB0ATPVQyLdfQ5k7mwW9ZLWWCCAA)

Using this library in conjunction with other libraries such as Firebase Firestore, which sometimes have recursive structures (in my case `DocumentReference->parent->parent->parent->parent->[...]`) can be rather frustrating due to the eager unpacking of path keys.

My proposed solution is to pass along a union of all traversed types, and see if the type we're about to go deeper into has already been included in the current stack of traversals.

### DIFF
--- a/src/types/path/eager.ts
+++ b/src/types/path/eager.ts
@@ -1,6 +1,7 @@
 import { FieldValues } from '../fields';
 import {
   BrowserNativeObject,
+  IsExactlyAssignable,
   Primitive,
   UnPackAsyncDefaultValues,
 } from '../utils';
@@ -9,29 +10,29 @@ import { ArrayKey, IsTuple, TupleKeys } from './common';
 
 /**
  * Helper type for recursively constructing paths through a type.
+ * This actually constructs the strings and recurses into nested
+ * object types.
+ *
  * See {@link Path}
  */
 type PathImpl<K extends string | number, V, TraversedTypes> = V extends
   | Primitive
   | BrowserNativeObject
   ? `${K}`
-  : V extends TraversedTypes
-  ? [Extract<V, TraversedTypes>] extends [Extract<TraversedTypes, V>]
-    ? `${K}`
-    : `${K}` | `${K}.${Path<V, TraversedTypes | V>}`
-  : `${K}` | `${K}.${Path<V, TraversedTypes | V>}`;
+  : // Check so that we don't recurse into the same type
+  // by ensuring that the types are mutually assignable
+  // mutually required to avoid false positives of subtypes
+  IsExactlyAssignable<V, TraversedTypes> extends true
+  ? `${K}`
+  : `${K}` | `${K}.${PathInternal<V, TraversedTypes | V>}`;
 
 /**
- * Type which eagerly collects all paths through a type
- * @typeParam T - type which should be introspected
- * @typeParam TraversedTypes - types which have already been traversed to avoid infinite
- * recursion
- * @example
- * ```
- * Path<{foo: {bar: string}}> = 'foo' | 'foo.bar'
- * ```
+ * Helper type for recursively constructing paths through a type.
+ * This obsucres the internal type param TraversedTypes from exported contract.
+ *
+ * See {@link Path}
  */
-export type Path<T, TraversedTypes = never> = T extends ReadonlyArray<infer V>
+type PathInternal<T, TraversedTypes = T> = T extends ReadonlyArray<infer V>
   ? IsTuple<T> extends true
     ? {
         [K in TupleKeys<T>]-?: PathImpl<K & string, T[K], TraversedTypes>;
@@ -42,6 +43,18 @@ export type Path<T, TraversedTypes = never> = T extends ReadonlyArray<infer V>
     }[keyof T];
 
 /**
+ * Type which eagerly collects all paths through a type
+ * @typeParam T - type which should be introspected
+ * @example
+ * ```
+ * Path<{foo: {bar: string}}> = 'foo' | 'foo.bar'
+ * ```
+ */
+// We want to explode the union type and process each individually
+// so assignable types don't leak onto the stack from the base.
+export type Path<T> = T extends any ? PathInternal<T> : never;
+
+/**
  * See {@link Path}
  */
 export type FieldPath<TFieldValues extends FieldValues> = Path<
@@ -50,6 +63,9 @@ export type FieldPath<TFieldValues extends FieldValues> = Path<
 
 /**
  * Helper type for recursively constructing paths through a type.
+ * This actually constructs the strings and recurses into nested
+ * object types.
+ *
  * See {@link ArrayPath}
  */
 type ArrayPathImpl<K extends string | number, V, TraversedTypes> = V extends
@@ -59,31 +75,23 @@ type ArrayPathImpl<K extends string | number, V, TraversedTypes> = V extends
   : V extends ReadonlyArray<infer U>
   ? U extends Primitive | BrowserNativeObject
     ? never
-    : V extends TraversedTypes
-    ? [Extract<V, TraversedTypes>] extends [Extract<TraversedTypes, V>]
-      ? never
-      : `${K}` | `${K}.${ArrayPath<V, TraversedTypes | V>}`
-    : `${K}` | `${K}.${ArrayPath<V, TraversedTypes | V>}`
-  : V extends TraversedTypes
-  ? [Extract<V, TraversedTypes>] extends [Extract<TraversedTypes, V>]
+    : // Check so that we don't recurse into the same type
+    // by ensuring that the types are mutually assignable
+    // mutually required to avoid false positives of subtypes
+    IsExactlyAssignable<V, TraversedTypes> extends true
     ? never
-    : `${K}.${ArrayPath<V, TraversedTypes | V>}`
-  : `${K}.${ArrayPath<V, TraversedTypes | V>}`;
+    : `${K}` | `${K}.${ArrayPathInternal<V, TraversedTypes | V>}`
+  : IsExactlyAssignable<V, TraversedTypes> extends true
+  ? never
+  : `${K}.${ArrayPathInternal<V, TraversedTypes | V>}`;
 
 /**
- * Type which eagerly collects all paths through a type which point to an array
- * type.
- * @typeParam T - type which should be introspected
- * @typeParam TraversedTypes - types which have already been traversed to avoid infinite
- * recursion
- * @example
- * ```
- * Path<{foo: {bar: string[], baz: number[]}}> = 'foo.bar' | 'foo.baz'
- * ```
+ * Helper type for recursively constructing paths through a type.
+ * This obsucres the internal type param TraversedTypes from exported contract.
+ *
+ * See {@link ArrayPath}
  */
-export type ArrayPath<T, TraversedTypes = never> = T extends ReadonlyArray<
-  infer V
->
+type ArrayPathInternal<T, TraversedTypes = T> = T extends ReadonlyArray<infer V>
   ? IsTuple<T> extends true
     ? {
         [K in TupleKeys<T>]-?: ArrayPathImpl<K & string, T[K], TraversedTypes>;
@@ -92,6 +100,19 @@ export type ArrayPath<T, TraversedTypes = never> = T extends ReadonlyArray<
   : {
       [K in keyof T]-?: ArrayPathImpl<K & string, T[K], TraversedTypes>;
     }[keyof T];
+
+/**
+ * Type which eagerly collects all paths through a type which point to an array
+ * type.
+ * @typeParam T - type which should be introspected.
+ * @example
+ * ```
+ * Path<{foo: {bar: string[], baz: number[]}}> = 'foo.bar' | 'foo.baz'
+ * ```
+ */
+// We want to explode the union type and process each individually
+// so assignable types don't leak onto the stack from the base.
+export type ArrayPath<T> = T extends any ? ArrayPathInternal<T> : never;
 
 /**
  * See {@link ArrayPath}
--- a/src/types/path/eager.ts
+++ b/src/types/path/eager.ts
@@ -11,28 +11,34 @@ import { ArrayKey, IsTuple, TupleKeys } from './common';
  * Helper type for recursively constructing paths through a type.
  * See {@link Path}
  */
-type PathImpl<K extends string | number, V> = V extends
+type PathImpl<K extends string | number, V, TraversedTypes> = V extends
   | Primitive
   | BrowserNativeObject
   ? `${K}`
-  : `${K}` | `${K}.${Path<V>}`;
+  : V extends TraversedTypes
+  ? [Extract<V, TraversedTypes>] extends [Extract<TraversedTypes, V>]
+    ? `${K}`
+    : `${K}` | `${K}.${Path<V, TraversedTypes | V>}`
+  : `${K}` | `${K}.${Path<V, TraversedTypes | V>}`;
 
 /**
  * Type which eagerly collects all paths through a type
  * @typeParam T - type which should be introspected
+ * @typeParam TraversedTypes - types which have already been traversed to avoid infinite
+ * recursion
  * @example
  * ```
  * Path<{foo: {bar: string}}> = 'foo' | 'foo.bar'
  * ```
  */
-export type Path<T> = T extends ReadonlyArray<infer V>
+export type Path<T, TraversedTypes = never> = T extends ReadonlyArray<infer V>
   ? IsTuple<T> extends true
     ? {
-        [K in TupleKeys<T>]-?: PathImpl<K & string, T[K]>;
+        [K in TupleKeys<T>]-?: PathImpl<K & string, T[K], TraversedTypes>;
       }[TupleKeys<T>]
-    : PathImpl<ArrayKey, V>
+    : PathImpl<ArrayKey, V, TraversedTypes>
   : {
-      [K in keyof T]-?: PathImpl<K & string, T[K]>;
+      [K in keyof T]-?: PathImpl<K & string, T[K], TraversedTypes>;
     }[keyof T];
 
 /**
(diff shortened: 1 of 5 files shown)


bluebill1049:
thanks for the PR, could you share some insight or a code example to demonstrate the existing constraint?

SimplyLinn:
@bluebill1049 

[Here's a TS playground with the original implementation](https://www.typescriptlang.org/play?#code/C4TwDgpgBAShDGBXATgZwJYDcIBVzQF4oBvAKCijGXQFt1gsIB+ALigDtEaAjCZAbnKVqdBtgCCyZKw5deyANoBdQRSq16jHIjAAbZmwWcefFUID23AFYyyFCpasAFEW2PzV9xy9qTkbuT5lTwdrHxptPQhDd1NPAF9PRz9bIVCrP3CAkwEheOChZAQUDHN2GTgkNC18QUTSUlBIKABRAA8AQxoopw7gAAsoIgAxdAhdABNegYAeSpKayAA+QSbodq6ovw6QacGRscnt3b7+ueLq7DxlwQaAegAqJ+eX19e70keoXuQGDt0oBAOgBzPgAOmAqCgDw+jze8LeH0+T3IDygAAlxpBkFA1lAAGbmHFFKoYbC6EBQeBlVDAZCIeAMdjAyinKEDZDmRDAwYdXH4MGoqAAZQg0GIAAFdOh2ABrb6neKoj54vYASW6uhmAGlAW1gBB2BMobTqMyoAAfWQ5AA0UAAakshg69QajaghFbwppsJ6oAAhTkAd1QfAAcn1GAB5awIYBCJhQAAGABJiNr4kmhGxU+nM5bk2mM2C03sZo7M7c4ULrtAg-10PBBkDQcgKVTzLp9Iyof8AWA2bj+pzubz+ZAhRK1j8ulAcFAALTjusNptQVD9LmTKC8KAyunmVCQRkQCaTiCdTUQIVJ29CsvEQnmNjEbgdfzruky4HxeJOogAORPgBBZAeY5hgm+yAATed7Qh8F5gESwDLgqsw4P+c6uoaxqwECExlBSxwzDK+J8A6SwJlAaqoJE+gzBh2Huri9LXvYUCJnY7H2AouoynOOj6NqEAgKgDFLEoC4yOqmo6lAABkn5msCdo4LxSgrGk9j5HREDCaJ4lKFpbAyXoMzHPpdqOtmJBaRQvF7uwUCyiJ5j4nOknSacGpmbqimmt+qnqZp7H5C5IBuR5VYotCIpiiQUoyvKexKvBpCIchqGjOMUynAx2WTPa-yIBAUIXm6uEFRMRW6CVqCYWWQgAKrsL08CyuIqAgOw8AACIQPiHSILowA1XV+WHNVxWlZRmnIg8QqYro2KoYSxIXGS4yUtS7CmgyTIsgOAzssOXI8lAfJrIKsWiuKiVylAxwpcqjT4I9Ug7KZWq6uVOEml+5pWrEyBWZh9pMcafremIbEFoG5ghuGkbYDGVhxlR7AQNgyA2eDv3MXAHQEewREfSAJHsGROJNZRFCJk1ENQtDjBw8GobIBGMOo+j7GJpj2PGYWeZJgWubFmmT15RWWYUDmRbxCWxCS7M0vRQtsW1lA9aNs2IJ8O21JdnGvZdqyx1DiO52XW92trkh+64uYF1Oe+yA7EKV2TtO76zvOS54rbgwbluEw7tA+6ckecanuel5RLBMtog+T4vlBbABcyyh2m+ABe2TyMov6YWBEFQSBVol5BHQ5zBsW3onCFtEhvyocrZyMUQ8747hhPE6Tbvk6R5HWXT1G0YJEDiYzLElVpnF2TxfFObp+liRhnlsG3PnfQpSmBXOwUhNpCgryJa8SYLW+yRZImgzZXHsQ5-HhZFOAb+9A9fXJ-kAypB-ahpI+YVXLuTfmrIUt0ErSgem3VKMJ0pN0yniKqbcJo5TGqVaeVUMH1WdKg5qrUOjtU6t1PqA0hojRwWgwq016qkDmvNBECJYRojVGGAAMmwlo1EACyTgowwBwMKNKcImHwkbs3FCyDJo4OdJUIkEwZgZz-h0dgIBNIZRbniG+lIiDA0EJoqRb0aK6QYtPXuhEQDEVUeozCwNp5qQAvoZkAwAJGVHoNXQoYbJ0hKgYxBWi3qnwMl3fUf08JE0sdYtRSxMLtHgLVCYk8hAv1ATaFJIDnYgGUPQ-xkjULM2wEMP0nAux+kQEaAaMoY4UCtMokpgQca1J3OBfQqi-RdR4J2P03B0DAn3HkpBb14aIw5sjCA3NGTOl6n0aAVpRj6HYegWkBYFkQEGYE5oLU2odS6j1fqg1hqjVodQqatUZrFIoDgbBtDp4AAoACUQwnQuHMHQUMFMqZQBphxb5UA2DXJkbQwQQA)

Typescript gets very sad at the recursion.

src/types/path/eager.ts
```diff
@@ -11,28 +11,34 @@ import { ArrayKey, IsTuple, TupleKeys } from './common';
...
  * See {@link Path}
  */
-type PathImpl<K extends string | number, V> = V extends
+type PathImpl<K extends string | number, V, TraversedTypes> = V extends
   | Primitive
   | BrowserNativeObject
   ? `${K}`
-  : `${K}` | `${K}.${Path<V>}`;
+  : V extends TraversedTypes
+  ? [Extract<V, TraversedTypes>] extends [Extract<TraversedTypes, V>]
```
felixschorer, line 19: I think this can be simplified to `V extends Extract<TraversedTypes, V>`.

- `Extract<V, TraversedTypes>` should always be `V` due to the preceding `V extends TraversedTypes`.
- `V` also can't be a union due to the preceding `V extends TraversedTypes`. Wrapping it in a tuple should be unnecessary.

Can you please add a comment why this conditional is required? I.e. to make sure that `V` is actually part of `TraversedTypes` and not only a subtype of any of the types in `TraversedTypes`.

src/types/path/eager.ts
```diff
@@ -46,33 +52,45 @@ export type FieldPath<TFieldValues extends FieldValues> = Path<
...
   | Primitive
   | BrowserNativeObject
   ? never
   : V extends ReadonlyArray<infer U>
   ? U extends Primitive | BrowserNativeObject
     ? never
-    : `${K}` | `${K}.${ArrayPath<V>}`
-  : `${K}.${ArrayPath<V>}`;
+    : V extends TraversedTypes
+    ? [Extract<V, TraversedTypes>] extends [Extract<TraversedTypes, V>]
```
felixschorer, line 63: See my other comment.

src/types/path/eager.ts
```diff
@@ -46,33 +52,45 @@ export type FieldPath<TFieldValues extends FieldValues> = Path<
...
+      ? never
+      : `${K}` | `${K}.${ArrayPath<V, TraversedTypes | V>}`
+    : `${K}` | `${K}.${ArrayPath<V, TraversedTypes | V>}`
+  : V extends TraversedTypes
+  ? [Extract<V, TraversedTypes>] extends [Extract<TraversedTypes, V>]
```
felixschorer, line 68: See my other comment.

src/types/path/eager.ts
```diff
@@ -11,28 +11,34 @@ import { ArrayKey, IsTuple, TupleKeys } from './common';
...
  * @typeParam T - type which should be introspected
+ * @typeParam TraversedTypes - types which have already been traversed to avoid infinite
+ * recursion
  * @example
  * ```
  * Path<{foo: {bar: string}}> = 'foo' | 'foo.bar'
  * ```
  */
-export type Path<T> = T extends ReadonlyArray<infer V>
+export type Path<T, TraversedTypes = never> = T extends ReadonlyArray<infer V>
```
felixschorer, line 34: Is there a benefit to exposing the `TraversedTypes` type argument to users? I would introduce a type alias to hide it. E.g. `Path<T>` -> `PathRecursive<T, TraversedTypes>` -> `PathRecursiveImpl<K, T, TraversedTypes>`

src/types/path/eager.ts
```diff
@@ -46,33 +52,45 @@ export type FieldPath<TFieldValues extends FieldValues> = Path<
...
  * @typeParam T - type which should be introspected
+ * @typeParam TraversedTypes - types which have already been traversed to avoid infinite
+ * recursion
  * @example
  * ```
  * Path<{foo: {bar: string[], baz: number[]}}> = 'foo.bar' | 'foo.baz'
  * ```
  */
-export type ArrayPath<T> = T extends ReadonlyArray<infer V>
+export type ArrayPath<T, TraversedTypes = never> = T extends ReadonlyArray<
```
felixschorer, line 84: See my other comment.

felixschorer (commented):
Can you please add some type tests [here](https://github.com/react-hook-form/react-hook-form/blob/master/src/__typetest__/path/eager.test-d.ts)? E.g. one test each (`Path` and `ArrayPath`) for a simple recursive type and one test each where the sub type check is covered.

bluebill1049:
Thanks a lot for the review @felixschorer would you be able to address those comments @SimplyLinn whenever you got a chance 🙏

SimplyLinn:
(I have not forgotten, nor am I ignoring this, I found an edge case while writing the tests, and while I have a plan and have mostly addressed the edge case, a corner case emerged from the edge case fix, and then I ended up busy for the weekend. I'll probably have an update on Monday)

bluebill1049:
thank you very much @SimplyLinn no pressure on the PR ❤️

src/types/utils.ts
```diff
@@ -77,6 +77,26 @@ export type IsAny<T> = 0 extends 1 & T ? true : false;
...
+ * IsEqual<string, 'foo'> = false
+ * IsEqual<'foo', string> = false
+ * IsEqual<'foo' | 'bar', 'foo'> = boolean // 'foo' is assignable, but 'bar' is not (true | false) -> boolean
+ * ```
+ */
+export type IsEqual<T1, T2> = T1 extends T2
+  ? (<G>() => G extends T1 ? 1 : 2) extends <G>() => G extends T2 ? 1 : 2
+    ? true
+    : false
+  : false;
```
SimplyLinn, line 98: I'm not 100% on the naming here. Well, for a while it was a strictly equals check, which is how it got its name, but I meant to rename it to something more descriptive of what it does, but couldn't figure out a good name for what it does.

To describe in long form:
It explodes potential unions of T1 and checks if they're mutually assignable to T2

Which means that if this resolves to `true`, this would be possible
```ts
let var1: T1 = null as any;
let var2: T2 = null as any;
function assignVar2ToVar1() {
  var1 = var2;
}
function assignVar1ToVar2() {
  var2 = var1;
}
```

if it resolves to `false`, the this would never be possible, and if it resolves to `boolean`, it's possible for some of the cases in the T1 union, but not others.

It does not support unions in T2.

All I know is that "IsEqual" is most likely not the right name, but I'm really struggling with concisely naming what this type does

felixschorer: I think `IsEqual`  is descriptive enough. I appreciate that you've added all those examples. One of the examples seems to be duplicated though.

Can you please give me some pointers how the type works? So far I could only figure out how to parse it 😅 
```ts
export type IsEqual<T1, T2> = T1 extends T2
  ? (<G>() => (G extends T1 ? 1 : 2)) extends (<G>() => (G extends T2 ? 1 : 2))
    ? true
    : false
  : false;
```
- [`(<G>() => (G extends T1 ? 1 : 2)) extends (<G>() => (G extends T2 ? 1 : 2))`](https://www.typescriptlang.org/play?ssl=7&ssc=52&pln=6&pc=1#code/C4TwDgpgBAKhDOwA8MB8UC8UkHFUAoBKTdfHKCAD2AgDsATeWKAfigEYoAuKAJkIBQAgMYB7Woig1E7HnERIA5ADNRoxeiy0ArgBtdUAIZNDtEEabzkiAE4BLWgHNUI8ZOnBechNeD2nmlA6+hZGZqFWSqrqLmISwFI+AMzeCipqGphBegbGYeZ5kekxrvGJiAAsqcimIIHBuSbhhT5RGbFuCR4ArNVtMVkNobURrbUuQA) seems to check for strict equality. How does it differ from [`((_: T1) => T1) extends ((_: T2) => T2)`](https://www.typescriptlang.org/play?#code/C4TwDgpgBAKhDOwBMAeGA+KBeKAKA+gFywCU2mMAUJQMYD2AdolMAsAIzFyKoDkAZnTq9MOBgFcANpKgBDeHIYg5C7shSIATgEsGAc3S1GzVjy5tUW3QexQJ0lYuXzYFlAKEijTYCzYBmcx53QWFROykZF1klRzU+UK96Hz9EABYg9RiQcPsohWy4tw8w7xM2AFZMhM9cyMdCl3iUbPQgA)? My first guess would have been `any`, but that's not it. The latter would not work as expected when strict mode is disabled, I guess?
- The only purpose of `T1 extends T2` seems to be distributing over `T1`?

You've mentioned some edge case in your previous comment. I guess the type is the way it is to address that edge case? Can you give an example? I would be good if you could add a type test to test that edge case.

SimplyLinn: The purpose for extends T1 extends T2 is also a wild guess at performance. Allowing TS to know that anything in T1 that isn't extending T2 can immediately be resolved into `false`.

Whether this has real world effects I do not know, but I feel like short circuiting that would be more performant than exploding every case into the more complex check.

As for the weird type itself, I think you're right. I never keep that equality check in my head and constantly have to look it up. I couldn't find my normal example when I searched this time, but this one worked fine. But I'll your neater variant later today

SimplyLinn: Actually, [I just checked your example equality check](https://www.typescriptlang.org/play?ts=4.9.4#code/C4TwDgpgBAKgFgSwHYHMoF4oG8oICYBcUAzsAE7IoDcUSAhgLYRGkWpQC+VAUMsBGQBmdAMbQAInWB1s3KLUbMS5Sj3nBEqYgH4iAQTJk6IADw56TFivYcAfGtz8GO-YeMn4le9w7duoSCgASWIAUQBHAFc6ABsAYTokPHwpCA8ARgAaWAAmWwxYdKgIAA9+JOJcuShtKAAKOroiGHSASgx8lvbS8rxKuoAjZpz29E6c6vla8kiISagiYRjiOflF2JWef3BoEIjo+MjDCCRgDOyYPIKW4rKTvqqp+pMAcVs60fyX297Km9qikQRj97pVXu9PlBvj1QbkalBAVAJvInjNVij1st0ZjNn4AtAYBBSAkkil+AU9lFYiTknhUiZJNJsgA5AD2SGZkRiMToAxiaUZdAA2gByDSUYgigC6tiFSEiDAGAhlPHxsCJwDiRzIJ2AFLCVMOx1ODKkdBZ7M53N5-NN0lF4q00tl8sVyu8QA) and it fails on one case relevant to this codebase. The weird implementation using the generics is working however, so I just found the actual reason I picked this version over any other!

And with this comment, my mental "to-do" list of this PR is concluded and I'm back in the state of awaiting feedback.

felixschorer (approved):

bluebill1049:
thanks very much for the hard work @SimplyLinn and the review from @felixschorer

cdbattags:
@bluebill1049 does this help at all with https://github.com/react-hook-form/react-hook-form/issues/9287?

PR #9555, opened 2022-12-08
fix(useWatch): default value for array of inputs

### LINKED ISSUE #9530: issue: error when using `useWatch` with a name array and default values as object
### Version Number

v7.40.0

### Codesandbox/Expo snack

https://stackblitz.com/edit/react-ts-beok4h?file=App.tsx

### Steps to reproduce

1. Modify the TS doc example for `useWatch` so it uses an array as `name`, as mentioned in the comment next to it
2. Update `defaultValues` to an object following TS directives

```ts
function FirstNameWatched({ control }: { control: Control<FormInputs> }) {
  const [firstName, lastName] = useWatch({
    control,
    name: ['firstName', 'lastName'],
    defaultValue: { firstName: 'default', lastName: 'default' },
  });

  return (
    <p>
      Watch: {firstName} {lastName}
    </p>
  );
}
```
3. See the runtime error
```
object is not iterable (cannot read property Symbol(Symbol.iterator))
```

No errors when passing an array as defaultValues but then it generates other TS errors. 

### Expected behaviour

Should not crash, or type should be updated 

_No response_

### Relevant log output

_No response_

### Code of Conduct

adesurirey:
Closes #9530

### DIFF
--- a/src/useWatch.ts
+++ b/src/useWatch.ts
@@ -181,12 +181,10 @@ export function useWatch<TFieldValues extends FieldValues>(
   });
 
   const [value, updateValue] = React.useState<unknown>(
-    isUndefined(defaultValue)
-      ? control._getWatch(name as InternalFieldName)
-      : control._getWatch(
-          name as InternalFieldName,
-          defaultValue as DeepPartialSkipArrayKey<TFieldValues>,
-        ),
+    control._getWatch(
+      name as InternalFieldName,
+      defaultValue as DeepPartialSkipArrayKey<TFieldValues>,
+    ),
   );
 
   React.useEffect(() => control._removeUnmounted());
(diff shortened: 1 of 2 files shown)


src/useWatch.ts
```diff
@@ -183,7 +183,10 @@ export function useWatch<TFieldValues extends FieldValues>(
   const [value, updateValue] = React.useState<unknown>(
     isUndefined(defaultValue)
       ? control._getWatch(name as InternalFieldName)
-      : defaultValue,
+      : control._getWatch(
+          name as InternalFieldName,
+          defaultValue as DeepPartialSkipArrayKey<TFieldValues>,
+        ),
   );
```
bluebill1049, line 190: I think we can simplify to the following

```tsx
const [value, updateValue] = React.useState<unknown>(
  control._getWatch(
    name as InternalFieldName,
    defaultValue as DeepPartialSkipArrayKey<TFieldValues>,
  ),
);
```

bluebill1049 (commented):
Looking great! thanks a lot for the fix.

bluebill1049 (approved):
Let's go! Thanks.

adesurirey:
Thank you @bluebill1049!

adesurirey:
@bluebill1049 when will it be released?

bluebill1049:
I am planning to do the final release of the year next week.

PR #9710, opened 2023-01-03
Remove ref.value check

### LINKED ISSUE #9709: issue: calling setValue breaks future onChange calls from a Controller
### Version Number

7.41.3

It appears this broke in 7.41.2, it works in 7.41.1

### Codesandbox/Expo snack

https://codesandbox.io/s/dawn-glade-g2bibl?file=/src/App.js

### Steps to reproduce

1. Go to https://codesandbox.io/s/dawn-glade-g2bibl?file=/src/App.js
2. Change the select dropdown from `one` to `two`
3. Enter a value for the username
4. Submit the form

An error appears that the username is required, however, if you look at the console output username has a value

### Expected behaviour

Calling setValue does not break future onChange calls

if I switch from `setValue` to `reset` then things work as expected, however, we make heavy use of setValue in a number of locations on some complex forms.

Chrome

### Relevant log output

_No response_

### Code of Conduct

Moshyfawn:
Closes: #9709

### DIFF
--- a/src/__tests__/logic/validateField.test.tsx
+++ b/src/__tests__/logic/validateField.test.tsx
@@ -36,13 +36,15 @@ describe('validateField', () => {
       },
     });
 
+    const input = document.createElement("input");
+
     expect(
       await validateField(
         {
           _f: {
             mount: true,
             name: 'test',
-            ref: { type: 'text', value: '', name: 'test' },
+            ref: input,
             required: true,
             valueAsNumber: true,
           },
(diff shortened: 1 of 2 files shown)


Moshyfawn:
@bluebill1049 I haven't updated the tests yet, since I'm not sure what this change was for, but it breaks the current func of `setValue`: the `ref.value` isn't updated

bluebill1049:
thanks @Moshyfawn I have fixed the issue, thanks for the patch will release it right away.

PR #9746, opened 2023-01-10
🌏 feature: support global error type

bluebill1049:
### `setError` support global error

- RFC: https://github.com/react-hook-form/react-hook-form/discussions/9691
- close feature request #7661

```tsx
const { setError } = useForm()

setError('root.serverError', { type: 404, message: 'Not Found'})
setError('root.userError', { type: 'IncorrectDataType', message: 'user provided incorrect type'})
```


src/types/errors.ts
```diff
@@ -33,16 +33,25 @@ export type DeepRequired<T> = T extends BrowserNativeObject | Blob
 export type FieldErrorsImpl<T extends FieldValues = FieldValues> = {
   [K in keyof T]?: T[K] extends BrowserNativeObject | Blob
     ? FieldError
+    : T[K] extends GlobalError
```
Chaoste, line 36: Unfortunately, this breaks error types in cases where `formState.error` is erroneously assumed to be a `GlobalError` while it's just a nested object.

In my case, I use
`const { formState: { errors } } = useFormContext<MyCustomField>()`
where `MyCustomField` has properties which are again objects.

When reading out a specific error, it looks somehow like this:
`const specificError = errors.definition?.configuration?.recipients as FieldError | undefined` 

This was already difficult before this PR which force me to add the user-defined type cast at the end. But now, it will fail already when reading `definition` with this error:
```
Property 'configuration' does not exist on type 'Partial<{ type: string | number; message: string; }>'.ts(2339)
```

Chaoste: @bluebill1049 for visibility

PR #9753, opened 2023-01-11
Fix formState.invalid not works at first render.

mango906:
https://github.com/react-hook-form/react-hook-form/discussions/9750

formState.invalid should be classified by whether the field is valid or not, not by error.

Now, We're setting this standard by an error right now. 
The error value changes only when there is an action, So invalid value is not correct at initial render.

By not using the invalid value through error, This value will be normal.

I don't think it needs to be deprecated anymore because this value is normal.

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -144,7 +144,7 @@ export function useController<
       {
         invalid: {
           enumerable: true,
-          get: () => !!get(formState.errors, name),
+          get: () => !get(formState.isValid, name),
         },
         isDirty: {
           enumerable: true,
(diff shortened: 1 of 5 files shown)


src/useController.ts
```diff
@@ -144,7 +144,7 @@ export function useController<
       {
         invalid: {
           enumerable: true,
-          get: () => !!get(formState.errors, name),
+          get: () => !get(formState.isValid, name),
```
mango906, line 147: I was wrong about this part. There is no object with valid field, I don't know how to figure this out.

mango906:
I found a new way. I'll open another PR.

mango906:
https://github.com/react-hook-form/react-hook-form/pull/9755

PR #9755, opened 2023-01-11
Add formState.invalidFields property

mango906:
formState.invalid should be classified by whether the field is valid or not, not by error.

Now, We're setting this standard by an error right now.
The error value changes only when there is an action, So invalid value is not correct at initial render.

By not using the invalid value through error, This value will be normal.

So, i add formState.invalidFields and calculate fieldState.invalid with that.
https://github.com/react-hook-form/react-hook-form/pull/9753#discussion_r1067029226

But there is one problem in my work. In useFieldArray, the function I added doesn't work properly. I would appreciate your help on this part.

Also, there seems to be some deficiencies in the code. I would appreciate your help in this part, too.

Thanks :) 

related PR: https://github.com/react-hook-form/react-hook-form/pull/9753
discussion: https://github.com/react-hook-form/react-hook-form/discussions/9750

### DIFF
--- a/src/types/controller.ts
+++ b/src/types/controller.ts
@@ -13,12 +13,6 @@ import {
 } from './';
 
 export type ControllerFieldState = {
-  /**
-   * @deprecated check `fieldState.error` instead
-   * ```jsx
-   * {fieldState.error && <p>{fieldState.error.message}</p>}
-   * ```
-   */
   invalid: boolean;
   isTouched: boolean;
   isDirty: boolean;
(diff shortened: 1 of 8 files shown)


mango906:
https://github.com/react-hook-form/react-hook-form/discussions/9750

src/types/controller.ts
```diff
@@ -13,12 +13,6 @@ import {
...
 
 export type ControllerFieldState = {
-  /**
-   * @deprecated check `fieldState.error` instead
-   * ```jsx
-   * {fieldState.error && <p>{fieldState.error.message}</p>}
-   * ```
-   */
   invalid: boolean;
   isTouched: boolean;
```
bluebill1049, line 17: Would you like to create a PR for this? We can merge this in first.

mango906: Sure

mango906: @bluebill1049 https://github.com/react-hook-form/react-hook-form/pull/9760 here is PR

PR #9758, opened 2023-01-12
🌏 feature: support global error type #9746

### LINKED ISSUE #7661: issue: handle global error in forms

bluebill1049:
- 🌏 feature: support global error type #9746

original PR: https://github.com/react-hook-form/react-hook-form/pull/9746

close #7661


bluebill1049:
@shartte `root` error does not persist, you can use custom error. 

`setError('nameNotRelatedToField', { type: 'error' })`

PR #9795, opened 2023-01-18
simple fix to allow numeric fields to use regex patterns

rekliner:
I saw you added external validation functions a few days ago but it seemed to me a less confusing path to simply treat numbers as strings if the user wants to use a pattern.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-hook-form",
   "description": "Performant, flexible and extensible forms library for React Hooks",
-  "version": "7.42.1",
+  "version": "7.42.2",
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.mjs",
   "umd:main": "dist/index.umd.js",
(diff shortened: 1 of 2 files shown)


bluebill1049:
Thanks for the PR, however by following the spec. Input number is not included.

> The pattern attribute is an attribute of the [text](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text), [tel](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel), [email](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email), [url](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url), [password](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/password), and [search](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search) input types.

rekliner:
I understand that that's a core goal of your code but it's never going to be a perfect match.
> Embraces native HTML form [validation](https://react-hook-form.com/get-started#Applyvalidation)
This will not interfere with any functionality from the spec, only extend it.  You are already extending it with attributes such as "valueAsNumber". 

This example is displayed first thing on your readme quickstart:
`<input {...register('age', { pattern: /\d+/ })} />` 
It implies support for handling a numeric input and is ambiguous about type.  Your solution, only mentioned in the changelog, is to suggest writing a custom validator function for any inputs using valueAsNumber.  Another solution would be to handle everything as text and manually convert those fields to numeric values on submission.  This is what your readme example would require.  They are both cumbersome and unintuitive compared to simply extending the pattern input.

PR #9809, opened 2023-01-19
stronger typing to document non-spec attribute uses

rekliner:
Per #9795 the intention is to disallow the use of the pattern attribute when using the valueAsNumber attribute.  Also the valueAsNumber and ValueAsDate attributes would never be simultaneously true.  This updates the typing to enforce those choices and prevent confusion.


bluebill1049 (approved):
LGTM! Thanks for this.

Issue #9844, opened 2023-01-31
Allow setting of formState.isSubmitted, formState.isSubmitSuccessful


craig-o-curtis:
**Is your feature request related to a problem? Please describe.**
This is a continuation of [https://github.com/react-hook-form/react-hook-form/issues/2859](https://github.com/react-hook-form/react-hook-form/issues/2859). onSubmit automatically sets the `isSubmitted`, `isSubmitSuccessful`, but we need control over these.
Using v7.40.0

**Describe the solution you'd like**
We need a way to programmatically specify `formState.isSubmitted` and `formState.isSubmitSuccessful` if we have things like confirmation prompts and they are cancelled.

The `formState.isSubmitted` flag should also be optionally controlled - if we don't want it set to `true` if a user goes back to the form and continues filling it out.

**Describe alternatives you've considered**
In this [forked codepen](https://codesandbox.io/s/react-hook-form-handlesubmit-v6-forked-wvry6b?file=/src/index.jsx) I am able to set `formState.isSubmitSuccessful` for example, but this directly on the object, whereas we should have an exported `setFormState` method to set what we need.

bluebill1049:
We should be able to have better control once the Form component's API is getting stable: 

```tsx
<Form
  onSubmit={() => {
    throw new Error('run time error') // isSubmitted or isSubmitSuccessful will be false
  }}
  reject={() => {
    // callback here
  }}
/>

```

PR #9849, opened 2023-01-31
improve error handle with handleSubmit callback

bluebill1049:
This PR is trying to address the handleSubmit runtime error.

```tsx
const onSubmit = () => {
  throw new Error('data')
}

const onError = (error, e, runTimeError) => {
  console.log(runTimeError)
}

handleSubmit(onSubmit, onError)
```

`isSubmitSuccessful` will return false if any runtime error 
`isSubmitting` wouldn't be stuck at true.

Somewhat related to this: #9844

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -490,7 +490,7 @@ export type Subjects<TFieldValues extends FieldValues = FieldValues> = {
 };
 
 // @public (undocumented)
-export type SubmitErrorHandler<TFieldValues extends FieldValues> = (errors: FieldErrors<TFieldValues>, event?: React_2.BaseSyntheticEvent) => any | Promise<any>;
+export type SubmitErrorHandler<TFieldValues extends FieldValues> = (errors: FieldErrors<TFieldValues>, event?: React_2.BaseSyntheticEvent, error?: unknown) => any | Promise<any>;
 
 // @public (undocumented)
 export type SubmitHandler<TFieldValues extends FieldValues> = (data: TFieldValues, event?: React_2.BaseSyntheticEvent) => any | Promise<any>;
@@ -787,8 +787,8 @@ export type WatchObserver<TFieldValues extends FieldValues> = (value: DeepPartia
 
 // Warnings were encountered during analysis:
 //
-// src/types/form.ts:97:3 - (ae-forgotten-export) The symbol "AsyncDefaultValues" needs to be exported by the entry point index.d.ts
-// src/types/form.ts:419:3 - (ae-forgotten-export) The symbol "Subscription" needs to be exported by the entry point index.d.ts
+// src/types/form.ts:98:3 - (ae-forgotten-export) The symbol "AsyncDefaultValues" needs to be exported by the entry point index.d.ts
+// src/types/form.ts:420:3 - (ae-forgotten-export) The symbol "Subscription" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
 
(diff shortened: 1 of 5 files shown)


amalitsky (commented):
I think this approach further couples the submission lifecycle with validation lifecycle which, IMO, is not helpful because this library is more about client form validation, not API requests. It makes everything more entangled and harder to reason about.

I'm saying this because now user is expected to handle not only validation errors in his `SubmitErrorHandler`, but also exceptions of `SubmitHandler` he is passing separately. This doesn't seem natural to me. After all, (s)he could have handled any exceptions in `SubmitHandler` if (s)he wanted to.

Oh, and one more thing (almost overlooked, lol) - now you are effectively back to swallowing exceptions again (unless user provides proper `SubmitErrorHandler` that is). Which is not a good thing.

bluebill1049:
Thanks for the feedback.

> I think this approach further couples the submission lifecycle with validation lifecycle which, IMO, is not helpful because this library is more about client form validation, not API requests. It makes everything more entangled and harder to reason about.

We are moving towards that direction.

bluebill1049:
> Oh, and one more thing (almost overlooked, lol) - now you are effectively back to swallowing exceptions again (unless user provides proper SubmitErrorHandler that is). Which is not a good thing.

I think it's probably better for this approach, especially the new Form component that i am building and it works nicely with error handling.

```tsx
<Form 
  action="/api/test"
  reject={(e) => { console.log(e) }}
/>
```

bluebill1049:
> Oh, and one more thing (almost overlooked, lol) - now you are effectively back to swallowing exceptions again (unless user provides proper SubmitErrorHandler that is). Which is not a good thing.

This actually gives the user better control over the isSubmitScuessful state, because the user can now throw inside the callback and prevent the submission from being successful. Of course, this will need to be documented.

```diff
handleSubmit(async () => {
  const reponse = await fetch()
  if (reponse.status > 200) {
    throw new Error('Error status') // which would mark isSubmitSuccessful false
  }
})
```

amalitsky:
> This actually gives the user better control over the isSubmitScuessful state, because the user can now throw inside the callback and prevent the submission from being successful. Of course, this will need to be documented.

What if the exception was not foreseen? No way for the library to know what kind of exception is good or bad. Ability to handle exceptions is a part of the language (and callback passed), not sure why library has to cater for it and what does it benefit from it. Aside from distinguishing between success and failure - the rest is, IMO, user's responsibility. Swallowing exceptions (or not) is another problem this library has to decide - like it or not. Mb config setting to allow swallowing - again, in order to push this decision to the user.

Hot take: May be, `isSubmitting` and friends don't belong here? Since library has virtually no control over it?

I guess it's your call whether or not you want to get into network requests game. But if you do want (and, I assume, users are actively asking for it), I would probably split these two completely (on the interface side of things). Mb separate packages even. Both are very complex subjects. Yes, there will be no compound effect but you will get robustness and clean API which is easy to reason about and subject specific.

_Low footprint is a feature, not a bug!_

amalitsky:
On a second thought - we can't really go by exception/unresolved promise route if we don't _swallow_ them. If that's a part of API we shouldn't punish user for using it.
Mb introduce explicit callbacks to notify form of being submitted and ultimate submission resolution instead? That brings up a question of why form has to be aware of those at all...

bluebill1049:
@amalitsky valid points, let's cancel this PR for now.

bluebill1049:
> On a second thought - we can't really go by exception/unresolved promise route if we don't swallow them. If that's a part of API we shouldn't punish user for using it.

I think it makes sense to address all of these in the Form component which does take care of the submission part, and it probably makes more sense to provide an error/reject callback there.

PR #9921, opened 2023-02-12
🗑️ remove JSX message check

bluebill1049:


lennerd:
@bluebill1049 are there any more information why JSX support for error messages was removed?

Btw. the documentation still says its okay to use JSX inside validate functions (via the code examples, when „Register with validation and error message“ is selected): https://react-hook-form.com/api/useform/register/

bluebill1049:
@lennerd will follow this up later: https://github.com/react-hook-form/core/issues/2

PR #9940, opened 2023-02-15
Catch errors in handleSubmit's onValid callback and reflect state

mryraghi:
This PR handles cases where the async function provided to `handleSubmit` throws an error, setting `isSubmitting` back to `false` and without increasing `submitCount.`

I stumbled upon a similar scenario when the API returned a 500 code, keeping the form's 'Submit' button disabled (since `isSubmitting = false`) and preventing the user from retrying.

### DIFF
--- a/src/__tests__/useForm/handleSubmit.test.tsx
+++ b/src/__tests__/useForm/handleSubmit.test.tsx
@@ -524,4 +524,28 @@ describe('handleSubmit', () => {
 
     await waitFor(() => expect(onSubmit).toBeCalled());
   });
+
+  it('should set isSubmitting back to false if onValid throws an error', async () => {
+    const { result } = renderHook(() =>
+      useForm<{
+        test: string;
+      }>(),
+    );
+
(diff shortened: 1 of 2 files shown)


Moshyfawn:
This is the expected behaviour updated in #9688 to avoid "swallowing" errors thrown in `onValid`.

All promises need to be handled in user-land inside `onValid`

mryraghi:
I get it, but how can I take advantage of `isSubmitting` (because I'm returning a promise to `handleSubmit`) while ensuring the error is caught?

Moshyfawn:
> I get it, but how can I take advantage of `isSubmitting` (because I'm returning a promise to `handleSubmit`) while ensuring the error is caught?

Here's a quick example: [sandbox](https://codesandbox.io/s/serene-sunset-5wn87z?file=%2Fsrc%2FApp.tsx)

mryraghi:
Thanks for the example, I appreciate it!

Issue #9945, opened 2023-02-15
issue: Submission state no longer reset on handleSubmit error.


adriantrunzo:
### Version Number

7.43.1

### Codesandbox/Expo snack

https://codesandbox.io/s/submit-form-state-tef0l3?file=/src/App.tsx

### Steps to reproduce

The 7.42.0 release introduced a regression by [removing a try/catch/finally block](https://github.com/react-hook-form/react-hook-form/pull/9603/files#diff-5a85c1643c71f45e050ce584d7c0db2bdf2e56f8a7dae86d74fedf01be2d2cabL1065) in the definition of `handleSubmit`. The call to `_subjects.state.next` in the finally block ensured that the form submission state (eg `isSubmitting` etc) was reset if the `onValid` handler encountered an error. Starting in 7.42.0, if the `onValid` handler throws an error, such as from a bad API request, the form incorrectly remains in the submitting state and does not update `isSubmitting`, `submitCount`, etc.

To reproduce, take a look at the attached CodeSandbox:
1. Verify that the text `isSubmitting: false, submitCount: 0` appears above the input. This shows the initial form state.
2. Click the submit button.
3. You should see the form state text above the input change to `isSubmitting: true, submitCount: 0`.
4. Wait for the error to be thrown/promise rejected.
5. Verify that the submit button remains disabled (because `isSubmitting` isn't reset to `false`).
6. Verify that the form state text continues to show `isSubmitting: true, submitCount: 0`.

### Expected behaviour

The form state should be updated no matter if the `onValid` handler throws an error or not. If you fork the CodeSandbox and change the installed version of RHF to 7.41.5, you will see the correct behavior: `isSubmitting` is updated to false, `submitCount` incremented, and the submit button no longer disabled.

### What browsers are you seeing the problem on?

Firefox, Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Moshyfawn:
Expected behaviour from #9688.

Instead of swallowing user-land errors like promise rejections, RHF let's you handle your errors now.

Moshyfawn:
@bluebill1049 people seem to be reporting this one quite a bit 🤔

Moshyfawn:
The way to handle promise rejections: https://github.com/react-hook-form/react-hook-form/pull/9940#issuecomment-1431585324

bluebill1049:
[...]

PR #10036, opened 2023-03-02
issue #9945 proposal to use callback to report runtime error

bluebill1049:
report a runtime error with an invalid callback.

related #9945


supervanya:
@bluebill1049 why was this one closed? I am realizing that `isSubmitting` form state is basically useless, and we used to rely on that functionality. https://github.com/react-hook-form/react-hook-form/issues/9945 seems to have a very abrupt closure as well.

PR #10088, opened 2023-03-10
fix: typos

ghost:


voxlol (approved):

bluebill1049 (approved):

kensaadi:
useFieldArray is great until you start nesting or adding dynamic rules.

At that point, the main issue becomes predictability:
- updates are not always intuitive
- validation timing can get tricky
- UI can desync from form state

Feels like this is where RHF shifts from being a helper to something you need to actively control.

bluebill1049:
> useFieldArray is great until you start nesting or adding dynamic rules.
> 
> At that point, the main issue becomes predictability:
> 
> * updates are not always intuitive
> * validation timing can get tricky
> * UI can desync from form state
> 
> Feels like this is where RHF shifts from being a helper to something you need to actively control.

I get where you’re coming from—nested useFieldArray + dynamic rules can definitely feel unpredictable at first.

That said, it’s hard to really evaluate or debug without something concrete to look at. A minimal reproducible CodeSandbox would help a lot here (report at main repo). Not your full project—just the smallest possible example that shows the issue.

From experience, a lot of the “desync” or “weird validation timing” cases usually come down to how keys, defaultValues, or controlled inputs are wired up rather than a fundamental limitation of RHF itself.

If you can strip it down and still reproduce the problem, that would make the discussion much more actionable.

Issue #9934, opened 2023-02-14
issue: formState.isLoading always returns true


crashtech:
### Version Number

7.43.1

### Codesandbox/Expo snack

https://codesandbox.io/s/nice-gould-tzmvfo?file=/src/App.js

### Steps to reproduce

1. Just load the page
2. Using `FormProvider` makes `isLoading` always return true
3. Even trying to override using `<FormProvider {...hook} formState={{ ...hook.formState, isLoading: false }}>` doesn't work

### Expected behaviour

The default value for `isLoading` should be false, especially if `defaultValues` hasn't been provided, or if it hasn't been assigned to a Promise.

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Moshyfawn:
Your sandbox link contains the default RHF template; no reproduction. Please update your sandbox with the reproduction code

crashtech:
Done!

Moshyfawn:
I'm a bit rusty with how the change to the `formState` due to the new async `defaultvalues`, but there's something going with [how the default `isLoading: true`](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L1231) and it's never handled if `defaultValues` are not async.

@bluebill1049, I might need your help on this one

bluebill1049:
thanks @Moshyfawn i will look into this one.

bluebill1049:
@crashtech If you are not using async defaultValues, what's the use case for subscribing to `isLoding`?

Moshyfawn:
> @crashtech If you are not using async defaultValues, what's the use case for subscribing to `isLoding`?

That's a good question, actually. I guess, it feels like `isSubmitting`?

bluebill1049:
Yea, if you are providing static `defaultValues`, then you don't need `isLoading`, and we can save the extra byte in the lib to skip the logic check.

crashtech:
I'm connecting react hook forms with Semantic UI, so I'm forwarding this prop to the UI component. It should be consistent, regardless of using asynchronous default values.

Plus, a form here doesn't have default values, but select options are loaded asynchronously.

Moshyfawn:
> Plus, a form here doesn't have default values, but select options are loaded asynchronously.

Select options loading don't have anything to do with `isLoading` unless they're a part of `defaultValues`

crashtech:
[...]

PR #10095, opened 2023-03-13
fix: default for isLoading form state

adamtowle:
Mentioned here https://github.com/react-hook-form/react-hook-form/issues/9934.

This pull request fixes an issue with the `isLoading` functionality. Currently, the default value for `defaultValues` is always set to `true`, even if it's a static value. I have a component abstraction that subscribes to `isLoading` to display a loading state, but the default value was causing issues with the component's behaviour.

To address this, I've updated the `defaultValues` default to `true` only if it's a function, which allows the component to support both async and static `defaultValues` without any unexpected behaviour.

### DIFF
--- a/src/__tests__/useForm.test.tsx
+++ b/src/__tests__/useForm.test.tsx
@@ -1907,6 +1907,14 @@ describe('useForm', () => {
     });
   });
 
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):
LGTM. thanks

PR #10125, opened 2023-03-17
fix(appendErrors): can take an array of errors

jorisre:
As discussed, `appendErrors` is used by resolvers and need to pass en array of errors (`Record<string, string[]>`) when `validateAllCriteria: true`. This PR fix the type and add a test for the case


jorisre:
@bluebill1049 Api Extrator failed, not sure what should I do 🤔

PR #10244, opened 2023-04-07
🐞 fix #10238 return default values in watch and useWatch when reset is called with empty object

### LINKED ISSUE #10238: issue: useWatch not triggered after form reset() to empty object
### Version Number

7.43.9

### Codesandbox/Expo snack

https://stackblitz.com/edit/react-ts-zogyhj?file=App.tsx,style.css,UseWatchComponent.tsx,WatchComponent.tsx,ValuesComponent.tsx

### Steps to reproduce

1. Go to https://stackblitz.com/edit/react-ts-zogyhj?file=App.tsx,style.css,UseWatchComponent.tsx,WatchComponent.tsx,ValuesComponent.tsx
2. Click on Reset To Values
3. Form has values and useWatch, and watch methods get triggered with values
4. Click on Reset to Nothing
5. The watch method from useFormContext appears to have been reset, but the useWatch doesn't get reset, even when defined before the reset methods. On second click of Reset to Nothing, useWatch gets triggered with empty object.

### Expected behaviour

I would expect useWatch to be triggered with the updated values that are used with the form.reset method the first time it is set to an empty object, in the same way that it is triggered when the form is set to an object with values.

Chrome, Edge

### Relevant log output

_No response_

### Code of Conduct

devakrishna33:
closes #10238

### DIFF
--- a/src/__tests__/useForm/reset.test.tsx
+++ b/src/__tests__/useForm/reset.test.tsx
@@ -18,6 +18,7 @@ import {
 import { useController } from '../../useController';
 import { useFieldArray } from '../../useFieldArray';
 import { useForm } from '../../useForm';
+import { useWatch } from '../../useWatch';
 
 jest.useFakeTimers();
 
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
Thank you 🙏 LGTM

PR #10300, opened 2023-04-18
src/useFormContext.tsx: remove unnecessary as unknown as cast

ryota-murakami:
Removed unnecessary as unknown as cast from `src/useFormContext.tsx` .


bluebill1049 (approved):
Oh wow, that's awesome! Thanks for the fix @ryota-murakami ❤️

PR #10357, opened 2023-04-28
chore: adding link to v7 examples folder in readme

nicholasgriffintn:
Noticed that the examples has a link to V6 but not to V7, this change just adds a link to the v7 folder in the examples readme.

(super simple change, I know, thought that I'd fix it when I noticed it).


bluebill1049:
Thanks for the PR, it's actually made so user now the active page that they are on.

nicholasgriffintn:
Ahh I see, I guess  I didn't clock that :)

PR #10395, opened 2023-05-13
feature: add resetField to Controller (#9857)

sanduluca:
This pull request introduces a new feature that was requested in [this discussion thread](https://github.com/orgs/react-hook-form/discussions/9857). 
#9857 

The feature adds support for accessing `resetField` method from `Controller` but without possibility to reset some other path outside Controller.

The method was added at root of Controller render and useController return. 

resetField was not added to `ControllerRenderProps` (field property) because it is not a valid DOM attribute and it will generate warning/error when developer will use spread syntax for `field`
Ex: 
```jsx
const { field } = useController();
<input  {...field} />
```

### Naming
I thought about method naming. I came up with 3 options: `reset`, `resetField`, `resetValue`. Decided to go with `resetField`

### Motivation

As RHF is a DX/UX library, reseting the field is a must have feature. Luckily the library provides such a method but at a form level. This feature comes to introduce `resetField` at Controller level with possibility to reset only controlled name/path.

Passing resetField from form level forces you to write more code, by passing the method via props or by using a new hook `useFormContext` in every component (which forces you to wrap your form with a provider). This 2 methods can encourage developers to create side effects in their controlled component by giving them access to reset any form field.

### Conclusion

This feature enhances the functionality of React Hook Form and addresses the specific need outlined in the discussion thread. The implementation follows the existing coding standards and conventions of the project.

Please review and provide feedback on the changes.


JatinAntier (approved):

JatinAntier (approved):

sanduluca:
@bluebill1049 Can you take a look to this PR and come with a feedback pls ?

sanduluca:
@bluebill1049 Can you take a look to this PR and come with a feedback pls ?

PR #10450, opened 2023-05-26
Require index/indices for useFieldArray#remove

andrew-atwood-infinitusai:
It's unclear to me what the use case is for calling `remove` without a specific index or set of indices.

Requiring this parameter can prevent the very easy mistake to make of binding `remove` directly to a handler without passing `index` from the field.

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -187,7 +187,7 @@ export function useFieldArray<
     });
   };
 
-  const remove = (index?: number | number[]) => {
+  const remove = (index: number | number[]) => {
     const updatedFieldArrayValues: Partial<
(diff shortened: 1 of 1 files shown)


Moshyfawn:
Calling the `remove` method without specifying an index value removes all fields.

andrew-atwood-infinitusai:
Strange– it only removes the `0` index field for me, hence feeling like a bug. Will investigate further.

Moshyfawn:
> Strange– it only removes the `0` index field for me, hence feeling like a bug. Will investigate further.

[Can't repro](https://codesandbox.io/s/cold-thunder-g54gut?file=/src/App.tsx)

PR #10463, opened 2023-05-30
Add field array context

Moshyfawn:
#7544 is very stale and hard to merge. Here's a PR with the most relative field array context updates to either merge instead of #7544 or to resolve & update the OG PR.

### DIFF
--- a/src/__tests__/useFormContext.test.tsx
+++ b/src/__tests__/useFormContext.test.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 
 import { useController } from '../useController';
+import { useFieldArray } from '../useFieldArray';
 import { useForm } from '../useForm';
 import { FormProvider, useFormContext } from '../useFormContext';
 import { useFormState } from '../useFormState';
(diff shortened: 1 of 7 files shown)


src/__tests__/useFormContext.test.tsx
```diff
@@ -229,4 +230,45 @@ describe('FormProvider', () => {
...
+      const testField = useFieldArray({
+        control: methods.control,
+        name: 'test',
+      });
+
+      return (
+        <FormProvider
+          {...methods}
+          fieldArrays={{
+            test: testField,
```
Moshyfawn, line 262: I'm getting a type error here
```bash
Type 'UseFieldArrayReturn<FieldValues, "test", "id">' is not assignable to type 'UseFieldArrayReturn<FieldValues>'.
  Types of property 'prepend' are incompatible.
    Type 'UseFieldArrayPrepend<FieldValues, "test">' is not assignable to type 'UseFieldArrayPrepend<FieldValues, string>'.
      Type 'string' is not assignable to type '"test"'.ts(2322)
```

It worked in the OG PR but the types were changed since then.

I can't seem to resolve it 😔

Moshyfawn: I'm not sure if adding additional field array names is a good idea

Moshyfawn: This way, `FormProviderProps` needs to be aware of the field array names path generics..

bluebill1049: Sorry, for the late reply. Let's hold this feature for now, perhaps we should find a better approach.

Moshyfawn: No worries! Yea, I fill a bit weird about the way it is currently. I'm dedicating some time to think it through, but no luck yet

stanbaker: Hey, is there any plan to revisit this feature?

evgeniyworkbel: we want this feature

evgeniyworkbel:
@Moshyfawn @bluebill1049 could you please add this feature to your tasks table? It is useful feature to handle useFieldArray more convenient

PR #10471, opened 2023-05-31
Fix: removed code from the README updater which was trying to read a …

protofarer:
…non-existent `docs/Helpers.yaml`. This was causing the corresponding build job to fail. For reference, this file was removed in commit b93c873603de9d68b9e4601723614d5997eed354 as part of a README cleanup effort.


bluebill1049 (approved):
LGTM! thanks

bluebill1049 (approved):
LGTM! thanks

PR #10487, opened 2023-06-01
Upgrade api-extractor version

Moshyfawn:
Running `pnpm api-extractor` version `^7.33.7` yields this warning which can lead to TS generated API discrepancies down the line.
```bash
Analysis will use the bundled TypeScript version 4.8.4
*** The target project appears to use TypeScript 5.0.2 which is newer than the bundled compiler engine; consider upgrading API Extractor.
```

This PR simply upgrades the `@microsoft/api-extractor` package to a greater version with bundled TS version that's equal to RHF TS version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
   "author": "<bluebill1049@hotmail.com>",
   "license": "MIT",
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
LGTM!

PR #10488, opened 2023-06-01
Use up-to-date terser plugin

Moshyfawn:
[`rollup-plugin-terser`](https://www.npmjs.com/package/rollup-plugin-terser) has been deprecated in favour of [@rollup/plugin-terser](https://www.npmjs.com/package/@rollup/plugin-terser)

This PR simply removes the former plugin and installs the latter. The API is identical, at least for the options used by RHF, so no further updates required.


bluebill1049 (approved):
LGTM.

PR #10490, opened 2023-06-02
Refresh value when name property changes in useWatch hook

### LINKED ISSUE #9732: useWatch to react to name property change

`useWatch` and `useFieldArray` are not reacting to name property change. That is, the name they receive on mount is the one they are subscribed to throughout the lifetime of the component. The issue was originally discussed [here](https://github.com/react-hook-form/react-hook-form/issues/9727). You can also check [this CSB](https://codesandbox.io/s/winter-voice-bnpmd6).

Remounting the child component based on `key` works fine, but it's particularly frustrating when you don't want the parent component to rerender in case of changes that would otherwise concern the child only. For such cases we use `jotai` and use the `atom` from `jotai` in the child only, so only it is subscribed to changes in the atom, which results in the parent not rerendering. Sure, you can wrap the child in another component (which is what we did), but I still feel this is unnecessary friction and noise in the code.

`useWatch` and `useFieldArray` to react to changes in the `name` property.

Wrapping the `useWatch`-dependent component in another one and remounting the latter with `key` from the original parent.

Using `useFieldArray` which results in the same result.

Strangely to me (I haven't delved into the RHF code), if I do this, it works fine:

```javascript
const Comp: React.FC<{ idx: number }> = ({ idx }) => {
  const values = useWatch<FormData, `steps.${number}.elements`>({ name: `steps.${idx}.elements` });
  const arrayValues = useFieldArray<FormData, `steps.${number}.elements`>({ name: `steps.${idx}.elements` });

  return <p>{JSON.stringify(values)}</p>;
};
```

However, removing the `useFieldArray` subscription or printing the values received from it (`<p>{JSON.stringify(arrayValues)}</p>`), with or without `useWatch` still does not work as one would expect.

**Additional context**

[This CSB](https://codesandbox.io/s/winter-voice-bnpmd6) illustrates it. In order to make it work, you need to change this:

```javascript
      <Comp idx={idx} />
```
to
```javascript
      <Comp key={idx} idx={idx} />
```

flash1293:
Fixes https://github.com/react-hook-form/react-hook-form/issues/9732

This PR makes `useWatch` response to changes of the `name` property - every time it changes (doing a deep equal check for arrays of field names), the current value is checked and the returned value is updated if the value changes due to this.


lmossman:
@bluebill1049 the issue that this PR fixes is causing us problems for our platform, so getting this merged would be very helpful, if you are able to take a look!

MichalIrzylowski:
This PR solves our issue too, it would be good to see it on master :)

MichalIrzylowski (approved):

flash1293:
@bluebill1049 is there anything missing on this PR to get it merged?

flash1293:
I would still like to get this one merged

PR #10493, opened 2023-06-03
Fix incorrect test case in useController.test.tsx

mrsekut:
I noticed that one of the test cases in `useController.test.tsx` was not correctly set up, causing it to be not executed as expected.
In the original setup, the test runner function was incorrectly nested inside the App component, resulting in the test not being executed as expected.

I've fixed the nesting issue and revised the test case.

In the revised test case, I've moved the test runner function out of the App component. Additionally, to eliminate the warning about an input changing from uncontrolled to controlled, I added defaultValue: '' to the options passed to useController.

These changes should make the test run correctly and remove the warning.

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -301,6 +301,7 @@ describe('useController', () => {
       const { field } = useController({
         control,
         name: 'test',
+        defaultValue: '',
       });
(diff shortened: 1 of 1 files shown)


bluebill1049 (approved):
Thanks!

Issue #10511, opened 2023-06-08
Focused field as part of form state


akmjenkins:
**Is your feature request related to a problem? Please describe.**
I want to be able to watch which field is focused from the react-hook-form so I don't have to write a lot of user-land code

**Describe the solution you'd like**
Add a property on the form state called `focused` which is a string that is the flattened path to the field name

**Describe alternatives you've considered**
Implementing it in userland

**Additional context**
I was going to try to PR this, but before I did, I wanted input from the maintainers about potential problems that might make this feature request/PR untenable.

akmjenkins:
@bluebill1049 sorry if I didn't provide enough detail here - I know this feature isn't currently a part of react-hook-form, is there other info I should have provided in the feature request?

bluebill1049:
Just putting as pending feature column not closing, will consider to include this feature.

PR #10514, opened 2023-06-09
First attempt at detecting focused field

akmjenkins:
Relates to https://github.com/react-hook-form/react-hook-form/issues/10511

Needs tests + consensus, as it does add an extra render because of the FormState


akmjenkins (commented):
Ok, got the concept of react-hook-form down now

I _**think**_ it's relatively straightforward to do this, but there's a caveat that I don't like - I would like to connect blur and focus events so I don't update twice (or preferably late) when a user is "tabbing" through the fields in the form.

I think an opinion has to be made about the behaviour, because there's benefits and drawbacks no matter which way you go

Updating focused state to a non-empty string immediately should be fine (focus), but updating focused state to an empty string (blur) could probably be delayed by a tiny amount, maybe even as a configuration option

Also a bunch of e2e tests are failing that are probably due to my naive implementation 😆

PR #10543, opened 2023-06-15
fix: useFormContext ReturnType

swallowtail62:
Thanks for providing a good library. It helps me a lot.

Now, I noticed that the return type of `useFormContext` is not proper, so I fixed it. Take a look at it when you have time.

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -621,7 +621,7 @@ export function useForm<TFieldValues extends FieldValues = FieldValues, TContext
 export type UseFormClearErrors<TFieldValues extends FieldValues> = (name?: FieldPath<TFieldValues> | FieldPath<TFieldValues>[] | readonly FieldPath<TFieldValues>[] | `root.${string}` | 'root') => void;
 
 // @public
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):

reports/api-extractor.md
```diff
@@ -621,7 +621,7 @@ export function useForm<TFieldValues extends FieldValues = FieldValues, TContext
 export type UseFormClearErrors<TFieldValues extends FieldValues> = (name?: FieldPath<TFieldValues> | FieldPath<TFieldValues>[] | readonly FieldPath<TFieldValues>[] | `root.${string}` | 'root') => void;
 
 // @public
-export const useFormContext: <TFieldValues extends FieldValues, TransformedValues extends FieldValues | undefined = undefined>() => UseFormReturn<TFieldValues>;
+export const useFormContext: <TFieldValues extends FieldValues, TContext = any, TransformedValues extends FieldValues | undefined = undefined>() => UseFormReturn<TFieldValues, TContext, TransformedValues>;
```
BatuhanW, line 624: Shouldn't this `TContext` added to the end of the params? Currently our builds are failing, I suspect this is the reason, because our 2nd parameter isn't context.

PR #10584, opened 2023-06-23
test(generated-id): setting window performance undefined without ts-ignore

mateoguzmana:
Setting window performance undefined without ts-ignore


bluebill1049 (approved):
LGTM

PR #10586, opened 2023-06-24
🧑‍🎓 close #10585 update controller value on each render

### LINKED ISSUE #10585: issue: reset Field Array in combination with `shouldUnregister: true` at `useForm` level and `Controller` doesn't give correct submission data
### Version Number

7.45.0

### Codesandbox/Expo snack

https://codesandbox.io/s/react-hook-form-reset-usefieldarray-forked-n3xksj

### Steps to reproduce

1. Click on 'Submit'
2. See submission data

### Expected behaviour

The submission data should be the reset values: `{ names: [{ firstName: 'Glenn' }, { firstName: 'Ania' }] }` instead of the default values: `{ names: [{ firstName: 'John' }, { firstName: 'Glenn' }] }`.

Chrome

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:


andyg-12:
Hey @bluebill1049, this change has broken the sync between _setValue_ and _watch_.
Since v7.45.1, my watched variables don't get refreshed with _setValue_.
My app works fine with the previous version 7.45.0.
Thanks!

PR #10589, opened 2023-06-24
feat: Add trigger to ControllerRenderProps

glekner:
For some fields, we need a method to trigger a revalidation `onChange`, while the form's mode is `onSubmit`.
This is very helpful since we can't dynamically change the `mode` of the form:


glekner:
@bluebill1049 thoughts?

glekner:
@bluebill1049 last try

PR #10590, opened 2023-06-25
fix: `trigger` should use passed names to construct resolver options

vonagam:
Documentation states that passing a name or names is a way to select what fields are getting validated, but builder for resolver options does not receive those names, always producing full fields and all names, so resolver cannot skip unrequested checks.

### DIFF
--- a/src/__tests__/useForm.test.tsx
+++ b/src/__tests__/useForm.test.tsx
@@ -1165,8 +1165,8 @@ describe('useForm', () => {
 
         expect(resolver).toHaveBeenCalledWith(defaultValues, undefined, {
           criteriaMode: undefined,
-          fields,
-          names: ['test.sub', 'test1'],
+          fields: { test: fields.test },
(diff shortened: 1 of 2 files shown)


src/__tests__/useForm.test.tsx
```diff
@@ -1165,8 +1165,8 @@ describe('useForm', () => {
 
         expect(resolver).toHaveBeenCalledWith(defaultValues, undefined, {
           criteriaMode: undefined,
-          fields,
-          names: ['test.sub', 'test1'],
```
bluebill1049, line 1169: what happens when you did not change this line?

vonagam: Then the test fails. That's what the PR changes. If you `trigger('field')` you get only that `'field'` in `options.names` and not all of them.

bluebill1049 (approved):

jorisre (approved):

PR #10633, opened 2023-07-04
fix(types): UseFieldArray Append, Prepend, Insert, Replace

southball:
Fix types of `UseFieldArray{Append, Prepend, Insert, Replace}` so they match those of https://github.com/react-hook-form/react-hook-form/blob/97a4ed9d90f3bbeb2f397f5a07fa43d292f8f0a3/src/useFieldArray.ts.

This PR allows these functions to be used with partial objects, so it should not break existing code unless one depends on the argument type of these functions directly.


glekner:
This won't work. if you prepend with a partial value, the new element will take the missing nested value from `[0]` which is not what you want.

southball:
> This won't work. if you prepend with a partial value, the new element will take the missing nested value from `[0]` which is not what you want.

Thank you for your response!

Kindly let me check the source code to see if this PR still makes any sense, and if it does not I will close the PR.

bluebill1049:
thanks for the PR @southball yea, inputs are uncontrolled and we need default value for them to be mounted. and undefined is currently not valid.

PR #10637, opened 2023-07-05
fix(types): async defaultValues await promise value

### LINKED ISSUE #10625: issue: useForm hook is not typed correctly when using async default values
### Version Number

7.45.1

### Codesandbox/Expo snack

https://codesandbox.io/s/reverent-pond-mqh73c?file=/src/App.tsx

### Steps to reproduce

Simply provide an async function as `defaultValues`.

For instance, `handleSubmit` is typed like so:
```ts
const handleSubmit: UseFormHandleSubmit<() => Promise<{
    firstName: string;
    lastName: string;
}>>
```

### Expected behaviour

I would expect `handleSubmit` to be typed like so:
```ts
const handleSubmit: UseFormHandleSubmit<{
    firstName: string;
    lastName: string;
}>
```

Chrome

### Relevant log output

_No response_

### Code of Conduct

milhamm:
Closes #10625 

Previously, when an async `defaultValues` was passed, the field type was inferred as follows:
```ts
useForm<() => Promise<{ firstName: string, lastName: string }>>
```

This caused a problem with the form methods because it treated the type as an async function instead of the value type. This pull request fixes the async `defaultValues` to return the value directly, similar to the behavior of non-async `defaultValues`.

```ts
useForm<{ firstName: string, lastName: string }>
```

### DIFF
--- a/app/src/app.tsx
+++ b/app/src/app.tsx
@@ -11,6 +11,7 @@ import SetValue from './setValue';
 import FormState from './formState';
 import ManualRegisterForm from './manualRegisterForm';
 import DefaultValues from './defaultValues';
+import DefaultValuesAsync from './defaultValuesAsync';
 import WatchDefaultValues from './watchDefaultValues';
 import Reset from './reset';
 import TriggerValidation from './triggerValidation';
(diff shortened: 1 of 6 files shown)


bluebill1049 (approved):
Thank you for fixing and improving this. 👍

PR #10707, opened 2023-07-25
🐞 fix: add exact option for array name in useWatch.

rabbitson87:
An unwanted update occurred when using useWatch as the array name. So I used the exact option, but it didn't work with the array name. The current exact option only works when it is not an array, so the condition has been changed to work on the array. Please review the corresponding update. Thank you.

### DIFF
--- a/src/logic/shouldSubscribeByName.ts
+++ b/src/logic/shouldSubscribeByName.ts
@@ -7,9 +7,11 @@ export default <T extends string | string[] | undefined>(
 ) =>
   exact && signalName
     ? name === signalName ||
-      convertToArrayPayload(name).some(
-        (currentName) => currentName && exact && currentName === signalName,
-      )
+      (typeof name !== 'string' &&
+        Array.isArray(name) &&
+        name.some(
+          (currentName) => currentName && exact && currentName === signalName,
+        ))
     : !name ||
       !signalName ||
       name === signalName ||
--- a/src/logic/shouldSubscribeByName.ts
+++ b/src/logic/shouldSubscribeByName.ts
@@ -6,7 +6,10 @@ export default <T extends string | string[] | undefined>(
   exact?: boolean,
 ) =>
   exact && signalName
-    ? name === signalName
+    ? name === signalName ||
+      convertToArrayPayload(name).some(
+        (currentName) => currentName && exact && currentName === signalName,
+      )
     : !name ||
       !signalName ||
       name === signalName ||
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks, i will review it soon. 🙏

rabbitson87:
> thanks, i will review it soon. 🙏

Thank you for accepting my opinion :).

bluebill1049 (dismissed):
So this PR would fix the array payload with exact right?

src/logic/shouldSubscribeByName.ts
```diff
@@ -6,7 +6,10 @@ export default <T extends string | string[] | undefined>(
   exact?: boolean,
 ) =>
   exact && signalName
-    ? name === signalName
+    ? name === signalName ||
+      convertToArrayPayload(name).some(
```
bluebill1049, line 10: perhaps for better perf, we just check if payload is Array or not?

rabbitson87: First, make sure it is an array and that the factors in it match exactly the event name

bluebill1049: `Array.isArray(signalName) ? x : y`, so we split check string and array.

rabbitson87: I don't think it's a problem whether it's a string or an array without having to distinguish. What do you think happens when there's a problem?

bluebill1049: save some perf without have running `.some` each time.

rabbitson87: All right, so can you see what it'll be like this?

Before
```
name === signalName ||
      convertToArrayPayload(name).some(
        (currentName) => currentName && exact && currentName === signalName,
      )
```

After
```
name === signalName ||
      (typeof name !== 'string' &&
        Array.isArray(name) &&
        name.some(
          (currentName) => currentName && exact && currentName === signalName,
        ))
```

bluebill1049: is `Array.isArray(name)` enough?

rabbitson87: I think it's enough.

bluebill1049: Sounds good, let's do that then, we can merge this in after the fix.

rabbitson87: I'll fix it right away and merge it! thank you :)

bluebill1049: 👍

bluebill1049 (approved):
thank you 🙏

PR #10791, opened 2023-08-12
fix: Set errors formState if defaultValues fail resolver validation

martinjlowm:
Validation errors weren't propagated to the form consumer if default values failed validation.

This PR changes the concern of `_updateValid` to also consider if errors should be set on `formState`. I presume `_updateValid` is a private interface hence the rename.

The provided resolver was invoked already, but the errors would be ignored, leaving registered inputs as if they would not be in an error state.

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -51,7 +51,7 @@ export type Control<TFieldValues extends FieldValues = FieldValues, TContext = a
     _getDirty: GetIsDirty;
     _resetDefaultValues: Noop;
     _formState: FormState<TFieldValues>;
-    _updateValid: (shouldUpdateValid?: boolean) => void;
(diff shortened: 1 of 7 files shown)


bluebill1049:
thanks for the PR, because react hook form is focused on field-level validation and it looks like this change would potentially update all errors state when user interactive with a single user input.

martinjlowm:
Just curious, seeing a bit of mixed messages here if changes to this PR is a wish. As for updating all errors, that's true... and that may be desired. My team has been a bit annoyed about fields that are dependent on one another where the validation must be forced to accomplish this.

I suppose one optimization that could be considered is to track the individual errors and preserve the old state if it remains unchanged. So instead of a deepEqual and reassigning all of the `errors`-field, mutate the existing formState with fields that need updating.

gustavoabel (approved):
LGTM

PR #10879, opened 2023-09-04
🐞 fix #10878 return `disabled` field state

### LINKED ISSUE #10878: issue: `field.disabled` in `Controller` is always `undefined`
### Version Number

7.46.0

### Codesandbox/Expo snack

https://codesandbox.io/s/suspicious-leaf-xqkt32

### Steps to reproduce

1. Make use of `Controller`, set the `disabled` props to `true`.
2. In the `render` function, pass the `field` to any components while logging `field.disabled`
3. Observe that `field.disabled` is always `undefined` despite `Controller` has `disabled = true` set

### Expected behaviour

`field.disabled` should have the same value as the `disabled` props of Controller

_No response_

### Relevant log output

_No response_

### Code of Conduct

Moshyfawn:
@bluebill1049 I've moved the `disabled` state from `field` to `fieldState` while fixing the issue because it makes more sense to me. Let me know of you disagree and I'll change it back _(or update the API docs and tests if you're down)_

Closes: #10878

### DIFF
--- a/src/types/controller.ts
+++ b/src/types/controller.ts
@@ -17,7 +17,6 @@ export type ControllerFieldState = {
   isTouched: boolean;
   isDirty: boolean;
   error?: FieldError;
-  disabled?: boolean;
 };
 
 export type ControllerRenderProps<
@@ -26,6 +25,7 @@ export type ControllerRenderProps<
 > = {
   onChange: (...event: any[]) => void;
   onBlur: Noop;
+  disabled?: boolean;
   value: FieldPathValue<TFieldValues, TName>;
   name: TName;
   ref: RefCallBack;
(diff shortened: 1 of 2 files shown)


Moshyfawn:
..if returning the `disabled` even makes sense since passing `disabled` to `useController` essentially makes it a "controlled" state at which point, the user should already have access to the `disabled`state

bluebill1049:
> ..if returning the `disabled` even makes sense since passing `disabled` to `useController` essentially makes it a "controlled" state at which point, the user should already have access to the `disabled` state

yes, that's correct. However, hook form won't be able to omit the value without been notified at `useController` level.

bluebill1049 (approved):

PR #10899, opened 2023-09-07
fix controller test is dirty

pierluigigiancola:
the test for the controller that should test that it's dirty was using isTouched and focus + blur instead of isDirty and changing the value

### DIFF
--- a/src/__tests__/controller.test.tsx
+++ b/src/__tests__/controller.test.tsx
@@ -719,7 +719,7 @@ describe('Controller', () => {
           render={({ field: props, fieldState }) => (
             <>
               <input {...props} />
-              {fieldState.isTouched && <p>Input is dirty.</p>}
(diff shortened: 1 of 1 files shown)


bluebill1049:
oh nice, could you duplicate the test for touch as well? thanks 👍

pierluigigiancola:
I'm sorry, but I don't understand your request.
The test for touch is already in **controller.test** and it seems okay to me.
Are you suggesting we create a test that asserts whether the input is touched after its value is changed?

bluebill1049 (approved):

bluebill1049:
> I'm sorry, but I don't understand your request. The test for touch is already in **controller.test** and it seems okay to me. Are you suggesting we create a test that asserts whether the input is touched after its value is changed?

That's fine then 👍 thanks for the fix.

pierluigigiancola:
Thanks for the amazing library 😄

PR #10924, opened 2023-09-12
#8020 Make handleSubmit return void instead of Promise<void>

paleite:
This PR makes handleSubmit return void instead of Promise<void>, but the tests are breaking, because they assume handleSubmit should reject a Promise instead.

How this should be handled needs to be decided by the maintainers. Either:

1. Update docs, so they catch consistently
  - i.e. `<form onSubmit={handleSubmit((data) => console.log(data))}>` would become `<form onSubmit={() => handleSubmit((data) => console.log(data))().catch(() => {})}>`
2. Update the tests, so they don't catch at all
3. Pass a rejectionHandler which is called in the `catch`-part of my PR

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1056,7 +1056,7 @@ export function createFormControl<
     );
 
   const handleSubmit: UseFormHandleSubmit<TFieldValues> =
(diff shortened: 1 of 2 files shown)


camsteffen:
If I leave a worthless comment here will stale bot reopen?

PR #10937, opened 2023-09-16
🙄 close #10932 omit disabled prop when not defined

### LINKED ISSUE #10932: issue: breaking change with controller disabled feature
### Version Number

7.46.1

### Codesandbox/Expo snack

https://codesandbox.io/s/nervous-leakey-vdz6hp

### Steps to reproduce

1. Add a controlled input with Controller
2. On the render of the Controller pass the "field" object as a spread
3. Add a "disabled" prop on the input before the field spread
4. See how disabled is passed as undefined to the component
5. See on the console how the field object has a "disabled" property with the value `undefined`
image

### Expected behaviour

For backwards compatibility on the same major version, the field object should not have the disabled property when it is undefined. This way the code that passes a disabled prop directly to the controlled component won´t be affected by this change.

Chrome

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
fix #10932

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -4,6 +4,7 @@ import getEventValue from './logic/getEventValue';
 import isNameInFieldArray from './logic/isNameInFieldArray';
 import cloneObject from './utils/cloneObject';
 import get from './utils/get';
+import isBoolean from './utils/isBoolean';
 import isUndefined from './utils/isUndefined';
 import { EVENTS } from './constants';
 import {
@@ -123,7 +124,7 @@ export function useController<
     field: {
       name,
       value,
-      disabled,
+      ...(isBoolean(disabled) ? { disabled } : {}),
       onChange: React.useCallback(
         (event) =>
           _registerProps.current.onChange({


Pumpuli:
With the addition of the "disable whole form" feature, it's now technically possible to receive `{ disabled: undefined }` again.

[The change](https://github.com/react-hook-form/react-hook-form/pull/10496/files#diff-3f5bdfa7d822b4562b8e1c72dd1996197fe93478b039779bb78853ab3ae2a3fe) makes this situation possible:

```tsx
const { control } = useForm({ disabled: false });

return <Controller
  // No `disabled` prop given here
  control={control}
  render={({ field }) => {
    // `field.disabled` is `undefined`
  }}
/>;
```

Maybe this is not critical, since it can only happen when `field.disabled` is _supposed to_ be `false` (in other words, that `undefined` is never going to override anything that's _supposed to_ be truthy).

An easy fix would be to use nullish coalescing operator here:

https://github.com/react-hook-form/react-hook-form/blob/0d565fcf944621f5510da4623129ffbd238f9530/src/useController.ts#L130

bluebill1049:
> With the addition of the "disable whole form" feature, it's now technically possible to receive `{ disabled: undefined }` again.
> 
> [The change](https://github.com/react-hook-form/react-hook-form/pull/10496/files#diff-3f5bdfa7d822b4562b8e1c72dd1996197fe93478b039779bb78853ab3ae2a3fe) makes this situation possible:
> 
> ```tsx
> const { control } = useForm({ disabled: false });
> 
> return <Controller
>   // No `disabled` prop given here
>   control={control}
>   render={({ field }) => {
>     // `field.disabled` is `undefined`
>   }}
> />;
> ```
> 
> Maybe this is not critical, since it can only happen when `field.disabled` is _supposed to_ be `false` (in other words, that `undefined` is never going to override anything that's _supposed to_ be truthy).
> 
> An easy fix would be to use nullish coalescing operator here:
> 
> https://github.com/react-hook-form/react-hook-form/blob/0d565fcf944621f5510da4623129ffbd238f9530/src/useController.ts#L130

Feel free to send a PR.

PR #10939, opened 2023-09-16
feat: added smart form component

alirezakarimidev:
This component appears to be a wrapper around the useForm hook.
This wrapper aims to simplify the usage of the form handling logic and provide a more structured and reusable way to manage forms in our application.
We can easily use this component without using useForm hook directly.

Well documented
Added type declaration for it
Added some basic tests

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -50,7 +50,8 @@
         ]
       }
     ],
-    "simple-import-sort/exports": "error"
+    "simple-import-sort/exports": "error",
+    "prettier/prettier":0
   },
   "overrides": [
     {
--- a/app/pnpm-lock.yaml
+++ b/app/pnpm-lock.yaml
@@ -1,74 +1,91 @@
-lockfileVersion: 5.4
+lockfileVersion: '6.0'
 
-specifiers:
-  '@hookform/resolvers': 2.8.8
-  '@material-ui/core': ^4.12.3
-  '@types/react': ^17.0.38
-  '@types/react-dom': ^17.0.9
-  '@types/react-router-dom': ^5.3.3
-  '@types/react-select': ^5.0.1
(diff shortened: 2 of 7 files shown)


hoomantalakian:
That's an interesting feature. I think it would be useful to incorporate this feature into this Library!

Moshyfawn:
Thank you for your time and the contribution, Alireza!

Unfortunately, I don't think we want to expand the RHF API with a component wrapper around the `useForm` hook as it can easily be done in user-land.

Unless @bluebill1049 thinks otherwise.

Moshyfawn:
Additionally, there's a new [`Form` component](https://react-hook-form.com/docs/useform/form) that encompasses additional request API logic on top of the standard `useForm` API in RHF

GetWrecked99:
Thank you for your response @Moshyfawn

Honestly, it doesn't significantly alter the manner in which you employ it. It hinges upon the developer's preferred approach for implementing the form's logic.
One of the reasons I developed this was its applicability to server components, particularly the recently introduced Next.js app directory feature. You need not concern yourself with utilizing the useForm hook in server components any longer.


form-props

Form-Methods

bluebill1049:
Thanks for the awesome contribution @GetWrecked99! Apologies for the late reply. Would you be keen to publish that as your own package? we can include that in the FAQ as an external package to be used by users.

GetWrecked99:
Thank you for your response @bluebill1049 and I appreciate your feedback on my contribution. While I'm glad to hear that you found the contribution valuable, I would like to continue improving the feature before considering it as a standalone package.

I plan to work on adding more features and refining the codebase to make it even more robust and user-friendly. Once I feel it's in a more polished state, I'll certainly consider publishing it as its own package.

I'll keep you updated on the progress.

Best Regards,
Alireza

Issue #11018, opened 2023-10-07
Lazily evaluate `FieldPath<TFieldValues>` for better TS performance when using `setValue`


jussisaurio:
**Is your feature request related to a problem? Please describe.**

Typechecking `setValue` invocations results in poor TS performance with `type UseFormSetValue` when `TFieldValues` is a complex type. This is especially noticeable as sluggish VSCode autocompletion times, but also in 'offline' compilation times.

The project I discovered this behavior in is closed source, so I'm -- for now -- unfortunately only including some screenshots of a TS performance traces captured using `tsc --generateTrace`, plus some rationale and code snippets.

**Performance trace**

This is how much time it takes to typecheck a single `setValue` call in this particular project (the type ID of the type in question points to `UseFormSetValue`:

---

<img width="234" alt="image" src="https://github.com/react-hook-form/react-hook-form/assets/22619982/efb62d3f-20f6-4d72-a7fd-e0a541a73a63">

---

I believe the culprit to be this line:
https://github.com/react-hook-form/react-hook-form/blob/0d565fcf944621f5510da4623129ffbd238f9530/src/types/form.ts#L529

My (perhaps unfounded) assumption is that the `extends` constraint results in the full union of `FieldPath<TFieldValues>` being eagerly evaluated every time. 

**Describe the solution you'd like**
I believe constraining `TFieldName` to just `string` would be an acceptable solution, because the type helpers `PathValue` and friends eventually will make conditional checks to narrow the type of `TFieldName` anyway. Constraining to `string` allows lazily evaluating whether the provided type parameter conforms to a valid path, and importantly -- terminating early when it either does or doesn't. Developer experience (`the type "foobar" is not assignable to "path | path.subpath | path.anotherPath ...."`) is preserved with a conditional check:

``` 
export type UseFormSetValue<TFieldValues extends FieldValues> = <
  TFieldName extends string,
>(
  name: FieldPathValue<TFieldValues, TFieldName> extends never
    ? FieldPath<TFieldValues> // <-- "the type "foobar" is not assignable to path | path.subpath | path.anotherPath ...."
    : TFieldName,
  value: FieldPathValue<TFieldValues, TFieldName>,
  options?: SetValueConfig,
) => void;
```

`PathValue` and `FieldPathValue` do not need to constrain `P` because they do conditional type checking wrt. `T` anyway and return `never` in error cases:

```
-export type PathValue<T, P extends Path<T> | ArrayPath<T>>
[...]

PR #11019, opened 2023-10-07
TS compilation perf: lazily evaluate FieldPath for setValue

jussisaurio:
Addresses #11018


prophetitaly:
This is exactly what helped me out of many problems with too deeply nested objects! Using the library with this patch is the only viable way!

PR #11028, opened 2023-10-10
chore: improve error when not passing `control` to `useController`

MeLlamoPablo:
## Proposed Changes

This PR improves the message of the error that is generated when you don't pass `control` to `Controller` or `useController`. This error occurs if you neither pass `control` as a prop directly nor provide `control` via `FormProvider`.

Currently, you would get a rather cryptic "methods is null" error:

image

After this PR, you get a more informative error:

Screenshot 2023-10-10 at 19 03 48

Interestingly, while the console stack trace is able to pinpoint `useController` as the source for both errors, Next.js can only understand that this error happens in the code of the library if it's explicitly thrown. Before this PR, Next.js thinks that the error happens in my code.

While to most users of the library it should be obvious that you need to somehow pass `control` to `Controller`, when developing you can easily get tunnel vision and forget about basic mechanisms, as happened to me. This PR should help developers get back on track without having to spend a lot of time understanding what they did wrong.

## Type of change

Please delete options that are not relevant.

## Checklist:

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -1012,4 +1012,24 @@ describe('useController', () => {
       }),
     );
   });
+
+  it("should throw a meaningful error if the control prop isn't passed", async () => {
+    const Component = () => {
+      useController({
+        name: 'test',
+        defaultValue: '',
+      });
+
+      return null;
+    };
+
+    // Don't log the thrown error to the console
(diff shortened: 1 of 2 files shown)


MeLlamoPablo:
Solved conflicts. This PR is still relevant, bumping to avoid having it auto closed.

PR #11139, opened 2023-11-04
fix: avoid exception by possible undefined event

calhaudev:
When a user fills out a validated form and hits ENTER in the keyboard, the lib throws an exception by undefined event in onChange handler. This PR just verifies if the event exists.

image


rafaelcalhau:
hi @bluebill1049, my first time trying to contribute here. This PR is very simple, is there another requirement for this to be merged, please let me know.

Issue #11106, opened 2023-10-27
issue: formState no longer updates when some inputs are disabled


jessypouliot98:
### Version Number

7.47

### Codesandbox/Expo snack

https://codesandbox.io/s/reverent-wilbur-l952z4?from-embed=&file=/src/App.js

### required things to trigger the issue

1. one or more inputs must have `{ disabled: true }` in their register options `register("foo", { disabled: true })`
2. `useForm()` has to have ran twice before the user tries to input anything. Otherwise the main issue doesn't trigger.

### Steps to reproduce

1. rerender the component to cause `useForm()` to run again (click the button on top of the page to do this in the codesandbox)
4. isDirty is initially false (expected)
5. edit an input
6. isDirty is not changed to true (unexpected) and component has not rerendered (unexpected)

### Expected behaviour

1. rerender the component to cause `useForm()` to run again (click the button on top of the page to do this in the codesandbox)
3. isDirty is initially false (expected)
4. edit an input
5. isDirty is changed to true and component will rerender

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cs25-esc:
could you please elaborate your issue?

jessypouliot98:
> could you please elaborate your issue?

I updated the issue/codesandbox with a better description and exact way to trigger the issue

jessypouliot98:
I've opened a PR with a jest test to reproduce this issue. This was difficult to figure out, but strictMode and a rerender must happen before updating the form to trigger the issue

PR #11143, opened 2023-11-05
🐞 #11106  fix disabled prop not update dirty formState

bluebill1049:
ref: https://codesandbox.io/s/funny-keldysh-8p5pmv?file=/src/App.js


villesau:
This breaks isDirty check for us. It resets back to true even with: `reset({}, { keepValues: true, keepDirty: false });`

PR #11164, opened 2023-11-07
Upgrade Node.js Test

kotarella1110:


kotarella1110:
The workflow that compares package sizes is failing because the composite action added in this PR does not exist in the base branch. Merging this PR will resolve the issue and allow the workflow to pass without any problems.

PR #11179, opened 2023-11-10
Fix isEmpty check in validateField for custom field components

rvetere:
With a custom built field component it can happen that the referenced element does not have a value attribute.
But the state is correctly set at the moment of validation.

In that case, the `validateField` function thinks the field is empty, because of this line:
```
isHTMLElement(ref) && ref.value === ''
```

I've added a test to show that behavior and it fails with the current implementation as it does in our case of a multi select field with downshift.js

I think we could improve the "is-empty-check" by increasing the priority of the `inputValue` - we don't need to check if the ref-element's value attribute is empty when we have a valid `inputValue`, right?

### DIFF
--- a/src/logic/validateField.ts
+++ b/src/logic/validateField.ts
@@ -67,14 +67,16 @@ export default async <T extends FieldValues>(
   const isRadio = isRadioInput(ref);
   const isCheckBox = isCheckBoxInput(ref);
   const isRadioOrCheckbox = isRadio || isCheckBox;
-  const isArrayType = Array.isArray(inputValue);
+  const isInputValueEmpty =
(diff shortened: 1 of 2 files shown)


src/logic/validateField.ts
```diff
@@ -67,13 +66,18 @@ export default async <T extends FieldValues>(
   const isRadio = isRadioInput(ref);
   const isCheckBox = isCheckBoxInput(ref);
   const isRadioOrCheckbox = isRadio || isCheckBox;
+  const isInputValueEmpty =
+    isUndefined(inputValue) ||
+    inputValue === null ||
```
bluebill1049, line 71: so the additional check is just check `null`?

PR #11183, opened 2023-11-10
refac: improved isObjectType and swap

ssi02014:
@bluebill1049 👋
I clearly specified the type of `isObjectType` for the `type guard` to work.

Task on `swap` Just a quick suggestion. I think we could use the ECMAScript 6 syntax `destructuring assignment` to make it more concise and readable. What do you think?


bluebill1049 (approved):
LGTM. thanks.

Issue #11239, opened 2023-11-28
issue: Field disabled state not correct on initial render


thomasdbock:
### Version Number

7.48.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/purple-frost-zfytqd

### Steps to reproduce

1. Go to https://codesandbox.io/p/sandbox/purple-frost-zfytqd
2. Notice the "First name" field which is currently enabled
3. Toggle the field state by pressing either button (toggle field)
4. Notice the field is still enabled
5. Toggle again (repeat step 3)
6. Notice the field is now correctly disabled

### Expected behaviour

Since the fieldState was disabled by default, I expect the field to have the disabled state when the page has finished rendering.

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari

### Relevant log output

```shell
Not a log, but manually mapping each field property except the ref fixes this issue.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

thomasdbock:
Not providing a `disabled` property in `useForm` also fixes this issue: https://codesandbox.io/p/sandbox/cranky-keldysh-pxd5hv

bluebill1049:
this is close to chicken and egg problem, we can either take field level > form level or form level > field level.

thomasdbock:
Hey @bluebill1049 
For me the case is as follows:
Form state is used as "edit mode on/off", field disable state is conditional with another checkbox controlling it (but default off).

I think if one of them is false, it should disable the field. 
Right now the field is saying it's disabled, but it really isn't. In my use case, this is causing UX side effects since I have no clear way of knowing if it is truly disabled.

Then again, I never expect `disabled: false` to explicitly "enable" the fields if some other part is disabling it.

PR #11241, opened 2023-11-28
fix(#11239): Set fields disabled state based on form and field disabl…

thomasdbock:
Quickfix on #11239 
Needs some love by somebody who knows the internals better than I do.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1084,11 +1084,7 @@ export function createFormControl<
         (ref, name) => {
           let requiredDisabledState = disabled;
           const currentField = get(_fields, name);
-          if (
-            currentField &&
-            isObject(currentField._f) &&
-            isBoolean(currentField._f.disabled)
-          ) {
+          if (currentField && isBoolean(currentField._f.disabled)) {
             requiredDisabledState ||= currentField._f.disabled;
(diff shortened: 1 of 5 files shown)


src/logic/createFormControl.ts
```diff
@@ -1081,8 +1081,11 @@ export function createFormControl<
       _subjects.state.next({ disabled });
       iterateFieldsByAction(
         _fields,
-        (ref) => {
-          ref.disabled = disabled;
+        (ref, name) => {
+          const currentField = get(_fields, name);
+          if (currentField) {
+            ref.disabled = currentField._f.disabled || disabled;
```
bluebill1049, line 1087: probably need to check `currentField._f.disabled` is defined as well.

bluebill1049 (commented):
Thanks very much for the PR! could you please include a unit test for this usecase well.

src/logic/createFormControl.ts
thomasdbock: Had a hard time figuring out if there is a standard way you like to check these, hope this is sufficient :) 
Added a test as well.

bluebill1049 (dismissed):
LGTM. 👍

bluebill1049 (approved):

PR #11245, opened 2023-11-29
Change type structure of UseFormHandleSubmit

### LINKED ISSUE #11244: UseForm is returning a submit handler without passing on the TFieldValues generic
### Version Number

7.48.2

### Codesandbox/Expo snack

n/a

### Steps to reproduce

- get `handleSubmit` method from `useForm`:
```ts
type MyData = {
  firstName?: string
  lastName?: string
}
const {handleSubmit} = useForm<MyData>({ //...
```

- see that when typescript infers the parameter type in the first callback in `handleSubmit` is `undefined`

image

**Note:** It looks like this the issue is [here](https://github.com/react-hook-form/react-hook-form/blob/c7b7eacb00cd11fe8fd075c958589b5b00b2c49d/src/types/form.ts#L622): 
```ts
export type UseFormHandleSubmit<
  TTransformedValues extends FieldValues | undefined = undefined,
> = (
  onValid: TTransformedValues extends FieldValues
    ? SubmitHandler<TTransformedValues>
    : SubmitHandler<TFieldValues>,
  onInvalid?: SubmitErrorHandler<TFieldValues>,
) => (e?: React.BaseSyntheticEvent) => Promise<void>;
``` 
It looks like `TTransformedValues extends FieldValues` always resolves to truthy, but I do not know why

### Expected behaviour

It should correctly infer the type in the submit handler

image

_No response_

### Relevant log output

_No response_

### Code of Conduct

mildfuzz:
Fixes https://github.com/react-hook-form/react-hook-form/issues/11244

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -668,7 +668,7 @@ export type UseFormGetValues<TFieldValues extends FieldValues> = {
 };
 
 // @public
-export type UseFormHandleSubmit<TFieldValues extends FieldValues, TTransformedValues extends FieldValues | undefined = undefined> = (onValid: TTransformedValues extends FieldValues ? SubmitHandler<TTransformedValues> : SubmitHandler<TFieldValues>, onInvalid?: SubmitErrorHandler<TFieldValues>) => (e?: React_2.BaseSyntheticEvent) => Promise<void>;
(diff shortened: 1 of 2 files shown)


mildfuzz:
FWIW It seems the original implementation *should* work, but doesn't....

bluebill1049 (approved):
thanks!

hamideslami32:
Thanks @mildfuzz @bluebill1049 . I found this just in time. Great job.

PR #11309, opened 2023-12-17
🏷️ improve CustomElement type

r-kataoka11:
Thanks for providing a good library. It helps me a lot.

I've noticed that the type of `Ref` is different from the actual one. You can use the major properties and Web API in a type-safe manner.

For example, scrolling to the form with validation errors:

```tsx
useEffect(() => {
  if (formState.errors.email?.ref) {
    formState.errors.email.ref.scrollIntoView({
      block: 'center',
      behavior: 'smooth',
    })
  }
}, [formState.errors.email?.ref])
```

This can also be effectively utilized when referring to attributes such as `id` or `aria` properties.


bluebill1049 (approved):
thank you very much! 🙏

PR #11333, opened 2023-12-22
fix: Change `DeepPartial` type definition to be compatible with `unknown`

RulerOfCakes:
With the previous type definition, some codes like the example below containing the `unknown` type caused a type error:
```typescript
interface InterfaceWithUnknown {
  a: unknown;
}

interface IFormInput {
  test: InterfaceWithUnknown;
}

  const x: IFormInput = {
    test: {
      a: 'hi',
    },
  };

  const { register, handleSubmit } = useForm<IFormInput>({
    // Type 'IFormInput' is not assignable to type '{ test?: { a?: {} | undefined; } | undefined; }'.
    defaultValues: x,
  });
  ...   
}

```
I found out that this could be very similar to a case that happened on the [redux](https://github.com/reduxjs/redux/issues/3368) repository.
This PR fixes the type definition to take account of such cases and remove the error.

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -135,7 +135,7 @@ export type DeepMap<T, TValue> = IsAny<T> extends true ? any : T extends Browser
 
 // @public (undocumented)
 export type DeepPartial<T> = T extends BrowserNativeObject | NestedValue ? T : {
-    [K in keyof T]?: DeepPartial<T[K]>;
+    [K in keyof T]?: T[K] extends object ? DeepPartial<T[K]> : T[K];
 };
 
 // @public (undocumented)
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):

bluebill1049:
thank you for your contribution.

jonahallibone:
This PR causes an issue where all sub keys in an object become required

eltonio450:
indeed it introduced a bug: nested defaultValues are not considered as partial anymore
image

here we can see that the key createLocation can be either NewLocationFormValues or undefined, where as it should be Partial<NewLocationFormValues> or undefined

before the update, it was:

image

Moshyfawn:
> indeed it introduced a bug: nested defaultValues are not considered as partial anymore

This has been addressed in the linked PR #11373

PR #11336, opened 2023-12-24
fix: trigger validation netsted field

### LINKED ISSUE #10862: issue: setValue an empty array to a nested field doesn't trigger validation when shouldValidate is true
### Version Number

7.43.9

### Codesandbox/Expo snack

https://codesandbox.io/s/dark-wave-y2zj85

### Steps to reproduce

I have a nested field and a custom component hooked to RHF via `useController`. The custom component has `string[]` as its value type. 

I need to set the value to `[]`. These are two ways that I tried.

a)
```tsx
setValue("item1.positions", [{ activities: [] }], {
   shouldValidate: true
});
```

b)
```tsx
setValue("item2.positions.0.activities", [], {
   shouldValidate: true
});
```
a) doesn't work well,
b) works as expected

There are two issues:
1. `useController` doesn't pick up on the state change because it has `exact` internally set to `true` by default without the option to set it. That is why the component doesn't even re-render upon the `setValue` call => I have found a solution to this - I'm using useWatch with `exact` set to `false`. 
2. `error` is not set. It is not just that `useController` doesn't cause the component to re-render (as I stated in my 1. point), the error isn't anywhere. I can even use `useFormState({ exact: false })` and it still doesn't show. 

I have been debugging the 2. issue for a long time and I may have found the reason in the `setValues` function.

```tsx
  const setValues = <
    T extends InternalFieldName,
    K extends SetFieldValue<TFieldValues>,
    U extends SetValueConfig,
  >(
    name: T,
    value: K,
    options: U,
  ) => {
    for (const fieldKey in value) {
      const fieldValue = value[fieldKey];
      const fieldName = `${name}.${fieldKey}`;
      const field = get(_fields, fieldName);

      (_names.array.has(name) ||
        !isPrimitive(fieldValue) ||
        (field && !field._f)) &&
      !isDateObject(fieldValue)
        ? setValues(fieldName, fieldValue, options)
        : setFieldValue(fieldName, fieldValue, options);
    }
  };
```

As it recursively goes through the field name, to parse through the name and use the correct value type, it encounters `[]` as the value and it never calls `setFieldValue`, because `[]` is not a primitive value. If the value to set was a primitive, this process wouldn't have been stopped in its tracks and it would have updated the value and `formState`. It doesn't register the error. I believe it would behave the same if I passed an empty object (to a different field of course) - as it isn't primitive either.

I have not found a solution for this issue. The only workaround was to call `trigger` immediately after setting the value. That works but I would like to avoid that.

### Expected behaviour

Empty array `[]` should be a valid value and it should correctly trigger validation when `shouldValidate` is set to `true`. 

Chrome

### Relevant log output

_No response_

### Code of Conduct

sungh0lim:
To fix #10862 .

To ensure that an empty array [] is considered a valid value and correctly triggers validation when shouldValidate is set to true, I add the condition `isEmptyNonPrimitive`.

I want to add tests for the cases of an empty object and an empty array in `setValue.test.tsx`, but I'm not sure what to compare. Can you help me with this?


bluebill1049 (dismissed):
LGTM! Thank you for the contribution. Sorry I have been busy and completely miss the PR>

bluebill1049:
looks like test is failed.

src/__tests__/useForm/setValue.test.tsx
```diff
@@ -1385,4 +1385,48 @@ describe('setValue', () => {
...
+          />
+          <input
+            {...register('item2.positions.0.activities')}
+            data-testid={'b'}
+          />
+
+          <button
+            type="button"
+            onClick={() => {
+              setValue('item1.positions', [{ activities: [] }]);
```
sungh0lim, line 1416: @bluebill1049 To check #10862 . How about this?

```typescript
setValue("item1.positions", [{ activities:[] }], {
  shouldValidate: true
});

setValue("item2.positions.0.activities",[], {
  shouldValidate: true
});
```

bluebill1049: yea, i did. I think it would be rather difficult to set inputs to empty values. if they are deeply nested. so it's better let the users at the app level doing that setValue within a loop. thanks a lot for your help and contribution tho.

sungh0lim: How about incorporating this change into`'should be able to set to empty array value'` tests? The existing test [failed](https://github.com/react-hook-form/react-hook-form/pull/11336#issuecomment-1890791672) because only one side was modified.
If you confirm, I will change it.

sungh0lim: @bluebill1049 Please check it.

bluebill1049: sorry, miss the pr. please update PR and I will review it again.

bluebill1049 (approved):

bluebill1049:
thank you for your contribution! <3

sungh0lim:
@bluebill1049 Can it be merged into the `main` branch? It's been a long time since I was approved.

PR #11353, opened 2024-01-02
🐛 fix(useController): check if disabled state was set by user

### LINKED ISSUE #11320: issue: regression with omit prop on field
### Version Number

7.49.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/fast-sun-d8m2tc?file=%2Fsrc%2FApp.tsx

### Steps to reproduce

Don't pass a disabled to useForm or Controller, then log the fields object.
It shows the disabled property as undefined.

### Expected behaviour
The issue was previously described at #10932. Then behavior was fixed with https://github.com/react-hook-form/react-hook-form/commit/5fe56141a5b756e1d928efa0fbf2d5635edd8b58

But it came back with the changes at https://github.com/react-hook-form/react-hook-form/commit/0d565fcf944621f5510da4623129ffbd238f9530#diff-3f5bdfa7d822b4562b8e1c72dd1996197fe93478b039779bb78853ab3ae2a3feR129

The code does a check for `isBoolean(formState.disabled)`, but `disabled` is defined as a getter and it returns false when `disabled` is not present on useForm.

Chrome

### Relevant log output

_No response_

### Code of Conduct

Moshyfawn:
The [`isBoolean`](https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/isBoolean.ts ) utility may not be suitable for checking if the `disabled` state is set, as it is primarily used for type-guards.

### DIFF
--- a/app/pnpm-lock.yaml
+++ b/app/pnpm-lock.yaml
@@ -7,7 +7,7 @@ settings:
 dependencies:
   '@hookform/resolvers':
     specifier: 2.8.8
-    version: 2.8.8(react-hook-form@7.48.2)
+    version: 2.8.8(react-hook-form@7.49.2)
   '@material-ui/core':
     specifier: ^4.12.3
     version: 4.12.4(@types/react@17.0.52)(react-dom@17.0.2)(react@17.0.2)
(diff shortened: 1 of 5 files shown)


bluebill1049 (approved):
LGTM. Thanks for the fix @Moshyfawn

PR #11359, opened 2024-01-03
Fix: watch and useWatch without parameters return type

### LINKED ISSUE #10734: issue: useWatch<T>() (parameterless) and watch type from the useFormContext have different types
### Version Number

7.43.9

### Codesandbox/Expo snack

https://codesandbox.io/s/adoring-shadow-zgr3rd?file=/src/App.tsx

### Steps to reproduce

Recreated from https://github.com/react-hook-form/react-hook-form/issues/10721 using the template as the GitHub "new issue" from discussion functionality skips your templates. 

1. Go to Example1
2. See the types comparison between `useWatch<Form>()` and `useFormContext<Form>()`
3. Go to Example2
4. See the workaround that gets rid of the deep partial

Description:
Using `useWatch<T>()` (without providing any parameters) returns a deep partial unlike the watch type from `useFormContext`. This results in having to either cast the result to the non-partial value (`T`) or @ts-ignore the type in order to get my defined type instead of the deep partial one.

I can't actually just use the `useFormContext#watch` because that results in infinite updates when I try to use it the same way I use `useWatch` and gives me performance issues so I've moved away from using it even though the typing there is what I need.

You can see the `Example2` file for the usage of a nested property to avoid getting deep partials. 

### Expected behaviour

I expect that the types would be the same between the two. Unless it's some design decision for some specific case, to me it doesn't make sense for it to be a deep partial. 

_No response_

### Relevant log output

_No response_

### Code of Conduct

eladattias1996:
- Fixes #10734 - `watch` and `useWatch` without parameters return types were different.
- Fix prettier in `app/src/app.tsx` and `app/src/disabledFields.tsx`.

### DIFF
--- a/app/pnpm-lock.yaml
+++ b/app/pnpm-lock.yaml
@@ -7,7 +7,7 @@ settings:
 dependencies:
   '@hookform/resolvers':
     specifier: 2.8.8
-    version: 2.8.8(react-hook-form@7.48.2)
+    version: 2.8.8(react-hook-form@7.49.2)
   '@material-ui/core':
     specifier: ^4.12.3
     version: 4.12.4(@types/react@17.0.52)(react-dom@17.0.2)(react@17.0.2)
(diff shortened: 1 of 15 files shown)


bluebill1049 (approved):

bluebill1049:
thanks @eladattias1996 very much for the contribution!

PR #11361, opened 2024-01-04
🐛 fix(formState, disabled): infer controlled disabled form state

### LINKED ISSUE #11360: issue:  Initial form context state does not match initial useForm props
### Version Number

7.49.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/twilight-mountain-ddvxd8

### Steps to reproduce

1. Go to (https://codesandbox.io/p/sandbox/twilight-mountain-ddvxd8)
2. Look at console to see formContext: { disabled }  state progress
3. See it starts as "false" and then turns to "true"

### Expected behaviour

Since `useForm({ disabled: true });` is always true, expect first render with formContext to contain disabled: true.

Chrome

### Relevant log output

```shell
Form State disabled false
Form State disabled true
```

### Code of Conduct

Moshyfawn:
Since the `disabled` state of the form can be controlled, we should inherit the `disabled` state passed from the `useForm` hook. This will prevent the form from flashing the wrong state when it is disabled.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -113,7 +113,7 @@ export function createFormControl<
     touchedFields: {},
     dirtyFields: {},
     errors: _options.errors || {},
-    disabled: false,
+    disabled: _options.disabled || false,
   };
   let _fields: FieldRefs = {};
   let _defaultValues =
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):

bluebill1049:
Thank you for the fix! <3

PR #11373, opened 2024-01-09
fix: Nested deep partial with unknown

jonahallibone:
Hi!

In the recent `7.49.3` release, there was a change to make the [DeepPartial compatible with `unknown`.](https://github.com/react-hook-form/react-hook-form/pull/11333).

The side effect of this PR, however, was to make nested properties on objects no longer optional. I believe this is because the `DeepPartial` checked for `extends object` which inherently excludes all other types like `string | null | undefined`. I am not a TS wizard, but I think I am close with that assumption.

My change is to check for `T extends never`, which should solve this issue with unknown attempting to be solved in the previous PR but also preserving the types of non-object keys

An example of the issue I am referring to can be found [here](https://codesandbox.io/p/devbox/hidden-meadow-crptl9?file=%2Fsrc%2FApp.tsx%3A9%2C12). The errors in questions will not appear on a version of this app with `react-hook-form` below `7.49.3`.  An example at `7.49.2` without the bug can be found [here](https://codesandbox.io/p/devbox/hidden-meadow-forked-yt559w?file=%2Fsrc%2FApp.tsx%3A28%2C26)

I want to reiterate I am not a TS wizard, and I very well could be wrong with this solution. However, I am pretty sure we could label the behavior introduced in `7.49.3` as a bug.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
   },
   "typescript.tsdk": "node_modules/typescript/lib",
   "jest.jestCommandLine": "pnpm test --",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": "explicit"
+    "source.fixAll.eslint": true
   },
   "typescript.tsdk": "node_modules/typescript/lib",
   "jest.jestCommandLine": "pnpm test --",
--- a/app/pnpm-lock.yaml
+++ b/app/pnpm-lock.yaml
@@ -7,7 +7,7 @@ settings:
 dependencies:
   '@hookform/resolvers':
     specifier: 2.8.8
-    version: 2.8.8(react-hook-form@7.49.3)
+    version: 2.8.8(react-hook-form@7.49.2)
   '@material-ui/core':
     specifier: ^4.12.3
     version: 4.12.4(@types/react@17.0.52)(react-dom@17.0.2)(react@17.0.2)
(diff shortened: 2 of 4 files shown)


bluebill1049 (approved):

RulerOfCakes:
Hi! I just found out about this PR. Sorry for the unintended bug in #11333.

To reinstate the problem, my previous PR didn't make nested properties on objects no longer optional *for all object types*. I was unaware that a union type containing both objects and non-object primitives didn't pass the `extends object` check, and thus it would have only failed in such cases like the example sandbox you provided in the PR.

However your solution seems to effectively revert the definition of `DeepPartial` back before #11333, as typically no type will satisfy `extends never` in this scenario. I have checked that this new definition will again cause errors with types containing `unknown` such as the previous example I gave in #11333, so it seems like we're back to square one here.

Before I upload a new PR, I'll post a new solution here to check for both cases: 
```
export type ExtractObjects<T> = T extends infer U
  ? U extends object
    ? U
    : never
  : never;

export type DeepPartial<T> = {
  [K in keyof T]?: ExtractObjects<T[K]> extends never ? T[K] : DeepPartial<T[K]>;
};
```
This will be able to extract any object types that are present in a union(if T[K] happens to be one) and effectively apply `DeepPartial` for only those objects. Accounts for `unknown` as well since `ExtractObjects<unknown>` will evaluate to `never` too. [Here is a quick sandbox link that I forked off to test compile errors.](https://codesandbox.io/p/devbox/hidden-meadow-crptl9?file=%2Fsrc%2FApp.tsx%3A9%2C12) WDYT about this change? @bluebill1049 @jonahallibone 

If all looks good, I'll upload a new PR with some type tests to account for all these cases we have found.

bluebill1049:
@RulerOfCakes please send a pr, would be good if we can include some type tests as well. 🙏

jonahallibone:
@RulerOfCakes Ah yes your solution is what I was aiming to do! I figured that `infer` would end up being part of the solution however in practice I've no experience with it. I thought from some (admittedly light) testing that my solution had somehow worked with `unknown` but I now see that it was more likely codesandbox having incredibly flaking lint warnings. Thanks for correcting me!

PR #11375, opened 2024-01-09
🐞 fix #11372 reset keepValues watch return values

### LINKED ISSUE #11372: issue: reset keepValues does not work
### Version Number

7.49.3

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/vibrant-vaughan-wszfpx?file=%2Fsrc%2FApp.js%3A30%2C14

### Steps to reproduce

1. Change any form input value, see the data change (tracked via watch)
2. Click on reset
3. See data is lost rather than keeping current values. Input fields correctly show current data.

### Expected behaviour

I'm expecting the watched data to remain the same after resetting since keepValues is true, I just want to clear the dirty state. Instead, it is removing the values. The input itself seems to keep the current value, but the watched data does not. 

Firefox, Chrome

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
ref: https://codesandbox.io/p/sandbox/musing-wind-3z75kh?file=%2Fsrc%2FApp.js

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1232,7 +1232,9 @@ export function createFormControl<
         : cloneObject(values);
 
       _subjects.array.next({
-        values: { ...values },
+        values: keepStateOptions.keepValues
+          ? cloneUpdatedValues
+          : { ...values },
       });
 
       _subjects.values.next({


Moshyfawn:
@bluebill1049 I'm not sure this is needed. Check out [this reply](https://github.com/react-hook-form/react-hook-form/issues/11372#issuecomment-1883882917).

bluebill1049:
oh oops thanks 🙏 @Moshyfawn

Moshyfawn:
> oh oops thanks 🙏 @Moshyfawn

No problem!

PR #11385, opened 2024-01-11
fix: change `DeepPartial` to account for object union types & `unknown`

RulerOfCakes:
From the previous enhancement to allow `unknown` to be compatible with `DeepPartial`(#11333), it introduced an unintended bug, being incompatible with union types containing both object & non-object types. Although this was resolved in #11373, it again made the definition of `DeepPartial` incompatible with any types containing `unknown` again.

With a new proposed definition that should be able to handle such cases, I have added some test cases discovered throughout #11333 and #11373 as type tests along with the new definition. By using `ExtractObjects` to extract any type that extends `object` from a potential union type, it successfully applies `DeepPartial` for such object types while supporting `unknown` as well.


bluebill1049 (approved):
Nice thanks for the PR and the type tests.

RulerOfCakes:
> This PR breaks types in global `watch()`
> 
> All data become optional

I think the issue you're mentioning is related to #11359 🤔

PR #11399, opened 2024-01-15
✨ feat(values): priorities values over defaultvalues

### LINKED ISSUE #11398: When using both `defaultVaules` and `values`, useFrom returns actual values with delay.
### Discussed in https://github.com/orgs/react-hook-form/discussions/11397

<div type='discussions-op-text'>

<sup>Originally posted by **t1m0n** January 15, 2024</sup>
Hello! 
I have a typical situation when I fetch some data and pass this data to the `useForm`.  I also have `defaultValues` set along with `values`. The problem is that when my request is loaded I show the page with a form and it displays `defaultValues` for a moment and replaces them with actual data after.

Even if using sync data, `useForm` anyway for a moment uses `defaultValues` instead of `values`

**Example**
```jsx
const Form = () => {
  const form = useForm({
    defaultValues: {
      name: 'name default',
    },
    values: {
      name: 'name current',
    },
  });

  console.log(form.getValues());
};
```
Live example - https://stackblitz.com/edit/stackblitz-starters-lkabsa?file=src%2FApp.tsx

The code above firstly logs `name default` and after `name current`.

So the question is, is where anyway to get `values` in the first place if both `defaultValues` and `values` are set?

Thank you!
</div>

Moshyfawn:


bluebill1049 (approved):
LGTM

bluebill1049:
yea, let's release a minor version for this update. 👍

PR #11401, opened 2024-01-16
Get new dirtyFields when reset keeping defaults and dirty values

### LINKED ISSUE #11386: issue: dirtyFields is incorrect when reset() with `keepDefaultValues` and `keepDirtyValues`
### Version Number

7.49.3

### Codesandbox/Expo snack

https://codesandbox.io/p/github/IanVS/reproduction-keep-dirty-and-default/main

### Steps to reproduce

1. Go to  https://codesandbox.io/p/github/IanVS/reproduction-keep-dirty-and-default/main
2. Click on the first button ('Reset with keepDirtyValues')
3. Notice that the first field was updated, but is not considered dirty, even though it does not match the defaultValue
4. Click on the first button ('Reset')
5. This works correctly, the field is updated again, and now it is considered to be dirty

### Expected behaviour

Sorry, I've got another sneaky potential `keepDefaultValues` bug.

I think I should be able to update the value of non-dirty fields in the form without changing the defaultValue, and dirty tracking should continue to work correctly.

I did notice that controlled form elements work correctly as I expect, but not uncontrolled as shown in the reproduction.

_No response_

### Relevant log output

_No response_

### Code of Conduct

IanVS:
Fixes #11386 

I'm not positive this is the best way to do this, but it seems to work, tests pass, and it works in my app without breaking anything else.

### DIFF
--- a/src/__tests__/useForm/reset.test.tsx
+++ b/src/__tests__/useForm/reset.test.tsx
@@ -724,6 +724,82 @@ describe('reset', () => {
           }),
         );
       });
+
+      it('should treat previously-undirty fields as dirty when keepDefaultValues is set', async () => {
+        let updatedDirtyFields: Record<string, boolean> = {};
+        let updatedDirty = false;
+
+        function App() {
+          const {
+            reset,
+            register,
+            handleSubmit,
+            formState: { dirtyFields, isDirty },
+          } = useForm({ defaultValues: { firstName: '', lastName: '' } });
+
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
Awesome! thanks @IanVS. LGTM. This marks the day we turn issues to 0 again.

IanVS:
It has always impressed me that you stay anywhere near 0 open issues.

Issue #6841, opened 2021-10-19
Add setValueAs options to useController


hpohlmeyer:
**Is your feature request related to a problem? Please describe.**  
I am mostly using `useForm` in combination with `register` for my inputs. If I need to parse a value I can use `setValueAs`, `valueAsNumber` and `valueAsDate`.

If I have a third party component that is a controlled component, I can use `useController` to control it, but `useController` does not have the handy parsing functions, so I can’t destructure the `field` property from `useController` into my component, but need to wrap the `onChange` method first.

**Describe the solution you'd like**  
I would like to have `setValueAs`, `valueAsNumber` and `valueAsDate` as options for `useController`. They should behave the same way as the ones in `useRegister`.

```tsx
const MyForm = () => {
  const { controller } = useForm<{ myNumber: number }>();
  const { field } = useController({ name: "myNumber", controller, valueAsNumber: true });

  return (
    <form>
      <FancyNumberInput {...field} />
  	</form>
  );
}
```
**Describe alternatives you've considered**  
It can be solved by wrapping the external input and allow the value to be parse before `onChange` is called.

```tsx
const WrappedFancyInput = ({setValueAs, ...controllerProps}) => {
  const handleChange = (event) => {
    const newEvent = valueAs === undefined
      ? event
      : { ...event, target: { ...event.target, value: setValueAs(value) } };
    controllerProps.onChange(newEvent);
  }

  return <FancyNumberInput {...controllerProps} onChange={handleChange} />;
};

const MyForm = () => {
  const { controller } = useForm<{ myNumber: number }>();
  const { field } = useController({ name: "myNumber", controller });
  const parseNumber = (value?: string) => Number.isNaN(Number(value)) ? undefined : Number(value);

  return (
    <form>
      <WrappedFancyInput {...field} setValueAs={parseNumber} />
    </form>
  );
};
```

PR #11447, opened 2024-01-28
Added value as number in useController 

deepgolani4:
issue #6841


deepgolani4:
@bluebill1049 Please review

deepgolani4:
I'm abandoning this PR because I might have found a better way to tackle the same task that aligns perfectly with the current code style.

darshkpatel:
> I'm abandoning this PR because I might have found a better way to tackle the same task that aligns perfectly with the current code style.

What'd you end up doing ?

PR #11455, opened 2024-01-31
Add noop Utils Function

ssi02014:
@bluebill1049 👋
I looked at the already defined `Noop` type and added a `noop` utility function that does nothing. (+test code)

```ts
type Noop = () => void;
```
```ts
export default function noop() {}
```

I thought it would be appropriate to use it as a meaningful function like `noop` rather than simply writing "() => {}".
I'd appreciate it if you could review it. 🙏


bluebill1049 (approved):

bluebill1049:
thanks you again!

ssi02014:
@bluebill1049 Thanks !!🎉

PR #11520, opened 2024-02-16
feat: Add support for the Date type in defaultValues

LoisDuplain:
Because it's a pain to have to convert dates each time into a string and then into an object again and again


LoisDuplain:
Up ?

PR #11522, opened 2024-02-17
⚡ feat: `createFormControl` and `subscribe` function

bluebill1049:
## Proposal

Expose `createFormControl` method and enable subscribe outside of the react component. This would solve a lot of subscriptions within child component issues.

⚡️ `createFormControl`

- Allow us to start subscribing before the react life cycle
- This is completely optional and once everything is working, we can switch to using the core library

```tsx
const { formControl, control } = createFormControl(props)

function App() {
  const { register } = useForm({
    formControl,
  })

  return <form />
}

function Test() {
  useFormState({
    control // no longer need context api
  })
}
```

⚡️ `subscribe`

- subscribe form state update without re-render
- subscribe outside of the react component

```tsx
const { formControl } = createFormControl(props)

control.subscribe({
  formState: { isDirty: true },
  callback: (formState) => {
    if (formState.isDirty) {
      // do something here
    }
  }
})

function App() {
  const { register } = useForm({
    formControl,
  })

  return <form />
}
```


ap0nia:
Hello, I was wondering if there was any updates on this! I'd love to contribute to the core library, which is currently awaiting on this dependency.

bluebill1049:
@ap0nia will sort this out over this weekend. sorry for the delay

Neschadin:
when-when-when?  =)

bluebill1049:
> when-when-when? =)

Going to release RC version soon. Have some bugs that need to be patched first.

PR #11550, opened 2024-02-26
Issue with Resolvers tranform 

jmtwr:
Hi, I found one annoying, some time ago the next feature was broken.

**PROBLEM:**
1. describing schema validation w/ transform function > most common usage it's trim string input
2. Trigger validation 
3. Expected to get string without spaces at the end & start of string
4. The Result > value still with spaces 

**CODE SANDBOX EXAMPLE** > https://codesandbox.io/p/sandbox/react-hook-forms-resolvers-issue-mpyd39

Perhaps there is a more graceful solution to this problem. Thanks!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
   },
   "typescript.tsdk": "node_modules/typescript/lib",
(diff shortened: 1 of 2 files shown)


ultrox:
Did you consider using preprocesor like what zod have?

https://stackblitz.com/edit/vitejs-vite-mwco7c


```js
const schema = z.object({
  name: z.preprocess((v) => (v as string).trim(), z.string()),
});
```

PR #11558, opened 2024-02-27
Update fieldArray.ts

1shaked:
update docs


Moshyfawn:
This doesn't seem right. The JSDocs definition you've updated to include the `remove` method description is for the `insert` methods instead.

1shaked:
you are correct I was supposed to change the position I will update this,

PR #11624, opened 2024-03-12
🐛 bug(validatingFields, trigger): handle all fields validation trigger

### LINKED ISSUE #11621: issue: validatingFields includes "undefined" as field name
### Version Number

7.51.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/react-hook-form-async-field-validation-forked-sswqwh?file=%2Fsrc%2Findex.js%3A44%2C5

### Steps to reproduce

1. Load sandbox and watch state output above the form
1. validatingFields includes "undefined" as a validating field name
1. this is caused by the `trigger()` call inside the `useEffect` callback

### Expected behaviour

Not sure if this is intentional or if I am doing something wrong here. But I came across this in my current application after updating to the newest version 7.51.0 which introduced this field. 

The issue in my code was that `isValidating` has never been reset to false. I solved it in my app but might lead to other bugs elsewhere if the `"undefined": true` bit is added unintentionally.  

_No response_

### Relevant log output

```shell
"validatingFields": {
    "undefined": true
  }
```

### Code of Conduct

Moshyfawn:
Closes #11621

### DIFF
--- a/app/pnpm-lock.yaml
+++ b/app/pnpm-lock.yaml
@@ -7,7 +7,7 @@ settings:
 dependencies:
   '@hookform/resolvers':
     specifier: 2.8.8
-    version: 2.8.8(react-hook-form@7.50.1)
+    version: 2.8.8(react-hook-form@7.51.0)
   '@material-ui/core':
     specifier: ^4.12.3
     version: 4.12.4(@types/react@17.0.52)(react-dom@17.0.2)(react@17.0.2)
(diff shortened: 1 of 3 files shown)


bluebill1049:
thanks @Moshyfawn let me double check this over the weekend.

Moshyfawn:
> thanks @Moshyfawn let me double check this over the weekend.

For sure! I'm a bit unsure if the `_mount` fields are the best way to implement this, but I haven't found any way to retrieve the necessary fields other than mapping over them at the moment.

BijayKumarSahoo:
image
Issue since, 7.51.1 version.
Happens when Devtool component is used, entry into any input fields lags and gives above error, doesn't allow inputting fast.

PR #11703, opened 2024-03-21
Update watch observer value type to use DeepPartialSkipArrayKey

LiahMartens:
I believe this is more accurate since `undefined` items in arrays should not be valid, please correct me if this is not correct


LiamMartens:
@bluebill1049 in case you're able to have a quick look for input :smile:

bluebill1049:
hi @LiamMartens could you share a actual use case?

beck-peterson:
I'm currently running into this issue. I can resolve it by casting the watched callback to a DeepPartialSkipArrayKey variant, but it would be nice to not have to.

PR #11782, opened 2024-04-15
exposed `scrollIntoView()` to ref created via `useController()`

oskarski:
Currently there is no way to access `scrollIntoView()` method of an element that is controlled via `useController()`. This is much needed when it comes to errors handling and will allow to scroll to the element with the error

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -159,6 +159,15 @@ export function useController<
           if (field && elm) {
             field._f.ref = {
               focus: () => elm.focus(),
+              scrollIntoView: (
+                arg?:
+                  | boolean
+                  | {
+                      inline?: 'center' | 'end' | 'nearest' | 'start';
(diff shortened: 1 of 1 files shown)


mchl18:
can I suggest using this as a type for arg? then it would be on par with the typescript definitions
```typescript
boolean
  | {
      inline: 'center' | 'end' | 'nearest' | 'start'
      block: 'center' | 'end' | 'nearest' | 'start'
      behavior: 'auto' | 'instant' | 'smooth'
    }
```

huang-yuwei:
Hi, is this pull request still pending for the TypeScript definitions? I need to access the ref or scrollIntoView feature for error handling. If needed, I would be happy to commit the types changes instead.

oskarski:
@huang-yuwei I've updated the types definition

Issue #566, opened 2019-11-25
How to submit the form outside the form 


eranabir:
we have a case that we want to submit the form outside the form element 
how can we do that ?

bluebill1049:
as long as your useForm() is outside

you can do 

````
const { handleSubmit } = useForm();

<button onClick={handleSubmit((d) => console.log(d))} />

<form>

</form>
```

please consider ask general question at our specturm channel

eranabir:
thanks but our use case is a little bit different 

we have a common component BaseForm : 

````
function BaseForm2({onSubmit, render,defaultValues}){
    const methods = useForm({defaultValues});
    return (
        <form onSubmit={methods.handleSubmit(onSubmit)}>
            {render(methods)}
        </form>
    )
}
````
and we use it like this : 

````
function CustomerForm({onSubmit, getFormMethods}){
    return (
        <BaseForm2
            getFormMethods={getFormMethods}
            onSubmit={onSubmit}
            render={({register}) =>
                <input name='name.firstName' type='text' ref={register}/>
            }/>);
}
````
now we want to submit the form not inside the render function

**Comment by bluebill1049** [comment]
you may want to consider to use `FormContext`. so you can move `useFrom` out from the inner Component.

**Comment by bluebill1049** [comment]
going to close this question in here, feel free to leave more questions or prefer to answer it in specturm then other community members can helps too.

**Comment by asciant** [comment]
For those who come across this issue like I did, and found it didn't quite address the question. You can submit a form programatically by passing a ref to the `form` DOM element and then dispatching a _cancelable_ event:

```
formRef.current.dispatchEvent(new Event('submit', { cancelable: true }))
```

This will trigger the `onSubmit` handler via the RHF `handleSubmit` method.

**Comment by chellongCoder** [comment]
thanks, you save my day @asciant

**Comment by dgtlmonk** [comment]
`getValues()` seems to work as well.

```js
onSubmit(formData)  {
 ....
}
```

neighborhood999:
I solved this issue by HTML attribute, e.g:

```jsx

const { handleSubmit } = useForm();
const onSubmit = data => {
	console.log(data);
};

//...

<form id="hook-form" onSubmit={handleSubmit(onSubmit)}>
	{/* your awesome form */}
</form>

<button type="submit form="form="hook-form"">Submit</button>
```

bluebill1049:
Thanks @neighborhood999 for sharing solutions.

[...]

PR #11801, opened 2024-04-19
feat: Imperative form `submit` method

### LINKED ISSUE #12030: Submit the form outside the form (imperative submission)

We've had use cases where we must submit a form using a button that is not inside the form's context, like from a modal or an external component.

Currently, one has to structure their components by wrapping items and buttons into the form context to gain access to the handleSubmit function and trigger it manually. This can be difficult to design and implement for complex Form components such as flows or other form setups and requires tightly coupling the UI to the Form scaffolding.

Ideally, one can reference the form by `id` or use the form itself to call a `submit` function to trigger the submit instead of depending on the handleSubmit callback, which requires having the `onSubmit` function.

The only known alternatives are shared [here](https://github.com/react-hook-form/react-hook-form/issues/566#issuecomment-612328736): 

1. If you don't have access to the handleSubmit method. You can submit a form programmatically by passing a ref to the form DOM element and then dispatching a cancelable event. This will trigger the onSubmit handler via the RHF handleSubmit method:
```
formRef.current.dispatchEvent(new Event('submit', { cancelable: true }))
```

Handling refs isn't the ideal React solution as it is usually cumbersome, verbose, and not dev-friendly.

Another alternative is to have yet _another_ context to pass around a function that sets up the handleSubmit and onSubmit functions and can be used imperatively.

bombillazo:
Hello.

We've created a patch for this library, which has worked wonderfully. This change allows users to submit forms imperatively with a new `submit` function returned by the form hook, a feature that users have long requested (#566). This is done by giving the form component a unique ID that is stored within the form context of the RHF instance.

Previously, one had to restructure their components by wrapping items and buttons into the form context to gain access to the handleSubmit function and trigger it manually. Now, we can do this:

```tsx

export const MyComponent: FC = () => {
  const form = useForm({ 
    id: 'my-new-form', // optional 
  });

  const onSubmit = () => {
  // some summit logic
  }

  return (
    <>
      <Button
        onClick={() => {
          form.submit();
        }}>
        Submit
      </Button>
      <form
        id={form.id}
        onSubmit={handleSubmit(onSubmit)}>
        ...
      </form>
    </>
    );

}
```

The beauty is that one can trigger the form from any component with the `form.id`, which is now returned by the `useFormContext();` hook. You can also specify your own ID value for the form if you want to control it, but if not, RHF will generate a UUID.

This resolves #12030


bombillazo:
It's failing due to an increased bundle size, up to 10.23KB. Not sure how to incorporate the UUID generation without affecting bundle size.

bombillazo:
Hey @bluebill1049 , I believe this is a feature that could benefit many users as well, I'd love your feedback for it when possible.

bluebill1049:
hi, could we just do this?

```
<form onSubmit={handleSubmit(....)}>
    <input type="text" name="name" />
</form>

<input type="submit" form="myform" value="Update"/>
```

bombillazo:
That assumes we want to trigger a submit only when we click a button. But we may want to submit a form programmatically by other means.

With this change, any logic can trigger a submit event.

bombillazo:
Hey @bluebill1049 , that assumes we only want to trigger a submit with a button click. But we may want to submit a form programmatically by other means. Also, having the submit input/button outside the form only works by setting up the form with the proper id  (this PR takes care of that). 

With this change, any logic can trigger a submit event by calling `form.submit()`.

bombillazo:
Any further thoughts @bluebill1049 ?

bombillazo:
Hey, any luck to consider this feature for merging? I have to reconcile all the changes to our fork and rebuild :( we've been using this in prod for months an so far it has worked as expected.

bombillazo:
Hey @bluebill1049 any luck considering this feature for merging? I keep having to reconcile all the changes to our fork and rebuild :( We've been using this in prod for months, and so far, it has worked as expected.

bombillazo:
Hey @bluebill1049 , is there more feedback? What can we do to help improve or move the needle for this feature to work natively with RHF?

bombillazo:
Hey @DASPRiD I did but it was failing some tests

ClockworkJustin:
Would love to see this land as I had a similar use case and had to wrap an entire chunk of UI in the "form" to get around this. Ideally I'd be able to do:
```
const { submit } = useFormContext();
///
submit();
```
without needing to make everything a submit button that needs to do this.

bombillazo:
Hey @ClockworkJustin, this PR solves exactly that, and I've been using it for months now in prod without issues. We can even submit forms from any location in our app as long as we have the form ID, which is also available in the form data.

I've been trying to get this merged for a while, but I need more support.

ClockworkJustin:
Tried to add this code and run it myself but it's not actually submitting the form.

```ts
  const methods = useForm({ values, defaultValues })

  const handler = (data) => console.log(data)
  
  return (
    <FormProvider {...methods}>
      <form id="testing-form" onSubmit={methods.handleSubmit(handler)}>
        {children}
      </form>
    </FormProvider>
  )
```

But then running:
```
  try {
      document?.getElementById('testing-form')?.dispatchEvent(
        new Event('submit', {
          bubbles: true,
          cancelable: true,
        }),
      )
    } catch (_) {}
```
doesn't make the form submit. Just having only this for testing and `onSubmit` is never called as there are no console logs. I first tried in my project and then just did this small test and same scenario. Is the FormProvider maybe doing something?

bombillazo:
How are you triggering the submit?

Do you see the form when running ` document?.getElementById('testing-form')` in your browser console?

ClockworkJustin:
Yep it definitely finds the form. But the event dispatch doesn't trigger the submit handler even if the submit handler just runs a quick console log.

So at least for me this solution didn't work.

bombillazo:
@ClockworkJustin  By any chance do you have nested forms? like an HTML form inside another?

bombillazo:
@ClockworkJustin Not sure how you have it setup, but its working for me :/

https://github.com/user-attachments/assets/ecd8cf51-3f17-4103-8070-9a18c773b572

ClockworkJustin:
Here's a code sandbox example. https://codesandbox.io/p/sandbox/react-dev-forked-vjywfl

If you click the "Test" button nothing happens. But if you click the "Real" button the handler runs and set some state to update text on the screen.

bombillazo:
@ClockworkJustin What browser are you using?

That sandbox works for me.

https://github.com/user-attachments/assets/8aa266c4-e1a5-4c04-a6b5-733089bc08ac

bombillazo:
@ClockworkJustin, could you explain what you mean by "figure something else out"? Is it working or not? 

If yes, we could get @bluebill1049 to review and give feedback to integrate this feature into the package. I've been running this feature in prod for months using patch-package with no issues.

ClockworkJustin:
Yeah that code sandbox works. But still doesn't work inside my project. No nested forms. And aside from some custom wrapper components it's basically a form with inputs. A `<button type="submit">Save</button>` inside the form submits it great, but dispatching the event doesn't trigger it.

I've already got a different workaround for my project so don't worry about. Didn't mean to pollute the PR! Hopefully this lands and I sort out what the heck is wrong in my project.

bombillazo:
No worries, I'm happy to help in any way I can.

bombillazo:
@bluebill1049  Any consideration for this feature? I recently am looking at React Final Form and [this](https://final-form.org/docs/react-final-form/faq#via-documentgetelementbyid) is pretty much what Ive implemented for RHF.

How can I submit a form from any other component without somehow prop drilling the handle submit?

bombillazo:
Close to use a proper feature branch to merge from

PR #11827, opened 2024-04-26
Fix: break out of recursive loops on first focus

DPflasterer:
I have a form structure like so, where I have a group depth of more than 2.

```
personal.name.first
personal.name.last
personal.phone-number

```
What was happening is when `iterateFieldsByAction` calls itself recursively it did call focus on `personal.name.first` but then it breaks out of `personal.name` and continues looping and calling focus on `personal.phone-number`.

This change will break out of the current loop and recursively break out of all loops.

### DIFF
--- a/src/__tests__/logic/iterateFieldsByAction.test.ts
+++ b/src/__tests__/logic/iterateFieldsByAction.test.ts
@@ -64,4 +64,143 @@ describe('focusFieldBy', () => {
       );
     }).not.toThrow();
   });
+
+  it('should focus on the first error it encounter and not the second', () => {
+    const focus = jest.fn();
+    iterateFieldsByAction(
+      {
+        first: {
+          _f: {
+            name: 'first',
+            ref: {
+              name: 'first',
+              focus,
+            },
+          },
+        },
+        second: {
+          _f: {
+            name: 'second',
+            ref: {
+              name: 'second',
+              focus,
+            },
+          },
+        },
+      },
+      (ref) => {
+        // @ts-ignore
+        ref.focus && ref.focus(ref.name);
+        return 1;
+      },
+    );
+    expect(focus).toBeCalledWith('first');
+    expect(focus).not.toBeCalledWith('second');
+  });
+
+  it('should recursively drill into objects', () => {
+    const focus = jest.fn();
+    iterateFieldsByAction(
+      {
+        test: {
+          name: {
+            first: {
+              _f: {
+                name: 'name.first',
+                ref: {
+                  name: 'first',
+                  focus,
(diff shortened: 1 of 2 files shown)


src/__tests__/logic/iterateFieldsByAction.test.ts
```diff
@@ -64,4 +64,145 @@ describe('focusFieldBy', () => {
...
+      },
+      (ref) => {
+        // @ts-ignore
+        ref.focus && ref.focus(ref.name);
+        return 1;
+      },
+    );
+    expect(focus).toBeCalledWith('first');
+    expect(focus).not.toBeCalledWith('second');
+  });
```
bluebill1049, line 99: existing logic pass this test as well.

src/__tests__/logic/iterateFieldsByAction.test.ts
```diff
@@ -64,4 +64,145 @@ describe('focusFieldBy', () => {
...
+          // @ts-ignore
+          ref.focus && ref.focus(ref.name);
+          return 1;
+        }
+        return;
+      },
+    );
+    expect(focus).not.toBeCalledWith('first');
+    expect(focus).toBeCalledWith('last');
+  });
```
bluebill1049, line 140: existing logic pass this test as well.

bluebill1049 (commented):
not sure the fix here, i thought the existing logic to break early as well

src/__tests__/logic/iterateFieldsByAction.test.ts
DPflasterer: That is correct. These first few tests were simply to improve code coverage by asserting focus isn't called on when it's not expected inputs. The`should should recursively drill into objects and break out of all loops on first focus` test fixes the issue solved by this PR.

src/__tests__/logic/iterateFieldsByAction.test.ts
DPflasterer: That is correct. These first few tests were simply to improve code coverage by asserting focus isn't called on when it's not expected inputs.

DPflasterer:
> not sure the fix here, i thought the existing logic to break early as well

The existing logic does break early out of the first loop but it is not breaking out of the recursion. The last test case added, fixes the issue which I'll  try to illustrate.

Before:
```
personal
    name
        first
        last <- called here and breaks out of 'name'
    phone <- called here and breaks out of 'personal'
    address
        line1
```

After:
```
personal
    name
        first
        last <- called here and returns false to 'personal' and breaks out of it
    phone
    address
        line1
```

PR #11872, opened 2024-05-08
fix: add info.value type to WatchObserver

peinguin:
Adjust `info` parameter of `WatchObserver` type to its actual value.


bluebill1049 (approved):
LGTM

PR #11880, opened 2024-05-13
🐞 fix #11821 set value with disabled false before mount

### LINKED ISSUE #11821: issue: radio buttons are not checked when "disabled" property is set
### Version Number

7.51.3

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/stupefied-sun-hhlh96?file=%2Fsrc%2FApp.js

### Steps to reproduce

1. Register radio buttons *but* with `{ disabled: false }` as an option like for example `{...register("theme", { disabled: false })}`
2. Specify a `defaultValues` for one of these radio buttons
3. Nothing is `checked` by default
4. Now remove `{ disabled: false }` like for example `{...register("theme")}`
5. The radio button is `checked` as it should be

So it seems that `disabled`, even when set to `false`, prevents the checked state to be set.

### Expected behaviour

`{...register("theme", { disabled: false })}` should behave the same as `{...register("theme")}` and a radio button should always be checked when `defaultValues` was set for it.

Chrome

### Relevant log output

_No response_

### Code of Conduct

erashu212:
When expression is evaluated as null the get method does not return defaultValue.

Ref:https://github.com/lodash/lodash/issues/5607

closes #11821

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -275,11 +275,11 @@ export function createFormControl<
     const field: Field = get(_fields, name);
 
     if (field) {
-      const defaultValue = get(
-        _formValues,
-        name,
-        isUndefined(value) ? get(_defaultValues, name) : value,
-      );
+      const fieldValue = get(_formValues, name);
+      const defaultValue =
+        isNullOrUndefined(fieldValue) || isNullOrUndefined(value)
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks @erashu212 for helping! 👍

PR #11888, opened 2024-05-15
Fix SetError to Preserve Existing Errors elsewhere in the object

SimonJTurner:
When calling setError, if a call was made to set an error on a descendant within a form (`foo.bar.baz`), then a later call was made to set an error on the ancestor (`foo.bar`), the previous error would be lost. 

This was the run of `pnpm test` after the first commit, which only contained the test:
image

Following the update to `setError`, we make sure you don't lose the errors that are further down the tree by including them in the call to set. To be fair, one /could/ consider just spreading in the entire `currentError`, but by destructuring and omitting `message` and `type` we make sure that they get overwritten by the current error.

### DIFF
--- a/src/__tests__/useForm/setError.test.tsx
+++ b/src/__tests__/useForm/setError.test.tsx
@@ -164,4 +164,91 @@ describe('setError', () => {
       screen.findByText('not found');
     });
   });
+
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):

bluebill1049:
thanks for the contribution.

PR #11903, opened 2024-05-19
Document utils and logic functions

ultrox:
As I'm learning this codebase I find it hard if I need to look at implementation of each function to understand what it does.
Some of them are really non obvious (get) as this code base prefer short naming.

Here I gave some effort to document what fn do. Some very obvious predicates are out of scope.

Aside of createSubject, I didn't touch any code. in case of createSubject since it's really bottom down, I caught myself multiple times stunt about the name.

I would really appreciate early feedback to see do I continue or this is not something you would like to pursue right now.


bluebill1049:
Thank you very much for this, i will take a look this weekend and come back to you.

Issue #11508, opened 2024-02-14
🐞issue: Controlled component is broken (TS-only) with `TTransformedValues` generic value being used


xsjcTony:
### Version Number

7.50.1

### Codesandbox/Expo snack

https://stackblitz.com/edit/vitejs-vite-zu9h3d?file=src%2Fcomponents%2FControlledInput.tsx,src%2FApp.tsx&terminal=dev

### Steps to reproduce

1. Go to 'App.tsx'
2. See error

### Expected behaviour

It should not error.

First of all, I appreciate all your hard work💚
This seems a regression in #11420 @bluebill1049 

I've been using this controlled component pattern for ages, everything works smoothly before `7.50.1`, but breaks after.

I've provided a minimal reproduction of 2 react-hook-form versions below:

Comparison:
`v7.50.1`: https://stackblitz.com/edit/vitejs-vite-zu9h3d?file=src%2Fcomponents%2FControlledInput.tsx,src%2FApp.tsx&terminal=dev
`v7.49.3`: https://stackblitz.com/edit/vitejs-vite-ym12kg?file=src%2FApp.tsx,package.json&terminal=dev

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

xsjcTony:
https://github.com/react-hook-form/react-hook-form/pull/11420/files#r1460438108 This looks a bit weird to me, I don't find any benefits of adding the third generic to the `Control` type.
Maybe I'm missing something since I don't have much insight into this. I'll temporarily ping the version to `v7.49.3` to resolve the issue.
Cheers!

bluebill1049:
I may just revert that original PR fix for now.

bluebill1049:
reverted for now, will release in the next patch.

mlabrum:
+1 hit this same issue.

I've now had to fix the dependency, any update on when this will be released?

PR #11420, opened 2024-01-20
Fix data param in handleSubmit when TTransformedValues is a union type

Ar4ys:
## Proposed Changes

Replaces current `UseFormHandleSubmit` implementation with the one that relies on [default values in generics](https://www.typescriptlang.org/docs/handbook/2/generics.html#generic-parameter-defaults). For more info see this comment in original issue: https://github.com/react-hook-form/react-hook-form/issues/11414#issuecomment-1900726167

Fixes #11414

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing tests pass locally with my changes

PR #11913, opened 2024-05-21
Default TTransformedValues generics to TFieldValues

kevinresol:
- Successor to #11420
- Basically `TTransformedValues extends FieldValues = TFieldValues` everywhere.
- Breaking change for people who write code that forwards generics manually (e.g. #11508). The migration is to add the new generic type parameter to the wrapper type and forward it as well. Example shown below:

```patch
interface ControlledInputProps<
  T extends FieldValues,
  N extends FieldPath<T>,
+  U extends FieldValues = T,
 >  {
-  control: Control<T>;
+  control: Control<T, any, U>;
}
```

### DIFF
--- a/reports/api-extractor.md
+++ b/reports/api-extractor.md
@@ -37,7 +37,7 @@ export type ChangeHandler = (event: {
 }) => Promise<void | boolean>;
 
 // @public (undocumented)
-export type Control<TFieldValues extends FieldValues = FieldValues, TContext = any> = {
+export type Control<TFieldValues extends FieldValues = FieldValues, TContext = any, TTransformedValues extends FieldValues = TFieldValues> = {
     _subjects: Subjects<TFieldValues>;
     _removeUnmounted: Noop;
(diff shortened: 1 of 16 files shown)


jedahu:
Having `TTransformedValues` extend `TFieldValues` means your schema resolver cannot parse to a different shaped type which is an undesirable limitation imo.

I have a PR here (https://github.com/react-hook-form/react-hook-form/pull/11951) which removes that constraint.

I've just added this PR's type test to it and it passes.

kevinresol:
> Having `TTransformedValues` extend `TFieldValues` means your schema resolver cannot parse to a different shaped type which is an undesirable limitation imo.
> 
> I have a PR here (#11951) which removes that constraint.
> 
> I've just added this PR's type test to it and it passes.

That is very true. But it is out of scope for this PR. Happy to update when #11951 gets merged.

kevinresol:
Not stale

PR #11917, opened 2024-05-21
fix: change info.values type in WatchObserver

peinguin:
I am so sorry. 
I made a typo.
The real property name is `values`. 
Here:
https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L738


bluebill1049:
aha no worries

bluebill1049:
could you update the API Extrator?

peinguin:
Done. Updated API extractor

bluebill1049 (approved):

PR #11938, opened 2024-05-25
Add validation in the cleanup process in `useController`

### LINKED ISSUE #11937: issue: cleanup process in `useController` crashes when used with `useFieldArray` in certain cases
### Version Number

7.51.5

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/react-hook-form-null-check-zqxx7v?file=%2Fsrc%2FApp.js

### Steps to reproduce

1. Go to the codesandbox.
2. Click on "remove 0th elem from array (will cause error)" button
3. See error

### Expected behaviour

The error should not occur.

The cause of the error is as follows:

- In the CodeSandbox, a form with the default value `{ arr: [{ k: 0 }, { k: 1 }] }` is rendered as follows:
   - The 0th element of the `arr` field is rendered by the `<Zeroth />` component where the `"arr.0"` path is controlled by `useController`.
   - The 1st element of the `arr` field is rendered by `<First />` component where the `"arr.1.k"` path is controlled by `useController`.
- When the "remove 0th elem" button is clicked:
   1. The 0th element is removed from the form's `control._fields`. Here, the resulting `control._fields` is `{ arr: [{ k: { _f: ... }] }` because `arr.1.k` is controlled by `useController`.
   2. Both `<Zeroth />` and `<First />` components start to unmount.
   3. When unmounting the `<Zeroth />` component, [the cleanup process in `useController`](https://github.com/react-hook-form/react-hook-form/blob/master/src/useController.ts#L103) for the `"arr.0"` path is executed. Here, this cleanup expects `control._fields` to be `{ arr: [{ k: ..., _f: ... }, ...] }`. However, as mentioned above, the `_f` field does not exist.
   4. This causes the app to crash.

Although the current example is too trivial to use `useController` (just the return values of `useFieldArray` are enough), in a complex form with heavily nested form values, using `useController` with `useFieldArray` can sometimes be more convenient.

Chrome

### Relevant log output

_No response_

### Code of Conduct

nakaakist:
fixes https://github.com/react-hook-form/react-hook-form/issues/11937.

In some cases, in the `updateMounted` function, the `field` does not have an `_f` field even if the `field` is truthy.

This PR adds extra validation to cover such cases.

Also, increased the size limit from 10.1 kB to 10.2 kB as the size limit check failed.

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -85,7 +85,7 @@ export function useController<
     const updateMounted = (name: InternalFieldName, value: boolean) => {
       const field: Field = get(control._fields, name);
 
-      if (field) {
+      if (field?._f) {
         field._f.mount = value;
       }
     };
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -85,7 +85,7 @@ export function useController<
     const updateMounted = (name: InternalFieldName, value: boolean) => {
       const field: Field = get(control._fields, name);
 
-      if (field?._f) {
+      if (field && field._f) {
         field._f.mount = value;
       }
     };
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "files": [
       {
         "path": "./dist/index.cjs.js",
-        "maxSize": "10.1 kB"
+        "maxSize": "10.2 kB"
       }
     ]
   },


src/useController.ts
```diff
@@ -85,7 +85,7 @@ export function useController<
     const updateMounted = (name: InternalFieldName, value: boolean) => {
       const field: Field = get(control._fields, name);
 
-      if (field) {
+      if (field?._f) {
```
bluebill1049, line 88:
```suggestion
      if (field && field._f) {
```

bluebill1049 (approved):

bluebill1049 (approved):

PR #11951, opened 2024-05-30
Allow resolver output type to truly differ

jedahu:
Removing the `TFieldValues` constraint from `TTransformedValues` allows for resolvers to parse to different shaped output types.

Defaulting `TTransformValues` to `TFieldValues` instead of `undefined` means some conditional types can be removed (like those on `UseFormHandleSubmit`).

I have changed the types only. There are no runtime changes in this PR.


Fazla-GroM:
Any progress on this ?

jedahu:
It was ready to go but didn't get merged so it's just a case of fixing the conflicts again.

jorisre:
Hi @jedahu
Thank you so much for this PR. I know it's been a while, but if you have some time to resolve the conflicts, I'll test it with the latest version of the resolvers. It could be a very nice improvement!

bluebill1049:
I will look at it, and see if I can resolve those conflicts.

bluebill1049:
@jorisre this has never pass the build 
image

Omelyan:
After the years authors still do not understand what a _Form_ is.
A Form is the transformation (and possibly validation) of an _Input_ into an _Output_!

Why is this code logically absolutely correct, yet contains a TypeScript error?!
Image

and, on the other hand, why does the following code, despite having no errors, not work?!
Image

I'm over it. Just don't use the lib.

Issue #11960, opened 2024-06-02
Provide Validation Event Type to `resolver` Function 


slaymance:
### Is your feature request related to a problem? Please describe.
A useful feature in `react-hook-form` would be the ability to run certain field validations on events other than the one specified with the `mode` prop passed to `useForm`. When using the `resolver` pattern, many validation libraries offer the ability to do conditional validation based on context values provided to the schema validation call. The request is for `react-hook-form` to pass the type of event that triggered the schema validation to the function provided as the `resolver` prop for `useForm`.

### Describe the solution you'd like
The `resolver` function which calls the schema validation method is invoked in the `_executeSchema` function within the [src/logic/createFormControl.ts](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L405) file. `_executeSchema` is invoked in various places throughout the file, namely the [onChange](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L753) and [handleSubmit](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L1137) callbacks.

The `resolver` function accepts [3 arguments](https://react-hook-form.com/docs/useform#resolver): `values`, `context`, and `options`. The proposal is for `_executeSchema` function to accept an optional `context` argument that is subsequently passed to the `resolver` function along with the `context` value passed to `useForm`. The `onChange` and `handleSubmit` functions can then pass the event type to `_executeSchema`. Finally, the `resolver` function may know the event type that has triggered the schema validation.

#### Example implementation
```typescript
// src/logic/createFormControl.ts

const _executeSchema = async (
  name?: InternalFieldName[],
  resolverContext?: ExecuteSchemaContext, // now accepts an optional "resolverContext"
) => {
  _updateIsValidating(name, true);
  const result = await _options.resolver!(
    _formValues as TFieldValues,
    { ...resolverContext, ..._options.context }, // passes "resolverContext" as 2nd arg to "resolver"
    getResolverOptions(
      name || _names.mount,
      _fields,
      _options.criteriaMode,
      _options.shouldUseNativeValidation,
    ),
  );
  _updateIsValidating(name);
  return result;
};

const handleSubmit: UseFormHandleSubmit<TFieldValues> =
  (onValid, onInvalid) => async (e) => {
    ...

    if (_options.resolver) {
      const { errors, values } = await _executeSchema(
        undefined,
[...]

PR #11961, opened 2024-06-02
feat: Provide Validation Event Type to `resolver` Function

slaymance:
**Please see #11960 for a description of the problem these changes attempt to solve.**

This pull request changes the `_executeSchema` function in the `src/logic/createFormControl.ts` file to accept an additional `context` argument. The `context` argument will then be passed to the `resolver` prop provided in a `useForm` call.

This allows context values to be passed to a validation library from `react-hook-form`. This pull request makes it so that when a schema is validated based on a user event, that event type is provided to the `resolver` function. This allows developers to have access to the event that initiated the validation call within the schema declaration itself.

#### Example Use Case with `Yup`
```jsx
import { yupResolver } from '@hookform/resolvers/yup';
import * as yup from 'yup';

const schema = yup.object({
  username: yup
    .string()
    .required()
    .min(3)
    .when('$validationEvent', {
      is: 'submit',
      then: (schema) => schema.test(
        'is-username-available',
        'Username is not available.',
        async (value) => (
          await fetch(`/usernames?username=${value}`)
        )?.usernameAvailable === true,
      ),
    }),
});

function UsernameForm() {
  const { handleSubmit, register } = useForm(
    mode: 'onChange',
    resolver: yupResolver(schema),
  );

  const onSubmit = (formData) => console.log(formData);

  return (
      <input {...register('username')} />
      <input type="submit" />
    </form>
  );
}
```

### DIFF
--- a/src/__tests__/useForm/resolver.test.tsx
+++ b/src/__tests__/useForm/resolver.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { act, fireEvent, render, screen } from '@testing-library/react';
 
+import { EVENTS } from '../../constants';
 import { useForm } from '../../useForm';
 import noop from '../../utils/noop';
 import sleep from '../../utils/sleep';
@@ -204,5 +205,53 @@ describe('resolver', () => {
     });
 
     expect(errorsObject).toEqual({});
+  });
+
+  it('should be called with the event that triggered validation', () => {
+    type FormValues = {
+      test: string;
+    };
+
+    const resolverSpy = jest.fn(() => ({ errors: {}, values: {} }));
+
+    const App = () => {
+      const { handleSubmit, register } = useForm<FormValues>({
+        mode: 'all',
+        resolver: resolverSpy,
+      });
+
+      return (
+        <form onSubmit={handleSubmit(() => null)}>
+          <input {...register('test')} />
+          <input type="submit" />
+        </form>
+      );
+    };
+
+    render(<App />);
+
+    const input = screen.getByRole('textbox');
+    const submit = screen.getByRole('button');
+
+    fireEvent.change(input, { target: { value: 'test' } });
+    expect(resolverSpy).toHaveBeenLastCalledWith(
+      expect.anything(),
+      expect.objectContaining({ validationEvent: EVENTS.CHANGE }),
+      expect.anything(),
+    );
+
+    fireEvent.blur(input);
+    expect(resolverSpy).toHaveBeenLastCalledWith(
+      expect.anything(),
+      expect.objectContaining({ validationEvent: EVENTS.BLUR }),
+      expect.anything(),
+    );
+
+    fireEvent.click(submit);
+    expect(resolverSpy).toHaveBeenLastCalledWith(
+      expect.anything(),
+      expect.objectContaining({ validationEvent: EVENTS.SUBMIT }),
+      expect.anything(),
+    );
   });
 });
--- a/src/__tests__/useForm/resolver.test.tsx
+++ b/src/__tests__/useForm/resolver.test.tsx
@@ -11,32 +11,21 @@ describe('resolver', () => {
       test: string;
     };
 
+    const resolverSpy = jest.fn(() => ({ errors: {}, values: {} }));
+    const testContext = { test: 'test' };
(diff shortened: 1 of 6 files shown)


src/__tests__/useForm/resolver.test.tsx
```diff
@@ -47,9 +37,11 @@ describe('resolver', () => {
...
     fireEvent.click(screen.getByRole('button'));
 
-    expect(
-      await screen.findByText('{"test":"test"}', undefined, { timeout: 3000 }),
-    ).toBeVisible();
+    expect(resolverSpy).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.objectContaining(testContext),
+      expect.anything(),
+    );
```
slaymance, line 44: The changes caused this test to fail. I took the opportunity to refactor it, so it asserts on the actual arguments passed to `resolver`.

src/__tests__/useForm/resolver.test.tsx
```diff
@@ -214,4 +206,52 @@ describe('resolver', () => {
...
+
+    fireEvent.click(submit);
+    expect(resolverSpy).toHaveBeenLastCalledWith(
+      expect.anything(),
+      expect.objectContaining({ validationEvent: EVENTS.SUBMIT }),
+      expect.anything(),
+    );
+  });
```
slaymance, line 256: This new unit test asserts on all of the possible validation event types by passing `mode: 'all'` to `useForm`. Willing to split into a single test per assertion if requested.

src/constants.ts
```diff
@@ -2,6 +2,7 @@ export const EVENTS = {
   BLUR: 'blur',
   FOCUS_OUT: 'focusout',
   CHANGE: 'change',
+  SUBMIT: 'submit',
```
slaymance, line 5: Since `submit` is an event type that can trigger schema validation and will be passed to the `resolver` function, I felt this was a good addition to the `EVENTS` constant.

src/logic/createFormControl.ts
```diff
@@ -400,11 +401,17 @@ export function createFormControl<
...
+    context?: ResolverContext,
+  ) => {
+    const resolverContext = context
+      ? ({ ...context, ..._options.context } as ResolverContext & TContext)
+      : _options.context;
     _updateIsValidating(name, true);
     const result = await _options.resolver!(
       _formValues as TFieldValues,
-      _options.context,
+      resolverContext,
```
slaymance, line 414: `_executeSchema` now accepts an additional `context` parameter which is passed as the second argument to the `resolver` function provided to `useForm` (along with the optional `context` prop provided to `useForm`).

src/logic/createFormControl.ts
```diff
@@ -750,7 +757,9 @@ export function createFormControl<
       !isBlurEvent && watched && _subjects.state.next({ ..._formState });
 
       if (_options.resolver) {
-        const { errors } = await _executeSchema([name]);
+        const { errors } = await _executeSchema([name], {
+          validationEvent: isBlurEvent ? EVENTS.BLUR : EVENTS.CHANGE,
+        });
```
slaymance, line 762: The `blur`, `change` and `submit` events take advantage of the new `context` argument that `_executeSchema` accepts so that the `resolver` function may know the event type that triggered the validation.

src/types/resolvers.ts
```diff
@@ -16,6 +18,13 @@ export type ResolverResult<TFieldValues extends FieldValues = FieldValues> =
   | ResolverSuccess<TFieldValues>
   | ResolverError<TFieldValues>;
 
+export interface ResolverContext {
+  validationEvent?:
+    | typeof EVENTS.BLUR
+    | typeof EVENTS.CHANGE
+    | typeof EVENTS.SUBMIT;
+}
```
slaymance, line 26: A new `ResolverContext` interface gives us the flexibility to provide additional context from `react-hook-form` to a validation library in the future, should it be a pattern that proves useful.

jgubman (approved):
This is awesome! Having access to the resolver's context would be so helpful

jgubman:
This would be super helpful

bombillazo:
Is this done now that #11960 is closed?

Issue #11985, opened 2024-06-06
issue: setValue with shouldDirty on nested objected throws TypeError


pincy:
### Version Number

7.51.5

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/priceless-morning-xsk5sw

### Steps to reproduce

1. Go to [codesandbox.io](https://codesandbox.io/p/sandbox/priceless-morning-xsk5sw)
2. Click on the 'Reset Customer' Button
3. See error

### Expected behaviour

The nested object to be set to null, without errors.

The most recent version, working with that usage was ~7.43.5.

### What browsers are you seeing the problem on?

Firefox, Chrome

### Relevant log output

```shell
Firefox:

ERROR
get(...)._f is undefined
updateTouchAndDirty@http://localhost:3000/static/js/bundle.js:42734:52
setFieldValue@http://localhost:3000/static/js/bundle.js:42878:53
setValue@http://localhost:3000/static/js/bundle.js:42909:101
handleClick@http://localhost:3000/static/js/bundle.js:45:10

Chrome:

ERROR
Cannot read properties of undefined (reading 'disabled')
TypeError: Cannot read properties of undefined (reading 'disabled')
    at updateTouchAndDirty (http://localhost:3000/static/js/bundle.js:42734:73)
    at setFieldValue (http://localhost:3000/static/js/bundle.js:42878:53)
    at Object.setValue (http://localhost:3000/static/js/bundle.js:42909:101)
    at handleClick (http://localhost:3000/main.9b447242c2bada3d5aa2.hot-update.js:44:10)
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #11986, opened 2024-06-06
fix issue:#11985 logic createFormControl check field before usage

pincy:
In the function `updateTouchAndDirty` the disabled field is accessed before checking the existence of the _f field.

This change adds the check for the _f field in fields before usage.


bluebill1049 (approved):
Thanks for the fix!

PR #12022, opened 2024-06-16
fix: form-level `disabled` not set for inputs that `register` after initial render

### LINKED ISSUE #12021: issue: Form-level `disabled` doesn't affect uncontrolled inputs that aren't rendered immediately
### Version Number

7.51.5

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/white-glitter-lqhxg3

### Steps to reproduce

1. Visit the repro (https://codesandbox.io/p/sandbox/white-glitter-lqhxg3)
2. Only the controlled input is disabled

image

### Expected behaviour

Both fields are disabled

### Code of Conduct

amitdahan:
Closes #12021


amitdahan:
Will turn into draft until I fix the failing tests (I think I made `useController` re-render when it shouldn't)

PR #12039, opened 2024-06-21
🐞 fix #12036 async defaultValues in React.StrictMode is called only once

### LINKED ISSUE #12036: issue: Async defaultValues function overwrites values in React.StrictMode
### Version Number

7.51.5

### Codesandbox/Expo snack

https://codesandbox.io/p/github/xemle/react-hook-form-async-defaultValues/main

### Steps to reproduce

For the Codesandbox:

1. The form field `isFirst` will have 3 different values
1. `undefined` on the initial load
2. `false` after one second by the faster second call of async `defaultValues` function
3. `true` at second 2 by the first call of async `defaultValues` function

#### Background

When using a async `defaultValues` function in `React.StrictMode` the `defaultValues` function is executed twice and the later resolve overwrites the form values. This can lead to invalid form values when values are updated between the calls.

While the controller is initiated only once by `useRef` in `useForm.ts`, the `_resetDefaultValues()` function is called twice and executes the `defaultValues` function. The `defaultValues` function returns just the a plain object the previous values are overwritten. Also the dirty, touch states are overwritten if not specified in the `resetOptions`

#### Workaround

`_resetDefaultValues()` calls `reset()` on resolve. By undocumented functionality of `reset(values)`, `values` can be also a function which receives the previous form values as argument. The async `defaultValues` function can return a function which checks if values are already set:

```
const methods = useForm({
  defaultValues: async () => {
    const values = await yourAsyncCall()
    return prev => Object.keys(prev).length ? prev : values
  },
  resetOptions: {
    keepTouched: true
  }
}
```

### Expected behaviour

If `defaultValues` is a function, it is only called once when the controller is created.

Chrome

### Relevant log output

_No response_

### Code of Conduct

xemle:
Fixes #12036

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -122,7 +122,8 @@ export function useForm<
       control._reset(props.values, control._options.resetOptions);
       _values.current = props.values;
       updateFormState((state) => ({ ...state }));
-    } else {
+    } else if (!props.values && !control._state.isLoadingValues) {
+      control._state.isLoadingValues = true;
       control._resetDefaultValues();
     }
   }, [props.values, control]);
--- a/src/__tests__/useForm.test.tsx
+++ b/src/__tests__/useForm.test.tsx
@@ -2075,6 +2075,55 @@ describe('useForm', () => {
     });
   });
 
+  it('should update async default values for controlled components only once in strict mode', async () => {
+    let calls = 0;
+
+    const formProps = {
+      defaultValues: async () => {
+        const test = calls == 0 ? 'test' : 'other';
+        const timeout = calls == 0 ? 100 : 50;
+        calls++;
+
+        await sleep(timeout);
+
+        return {
+          test,
+        };
+      },
+    };
+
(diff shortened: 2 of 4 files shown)


alexandermirzoyan (approved):
Nice one!

bluebill1049:
thanks if you could update api extator

src/useForm.ts
```diff
@@ -122,7 +122,8 @@ export function useForm<
       control._reset(props.values, control._options.resetOptions);
       _values.current = props.values;
       updateFormState((state) => ({ ...state }));
-    } else {
+    } else if (!props.values && !control._state.isLoadingValues) {
+      control._state.isLoadingValues = true;
```
bluebill1049, line 126: do you need this to be reset after reset API?

bluebill1049: Yea, so when it's ever test back to false? I can see only set to true here

PR #12051, opened 2024-06-26
feat: render multiple errors on change

Cbridger-mywellabee:
Experimental demonstration which allows errors to render for other fields aside from the one that was touched.

pass `experimentalMultiError: true` to useForm

currently doesn't support delayed errors and hasn't been tested against any other resolver aside from yup


Cbridger-mywellabee:
@bluebill1049 Unsure how to get API Extractor step to pass, I ran `yarn api-extractor:build` as specified and committed the file but it still failed the check.

PR #12054, opened 2024-06-26
fix: skip call `executeBuiltInValidation` if no sub-fields left

newsiberian:
We've had a performance issue with `useFieldArray` and too many rows with validated fields. I've noticed that with function do some extra recursion with empty an `fieldValue` which mostly the reason of our issues

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -481,7 +481,7 @@ export function createFormControl<
               : unset(_formState.errors, _f.name));
         }
 
-        Object.keys(fieldValue).length &&
+        !isEmptyObject(fieldValue) &&
           (await executeBuiltInValidation(
             fieldValue,
             shouldOnlyCheckValid,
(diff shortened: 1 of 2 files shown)


src/logic/createFormControl.ts
```diff
@@ -481,7 +481,7 @@ export function createFormControl<
               : unset(_formState.errors, _f.name));
         }
 
-        fieldValue &&
+        Object.keys(fieldValue).length &&
```
bluebill1049, line 484: you can use `isEmptyObject` function to check

bluebill1049:
thanks, could you plz add a test case for this as well? so it validates a correct ussecase.

newsiberian:
Sure! I'm a bit busy atm. Will try to add tests during the next week

newsiberian:
@bluebill1049 done. Please, check. Kinda looks strange, but I was unable to directly mock `executeBuiltInValidation` and that's the way that came into my mind.

bluebill1049 (approved):

bluebill1049:
thanks for the improvement.

PR #12073, opened 2024-07-03
chore: update rollup to v4

abnud11:
This PR upgrades Rollup from v2 to v4, it also upgrades its plugins to the latest version in order to be compatible with Rollup 4.

Some Notes:

1. rollup-plugin-peer-deps-external plugin is removed and replaced with simple `external` config option.
2. rollup-plugin-sourcemaps is removed, I compared dist folder before upgrading rollup and after the upgrade and removal of this plugin, I found source maps with almost identical output so I'm not sure why this plugin was ever included.
3. We have verified that the output bundles are almost identical to the previous bundles in terms of output, we also ensured that size didn't increase.
4. since Rollup 4 requires minimum Node.js version 18, we updated `engines` field in `package.json`, this can be considered a breaking change for developers of react-hook-form who are still using Node.js older than 18(for consumers of the library, only a warning will be shown if they don't use Node.js 18 or later). I think this is a reasonable breaking change since the CI uses Node.js 20 and all libraries are requiring Node.js 18 or later in their newest version.

The change should have resulted in better build performance but to be honest on my local machine at least it didn't, the performance didn't degrade though, just not improved either.

But the upgrade is deemed required since Rollup v2 is a very old version.


bluebill1049 (commented):
LGTM! Awesome stuff.

bluebill1049 (approved):

dr3:
> this can be considered a breaking change for developers of react-hook-form who are still using Node.js older than 18(for consumers of the library, only a warning will be shown if they don't use Node.js 18 or later).

Im not sure this is true for users of `yarn 1`, it throws an error when consuming

```
error react-hook-form@7.53.0: The engine "node" is incompatible with this module. Expected version ">=18.0.0". Got "16.14.2"
error Found incompatible module.
```

abnud11:
@dr3 sorry for the inconvenience, anyway the CI used Node.js 20 even before this PR and also Node.js 16 is considered EOL now.

Also I think yarn v1 is EOL too, right?

abnud11:
And you can use `yarn install --ignore-engines`

PR #12079, opened 2024-07-05
chore: upgrade pnpm to v9

abnud11:
pnpm v8 is now considered a legacy version, we need to upgrade to v9.

This PR upgrades `pnpm-lock.yaml` files to version supported by `pnpm` 9 and also updates the pnpm version used in Github workflows.

This isn't considered a breaking change for consumers of `react-hook-form` but it's a breaking change for those who do development work on `react-hook-form`, `CONTRIBUTING.md` got updated to include info about pnpm version used and Node.js version too.

### Note
a `.npmrc` is created inside `app` folder with `auto-install-peers=true` content, this was required in order for E2E github workflow to work, otherwise a pnpm error occurs about mismatch between lockfile and settings.

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -5,9 +5,9 @@ runs:
   using: composite
   steps:
     - name: Install pnpm
-      uses: pnpm/action-setup@v3
+      uses: pnpm/action-setup@v4
       with:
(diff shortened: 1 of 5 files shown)


bluebill1049 (approved):
LGTM! awesome, thanks again!

PR #12082, opened 2024-07-05
chore: upgrade cypress to v13

abnud11:
This PR upgrades cypress from v10 to v13, a lot of breaking changes happened between these 2 versions but luckily the E2E tests weren't affected by any of them except the change about Test isolation.

E2E tests are now fully isolated, on each test run, page goes blank, cookies are cleared along with localStorage and any state.

Because of this, one of the tests failed since it relied on state preserved from previous state, this PR fixes that to make tests not depend on each other.

We are lucky that only one test needed to change.

### DIFF
--- a/.github/workflows/automation.yml
+++ b/.github/workflows/automation.yml
@@ -13,7 +13,7 @@ jobs:
         uses: ./.github/actions/install-dependencies
 
       - name: Cypress run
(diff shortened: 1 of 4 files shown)


bluebill1049 (approved):
thanks again for uplift the dev dep! awesome stuff

PR #12085, opened 2024-07-06
chore: upgrade react testing library

abnud11:
This PR upgrades `@testing-library/react` and `@testing-library/jest-dom` to the latest version, since `renderHook` is now exported from `@testing-library/react` itself, there is no need for `@testing-library/react-hooks` (as matter of fact, the latter isn't compatible with React 18 contrary to the former).

That's why this PR removes `@testing-library/react-hooks` and replace its usage with `renderHook` exported from `@testing-library/react`.


bluebill1049:
please fix the conflict, thanks.

abnud11:
> please fix the conflict, thanks.

conflicts fixed

bluebill1049 (approved):
LGTM! thanks

PR #12086, opened 2024-07-06
chore: upgrade msw

abnud11:
`msw` is used only in one test but it's good to upgrade it from v1 to v2.


bluebill1049 (approved):
LGTM! thanks

PR #12090, opened 2024-07-08
chore: upgrade app dependencies

abnud11:
This PR upgrades all dependencies of `app` folder except React.

React version is preserved at `17.0.1` since `react-hook-form` should be compatible with react lower than 18 for now and testing against react 17 is good, once we drop support for react lower than 18 we will need to update `app` folder to use React 18(or 19 maybe).

Also `material-ui` is upgraded to v5 with no issues what so ever.

The verification of the upgrade goes by E2E tests, if they succeed then the upgrade is good.


bluebill1049 (approved):

bluebill1049:
thanks!

PR #12130, opened 2024-07-24
Fix integer default value not working for radio inputs

bgirschig:
Currently, setting default values with integers does not work for radio inputs:

```tsx
import React from "react";
import ReactDOM from "react-dom";

function App() {
  const choices = [12, 17, 192, 10];
  const defaultValues = { myNumber: 192 };

  const { register } = useForm({ defaultValues });

  return (
    <form>
      {choices.map((choice) => {
        return (
          <label key={choice} style={{display: "block"}}>
            <input
              type="radio"
              value={choice}
              {...register("myNumber")}
            />
            {choice}
          </label>
        );
      })}
    </form>
  );
}

const rootElement = document.getElementById("root");
ReactDOM.render(<App />, rootElement);
```

Converting the defaultValue to a string does work:
```tsx
const defaultValues = { myNumber: "192" };
```

This PR makes this conversion within react hook forms, in a way that (I believe) is consistent with what a user would expect.

Note: Please keep in mind that this is my fist PR here and I'm quite new to react as well as react-hook-form, so you can safely assume I have no idea what I'm doing here.

Checklist:
[✓] create tests
[✓] pass tests
[✓] lint
[✓] automation suite
[✓] build
[✓] api-extractor

### DIFF
--- a/src/__tests__/useForm/setValue.test.tsx
+++ b/src/__tests__/useForm/setValue.test.tsx
@@ -1392,4 +1392,45 @@ describe('setValue', () => {
     expect(screen.getByText('dirty')).toBeVisible();
     expect(screen.getByText('touched')).toBeVisible();
   });
+
+  it('should handle integer value as defaultValue', () => {
+    function App() {
+      const { register } = useForm({
+        defaultValues: { myField: 12 },
+      });
+
+      return (
+        <form>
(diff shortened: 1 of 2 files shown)


bgirschig:
fixes #12172

PR #12150, opened 2024-08-01
chore: upgrade eslint to v9

abnud11:
This PR upgrades ESLint to v9 along with using flat config instead of legacy config.

After the upgrade there should be no new errors or warnings which is verified in this PR.

Fortunately the plugins we use support flat config except `eslint-react-hooks-plugin` which actually doesn't support v9 but I got it to work with `@eslint/compat` package. According to this issue:
https://github.com/facebook/react/issues/28313 ESLint 9 is planned in Canary release so in the future we should expect to get rid of `@eslint/compat` but for now this is necessary.

For `typescript-eslint`, v8 got released today but upgrading it resulted in many new errors, so I made this PR to upgrade to ESLint v9 and will make another PR to upgrade `typescript-eslint` to v8.


bluebill1049 (dismissed):

carakessler:
thank you for this work! plz merge :)

abnud11:
@bluebill1049 it seems you forgot to merge? since it's approved and all checks pass

bluebill1049:
thanks, could you take a look at some of the warnings?

abnud11:
I think I fixed the warnings, I found many `@ts-expect-error` to be unnecessary, some of them only needed an explanation.

I have changed `FieldRefs` type though, it can be a breaking change but it's more correct since it seems the implementation of the methods which uses it either expect an object of fields or object of fields refs.

bluebill1049:
sorry, i forgot to merge ... could you resolve the conflicts?

abnud11:
@bluebill1049 conflicts resolved

package.json
```diff
@@ -138,5 +138,6 @@
   },
   "engines": {
     "node": ">=18.0.0"
-  }
+  },
+  "packageManager": "pnpm@9.12.1+sha512.e5a7e52a4183a02d5931057f7a0dbff9d5e9ce3161e33fa68ae392125b79282a8a8a470a51dfc8a0ed86221442eb2fb57019b0990ed24fab519bf0e1bc5ccfc4"
```
bluebill1049, line 142: Do we need this? what's this for?

abnud11: this is to determine the `pnpm` version used by the project, tools like `corepack` uses this information to install the appropriate package manager and version, the github action also uses this same information.

Having a central place for versioning `pnpm` in `package.json` is preferrable, people will know what version of `pnpm` is installed just by looking at `package.json` and using tools like `corepack`, it will install the correct version automatically.

bluebill1049: could we remove it, plz? i don't have enough knowledge on this, don't want to include any security-related* stuff to the package.

abnud11:
@bluebill1049 pnpm version from package.json is removed now, you can check the PR again.

bluebill1049 (approved):
LGTM! thank you

PR #12267, opened 2024-09-21
☸️ fix #11910 watch reference update on each re-render

### LINKED ISSUE #11910: issue: Watch don't work with React Compiler (React 19)
### Version Number

7.51.4

### Codesandbox/Expo snack

https://github.com/Maxou44/issue-react-compiler-react-hook-form

### Steps to reproduce

1. Clone `https://github.com/Maxou44/issue-react-compiler-react-hook-form`
2. Run `npm i --force && npm run start`
3. Open a web browser on `http://localhost:5173/`
4. Type something in the input field, the watched value is not displayed on the page

To disable the React compiler and confirm it works without the compiler, comment the line `plugins: [["babel-plugin-react-compiler", ReactCompilerConfig]],` in vite.config.mjs, relaunch `npm start` and type something in the input field

### Expected behaviour

The use of React Compiler don't allow the use of the watch() function, it should work.

Firefox, Chrome, Safari, Edge

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:


lauri865:
@bluebill1049, when will this be released you think? Pretty major blocker for us in migrating to React 19. Thanks! 🙏

lauri865:
Actually, the fix for React 19 / Compiler is even more benign.

`watch` is for all intents and purposes a hook. And rules of hooks say that:
> Hook names must start with *use* followed by a capital letter

Since this rule has been broken within react-hook-form, compiler memoizes it away so that it's not reactive anymore.

by simply changing the following:
```typescript
const { watch } = useForm(...);
watch("fieldName");
```

To:
```typescript
const { watch: useWatch } = useForm(...);
useWatch("fieldName");
```

Et voila, everything is working again on React 19 + react-compiler + react-hook-form 7.53.0

So, I think for backwards compatibility, `useForm` and `useFormContext` could return both `watch` and `useWatch` with deprecation notice for the `watch` function, saying that it doesn't work with react-compiler. What do you think @bluebill1049?

PR #12278, opened 2024-09-27
fix: Trigger validation on Date field change

peinguin:
The problem is that validation does not triggers on fields with value of type Date.
Is it intentionally?
How it supposed to work with other non-primitive values?


peinguin:
Not relevant

Issue #9289, opened 2022-10-28
Introduce useLens hook


aspirisen:
**Is your feature request related to a problem? Please describe.**
* Currently we do not have single type for register binding
* Working with arrays is hard as you have to keep track of index. If you have nested arrays you have to create component wrapper as you have to use `useFieldArray`  hook.
* Lack of TypeScript support, `register()` is not generic.
* Hooks like useFieldArray, useFormState, useController do not know what TypeScript type they are going to operate. There is no internal way for type inference

**Describe the solution you'd like**
Here I suggest to introduce new type "Lens" which you can use to define field structure. It will help you to wire form elements with good TypeScript support. You can define a component that has Lens with explicit type in its props.

Lens consist all type information it needs and can be explicitly transferred via props. Inside child component type inference is working.

Here is API example how it can be used:
```
function Input(props: { lens: Lens<string> }) {
  return <input {...props.lens.register()} />;
}

interface FormData {
  name: string;
  children: { firstName: string; lastName: string }[];
}

function Example() {
  const form = useForm<FormData>();
  const lens = useLens(form);

  return (
    <div>
      <Input lens={lens.prop("name")} />
      {lens.prop("children").map((l) => (
        <div>
          <Input lens={l.prop("firstName")} />
          <Input lens={l.prop("lastName")} />
        </div>
      ))}
    </div>
  );
}
```
Here is rough implementation of this approach with more complex example: https://codesandbox.io/s/react-hook-form-lens-xn1p3z?file=/src/TestForm.tsx
It uses `react-hook-form` internals a lot.

bluebill1049:
thanks for the feature request, pinging to the project feature board, would it be possible to make this an external hook library?

aspirisen:
@bluebill1049 thanks, i am not sure that it is worth to make it as external hook. In the POC from codesandbox you can see that it uses a lot of internal stuff from react-hook-form library (and some copy-pastes). Any change of internals will make the possible external hook non-working. For example in the codesandox example i do not use the latests version as there are some changes in latest version of react-hook-form library.

PR #12284, opened 2024-09-30
Lenses support to help with reusable components

aspirisen:
Hi, here is a rough demo on how Lenses approach can help with reusable components. I had a feature request some time ago https://github.com/react-hook-form/react-hook-form/issues/9289, but making lenses stateful brought a lot of difficulties.

In this PR lenses are stateless, and in fact they just wrap controller and path together with reusable and shareable typings.

Related discussions: 
https://github.com/orgs/react-hook-form/discussions/7354
https://github.com/orgs/react-hook-form/discussions/10419
https://github.com/orgs/react-hook-form/discussions/8311
https://github.com/orgs/react-hook-form/discussions/10176
https://github.com/orgs/react-hook-form/discussions/7948
https://github.com/orgs/react-hook-form/discussions/10873

I added two examples: first with basic usage, and the second with transform support - when you have a generic component that uses fields with different names from your parent form. In order to make it properly work we need internal support for some kind of "transformed controller"

Here is a small usage example:
```tsx
function StringInput(props: { label: string; lens: Lens<string> }) {
  return (
    <>
      <label>
        {props.label}
        <input {...props.lens.register({})} />
      </label>
    </>
  );
}

function PersonForm({
  lens,
}: {
  lens: Lens<{
    name: string;
    surname: string;
  }>;
}) {
  return (
    <>
      <StringInput label="Name" lens={lens.focus('name')} />
      <StringInput label="Surname" lens={lens.focus('surname')} />
    </>
  );
}

export type FormValues = {
  title: string;
  children: { name: string; surname: string }[];
  firstName: string;
  lastName: string;
};

export default function DemoForm() {
  const lens = useLens({ control });

  return (
    <form>
      <StringInput label="Title" lens={lens.focus('title')} />
      <PersonForm
        lens={lens.transform((l) => ({
          name: l.focus('firstName'),
          surname: l.focus('lastName'),
        }))}
      />
      {lens.focus('children').map((l) => (
        <PersonForm lens={l} />
      ))}
    </form>
  );
}
```


bluebill1049:
Thanks for the PR, i wonder if this could be a separate package? which the documentation can promote and include?

aspirisen:
Theoretically that can be a separate package, but will require additional work in `react-hook-form`, we should have something like `TransformedController`, to cover case when reusable component uses different shape and field names (such problem can happen and if you do not use lenses).  Maybe there can be additional pitfalls that will require internals of `react-hook-form`.

However, according to the related discussions I can think that people want in-house solution for reusable form components.

bluebill1049:
Would you like to publish that package first? we can include the documentation. Can we bring it into the main lib afterwards?

aspirisen:
@bluebill1049 I could create a package in personal repo, but I am not sure that I could handle this new api in solo since there definitely will be complex cases i.e. with arrays

I just thought if it is possible to create a community repo similar to https://github.com/visgl/deck.gl-community 
That can have packages like `resolvers`, `error-message` and other packages under `@hookform/*` namespace. 

By that it will be add easier to add packages, they will be under community support and there will be less worry about maintenance 

If it is too much I could try to go by personal package approach.

bluebill1049:
yea sounds good @aspirisen, i will create a repo now we can move it there. https://github.com/react-hook-form/lenses

aspirisen:
Thanks, but I in my opinion I thought that one monorepo with community packages will suite a bit better instead of creating repo per packages. By that we could share i.e. ci configs, define single eslint/prettier rules, tests, [peer dependencies](https://pnpm.io/catalogs) etc.

bluebill1049:
> Thanks, but I in my opinion I thought that one monorepo with community packages will suite a bit better instead of creating repo per packages. By that we could share i.e. ci configs, define single eslint/prettier rules, tests, [peer dependencies](https://pnpm.io/catalogs) etc.

yea, requires more effort on that.

aspirisen:
@bluebill1049 could you please review https://github.com/react-hook-form/lenses/pull/1

PR #12308, opened 2024-10-07
feat: add support for opaque fields

DASPRiD:
Adds a new Opaque<T> type which can be used to mark certain property values as opaque. This prevents unwanted expansion of properties which RHF should consider opaque.

As an example, if you have a form field for a date picker, and the picker value is a `Dayjs` value, right now RHF will recommend e.g. `date.clone` as a valid field path. By wrapping it with the new `Opaque<Dayjs>`, RHF will then treat it like any other special type like `Date` or `File`.

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -131,19 +131,19 @@ export type DeepMap<T, TValue> = IsAny<T> extends true ? any : T extends Browser
 } : TValue;
 
 // @public (undocumented)
-export type DeepPartial<T> = T extends BrowserNativeObject | NestedValue ? T : {
+export type DeepPartial<T> = T extends BrowserNativeObject | NestedValue | Opaque<unknown> ? T : {
     [K in keyof T]?: ExtractObjects<T[K]> extends never ? T[K] : DeepPartial<T[K]>;
 };
 
 // @public (undocumented)
-export type DeepPartialSkipArrayKey<T> = T extends BrowserNativeObject | NestedValue ? T : T extends ReadonlyArray<any> ? {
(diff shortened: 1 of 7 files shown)


mestir (approved):

DASPRiD:
@mestir Is something blocking this from being merged?

bluebill1049:
Sorry i will come back to this, I think we will need to update the doc as well if we are going to roll out this type https://react-hook-form.com/ts

bluebill1049:
Could you create a PR under the doc as well? https://react-hook-form.com/ts https://github.com/react-hook-form/documentation that way ppl know such type exists and know how to use it.

bluebill1049:
@kotarella1110 would be able to review this one?

kotarella1110:
@DASPRiD 

This change seems quite similar to NestedValue from React Hook Form. How does the new Opaque type differ from NestedValue?

https://react-hook-form.com/ts

DASPRiD:
@kotarella1110 I looked through the code and all occurrences of `NestedValue`, but that type does not seem to do the same thing as my proposed `Opaque`, which is preventing RHF to see the inner object properties as possible field properties.

Apart from that, `NestedValue` is also deprecated :)

bluebill1049:
> @kotarella1110 I looked through the code and all occurrences of `NestedValue`, but that type does not seem to do the same thing as my proposed `Opaque`, which is preventing RHF to see the inner object properties as possible field properties.
> 
> Apart from that, `NestedValue` is also deprecated :)

We can remove the deprecated if NestedValue can achieve what we after and combine the type if so.

DASPRiD:
Looking at the original pull request, `NestedValue` solves a completely different use case:
https://github.com/react-hook-form/react-hook-form/pull/1503

Making it work like the proposed `Opaque` type would break BC as far as I can see.

DASPRiD:
Not stale.

DASPRiD:
@bluebill1049 Can you re-open this please?

PR #12317, opened 2024-10-08
🐞 fix #12316 setValue should work for arrays of primitives to handle checkboxes (#12316)

### LINKED ISSUE #12316: issue: setValue() doesn't update the state of multiple checkbox values (arrays)
### Version Number

7.53.0

### Codesandbox/Expo snack

https://codesandbox.io/s/eager-sun-jt4df?file=/package.json

### Steps to reproduce

1. Create a form with a nested multi-checkbox type, e.g.

```tsx
        <label className="label"><input {...register("fields.checkbox")} type="checkbox" name="checkbox" value="1" /> Checkbox one</label>
        <label className="label"><input {...register("fields.checkbox")} type="checkbox" name="checkbox" value="2" /> Checkbox two</label>
        <label className="label"><input {...register("fields.checkbox")} type="checkbox" name="checkbox" value="3" /> Checkbox three</label>
        <label className="label"><input {...register("fields.checkbox")} type="checkbox" name="checkbox" value="4" /> Checkbox four</label>
        <button type="button" onClick={() => setValue("fields", { checkbox: ['1', '3'] })}>Check odd numbers</button>
        <input type="submit" />
      </form>
```

Click the button to call setValue with the checkbox fields nested inside.

### Expected behaviour

The checkboxes should be checked. The internal value is updated, but the actual form state isn't updated because in `setValue` we call `setValues` instead of `setFieldValue` if the value isn't primitive - this looks like it was only intended to cover plain objects but it also catches arrays. This means that RHF tries to setFieldValue on `fields.checkbox.0` (etc) which won't have a matching _fields ref.

_No response_

### Relevant log output

_No response_

### Code of Conduct

matmannion:
Fix #12316 by making the type check stricter for whether we should recurse into nested values or set values at that level. The original `!isPrimitive` check matched all object types (including arrays)

### DIFF
--- a/src/__tests__/useForm/setValue.test.tsx
+++ b/src/__tests__/useForm/setValue.test.tsx
@@ -158,6 +158,81 @@ describe('setValue', () => {
     });
   });
 
+  it('should set value of multiple checkbox input correctly as a child', async () => {
+    const { result } = renderHook(() =>
+      useForm<{ parent: { test: string[] } }>(),
+    );
+
+    const { ref } = result.current.register('parent.test');
+
+    const elm = document.createElement('input');
+    elm.type = 'checkbox';
+    elm.name = 'test';
+    elm.value = '2';
+
+    document.body.append(elm);
+    isFunction(ref) && ref(elm);
+
+    const { ref: ref1 } = result.current.register('parent.test');
+
+    const elm1 = document.createElement('input');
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
LGTM

bluebill1049:
thanks for the fix 👍

PR #12332, opened 2024-10-12
Fix: add `FileList` availability check for environments without `FileList` support

alexandredev3:
This PR improves the `cloneObject` utility function by checking if the `FileList` object is available in certain environments. In some cases, such as Adobe Photoshop Plugin development (UXP), the `FileList` object is not defined, which cause the application to crash. To address this issue, this PR includes a conditional check `(typeof FileList !== "undefined" ? data instanceof FileList : false)` to ensure the function can safely clone objects without throwing errors in weird environments.


bluebill1049:
Thanks for the PR.

>  in certain environments.

could I get a better understand of under which environments?

Also would be great to add some unit test coverage if the above is valid. 👍

alexandredev3:
Hello there!

> could I get a better understand of under which environments?

I spot this issue while developing a plugin for Adobe Photoshop. In the Adobe ecosystem plugin environment, JavaScript support is limited: there’s no `FileList` object or `<input type="file" />`. Because of this, using the library throws an error: `"FileList" is not defined`. To work around this, I added a patch with an extra check: `typeof FileList !== 'undefined'` so I could work with the library.

> Also would be great to add some unit test coverage if the above is valid.

I can add unit tests if this addition is valid.

bluebill1049:
Interesting, if you could include a test case that would be nice.

alexandredev3:
> Interesting, if you could include a test case that would be nice.

Will do.👍

alexandredev3:
Hey @bluebill1049 !

I just added a test case for `cloneObject` when `FileList` is not defined.

bluebill1049 (approved):

PR #12309, opened 2024-10-07
test: Add task running unit tests with react-compiler

stevoland:
Related: https://github.com/react-hook-form/react-hook-form/issues/12298

Adds `test:react-compiler` task which runs the unit test suite with the test compiled (currently failing)

Doesn't touch production code but fixes test code that the compiler or babel-jest breaks

PR #12333, opened 2024-10-12
🔬 V8 (#12309)

bluebill1049:
# v8

- [x]  Support react-compiler
- [x] Support flat field array
- [x] ⚠️ Breaking change: pass input ref instead of partial (https://github.com/react-hook-form/react-hook-form/issues/12773)
- [x]  ⚠️ Breaking change: rename `useFieldArray` 
  - `id` to `key`
  - `keyName` props removed for `useFieldArray`

```tsx
const { fields, append } =useFieldArray()

// data will be part of the submission 
append({
  key: 'key',  // does not affect render key
  id: 'key',  // does not affect render key
})

field[0].key // will still be uniq id for re-render
```
- [x]  ⚠️ Breaking change: rename `<Watch />` rename `names` to `name`
- [x]  ⚠️ Breaking change: `watch` call back api removed
- [x]  ⚠️ Breaking change: `setValue` no longer direct update useFieldArray, instead use `replace` api 
https://react-hook-form.com/docs/useform/setvalue
> Important: use replace from useFieldArray instead, update entire field array with setValue will be removed in the next major version.

### install

```
npm i react-hook-form@8.0.0-beta.0
```

related #12990


ellemedit:
could I know what's blocking this?

bluebill1049:
> could I know what's blocking this?

I will need more time to get to it and then release an RC version of it later.

ybelakov:
Any updates here?

Maxou44:
How can we help you with this PR ?

Maxou44:
Just a little message to say I tried version "8.0.0-alpha.5", and it solved all my issues with watch() and the React Compiler 😀

Thanks for the fix! 🚀

juozas-rimantas-tg:
Is there an ETA for a stable release of version 8?

bluebill1049:
> Is there an ETA for a stable release of version 8?

We’ll let the alpha version run for a while to identify any potential issues before moving forward with the major release.

Maxou44:
The latest version of react compiler ignore files with a watch(), it will be complex to test it in real conditions

BrendanC23:
This version has a breaking change that renames the React key value returned from useFieldArray from `id` to `key`.

https://github.com/react-hook-form/react-hook-form/blob/7c403add0639273cfea2c1cc0a42834d9d1b4fa7/src/useFieldArray.ts#L99

As mentioned in [this discussion](https://github.com/orgs/react-hook-form/discussions/8935), many users have run into issues with `id` overwriting their data. I don't think switching the name to `key` will actually solve this. It just changes the name of the field that will be overwritten.

As this is version 8.0.0 and there will be breaking changes anyway, I wonder if there's a better solution that would be a more permanent fix (and would not require users to provide a separate `keyName` prop).

1. Use a more unique default than `key` that is unlikely to overwrite the user's data (for example, `rhfKey`)
1. Add a `getValues()` method to the `field` value (see [this comment](https://github.com/orgs/react-hook-form/discussions/8935#discussioncomment-12838703).
1. Change the type of `field` to `{ key: string, fieldValue: FormData }`. This would allow separation of the key value and the user's form data. See [this comment](https://github.com/orgs/react-hook-form/discussions/8935#discussioncomment-7012063).

As mentioned in [this comment](https://github.com/orgs/react-hook-form/discussions/8935#discussioncomment-4002113):

> The prop fields returned from hook useFieldArray is not a part of form values, it's an controlled array to to work with dynamic fields via loop.
>If you want to get form values which includes the dynamic field values belongs to this prop "fields", you should use method getValues from useForm or via onValid callback of method handleSubmit.

Option 2 would help solve this. The `field` value not being reactive is a common source of confusion  that this would fix.

BrendanC23:
I came across [this discussion](https://github.com/orgs/react-hook-form/discussions/7558) about `watch()` vs `useWatch()` which had some valuable comments. It seems like there's a lot of confusion around when to use each (that `watch` would trigger re-renders at the root of the form and that `useWatch` would only re-render the calling component). It doesn't seem like there's any use-case where `watch` is better than `useWatch`.

If V8 is going to include breaking changes, I wonder if it's worth it to remove or deprecate the `watch("name")` API. For performance reasons, `useWatch` is preferable (the linked discussion doesn't list any instances where this isn't true). This would help eliminate a common source of confusion.

When retrieving values outside of a component body (where a hook can't be used, such as in an `onChange` callback or `useEffect`), `getValues()` works well and won't trigger a re-render.

The callback form of `watch` is still useful, of course, and should be kept.

---

Sidenote: Is this the best place to discuss RHF V8? Would you prefer I make a discussion post instead?

BrendanC23:
I noticed that the [watch() documentation](https://react-hook-form.com/docs/useform/watch) was updated to mark the callback overload as deprecated:

>2. Deprecated: consider use or migrate to [subscribe](https://react-hook-form.com/docs/useform/subscribe). Watching with callback fn watch(callback: (data, { name, type }) => void, defaultValues?: {[key:string]: unknown}): { unsubscribe: () => void }

Are there plans to remove this? My current usage of `watch` involves calling `setState()`. According to the [subscribe docs](https://react-hook-form.com/docs/useform/subscribe), this isn't supported:

>This function is intended for subscribing to changes only; dispatching state updates or triggering re-renders is not allowed. eg setValue or reset

BrendanC23:
The [setValue docs](https://react-hook-form.com/docs/useform/setvalue) mention the following:

>Important: use replace from useFieldArray instead, update entire field array with setValue will be removed in the next major version.

Is this still a planned change for version 8?

bluebill1049:
> The [setValue docs](https://react-hook-form.com/docs/useform/setvalue) mention the following:
> 
> > Important: use replace from useFieldArray instead, update entire field array with setValue will be removed in the next major version.
> 
> Is this still a planned change for version 8?

yes

bluebill1049:
> This version has a breaking change that renames the React key value returned from useFieldArray from `id` to `key`.
> 
> https://github.com/react-hook-form/react-hook-form/blob/7c403add0639273cfea2c1cc0a42834d9d1b4fa7/src/useFieldArray.ts#L99
> 
> As mentioned in [this discussion](https://github.com/orgs/react-hook-form/discussions/8935), many users have run into issues with `id` overwriting their data. I don't think switching the name to `key` will actually solve this. It just changes the name of the field that will be overwritten.
> 
> As this is version 8.0.0 and there will be breaking changes anyway, I wonder if there's a better solution that would be a more permanent fix (and would not require users to provide a separate `keyName` prop).
> 
> 1. Use a more unique default than `key` that is unlikely to overwrite the user's data (for example, `rhfKey`)
> 2. Add a `getValues()` method to the `field` value (see [this comment](https://github.com/orgs/react-hook-form/discussions/8935#discussioncomment-12838703).
> 3. Change the type of `field` to `{ key: string, fieldValue: FormData }`. This would allow separation of the key value and the user's form data. See [this comment](https://github.com/orgs/react-hook-form/discussions/8935#discussioncomment-7012063).
> 
> As mentioned in [this comment](https://github.com/orgs/react-hook-form/discussions/8935#discussioncomment-4002113):
> 
> > The prop fields returned from hook useFieldArray is not a part of form values, it's an controlled array to to work with dynamic fields via loop.
> > If you want to get form values which includes the dynamic field values belongs to this prop "fields", you should use method getValues from useForm or via onValid callback of method handleSubmit.
> 
> Option 2 would help solve this. The `field` value not being reactive is a common source of confusion that this would fix.

`id` has made most of the complain over the years, `key` is a reserved keywords for react props make more sense to migrate over.

bluebill1049:
> If V8 is going to include breaking changes, I wonder if it's worth it to remove or deprecate the `watch("name")` API. For performance reasons, `useWatch` is preferable (the linked discussion doesn't list any instances where this isn't true). This would help eliminate a common source of confusion.

Not sure at the moment, as `watch` has the best DX imo.

BrendanC23:
> id has made most of the complain over the years, key is a reserved keywords for react props make more sense to migrate over.

I agree that a user-supplied object is more likely to have an `id` value than a `key` value, but it's definitely possible for it to have the latter. If there's going to be a breaking change anyway, I think it's a good idea to redesign `useFieldArray` to prevent it from modifying user-supplied objects. Changing to `key` would probably reduce the frequency of the problem, but wouldn't eliminate it entirely.

eugene-mohc:
I have found an issue in the latest beta version where `formState` from `useFormContext` and `useFormState` inside of a nested component is memoized incorrectly by the compiler and doesn't trigger re-renders.

Minimal example to show the issue: https://codesandbox.io/p/devbox/q3kgmx?file=%2Fapp%2Fpage.tsx

Press the submit button. The form shouldn't submit because of an error, but the `Component` doesn't reflect that there's an error.

bluebill1049:
> I have found an issue in the latest beta version where `formState` from `useFormContext` and `useFormState` inside of a nested component is memoized incorrectly by the compiler and doesn't trigger re-renders.
> 
> Minimal example to show the issue: https://codesandbox.io/p/devbox/q3kgmx?file=%2Fapp%2Fpage.tsx
> 
> Press the submit button. The form shouldn't submit because of an error, but the `Component` doesn't reflect that there's an error.

`<FormProvider {...form} formState={formState}>` we will have to do this unfortunately.

ivan-palatov:
> `id` has made most of the complain over the years, `key` is a reserved keywords for react props make more sense to migrate over.

`key` is reseved in react, yes, but it isn't reserved in js or on the server. I think that just doing `{key: string; value: T}` instead of `T & {key: string}` is a more elegant and robust solution overall.

Also, I wonder if its the only reason why in the current implementation you coundn't use `useFieldArray` on an array of primitives? If so, it would be cool to get this functionality. We have multiple places in our apps, where we have to first make array of objects out of array of primitives to pass it to default values and than do the reverse on validation step. It is one of the pain points of using this library for us.

It would be cool to see this fixed in v8

ivan-palatov:
as for `watch()`, we have a relatively large project which uses `rhf` heavily and we've been able to migrate all our code to use `useWatch` in components bodies and `getValues` in callbacks and effects. Maybe its just our experience, but I cant imagine a case where it would be beneficial to use `watch()`. And I wounder if those potential cases could be easily and more idiomatically accurately be solved with `formData.subscribe()`?

lukasvice:
> > `id` has made most of the complain over the years, `key` is a reserved keywords for react props make more sense to migrate over.
> 
> `key` is reseved in react, yes, but it isn't reserved in js or on the server. I think that just doing `{key: string; value: T}` instead of `T & {key: string}` is a more elegant and robust solution overall.
> 
> Also, I wonder if its the only reason why in the current implementation you coundn't use `useFieldArray` on an array of primitives? If so, it would be cool to get this functionality. We have multiple places in our apps, where we have to first make array of objects out of array of primitives to pass it to default values and than do the reverse on validation step. It is one of the pain points of using this library for us.
> 
> It would be cool to see this fixed in v8

I fully agree. I also think it’s more intuitive to return the defined value rather than a modified object with an added key. This way, the value reference can be preserved as well (though that would also be possible when adding the key property).

BrendanC23:
Are there any more plans for an alternative to the `watch` callback function that is being removed in version 8? The new `subscribe` function is great for observing updates, but the docs say

>This function is intended for subscribing to changes only; dispatching state updates or triggering re-renders is not allowed. eg setValue or reset

So it isn't able to some use-cases that the `watch` callback supported. I've found it useful to centralize inter-field dependency management. For example, a form might have a `total` field that performs a complex calculation based numerous other fields. Instead of requiring each of the dependencies to have an `onChange` handler that calls  `setTotal`, the total can be synced from the `watch` callback (which would involve calling `setValue` from the callback function).

Also see https://github.com/react-hook-form/react-hook-form/issues/13208.

bluebill1049:
> Are there any more plans for an alternative to the `watch` callback function that is being removed in version 8? The new `subscribe` function is great for observing updates, but the docs say
> 
> > This function is intended for subscribing to changes only; dispatching state updates or triggering re-renders is not allowed. eg setValue or reset
> 
> So it isn't able to some use-cases that the `watch` callback supported. I've found it useful to centralize inter-field dependency management. For example, a form might have a `total` field that performs a complex calculation based numerous other fields. Instead of requiring each of the dependencies to have an `onChange` handler that calls `setTotal`, the total can be synced from the `watch` callback (which would involve calling `setValue` from the callback function).
> 
> Also see #13208.

`useWatch` already has `compute` is that what you refer to?

BrendanC23:
I'm not talking about `compute`, which is used for restricting when `useWatch` will trigger a re-render. Instead, I have a use-case like this: https://codesandbox.io/p/sandbox/rhf-form-watch-subscription-4yg9ls

There is a complex form with many interconnected fields. When one of the fields changes, the values of many others need to be recomputed. While this can be done using an `onChange` callback within every field, that gets unwieldy and it's easy to forget to add it.

In my real codebase, that would also require prop-drilling the various `onChange` callbacks through many layers of nested components. Often, changing one field will require another in a completely different sub-tree to be updated. I'm using RHF lenses to build generic reusable components, and some inputs don't have information about the wider form design (this would like like an `<Item />` component in the form above that only has access to the individual `item` and doesn't have access to the main `totals` field).

The `watch` callback API allows all of the re-computation to be standardized in one location. There's no need for individual `onChange` callbacks everywhere. All of the values get recomputed when an item changes or is appended without needing to wire each input independently.

bluebill1049:
> The watch callback API allows all of the re-computation to be standardized in one location. There's no need for individual onChange callbacks everywhere. All of the values get recomputed when an item changes or is appended without needing to wire each input independently.

There is `subscribe` api just for this now: https://react-hook-form.com/docs/useform/subscribe

BrendanC23:
The docs say that it isn't safe to call `setValue` from the `subscribe` callback.

> This function is intended for subscribing to changes only; dispatching state updates or triggering re-renders is not allowed. eg setValue or reset

bluebill1049:
> The docs say that it isn't safe to call `setValue` from the `subscribe` callback.
> 
> > This function is intended for subscribing to changes only; dispatching state updates or triggering re-renders is not allowed. eg setValue or reset

I will improve on this and eventually remove that note. feel free to send a feature request as well in case i forgot.

woosanghun:
Instead of changing default from **id** to **key**, can you make it **_key** so it is less likely to conflict with anyone's structure? both id and key are probably pretty commonly used field names.

bluebill1049:
> Instead of changing default from **id** to **key**, can you make it **_key** so it is less likely to conflict with anyone's structure? both id and key are probably pretty commonly used field names.

The key name only affects the internal identifier we use; it won’t affect the form values at all. The submitted form data and user-facing fields will remain unchanged, so using `_key` would only help avoid potential conflicts without changing the form structure or behavior.

audun-cobrief (approved):

xiongemi (approved):

xiongemi:
We just migrated a large production monorepo (~4,100 test files / 40k tests across 6 apps) from `7.66.1` to `8.0.0-beta.3`. The checklist items above (`keyName` removal, `id` → `key`, `watch` callback → `subscribe`) were quick — sharing two things we hit that aren't on the list, in case they're useful for the checklist / migration guide:

**1. `useForm`'s return value is no longer referentially stable.** v8 re-spreads `_formControl.current` inside `useMemo(..., [formState, control])` and re-binds `getFieldState` / `watch` / `getValues` / `register`, so the methods object *and* those four methods get a new identity on every form-state change. Any `useEffect` that lists one in its dependency array now re-runs on every keystroke — and loops forever if it also writes to the form. We found 18 effects in our codebase that became infinite `Maximum update depth exceeded` loops under beta.3 (simplest case: `useEffect(() => { form.reset(x) }, [form, x])`, which was fine in v7). Is the identity churn intentional? Either way it may deserve a ⚠️ entry, since v7's docs describe these as memoized and the dep-array pattern is very common.

**2. Built-in `required` now fires on controlled inputs whose text input stays empty.** `useController` now stores the real DOM node (`_proxyRef.current = elm`, vs. v7's `{focus, select, setCustomValidity, reportValidity}` proxy), which makes `validateField`'s `isEmpty` clause `isHTMLElement(ref) && ref.value === ''` reachable for Controller fields for the first time. Wiring `field.ref` to e.g. a multi-select / chips input whose text `value` is always `''` makes `required` fail even while form state holds a non-empty value. We reduced this to a plain `<input>` repro — the identical spec passes on `7.87.0` and fails on `8.0.0-beta.3`. Resolver-based forms are unaffected since they bypass `validateField`. Happy to file a separate issue with the repro if that's helpful.

bluebill1049:
> We just migrated a large production monorepo (~4,100 test files / 40k tests across 6 apps) from `7.66.1` to `8.0.0-beta.3`. The checklist items above (`keyName` removal, `id` → `key`, `watch` callback → `subscribe`) were quick — sharing two things we hit that aren't on the list, in case they're useful for the checklist / migration guide:
> 
> **1. `useForm`'s return value is no longer referentially stable.** v8 re-spreads `_formControl.current` inside `useMemo(..., [formState, control])` and re-binds `getFieldState` / `watch` / `getValues` / `register`, so the methods object _and_ those four methods get a new identity on every form-state change. Any `useEffect` that lists one in its dependency array now re-runs on every keystroke — and loops forever if it also writes to the form. We found 18 effects in our codebase that became infinite `Maximum update depth exceeded` loops under beta.3 (simplest case: `useEffect(() => { form.reset(x) }, [form, x])`, which was fine in v7). Is the identity churn intentional? Either way it may deserve a ⚠️ entry, since v7's docs describe these as memoized and the dep-array pattern is very common.
> 
> **2. Built-in `required` now fires on controlled inputs whose text input stays empty.** `useController` now stores the real DOM node (`_proxyRef.current = elm`, vs. v7's `{focus, select, setCustomValidity, reportValidity}` proxy), which makes `validateField`'s `isEmpty` clause `isHTMLElement(ref) && ref.value === ''` reachable for Controller fields for the first time. Wiring `field.ref` to e.g. a multi-select / chips input whose text `value` is always `''` makes `required` fail even while form state holds a non-empty value. We reduced this to a plain `<input>` repro — the identical spec passes on `7.87.0` and fails on `8.0.0-beta.3`. Resolver-based forms are unaffected since they bypass `validateField`. Happy to file a separate issue with the repro if that's helpful.

Great feedback, thanks for sharing this. Both points are really useful, especially the `useForm` identity change and the potential infinite effect loops. Definitely worth documenting in the migration guide.

And yes, a separate issue with the repro for the `required`/controlled input behavior would be really helpful.

greeeg:
@bluebill1049 Thanks for your answer! Do I get correctly that the identity change on every form update is intended in v8? What would be your recommended approach for `useEffect` leveraging one of the exposed functions?

xiongemi:
@bluebill1049 i created a repo here: https://github.com/xiongemi/rhf-v8-useform-identity-repro

bluebill1049:
> @bluebill1049 Thanks for your answer! Do I get correctly that the identity change on every form update is intended in v8? What would be your recommended approach for `useEffect` leveraging one of the exposed functions?

Yes, that's intended — but only for `watch`, `getValues`, `getFieldState`, and `register`.

On every form update, `useForm` rebinds these methods, so their references change. Everything else — `setValue`, `trigger`, `reset`, `handleSubmit`, etc. — remains referentially stable.

This is required for React Compiler compatibility. 

If you need reactive value updates, use `useWatch({ control, name })` rather than relying on `watch`.

bluebill1049:
> @bluebill1049 i created a repo here: https://github.com/xiongemi/rhf-v8-useform-identity-repro

Cool, thanks for the repo. Issue will be fixed.

PR #12372, opened 2024-10-24
Fix type of FieldErrors.root

### LINKED ISSUE #12371: issue: errors.root cannot accept GlobalError in type-level.
### Version Number

7.53.1

### Codesandbox/Expo snack

https://github.com/LumaKernel/react-hook-form-issue-1

### Steps to reproduce

1. Clone https://github.com/LumaKernel/react-hook-form-issue-1 (if you like online editor, you may use any, or GitHub Copespace)
2. npm i
3. See editor error. (Or run `npx tsc -p tsconfig.app.json --noEmit`)

Then got the error:

```
src/App.tsx:12:7 - error TS2322: Type '{ message: string; }' is not assignable to type 'Record<string, Partial<{ type: string | number; message: string; }>> & Partial<{ type: string | number; message: string; }>'.
  Property 'message' is incompatible with index signature.
    Type 'string' has no properties in common with type 'Partial<{ type: string | number; message: string; }>'.

12       errors.root = {
         ~~~~~~~~~~~


Found 1 error in src/App.tsx:12
```

The main code is following.

```ts
import { useForm, ResolverError } from "react-hook-form"


type Inputs = {
  example: string
  exampleRequired: string
}

function App() {
  const {} = useForm<Inputs>({
    resolver: () => {
      const errors: ResolverError<Inputs>['errors'] = {};
      errors.root = {
        message: 'This is a root error',
      };

      // Same Error
      // const errors: ResolverError<Inputs>['errors'] = {
      //   root: {
      //     message: 'This is a root error',
      //   },
      // };

      return {
        values: {},
        errors,
      };
    },
  });

  return <div>App</div>;
}

export default App;
```

### Expected behaviour

No type error.

_No response_

### Relevant log output

_No response_

### Code of Conduct

LumaKernel:
Resolves https://github.com/react-hook-form/react-hook-form/issues/12371

### DIFF
--- a/src/types/errors.ts
+++ b/src/types/errors.ts
@@ -50,7 +50,7 @@ export type FieldErrors<T extends FieldValues = FieldValues> = Partial<
     ? any
     : FieldErrorsImpl<DeepRequired<T>>
 > & {
-  root?: Record<string, GlobalError> & GlobalError;
+  root?: Record<string, GlobalError> | GlobalError;
 };
 
 export type InternalFieldErrors = Partial<
--- a/src/types/errors.ts
+++ b/src/types/errors.ts
@@ -51,11 +51,9 @@ export type FieldErrors<T extends FieldValues = FieldValues> = Partial<
     : FieldErrorsImpl<DeepRequired<T>>
 > & {
   root?: {
-    [P in string]?: P extends 'key'
-      ? string | number
-      : P extends 'message'
-        ? Message
-        : GlobalError
+    type?: string | number;
+    message?: Message;
+    [key: string]: string | number | Message | GlobalError | undefined;
   };
 };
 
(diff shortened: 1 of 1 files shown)


LumaKernel:
Quick explanation:
Just with simplified example. `Record<string, { foo: string }> & { foo: string }` is including `{ foo: { foo: string } } & { foo: string }`. And they are merged as `{ foo: { foo: string } & string }`. It concludes `string` is not assignable to top `.foo` property.

And another note. `(): SomeType => ({ ... })` and `() => {const v: SomeType = { ... }; return v; }` is behaving differently. This is, because of the TypeScript subtle trick. See below:

```ts
const v: { a: number, b?: number } = { a: 1, c: 3 }; // error
```

We can't write like above.But, potentially, that can be assigned.
If we rewrite the code as follows, the error would not happen.

```ts
const w = { a: 1, c: 3 };
const v: { a: number, b?: number } = w;
```

The original one, the similar trick could happen. That's why `resolver: () => ({ values: {}, errors: { root: { message: 'hi' } } })` would not report an error.

LumaKernel:
Retired. Type definition is potentially contradicts. It needs re-design of that. I'm happy if this is considered in next major version. Thanks.

PR #12387, opened 2024-10-31
allow field names that single numeric strings to process as strings, not array indexes

f0rk:
when using field names like "0", "1", etc. the output is converted to an array, even when mixed names (e.g. "A" and "1") are used.


f0rk:
I'll flesh out some tests and so on in the near future if there's any interest in this idea

PR #12389, opened 2024-11-02
Fix: hasPromiseValidation return true or false appropriately.

developer-bandi:
In the `hasPromiseValidation` function, `fieldReference` or `fieldReference.validate` is false and `fieldReference.validate` cannot be a function or an object, so the function always returns false. 

Therefore, i modified the first condition to return true or false appropriately.

Also, I wrote an additional unit test for that function because it doesn't seem to exist.

+ Due to the impact of this bug, the `isValidating` value appears to not be updated properly even if a promise function is used in validation option. 

```typescript
const isPromiseFunction = field._f && hasPromiseValidation((field as Field)._f);

if (isPromiseFunction && _proxyFormState.validatingFields) {
     _updateIsValidating([name], true);
}
```

### DIFF
--- a/src/__tests__/logic/hasPromiseValidation.test.ts
+++ b/src/__tests__/logic/hasPromiseValidation.test.ts
@@ -1,0 +1,69 @@
+import hasPromiseValidation from '../../logic/hasPromiseValidation';
+
+const commonParam = {
+  mount: true,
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
Thank you.

PR #12418, opened 2024-11-15
🐞 fix #12415 issue with flatten object with null value

### LINKED ISSUE #12415: issue: isObjectType(obj[key]) returns true on null, tries to recurse on null values
### Version Number

7.53.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/condescending-mcclintock-ndy5wx

### Steps to reproduce

The flatten code from this commit: https://github.com/react-hook-form/react-hook-form/commit/087433d0045af612a2b45fe52b85bf66f4a369d6
crashes if you have a null value in defaultValues, or presumably in form values at the time of submission.

It throws because typeof null === "object" is true. To reproduce, you can submit on the codesandbox. If you remove the null defaultValue & reset the form it works fine.  I feel like explicitly checking if it's isObjectType && != null would be good here.

### Expected behaviour

It should not attempt to flatten nulls.

_No response_

### Relevant log output

_No response_

### Code of Conduct

bluebill1049:
https://codesandbox.io/p/sandbox/frosty-poincare-8t87vn?file=%2Fsrc%2Fsrc%2Futils%2Fflatten.ts

### DIFF
--- a/src/__tests__/utils/__snapshots__/flatten.test.ts.snap
+++ b/src/__tests__/utils/__snapshots__/flatten.test.ts.snap
@@ -4,7 +4,10 @@ exports[`flatten should flatten form values into flat form data 1`] = `
 {
   "array.0.test": "1",
   "array.0.test2": "2",
+  "array.0.test3": null,
   "hey": "test",
   "test.nested.test": "bill",
+  "test.nested.test3": null,
+  "test1": null,
 }
 `;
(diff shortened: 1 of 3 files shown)


vixalien:
Would be nice to have a release for this, since without it you can't submit forms with `null` as one of the values.

MiccWan:
+1
Need a new release to fix this bug.

PR #12424, opened 2024-11-17
fix: useForm should return a new object on formState changes

### LINKED ISSUE #12423: issue: useForm shouldn't return the same object - react-complier/memo issue
### Version Number

7.53.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/thirsty-keller-zdcy4q

### Steps to reproduce

1. Open https://codesandbox.io/p/sandbox/thirsty-keller-zdcy4q
2. Enter something to age  like 999
3. See that no validation error is shown

Or see 
https://next-15-form-git-main-miklosfazekas.vercel.app

Which has 3 forms:
1. Default ShadCN form - [src](https://github.com/mfazekas/next-15-form/blob/main/app/profile-form.tsx)
2. Default ShadCN form with react compiler disabled - [src](https://github.com/mfazekas/next-15-form/blob/main/app/profile-form-no-rc.tsx)
3. Default ShadCN form and forcing new object form form  - [src](https://github.com/mfazekas/next-15-form/blob/main/app/profile-form-workaround.tsx)

image

### Expected behaviour

Validation error should be shown

Chrome, Safari

### Relevant log output

_No response_

### Code of Conduct

mfazekas:
Fixes: #12423

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -154,7 +154,13 @@ export function useForm<
       });
   }, [props.shouldUnregister, control]);
 
-  _formControl.current.formState = getProxyFormState(formState, control);
-
-  return _formControl.current;
+  const formControlCurrent = _formControl.current;
+  return React.useMemo(
+    () =>
+      ({
+        ...formControlCurrent,
+        formState: getProxyFormState(formState, control),
+      }) as UseFormReturn<TFieldValues, TContext, TTransformedValues>,
+    [formState, control, formControlCurrent],
+  );
 }
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -154,13 +154,12 @@ export function useForm<
       });
   }, [props.shouldUnregister, control]);
 
-  const formControlCurrent = _formControl.current;
   return React.useMemo(
     () =>
       ({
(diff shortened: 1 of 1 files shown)


bluebill1049 (approved):

Ganbin:
Any ETA on the next release that will contain the fix?

kelkes:
How is that a fix? It's a breaking change that crashed almost all of my applications :)

cahilfoley:
Also landed here because this update broke every form in our application (forms being used in effects etc), this probably should have been a breaking change.

controversial:
This was a large breaking change for our application as well.

pablo-psi:
If anyone was doing form validation like this before:

```tsx
const formMethods = useForm()

formMethods.trigger().then(() => {
    if(formMethods.formState.isValid) {
        // ...
    }
})
```

After upgrading react-hook-form, better do it like this (fixed newly introduced validation bugs for me):

```tsx
const formMethods = useForm()

formMethods.trigger().then((isValid) => {
    if(isValid) {
        // ...
    }
})
```

Issue #8048, opened 2022-03-18
Watch & UseWatch Event Type/ Custom Event Types


Adrian-SolM:
**Is your feature request related to a problem? Please describe.**

Right now the only way to get the event type on input change is using the useffect watch callback method. ( I haven't found any other method yet at least. ) 

The frustration is that in scenarios like having a New Price Input and a Discount % input where updating one value should update the other becomes more difficult, as a user typing in a value triggers an update, as well as a setValue, however there is no way to tell the difference between the update type unless you are using the callback method, and even then type is either Change or Undefined. 

**Describe the solution you'd like**

The solution I believe would be useful would be to add an option flag to watch and useWatch where you can get the change event type in the data returned. 

For Example something like:

```
const { Value: FirstName, Type: FirstNameChangeEvent } = useWatch({
    control,
    name: "firstName",
    type: true
  });
```

With this, a change I also believe would coinside with this is custom Events as well, utilizing setValue.

So some form of registering new events and then being able to utilize those new events when setting a value

For example:
  
  New Event is: CustomEvent.NewPriceChange
  
  then when setting a value you would
  
  ```
  setValue(Name, Value, EventType)
  
  setValue("FirstName", "TestName", CustomEvent.NewPriceChange); 
```

**Describe alternatives you've considered**

One method I figure that can be used is setting the value to an object manually where you have to maintain the value and type.

For example 

```
const defaultValue = {
FirstName: { Value: "TestName", Type: ChangeEventType }
}

and when you set value

setValue("FirstName", { Value: "NewName", Type: CustomEventType } )
```

However I feel this adds more complexities than needed and require more workload to manage and maintain. 

**Additional context**

I began looking through the source, however I am not familiar with TypeScript, so it would be a little difficult for me to guage the workload required for such updates. Please do let me know your input/feedback on the idea and its feasibility. Also let me know if I should elaborate more. Hopefully I filled everything out with enough detail to get the idea across.

PR #12468, opened 2024-12-11
feat: add `type` option in `useWatch`

pushys:
Unlike in `watch` it's impossible to get an event type in `useWatch`. This small addition allows to subscribe to changes of a specific type.

Potentially can help with #8542 or #8048.

### DIFF
--- a/src/__tests__/useWatch.test.tsx
+++ b/src/__tests__/useWatch.test.tsx
@@ -340,6 +340,46 @@ describe('useWatch', () => {
     expect(screen.getByText('345')).toBeVisible();
   });
 
+  it('should subscribe to specific event type', () => {
+    const App = () => {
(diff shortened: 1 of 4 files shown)


src/__tests__/useWatch.test.tsx
```diff
@@ -340,6 +340,46 @@ describe('useWatch', () => {
...
+      target: {
+        value: '1234',
+      },
+    });
+
+    expect(screen.getByText('1234')).toBeVisible();
+
+    fireEvent.click(screen.getByRole('button'));
+
+    expect(screen.getByText('1234')).toBeVisible();
```
alexandermirzoyan, line 380: why this should not be '12345'?

alexandermirzoyan (approved):

PR #12475, opened 2024-12-12
Revert "fix: useForm should return a new object on formState changes"

bluebill1049:
Reverts react-hook-form/react-hook-form#12424


SimonFuetCG:
Hi!

I see you've reverted this change, what's the long term plan for this ? keep it the way it is ? bring it back in a major version as a breaking change ? ...

Thanks

Issue #12463, opened 2024-12-09
issue: useForm response: memoization behavior change in v7.54.0


JPStrydom:
### Version Number

7.54.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/clever-cache-76f5tj

### Steps to reproduce

The steps to reproduce the issue can be found in the provided code sandbox, but the steps are roughly the following:

1. Install react-hook-form version `7.54.0`
2. Create a basic form component as such:
    ```
    export const Form = ({ data, children }) => {
      const formMethods = useForm();

      useEffect(() => {
        if (data) {
          console.log('in Form useEffect');
          formMethods.reset(data);
        }
      }, [data, formMethods]);

      return <FormProvider {...formMethods}>{children}</FormProvider>;
    };
    ```
3. Use the component as such:
    ```
    const formData = { test: 'test' }
    export default () => (
      <div className="App">
        <Form data={formData} />
      </div>
    );
    ```
4. Notice how the `'in Form useEffect'` log get printed over and over until the app crashes as the `formValues` are considered "new" on every re-render.

### Expected behaviour

As was the case in react-hook-from version 7.53.2, the value returned from `useForm` should be memoised. 

In the provided code-sandbox, when reverting to react-hook-from version 7.53.2 (remember to exclude the range specifiers when downgrading), the issue disappears.

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

JPStrydom:
My current workaround for this is to either de-structure the values:
```
const { reset, ...formMethods }= useForm();

useEffect(() => {
  if (data) {
    reset(data);
  }
}, [data, reset]); 

// The de-structured reset seems to be memoised, but this requires manually passing reset={reset} to the FormProvider

return <FormProvider reset={reset} {...formMethods}>{children}</FormProvider>;
```

or to manually memoise them:
```
const formMethods = useForm();

const reset = useMemo(() => formMethods.reset, [formMethods.reset]);

useEffect(() => {
  if (data) {
    reset(data);
  }
}, [data, reset]); 

// reset can also be manually memoised, which at least doesn't require a change to the FormProvider

return <FormProvider {...formMethods}>{children}</FormProvider>;
```

JPStrydom:
> Most likely due to this change: #12424
> 
> To be fair, the docs for `useForm` always destruct the form methods, e.g.
> 
> ```ts
[...]

PR #12476, opened 2024-12-13
🧪 test(useWatch): destructure setValue from useForm

marcalexiei:
Related to #12463.

I searched for all `useEffect` in the codebase and I was able to find only one that was receiving the whole `useForm` result in dependencies array.


bluebill1049:
thank you very much@

bluebill1049 (approved):

PR #12491, opened 2024-12-19
⏰  track disabled fields and only omit data on submit

bluebill1049:
ref: https://codesandbox.io/p/sandbox/gs5hjr?file=%2Fsrc%2Fsrc%2Flogic%2FcreateFormControl.ts%3A1486%2C19


IanVS:
The sandbox in the ref isn't found (is it private?).  Can you maybe explain a little more about the purpose of this change?

PR #12495, opened 2024-12-20
feat: add FieldArray component

tran-simon:
Add `FieldArray` component, a component based on `useFieldArray` hook to work with controlled component.

`FieldArray` complements `useFieldArray` in the same way that `Controller` complements `useController`


lamy:
Yes i like

tran-simon:
@bluebill1049

bluebill1049:
thanks for the PR, i think you can consider keep this component at your app level.

PR #12503, opened 2024-12-28
🧮 feat: compute prop for useWatch subscription

### LINKED ISSUE #12501: useWatch computed value performance
Relevant when dealing with big forms and conditional fields

## **The scenario:**
Imagine you have a form with an input that fires many change events, like a simple text input (call it `userName`). And then you have another input that you conditionally render or change depending on the value of your first text input. Imagine the condition here is, when `userName` is of length > 10, also render `nickname` field

The way of solving this would be to use `useWatch` to watch the value of `useName`. Calculate if `usernName.length` > 10, then if true render `nickname`:

```tsx
const MyFormFields = () => {
  const { register } = useFormContext();

  const userNameValue = useWatch({ name: 'userName' });

  const displayNicknameField = userNameValue.length > 10;

  return (
    <>
      <input {...register('userName')} />

      {displayNicknameField && <input {...register('nickname')} />}
    </>
  );
};
```

## **The problem:**
Every keystroke on `userName` will cause a re-render. But in reality in most cases we wouldn't need to, we really only care of `displayNicknameField` changing

## **The proposal:**
Would it be possible to have `useWatch` accept a callback to compute a value (quite similar to `useMemo`) but only cause a re-render when said computed value changes?

Combining this
```ts
const userNameValue = useWatch({ name: 'userName' });
const displayNicknameField = userNameValue.length > 10;
```

Into this
```ts
const displayNicknameField = useWatch({ name: 'userName', compute: currentValue => currentValue.length > 10 });
```

For the first 9 keystrokes of `userName`, `displayNicknameField` would just be false, resulting in no re-renders. Only at the 10th when it actually changes to true, it would cause its first re-render

Same low-re-render behavior can probably be achieved by combining `watch` with a callback subscription and a `setState`, but it would be great if this was built in within watch into a simply one-liner

bluebill1049:
close #12501

- subscribe to the entire form but only return updated value with certain condition

```tsx
type FormValue = {
  test: string;
}

const watchedValue = useWatch({
  control: methods.control,
  compute: (data: FormValue) => {
    if (data.test?.length) {
      return data.test;
    }

    return '';
  },
});
```

- subscribe to a specific form value state

```tsx
type FormValue = {
  test: string;
}

const watchedValue = useWatch({
  control: methods.control,
  name: 'test',
  compute: (data: string) => {
      return data.length ? data : '';
  },
});
```


bluebill1049:
> Do you think there's value in including a name prop? This would ensure that the compute() function isn't running on every change to the form value. For example, from https://github.com/react-hook-form/react-hook-form/issues/12501

yea, can looking into that.

eslint.config.mjs
```diff
@@ -50,6 +50,7 @@ export default tseslint.config(
       curly: 'error',
       'no-extra-boolean-cast': 'error',
       'cypress/unsafe-to-chain-command': 'off',
+      '@typescript-eslint/no-unnecessary-type-constraint': 'warn',
```
kotarella1110, line 53: This line is no longer necessary since https://github.com/react-hook-form/react-hook-form/pull/12771 has been merged!

bluebill1049: oh nice, i will remove it.

bluebill1049: Going to get an rc version out this weekend. It should be a good feature for all.

PR #12510, opened 2024-12-30
Change Path type check in PathValue type to single check at start

WouterLVV:
I've noticed that in a project I'm working on, the intellij typescript service was struggling with react-hook-form typing, when using large types.

I started investigating and found that the `Path<>` calls in the `PathValue` are very intense for large input types, and they seemed redundant. As such, in this PR, I propose that the type only does a single `extends Path<> | ArrayPath<>` at the start, and remove the checks in the recursive part of the type.

I believe this is safe since the type would end up at `never` eventually even if the path was somehow invalid. However, since there is now still a check at the start, nothing should change outwardly.

The only internal issue I encountered was in `resetField`, where the type cast of the `defaultValue` was no longer correct. I believe this has to do with how typescript inferences certain types and I initially made a solution using the new `NoInfer` type, but this would raise the minimum typescript to use hook form to 5.4. Instead I chose the solution you see below.

This change has made a enormous difference for the responsiveness of Intellij and has also cut the type-checking of my project by about 10 seconds (from 50s to 40s, even while hook form is only used in specific parts of the application)

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -454,10 +454,10 @@ export type Path<T> = T extends any ? PathInternal<T> : never;
 // @public
 export type PathString = string;
 
-// Warning: (ae-forgotten-export) The symbol "ArrayKey" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "PathValueImpl" needs to be exported by the entry point index.d.ts
 //
 // @public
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):
Thanks, LGTM.

maxkostow:
This is an incredible find @WouterLVV; absolutely huge improvement. I'm `patch-package`ing this immediately 🐎

PR #12549, opened 2025-01-22
added try catch

Par222:


bluebill1049:
already have check in place

PR #12566, opened 2025-01-28
fix: acknowledge root errors in onChange

chrisgarber:
When changing a field, particularly arrays, new errors can be created at root or nested root level. These should count as a net new error and should cause an update to the errors object. Currently, if you `compact` an array using something like a yup schema, root errors aren't added to the errors object, even though their children are dirty/changed.

[Here is a CodeSandbox](https://codesandbox.io/p/sandbox/recursing-edison-hmnksj) that demonstrates the behavior I'm trying to fix. If you add whitespace to any of the fields (they're trimmed so whitespace doesn't count), you'll note that the errors are empty. This change would include the root error in the errors object.

This is somewhat of a regression, this used to work before the resolvers made a change to put this errors under the `root` instead of straight on the error object for arrays.

### DIFF
--- a/src/__tests__/logic/schemaErrorLookup.test.ts
+++ b/src/__tests__/logic/schemaErrorLookup.test.ts
@@ -232,4 +232,51 @@ describe('errorsLookup', () => {
       name: 'test.testXYZ',
     });
   });
+
+  it('should find the nearest root error', () => {
+    const errors = {
+      test: {
+        0: {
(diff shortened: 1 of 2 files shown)


chrisgarber:
Dang, this is sad, I was out on vacation for a few days and now it's closed and I need to open a whole new PR? It's stale because no one looked at it to give it approvals 😞

PR #12568, opened 2025-01-29
feat: introduce isReady state for subscription

bluebill1049:
## Ready state

wait for subscription ready before start setting value

```tsx
const { formState: { isReady }, setValue } = useForm()

useEffect(() => {
  if (!isReady) return {}

 setValue('data', 'isReady')
}, [isReady])
```


src/useForm.ts
```diff
@@ -75,15 +76,17 @@ export function useForm<
...
-  );
+  React.useEffect(() => {
+    control._subscribe({
+      formState: control._proxyFormState,
+      callback: () => updateFormState({ ...control._formState }),
+      reRenderRoot: true,
+    });
+    updateFormState((data) => ({
+      ...data,
+      ready: true,
```
srkacha, line 87: should this be "isReady" and not "ready"?

AleksaMCode: Probably a typo. bluebill1049

bluebill1049: yea i will come back to this, thanks guys

PR #12612, opened 2025-02-15
🐞 fix #12580 valid state update with onBlur mode

### LINKED ISSUE #12580: issue: onBlur validation mode doesn't work in version 7.50.1 and in the latest version
### Version Number

7.50.1 and the latest one

### Codesandbox/Expo snack

https://snack.expo.dev/@lukasrzavsky/react-hook-form-onblur-issue

### Steps to reproduce

1. Enter some value to first field
2. Enter some value to second field but do not lose focus
3. You can see that the submit button is enabled, but it should be enabled only in case you lose focus of the second field

### Expected behaviour

The form shouldn't change `isValid` value on every change. It should change the value only in case all fields are valid and user lose focus of the last input with validation.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

bluebill1049:


ivankrzak:
@bluebill1049 
Hellou, 
When will be this fix available in package ?

PR #12622, opened 2025-02-20
fix: add condition to prevent infinite render

### LINKED ISSUE #12599: issue: Infinite re-render when init with `errors`
### Version Number

7.53.0

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/interesting-spence-kgg47d

### Steps to reproduce

Careful the link in codesandbox can crash your tab.
You have to open the dev console to see the infinite re-render

I think what I'm doing is "wrong" in the sense that it does not make sense (and I have now fixed my usage) but I don't understand what's happening and I think anyway there should be some protection against it.

It seems that
- if you initialize your `useForm` with `errors: {}`
- monitor the errors state with `{formState: {errors}} = useFormContext()`
You get an infinite re-render

```typescript
export const FormAmountShowcase = () => {
    const formProps = useForm<{ foo: string }>({
        defaultValues: {foo: 'bar'},
        errors: {},
    })

    return (
        <FormProvider {...formProps}>
            <FormComp/>
        </FormProvider>
    )
}

const FormComp = () => {
    console.log('rendering')
    const {formState: {errors}} = useFormContext<{ foo: string }>()

    return <>Foo</>
}
```

### Expected behaviour

no infinite re-render.
Either the input is accepted, or it should fail when this value is used.

Chrome

### Relevant log output

```shell

```

### Code of Conduct

candymask0712:
to fix #12599 

**as-is:** if I initialize useForm with `errors: {}`, I get an infinite re-render.
**to-be:** I add condition to prevent infinite render

Thank you!

### DIFF
--- a/app/pnpm-lock.yaml
+++ b/app/pnpm-lock.yaml
@@ -205,139 +205,139 @@ packages:
     resolution: {integrity: sha512-snKqtPW01tN0ui7yu9rGv69aJXr/a/Ywvl11sUjNtEcRc+ng/mQriFL0wLXMef74iHa/EkftbDzU9F8iFbH+zg==}
 
   '@esbuild/aix-ppc64@0.21.5':
-    resolution: {integrity: sha512-1SDgH6ZSPTlggy1yI6+Dbkiz8xzpHJEVAlF/AM1tHPLsf5STom9rwtjE4hKAF20FfXXNTFqEYXyJNWh1GiZedQ==}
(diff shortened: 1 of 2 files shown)


app/pnpm-lock.yaml
bluebill1049, line 1: can we revert this change?

bluebill1049 (changes requested):
Thanks for the PR 🙏

src/useForm.ts
```diff
@@ -112,7 +112,7 @@ export function useForm<
   }, [props.values, control]);
 
   React.useEffect(() => {
-    if (props.errors) {
+    if (props.errors && Object.keys(props.errors).length > 0) {
```
bluebill1049, line 115: there is `isEmptyObject` function

app/pnpm-lock.yaml
candymask0712: Sure. I also feel strange because changes occurred after I completed the contribution procedure.

src/useForm.ts
candymask0712: Oh, thanks. now I found `isEmptyObject`

bluebill1049 (approved):

PR #12642, opened 2025-02-28
🐞 fix #12580 setError in useEffect does not work when used inside the FormProvider context

### LINKED ISSUE #12580: issue: onBlur validation mode doesn't work in version 7.50.1 and in the latest version

### LINKED ISSUE #12632: issue: setError in useEffect does not work when used inside the FormProvider context
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/rhf-context-seterror-gnry6j

### Steps to reproduce

1. Go to the sandbox
2. Check out the first input on the preview panel
3. It must display an error below the first field, but it's not

### Expected behaviour

The `formState.errors` should contain the error that we set in the `useEffect`.

The problem only occurs if we use `useFormContext`.

If we wrap the `setError` in a `requestAnimationFrame` (or `setTimeout`), it will work as expected.

Chrome

### Relevant log output

```shell

```

### Code of Conduct

candymask0712:
to resolve #12632 

Hello. I have created this PR to fix the issue related to setError. I learned a lot by following the subscribe flow to identify the cause. 😁 Detailed information is provided below.

### Issue
setError in useEffect does not work when used inside the FormProvider context

### Cause
**The `observer` is registered too late, so when `setError` attempts to call the observers in the list, the list is empty.** As a result, if you delay the timing using `requestAnimationFrame` or `setTimeout`, it works correctly. however, if we do not use such special techniques, the error is not registered.

### Solution
I considered two solutions:

1. Register the `observer` faster than the current implementation.
2. Delay the execution timing of `setError`.

From a user experience perspective, I believe that `option 1` is more advantageous, so I applied the approach of using `useLayoutEffect` instead of `useEffect` to register faster than before.

### Changes
- Changed the execution timing of `control._subscribe` by using `useLayoutEffect` instead of useEffect.
- Modified the test cases in `useFieldArray.test.ts` accordingly.
- Added a new test case that reproduces this issue.

Thank you.

### DIFF
--- a/src/__tests__/useFieldArray.test.tsx
+++ b/src/__tests__/useFieldArray.test.tsx
@@ -2609,9 +2609,9 @@ describe('useFieldArray', () => {
           names: [{ name: 'will' }, { name: 'Mike' }],
         },
       });
-      const { setValue } = methods;
+      const { setValue, watch } = methods;
 
-      result.push(methods.watch());
+      result.push(watch());
 
       return (
         <form>
@@ -2657,6 +2657,16 @@ describe('useFieldArray', () => {
 
     // Let's check all values of renders with implicitly the number of render (for each value)
     expect(result).toEqual([
+      {
+        names: [
+          {
+            name: 'will',
+          },
+          {
+            name: 'Mike',
+          },
+        ],
+      },
       {
         names: [
           {
@@ -2753,6 +2763,16 @@ describe('useFieldArray', () => {
 
     // Let's check all values of renders with implicitly the number of render (for each value)
     expect(watchedValues).toEqual([
+      {
+        test: [
+          {
+            value: 'test',
+          },
+          {
+            value: 'test1',
+          },
+        ],
+      },
       {
         test: [
           {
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):
LGTM. That's a good idea to shift from `useEffect` to `useLayoutEffect`.

bluebill1049:
btw, you may be interested in this API: https://react-hook-form.com/docs/createFormControl

candymask0712:
@darekrossman 
Thanks for comment. I upload new PR for this #12738

PR #12645, opened 2025-02-28
Allow use of handleSubmit with native events

joshkel:
While looking into using a third-party library to invoke react-hook-form's `handleSubmit`, I noticed that `handleSubmit`'s types require a React SyntheticEvent (and the third-party library uses native HTML events). This doesn't appear to be needed by the actual implementation - the implementation appears to work with either a native HTML Event object or a React SyntheticEvent.

Broadening the types seems like a straightforward change. However, if this is a bad idea or isn't the direction the project wants to go, feel free to close this PR.


kotarella1110 (approved):
LGTM 👍  
By allowing both native events and `SyntheticEvent`, this change aligns the type definitions with the actual behavior and helps prevent type errors. This also improves cases like:  

```ts
document.addEventListener("submit", handleSubmit(onSubmit));
```  

where type errors previously occurred. Thanks for your work!

bluebill1049:
Awesome thanks @kotarella1110 for the review!

Issue #12634, opened 2025-02-26
issue: wrapping `form` in `useMemo` stops `field.onChange` from working


jtomaszewski:
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/festive-wind-7zs3h7

### Steps to reproduce

**The bug can't be reproduced on a development build.** You need to make a production build to see the bug.

1. Download code from https://codesandbox.io/p/devbox/festive-wind-7zs3h7
2. Run `npm install && npm run build && serve -s build`
3. Try changing fields in the first form (that uses `React.useMemo` under the hood). Notice that first input doesn't react to any keystrokes, and changing value of the second input doesn't update form values neither.

The second form (that doesn't use `React.useMemo`) works correctly.

https://github.com/user-attachments/assets/238c24d4-827c-475d-84d8-a0c167e184b0

### Expected behaviour

Wrapping `form` or `control` in `useMemo` shouldn't prevent the form from responding to `control.field.onChange` calls.

So that one can pass the form and control objects downwards through react context, and benefit from `useMemo` and `memo` React functions.

Right now it's one of the other. If I use react-hook-form, I can't wrap some of my components with `memo`.

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jtomaszewski:
I forked the react-hook-form repo and added a test scenario that fails: https://github.com/react-hook-form/react-hook-form/compare/master...fullstackhouse:react-hook-form:react-hook-form/react-hook-form%2312634 .

PR #12652, opened 2025-03-03
fix(createFormControl): don't unregister fields on reset

jtomaszewski:
After https://github.com/react-hook-form/react-hook-form/pull/12635 , unfortunately in my app an issue similar to https://github.com/react-hook-form/react-hook-form/issues/12634 is still existing.

After further debugging, it appeared that [this line](https://github.com/react-hook-form/react-hook-form/blob/master/src/logic/createFormControl.ts#L1343) is the cause. In some scenarios, when form values are reset, fields seem to behave as if they were unregistered, even if they component haven't been dismounted.

And assuming that the child components (which call the `useController` hook) don't get rerendered after this (which can happen, if they are memoized using react's `memo` or `useMemo` functions), then the controls won't get reregistered.

Causing the existing controls to still appear in the DOM, but e.g. their `onChange` will do nothing when being called.

This PR fixes this - instead of completely clearing the internal `_fields` object, it just resets the values of currently registered (mounted) fields.

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -56,6 +56,45 @@ describe('useController', () => {
     expect(input.value).toBe('abc');
   });
 
+  it("setting values doesn't cause fields to be unregistered", async () => {
+    function App() {
+      const [values, setValues] = useState<{ login: string } | undefined>();
+
+      const form = useForm({
+        values,
+      });
+
+      useEffect(() => {
+        setTimeout(() => {
+          setValues({ login: 'john' });
+        }, 100);
+      }, []);
+
+      return useMemo(
+        () => values?.login && <LoginField form={form} />,
+        [values, form],
+      );
+    }
+
+    function LoginField({ form }: { form: UseFormReturn<{ login: string }> }) {
+      const ctrl = useController({
+        name: 'login',
+        control: form.control,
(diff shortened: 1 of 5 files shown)


src/__tests__/useController.test.tsx
```diff
@@ -56,6 +56,45 @@ describe('useController', () => {
...
+      });
+
+      useEffect(() => {
+        setTimeout(() => {
+          setValues({ login: 'john' });
+        }, 100);
+      }, []);
+
+      return useMemo(
+        () => values?.login && <LoginField form={form} />,
```
jtomaszewski, line 74: note: this lazy rendering of the control, the `defaultValue` prop in the `LoginField`, and the `useMemo` wrapping are all required here for the issue to appear (in the old code).

bluebill1049 (approved):

bluebill1049:
could you help resolve the conflicts?

jtomaszewski:
> could you help resolve the conflicts?

Sure.

However I see that you've had reverted https://github.com/react-hook-form/react-hook-form/issues/12634 , including tests added there that were valid.

Tus I will recreate a PR that adds these tests, and tests from this PR, and a test for issue found in #12665 (as you haven't added a test for that scenario), and a solution that will make RHF work for all of these scenarios.

PR #12638, opened 2025-02-26
feat: infer resolver output types

jorisre:
This PR adds automatic output type inference for resolver functions. It will unlock [corresponding changes in the resolvers repository.](https://github.com/react-hook-form/resolvers/pull/753)

For testing, I've included a temporary `mockZodResolver` until the updated resolvers are released.

PR #12664, opened 2025-03-06
Fix TTransformedValues inference for #12638

controversial:
cc @jorisre 

fixes types as part of #12638 

## Why

- Using ExtractTransformedValues<> from `(typeof props)['resolver']` does not work because within the implementation signature, the value in `typeof props` does not capture any specific type information that is not a part of the `useForm` function’s generic type parameters. 
- Since `useForm` is only generic over TFieldValues/TContext/TTransformedValues, there is no type information to be gained from `(typeof props)['resolver']` because the compiler will *never* see the type `(typeof props)['resolver']` as anything more specific than `Resolver<TFieldValues, TContext, TTransformedValues>`.
- Therefore, we should just use TTransformedValues
    - The compiler already checks that the type of `resolver` matches TTransformedValues
    - moreover, the compiler can already *infer* the type of TTransformedValues based on the type of `props.resolver`, without using the broken ExtractTransformedValues type helper

## Effect

This PR fixes the broken types in the `handleSubmit` tests in https://github.com/react-hook-form/resolvers/pull/753

Before, `form` is erroneously inferred with a TTransformedValues parameter of `any`, and the test fails
Screenshot 2025-03-06 at 1 21 19 PM

After, `form` infers the correct type for TTransformedValues and there is no type error; the test passes
Screenshot 2025-03-06 at 1 23 15 PM

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -654,10 +654,8 @@ export type UseFieldArraySwap = (indexA: number, indexB: number) => void;
 // @public
 export type UseFieldArrayUpdate<TFieldValues extends FieldValues, TFieldArrayName extends FieldArrayPath<TFieldValues> = FieldArrayPath<TFieldValues>> = (index: number, value: FieldArray<TFieldValues, TFieldArrayName>) => void;
 
-// Warning: (ae-forgotten-export) The symbol "ExtractTransformedValues" needs to be exported by the entry point index.d.ts
-//
 // @public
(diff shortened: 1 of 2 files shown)


bluebill1049:
aweomse thanks! could you update the API extractor?

bluebill1049 (approved):

controversial:
Now updated; sorry for the delay! @bluebill1049

jorisre:
@controversial Thanks for your assistance. Could you please take a look at PR #12638 when you get a chance?

PR #12677, opened 2025-03-11
Fix #12655: Unmounted Clean Nested Field Incorrectly Update On Reset with keepDirtyValues

### LINKED ISSUE #12655: issue: nested changes to the `defaultValues` using `reset` are dropped in combination with `keepDirtyValues`
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/react-hook-form-reset-dynamic-array-z5xg3p

### Steps to reproduce

1. Make a change to the text input to make the form dirty
2. click on the reset button
3. optional: click on the updateState button to inspect the form state

The Form State is the following:
```ts
type FormState = {
  nested: { obj: { value1: string; value2?: string } };
  topLevel?: string;
};
```
If the form is dirty and the default value of `nested.obj.value2` is adjusted within a `reset()` call, the value won't be updated correctly. The problem seems to be lying here:
https://github.com/react-hook-form/react-hook-form/blob/bfcd68f01f7104e4c56caebc5ea26cc8a0c3c29b/src/logic/createFormControl.ts#L1313
Just adding the keys of the dirty fields is not sufficient, imho. Instead, the names of the dirty properties should be extracted and added to the set rather than the whole object key.

### Expected behaviour

The issue seems to have been introduced with #12211. I would expect, that no data is lost when using `keepDirtyValues` in combination with `reset`, which is currently not the case for nested values.

Chrome

### Relevant log output

```shell

```

### Code of Conduct

DavidIMk:
Fix for: #12655 

**Issue**
For a  form that have nested field, if one of the nested fields attribute is dirty when resetting, other attribute in that fields are not resetting into the intended values.

**Cause**
In `_reset` function, there are a logic to list all of the fields to be checked whether they are dirty or not, and if true it will trigger value change. However the logic using `Object.keys` only target  outermost attribute/field and not nested fields.

**Solution**
Change the `Object.keys` implementation in that line into an utils function to collect all the leaf attribute/field

Thank you and please CMIIW

### DIFF
--- a/src/__tests__/useForm/reset.test.tsx
+++ b/src/__tests__/useForm/reset.test.tsx
@@ -1653,6 +1653,99 @@ describe('reset', () => {
     });
   });
 
+  it('should reset unmounted non-dirty value in a nested field after reset with keepDirtyValues', async () => {
+    function App() {
+      const { getValues, reset, setValue, register } = useForm<{
+        nested: {
+          obj: {
+            value1: string;
+            value2: string;
+          };
+        };
+        topLevel: string;
+      }>({
+        mode: 'onChange',
+        defaultValues: {
+          nested: { obj: { value1: 'defaultValue1' } },
+          topLevel: 'defaultTopLevel',
+        },
+      });
+
+      return (
+        <form>
+          <input data-testid="value1Input" {...register('nested.obj.value1')} />
+          <input data-testid="topLevelInput" {...register('topLevel')} />
+
+          <p>value1: {getValues('nested.obj.value1')}</p>
+          <p>value2: {getValues('nested.obj.value2')}</p>
+          <p>topLevel: {getValues('topLevel')}</p>
+
+          <button
+            data-testid="dirtyValue1Button"
+            type="button"
+            onClick={() =>
+              setValue('nested.obj.value1', 'updatedValue1', {
+                shouldDirty: true,
+              })
+            }
+          >
+            Dirty Value 1
+          </button>
+
+          <button
+            data-testid="dirtyTopLevelButton"
+            type="button"
+            onClick={() =>
+              setValue('topLevel', 'updatedTopLevel', {
+                shouldDirty: true,
+              })
+            }
+          >
+            Dirty Top Level
+          </button>
+          <button
+            data-testid="resetButton"
+            type="button"
+            onClick={() =>
+              reset(
+                {
+                  nested: {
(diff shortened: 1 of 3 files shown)


bluebill1049:
thanks for the PR, i wonder if we can just reuse `_names.mount` which has all the mounted input filed names

flokol120:
@bluebill1049 Are names which were modified using `setValue()` and are not registered via a `Controller` or `register()` included in `_names.mount`? If no, those values would be lost when setting `keepDirtyValues` to `true`.

DavidIMk:
@bluebill1049 @flokol120 I think for this case the field in questions `nested.obj.value2` did not included in `_names.mount`, as it was not displayed/registered

bluebill1049:
Nice one, i will check it over the weekend. 🙏

bluebill1049:
I don't think we will cate for setValue API, it's just the whole thing more complicated than it needs to be.  Only for register/mounted inputs.

DavidIMk:
Hi, IMO the change isn't significant, as it only change 1 specific variables inside `_reset`. But sure, should we close the PR and the issue instead then?

bluebill1049:
> Hi, IMO the change isn't significant, as it only change 1 specific variables inside `_reset`. But sure, should we close the PR and the issue instead then?

hi, let me have a play a bit more, don't want t

bluebill1049:
Thanks for the PR, going throught the object to produce the path just seem more complicated than it needs to. I did test setValue with shouldDirty flag which still preserved the input value correct.

flokol120:
@bluebill1049 `reset` with `keepDirtyValues: true` won't produce a correct form state in this example:
https://codesandbox.io/p/sandbox/react-hook-form-reset-dynamic-array-z5xg3p

1. Make a change to the text input to make the form dirty
2. click on the reset button
3. optional: click on the updateState button to inspect the form state

The output of the form state is now:
```json
{"nested":{"obj":{"value1":"default1"}},"topLevel":"topLevel"}
```

However, I am resetting the form state with this:

```ts
  const newDefault = _.cloneDeep(defaultValues);
  // Setting this value won't be added to the defaultValues
  newDefault.nested.obj.value2 = "default2";
  // The top level value will be added to the defaultValues
  newDefault.topLevel = "topLevel";
  reset(newDefault, {
    keepDirtyValues: true,
  });
```

So `newDefault.nested.obj.value2` is not correctly added to the form state.

PR #12690, opened 2025-03-16
Fix #12680: Update Fieldarray Unmount Status

### LINKED ISSUE #12680: issue: UseFieldArray errors on form state after array field is unmounted
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/angry-fermi-ssm9j8

### Steps to reproduce

1. Go to [codesandbox link](https://codesandbox.io/p/sandbox/angry-fermi-ssm9j8)
2. Click on Submit
3. Both the text and array fields have errors
4. Click on toggle text field
5. The form-state is updated and there are no errors for the text field
6. Click on toggle array
7. Array field errors are still present on the form-state

### Expected behaviour

Array field errors should not be present on form state when array field is not mounted. I could use the `shouldUnregister` parameter on the `useArrayField` hook but for my use case I need to keep the array values on the form when the array field is unmounted.

Chrome

### Code of Conduct

DavidIMk:
Fix for: https://github.com/react-hook-form/react-hook-form/issues/12680

Issue
Error for a FieldArray root still persist after unmounting

Cause
When unmounting a FieldArray, currently there's no handler to change `field._f.mount` to false

Solution
Add a handler to change field mount attribute to false when unmounting in `useFieldArray`, similar to one in `useController`

Thank you and please CMIIW

### DIFF
--- a/src/__tests__/useFieldArray.test.tsx
+++ b/src/__tests__/useFieldArray.test.tsx
@@ -3460,6 +3460,89 @@ describe('useFieldArray', () => {
       expect(screen.queryByAltText('Max length should be 2')).toBeNull();
     });
 
+    it('should no longer validate when unmounted', async () => {
+      const ArrayField = () => {
+        const { fields } = useFieldArray({
+          name: 'array',
+          rules: {
+            required: {
+              value: true,
+              message: 'This is required',
+            },
+            minLength: {
+              value: 2,
+              message: 'Min length should be 2',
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
Awesome! thanks again for another fix 🙏

lamy (commented):
[]()[]()

PR #12731, opened 2025-04-08
fix: Changes setValue to skip values that are not in ownProperties

### LINKED ISSUE #12730: issue: setValue triggers Maximum Call Stack error
### Version Number

7.55.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/tctl4v?file=%2Fsrc%2FApp.tsx

### Steps to reproduce

1. Run the included codesandbox
2. The page hangs, and results in Maximum Call Stack error after a few seconds.
3. This happens when I call `setValue` to update the value of the form.

#### Reason
The error happens when I include MooTools (it is included in the `<script...` tag in the `index.html`. If I remove Mootools, it works fine.

#### Diagnosis
MooTools modifies the `Object.prototype` which includes a self reference to `$family` object. `setValue` is using `for...in` loop which goes over the `prototype` properties as well, and hence gets stuck in the loop of self-referencing objects.
The issue happened as a part of this commit https://github.com/react-hook-form/react-hook-form/commit/c03c66eb8d650a9a881fe47d3e968b80ec7b3918#diff-5a85c1643c71f45e050ce584d7c0db2bdf2e56f8a7dae86d74fedf01be2d2cab. 

Earlier `setValue` was using `Object.entries` to go over the `values` properties (which only takes `ownProperties` into account). But with the `for...in` change, it takes prototype properties as well.
https://github.com/react-hook-form/react-hook-form/blob/a66b52a4f0d5c0de3c09307c074371e3e881ce73/src/logic/createFormControl.ts#L659-L681

### Expected behaviour

The Maximum Call Stack error should not happen.

#### Potential fix
There are two potential fixes:
- Either Change the `for...in` to `Object.keys` in `setValues` to iterate over the `values` object.
```diff
- for (const fieldKey in value) {
+ Object.keys(value).forEach((fieldKey) => {
```
https://github.com/react-hook-form/react-hook-form/blob/56a1fbcdb062ea4beaed5cfb7535bd7d62a908b8/src/logic/createFormControl.ts#L583-L607
- Or add a `hasOwnProperties` check inside the loop to filter out the prototype properties. 
```diff
+ if (!value.hasOwnProperty(fieldKey) {
+   return;
+ }
```

Chrome

### Relevant log output

```shell
Uncaught RangeError: Maximum call stack size exceeded
    at get hasIndices (<anonymous>)
    at get flags (<anonymous>)
    at [Symbol.split] (<anonymous>)
    at String.split (<anonymous>)
    at get (index.esm.mjs:142:31)
    at setValues (index.esm.mjs:1255:21)
    at setValues (index.esm.mjs:1256:111)
    at setValues (index.esm.mjs:1256:111)
    at setValues (index.esm.mjs:1256:111)
    at setValues (index.esm.mjs:1256:111)
    at setValues (index.esm.mjs:1256:111)
    at setValues (index.esm.mjs:1256:111)
    at setValues (index.esm.mjs:1256:111)...
```

### Code of Conduct

piyushkmr:
To prevent infinite call stack issue due to self-referencing proto properties.

This is the issue I am trying to fix: https://github.com/react-hook-form/react-hook-form/issues/12730

[Codesandbox repo](https://codesandbox.io/p/sandbox/tctl4v?file=%2Fsrc%2FApp.tsx) for reproduction.


bluebill1049:
thanks for the pr. Could you share the codesandbox, I would like to see the usecase.

bluebill1049:
> @bluebill1049 , the codesandbox link is in the PR description and on the linked issue as well. Here you go: https://codesandbox.io/p/sandbox/tctl4v?file=%2Fsrc%2FApp.tsx

Could you make sure your CSB is public?

bluebill1049:
if diffcult to share the CSB, can you include a unit test for this case?

bluebill1049:
nice one, thanks i will try to get a test up later as well, then merge it in.

PR #12737, opened 2025-04-09
fix: remove stale keys when updating rules

### LINKED ISSUE #12729: issue: unexpected behavior when conditionally setting rules based on react state
### Version Number

7.22.5

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/shy-butterfly-3plpqf

### Steps to reproduce

1. Go to https://codesandbox.io/p/sandbox/shy-butterfly-3plpqf
2. Fill out the input field with less than 5 characters e.g.: 123
3. Click submit
4. "Data" is being displayed because of successful submit
5. Click "enable rules"
6. Click submit
7. See error
8. Click "disable rules"
9. Click submit
10. Still see error even though the rules should be disabled

### Expected behaviour

When setting the `rules` object on the `Controller` component based on some state or passed props, I expect the `Controller` to re-evaluate the rules prop when the component re-renders.

I tried to provide a minimal reproducible example that reflects the current problem we're facing with `react-hook-form`. In one of our components, we receive a boolean prop from the parent component. This prop is `true` for the first two renders and `false` for all subsequent renders. Based on this prop, we either apply some rules or we don't.

The issue we're encountering is that sometimes the `Controller` reacts to the state changes and removes the rules, which is what we expect. However, in some cases, it doesn't remove the rules, and we receive validation errors even though no rules were provided to the `Controller`.

For context, during the `first two renders`, the boolean is `true`, and we provide rules to the `Controller`. Starting from the `third render`, the boolean evaluates to `false`, and we set the rules prop to `undefined` because we don't want any rules to be applied. The problem is that the behavior is inconsistent: sometimes the `Controller` correctly reacts to the prop changes and removes the rules, but other times it doesn't, and the rules remain applied.

If I understand this [comment](https://github.com/react-hook-form/react-hook-form/issues/1749#issuecomment-645188955) correctly, this behavior should have already been addressed in `V6`.

We can apply the workaround with `unregister` the field in a `useEffect`, but we are not really happy with this hacky solution.

Chrome

### Relevant log output

```shell

```

### Code of Conduct

DaveFz:
This PR fixes #12729.

This PR is definitely not perfect — for example, some tests are still missing. I wasn’t able to get everything set up quickly or easily, so I wanted to open this initial draft PR before investing more time, especially since I’m not even sure if you agree with the issue or the changes yet. 🙂

After digging a bit deeper into the code, I noticed that when updating the rules object passed to the Controller, we don’t remove old keys — we just spread the new object. This leads to a situation where, for example, if you initially provide minLength and then later try to remove it by providing `undefined`, the key is still present and not properly unset.

Also please see the example `controller.tsx` I adjusted, in which you can easily see the fix in action, since I did not really provide any meaningful tests, yet ;)

### DIFF
--- a/app/src/controller.tsx
+++ b/app/src/controller.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { useForm, Controller, ValidationMode } from 'react-hook-form';
 import ReactSelect from 'react-select';
 import {
@@ -23,7 +23,7 @@ const options = [
 
 const defaultValues = {
   Native: '',
-  TextField: '',
+  TextField: 'asd',
   Select: '',
   ReactSelect: '',
   Checkbox: false,
@@ -61,15 +61,18 @@ export default function Field() {
 
   const rerender = () => setRerender(Math.random());
 
+  const [isRulesEnabled, setIsRulesEnabled] = useState(false);
+
   return (
     <form onSubmit={handleSubmit(() => {})}>
       <div className="container">
-        <section id="input-checkbox">
+        {/* <section id="input-checkbox">
           <label>MUI Checkbox</label>
           <Controller
             name="Checkbox"
             control={control}
-            rules={{ required: true }}
+            rules={isRulesEnabled ? { required: true } : undefined}
+            //rules={{required: isRulesEnabled}}
             render={({ field: props }) => (
               <Checkbox
                 {...props}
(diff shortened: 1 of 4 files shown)


bombillazo:
Hey @DaveFz , I, too, see the value in this enhancement. If you'd like to contribute to RHF+, I am tracking this in [this issue](https://github.com/bombillazo/rhf-plus/issues/38).

PR #12771, opened 2025-04-23
improve useWatch type when compute prop is specified

kotarella1110:
Overview:

- Introduced function overloads to useWatch to provide more precise typings based on the presence and type of the name prop.
- Enabled conditional typing of the compute function and useWatch's return type depending on how name is specified.
- Enabled better inference of return types based on name prop variations via function overloads.

### DIFF
--- a/src/types/form.ts
+++ b/src/types/form.ts
@@ -895,7 +895,7 @@ export type UseWatchProps<TFieldValues extends FieldValues = FieldValues> = {
     | readonly FieldPath<TFieldValues>[];
   control?: Control<TFieldValues>;
   exact?: boolean;
-  compute?: <T>(formValues: T) => T;
+  compute?: (formValues: any) => any;
 };
 
 export type FormProviderProps<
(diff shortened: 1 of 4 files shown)


src/types/form.ts
```diff
@@ -895,7 +895,7 @@ export type UseWatchProps<TFieldValues extends FieldValues = FieldValues> = {
     | readonly FieldPath<TFieldValues>[];
   control?: Control<TFieldValues>;
   exact?: boolean;
-  compute?: <T>(formValues: T) => T;
+  compute?: (formValues: any) => any;
```
bluebill1049, line 898: Curious about the reason why this has to go with `any`?

bluebill1049 (approved):

src/types/form.ts
kotarella1110: @bluebill1049 

Using `<T>(formValues: T) => T` as the type for the compute function causes incompatibility with the following overloads:

https://github.com/react-hook-form/react-hook-form/pull/12771/files#diff-ff419364ba61880799eb26586378725c10c0b87e2ed7b9954cafafd30d256805R208-R223

Also, since the return type of the compute function doesn’t always match the type of its argument, I believe `<T>(formValues: T) => T` is not accurate.

That’s why I changed it to `any`.

bluebill1049: aha i see thanks 🙏

bluebill1049:
thanks a lot @kotarella1110 🙏

PR #12775, opened 2025-04-24
📭 close #12773 pass input ref instead partial

### LINKED ISSUE #12773: Passing whole element to ref in Controller
Why isn’t the entire element passed to the ref in Controller? Right now, only methods like focus(), select(), setCustomValidity(), and reportValidity() are forwarded. I also need access to appendChild(), and I noticed a closed issue suggesting the use of onBlur() here: https://github.com/react-hook-form/react-hook-form/issues/12272.

bluebill1049:


terrymun:
@bluebill1049 Do you foresee this feature being restored? There are some use-cases that require access to the element referenced in the `ref`, and at this moment components that are set up using the controller approach does not have access to this ref.

For example, I am working on implementing a pattern where I use the field errors object to render an error summary: I'd like to have access to the label text associated with the element, and the only way to do it programmatically is to have access to the `ref` and then attempt to access its [`labels` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels).

bluebill1049:
> @bluebill1049 Do you foresee this feature being restored? There are some use-cases that require access to the element referenced in the `ref`, and at this moment components that are set up using the controller approach does not have access to this ref.
> 
> For example, I am working on implementing a pattern where I use the field errors object to render an error summary: I'd like to have access to the label text associated with the element, and the only way to do it programmatically is to have access to the `ref` and then attempt to access its [`labels` property](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/labels).

I’d love to support it, but it’s tricky to make everything work smoothly with all UI libraries without introducing major breaking changes. It’s definitely something we can consider for the next major release, though.

terrymun:
Thanks for the quick response @bluebill1049, really appreciate it. I'll run with using a `pnpm patch` for now, but I can't wait to see what the next major release has to offer 👍

bluebill1049:
yes let do in v8

PR #12780, opened 2025-04-25
🐞 fix #12772 reset form useWatch to utilize ref for defaultValue and …

### LINKED ISSUE #12772: issue: reset form
### Version Number

7.56.0

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/rhf-context-seterror-forked-sdvnx5?workspaceId=ws_4Vb4LVtiikCfhSM2yZ7Krg

### Steps to reproduce

1. Select a gender
2. Click on 'Reset' button
3. Selected value is not reset
4. Click on 'Reset' button again
5. Selected value is reset

### Expected behaviour

User must click twice to reset form.
Controller keep selected value.

Use a previous version of RHF (before 7.56.0), it works

Chrome

### Relevant log output

```shell
- After first reset click

watch gender: undefined
controller value: M

- After second reset click

watch gender: undefined
controller value: undefined
```

### Code of Conduct

Han5991:
🐞 fix #12772 reset form

**Problem:**

Previously, the `useWatch` hook could sometimes fail to immediately reflect the form values after a `reset` operation, particularly when `defaultValue` was provided and used in conjunction with `Controller`. This could lead to a scenario where a user had to trigger the reset action twice for the `useWatch` value to update correctly to the default value

The likely cause was related to how `defaultValue` was managed and included in the dependency array of `useDeepEqualEffect`. When the form was reset, the internal handling of default values could potentially cause `useWatch`'s internal subscription to briefly become out of sync with the form's actual state, as the `useDeepEqualEffect` might trigger a re-subscription based on a perceived change in the `defaultValue` object reference or content during the reset process. This timing mismatch meant the hook might not pick up the newly reset value immediately.

**Solution:**

This change modifies how `defaultValue` is handled within the `useWatch` hook. Instead of including `defaultValue` in the `useEffect` dependency array (previously via `useDeepEqualEffect`), `defaultValue` is now stored in a `useRef`.

By storing `defaultValue` in a `useRef`, its value can be accessed within the effect callback without being part of the dependency array.

**Impact:**

This ensures that the subscription established by `useWatch` is not unnecessarily torn down and re-established solely due to potential changes or re-creations of the `defaultValue` object during the form reset process. The subscription remains stable, allowing `useWatch` to reliably pick up and reflect the form's state, including the correctly applied default values, immediately after `reset` is called. This resolves the issue where a second reset was sometimes required.

### DIFF
--- a/src/__tests__/useWatch.test.tsx
+++ b/src/__tests__/useWatch.test.tsx
@@ -8,6 +8,7 @@ import {
   within,
 } from '@testing-library/react';
 
+import { Controller } from '../controller';
 import {
   Control,
   UseFieldArrayReturn,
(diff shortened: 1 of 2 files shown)


bluebill1049:
oh nice, any chance we could follow with a unit test for this? just so this can be prevented next time.

Han5991:
> oh nice, any chance we could follow with a unit test for this? just so this can be prevented next time.

Of course, I'll get a unit test ready to make sure this doesn't happen again.

bluebill1049 (dismissed):

bluebill1049 (approved):

Issue #12673, opened 2025-03-08
issue: defaultValue is not getting updated when we call resetField in useEffect


harshagw:
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/wwtgxr

### Steps to reproduce

When calling resetField within a useEffect hook to update the default value of a form field, the default value doesn't get updated as expected. Although the field's value changes, the default value remains the same. However, when wrapping the resetField call inside a setTimeout with a delay of 0 milliseconds, the issue is resolved, and the default value is updated correctly.

Here’s the scenario:

https://github.com/user-attachments/assets/d4cab361-8814-4864-b47d-4ad8d9432f1c

1. Without setTimeout: When we click the submit button, even though the field's value is updated (as seen in the form), the defaultValue doesn’t update immediately.

2. With setTimeout: Wrapping the resetField call inside setTimeout with a delay of 0 milliseconds fixes the issue. This ensures that the default value is correctly updated.

### Expected behaviour

default value and value should both be updated without using setTimeout. 

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #12786, opened 2025-05-01
fix: use layoutEffect in useWatch for subscription similar to useForm 

harshagw:
Replaced useEffect with useIsomorphicLayoutEffect inside useWatch

🧩 Context & Motivation
Previously, useWatch used useEffect to subscribe to form state changes. However, since useEffect runs after the DOM paint, this introduced a timing issue during initial render.

Scenario:
1. Component A (a sibling) resets the form and updates default values inside its own useEffect.
2. Meanwhile, another sibling using useWatch hasn’t yet subscribed, since its useEffect hasn’t executed.
3. Result: useWatch sees stale or initial values, not the updated ones.

🔗 CodeSandbox Example: [https://codesandbox.io/p/sandbox/drs4ll?file=%2Fsrc%2Fform-header.tsx%3A4%2C1](https://codesandbox.io/p/sandbox/drs4ll?file=%2Fsrc%2Fform-header.tsx%3A4%2C1)

This led to inconsistent behavior depending on component render order and positioning.

✅ Solution
By switching to useIsomorphicLayoutEffect, the subscription now occurs before the browser paints, ensuring useWatch reliably captures form state updates during the first render.

It also solves - https://github.com/react-hook-form/react-hook-form/issues/12673


bluebill1049:
thanks we already have that hook can you export it and reuse it?

harshagw:
Sure @bluebill1049. I have updated the PR. Since `useDeepEqualEffect` wasn't getting used anywhere now, I have removed it as well.

bluebill1049 (approved):

PR #12799, opened 2025-05-07
Add children render prop support to Controller component

qurugi0347:
## Changes

- Added support for using children prop in the Controller component.
- Controller can now be rendered using either the render prop or a children function.
- Modified ControllerProps type to require either render or children prop.
- Extracted ControllerRender type to eliminate code duplication and improve readability.

## Motivation

- Provides more intuitive usage that aligns with common React component patterns by supporting children prop.
- Allows users to use the Controller component in their preferred way (render prop or children function).
- Maintains consistency with other React libraries and provides a better developer experience.

## Compatibility

- Existing render prop is still supported, ensuring compatibility with existing code.
- This change is additive only and doesn't affect existing users.

## Usage
```
<Controller
   control={control}
   name="test"
 >
   ({ field: { onChange, onBlur, value, ref }, formState, fieldState }) => (
      <>
         <input
            onChange={onChange} // send value to hook form
            onBlur={onBlur} // notify when input is touched
            value={value} // return updated value
            ref={ref} // set ref for focus management
         />
         <p>{formState.isSubmitted ? "submitted" : ""}</p>
         <p>{fieldState.isTouched ? "touched" : ""}</p>
      </>
   )
 </Controller>
```

from #12696


bombillazo:
Hey @qurugi0347, I, too, see the value in this enhancement. If you'd like to contribute to RHF+, I am tracking this at https://github.com/bombillazo/rhf-plus/issues/34.

bluebill1049:
Thanks for the PR, I think this is probably something you can leave at your app code base, or perhaps go to RHF+ <3

Issue #12797, opened 2025-05-06
issue: Validation mode can't be changed dynamically


dusan233:
### Version Number

7.56.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/rhf-validationmode-issue-cxmtry

### Steps to reproduce

1. Go to [example](https://codesandbox.io/p/sandbox/rhf-validationmode-issue-cxmtry)
2. Under the **Change Mode:** heading, click the onChange button
3. In the **First Name** input, delete a single character
4. Notice that validation does not run and no error message appears
5. Clicking submit button will trigger validation and error message will appear

### Expected behaviour

After dynamically changing mode to "onChange", validation should be triggered on every input change. 

From what I can tell, the `onChange` handler only uses the cached values from `validationModeBeforeSubmit` and `validationModeAfterSubmit`, which are derived just once when creating form control, so it never picks up any subsequent changes to `mode` or `reValidateMode`.

### What browsers are you seeing the problem on?

Chrome, Firefox, Edge, Safari

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #12803, opened 2025-05-07
fix: make `mode` and `reValidateMode` reactive

dusan233:
Moved `validationModeBeforeSubmit` and `validationModeAfterSubmit` into the `onChange` handler so that we re-compute both modes on every event trigger. This ensures the handler always respects the latest mode and reValidateMode settings instead of relying on stale, one-time values.

Related issue: #12797 
## Type of change

### DIFF
--- a/src/__tests__/useForm.test.tsx
+++ b/src/__tests__/useForm.test.tsx
@@ -15,6 +15,7 @@ import {
   FieldErrors,
   FieldValues,
   FormState,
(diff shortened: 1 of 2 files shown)


bluebill1049:
could you include a unit test for this?

dusan233:
Yes, will do

bluebill1049 (approved):

Issue #12794, opened 2025-05-06
issue: errors prop in useForm doesn't trigger focus


jkbach:
### Version Number

7.56.2

### Codesandbox/Expo snack

https://snack.expo.dev/@jkbach/react-hook-form-focus-issues

### Steps to reproduce

1. Go to code sandbox.
3. Click on "Submit" without typing anything.
4. Observe that `field2Focus` is correctly focused because it fails the "required" rule validation (expected).
5. Type something in `field2Focus` so it passes validation.
6. Resubmit, which will create an error for `field1Focus` that will be supplied to `useForm` via the `errors` prop.
7. Observe that the error message appears under `field1NoFocus`, but the field does not focus (expected behavior: field focuses).

### Expected behaviour

`field1NoFocus` should focus when it enters error state when errors are supplied via the `errors` prop on `useForm`.

### What browsers are you seeing the problem on?

Firefox

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #12805, opened 2025-05-08
feat: Focus form field for errors supplied by errors prop

jkbach:
I originally filed #12794 as an issue, but now I understand that this was never supported functionality. This PR adds the functionality, although the bigger question is whether or not this is something that we want in the first place.

Note: this does change the "default" behavior for errors supplied via the `errors` prop, since before they wouldn't cause the field to be focused, but after this PR they will. If this change in default behavior isn't acceptable, I could add a new optional prop `shouldFocusErrorFromProps` to `useForm` and have it default to `false` so that the default behavior is unchanged. Personally I think that having both validation _and_ server error focus controlled by the same prop makes sense, since the form state doesn't distinguish errors based on their source.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -115,10 +115,14 @@ export function useForm<
     if (props.reValidateMode) {
       control._options.reValidateMode = props.reValidateMode;
     }
+  }, [control, props.mode, props.reValidateMode]);
+
+  React.useEffect(() => {
     if (props.errors && !isEmptyObject(props.errors)) {
       control._setErrors(props.errors);
+      control._focusError();
     }
-  }, [control, props.errors, props.mode, props.reValidateMode]);
+  }, [control, props.errors]);
 
   React.useEffect(() => {
     props.shouldUnregister &&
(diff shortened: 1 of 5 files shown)


bluebill1049:
i think we probably only need focus field, and that would cover all the usecase for focus management.

bluebill1049 (approved):

src/useForm.ts
```diff
@@ -115,10 +115,14 @@ export function useForm<
...
     }
+  }, [control, props.mode, props.reValidateMode]);
+
+  React.useEffect(() => {
     if (props.errors && !isEmptyObject(props.errors)) {
       control._setErrors(props.errors);
+      control._focusError();
     }
-  }, [control, props.errors, props.mode, props.reValidateMode]);
+  }, [control, props.errors]);
```
terrymun, line 125: Since `props.errors` is an object, wouldn't that make `useEffect()` run at every render since the object reference is not the same? This is probably not related to the PR itself per se since it's already present before, but might be worth looking into.

PR #12809, opened 2025-05-10
Optimize deepEqual for TypedArrays

pscanf:
I've noticed that when setting the value of a field to a large `Uint8Array` (e.g., more than 1MB), the browser chokes while running the `deepEqual` function, which performs poorly when dealing with that type of value. This PR fixes that problem. :)

### DIFF
--- a/src/__tests__/utils/deepEqual.test.ts
+++ b/src/__tests__/utils/deepEqual.test.ts
@@ -56,7 +56,7 @@ describe('deepEqual', () => {
     ).toBeFalsy();
   });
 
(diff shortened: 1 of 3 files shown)


bluebill1049:
Thanks for the PR, we are not considering supporting TypedArrays atm, something we may do later on.

pscanf:
Sure, no worries. Hope it'll make it in at some point. :)

Just as a note, TypedArrays _already_ work: if a field's value is set to a TypedArray it works just fine. The only issue is that when the array grows in size, the current implementation that compares the values takes a lot of time, hence the optimization.

PR #12833, opened 2025-05-21
fix: Enable getDirtyField to support Sets as a field type

lukejgaskell:
When using `Set` as a field type, the `getDirtyField` does not recognize changes to the field. This is because it attempts to iterate the field with `for ... in` which doesn't work on a `Set` type.

This PR excludes `Set` objects from being considered objects and are treated more or less like primitives when doing the `getDirtyField` traversal. The deep equals then treats the `Set` as an array, comparing order as well as values.

This PR does not support a `Set` being the top level object of a form.

### DIFF
--- a/src/__tests__/logic/getDirtyFields.test.ts
+++ b/src/__tests__/logic/getDirtyFields.test.ts
@@ -341,4 +341,28 @@ describe('getDirtyFields', () => {
       ],
     });
   });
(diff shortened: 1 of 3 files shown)


bluebill1049:
Thanks for the PR. We are not considering Set, Map, Class, etc. for now, as it would probably need more consideration for all API across the board. So I am going to hold onto merging the PR just yet.

PR #12839, opened 2025-05-22
🫚 feat: root errors count in schema error lookup

chrisgarber:
This is just reopening #12566. I'll note that it's very frustrating that that was closed as stale while I was on vacation. I think this has value and it would be nice to incorporate into the codebase.

When changing a field, particularly arrays, new errors can be created at root or nested root level. These should count as a net new error and should cause an update to the errors object. Currently, if you compact an array using something like a yup schema, root errors aren't added to the errors object, even though their children are dirty/changed.

[Here is a CodeSandbox](https://codesandbox.io/p/sandbox/recursing-edison-hmnksj) that demonstrates the behavior I'm trying to fix. If you add whitespace to any of the fields (they're trimmed so whitespace doesn't count), you'll note that the errors are empty. This change would include the root error in the errors object.

This is somewhat of a regression, this used to work before the resolvers made a change to put this errors under the root instead of straight on the error object for arrays.


bombillazo:
Hey @chrisgarber , I, too, see the value in this enhancement. In case you'd like to contribute to RHF+, I am tracking this in [this issue](https://github.com/bombillazo/rhf-plus/issues/33).

bluebill1049 (approved):

PR #12850, opened 2025-05-26
🐞 fix: add proper types to `form.subscribe`

CertainlyAria:
**Problem:**
[Documentation says](https://react-hook-form.com/docs/useform/subscribe) that `useForm().subscribe` can be used either by passing `name: string` or `name: string[]`.

1. With current type definitions passing `string[]` causes a type error.
2. Other APIs (for example `useForm().watch`) have proper autocomplete based on FieldPath but ` useForm().subscribe` lacks autocomplete

**Solution:**
Properly infer field name based on form shape


bluebill1049 (approved):
LGTM

PR #12858, opened 2025-05-31
📇 feat #12813 use stringToPath to prevent error at field name with quotes

### LINKED ISSUE #12813: issue: Error thrown when field name contains single or double quotes
### Version Number

7.56.3 (latest)

### Codesandbox/Expo snack

https://stackblitz.com/edit/react-4lphnvm2?file=src%2FApp.js

### Steps to reproduce

The stackblitz example clearly demonstrates in which this error happens:

1. Create a form with a field which name contains either single or double quote
2. Try to call `register` for this field
3. `Cannot read properties of undefined (reading '_f')` error is thrown

### Expected behaviour

There should be no runtime error no matter which field name you use.

Seems like this is related to the difference in how `set` and `get` function works with provided path. `set` replaces all double/single quotes using `stringToPath` function ([set code](https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/set.ts#L13), [stringToPath code](https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/stringToPath.ts#L4)). However `get` [doesn't do that](https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/get.ts#L15) which leads to it not being able to retrieve a key that was previously defined using `set`. 

Firefox, Chrome, Safari, Edge

### Relevant log output

```shell
Error in /turbo_modules/react-hook-form@7.56.3/dist/index.cjs.js (1:18733)
Cannot read properties of undefined (reading '_f')
```

### Code of Conduct

candymask0712:
to resolve #12813

Hello Bill 😁
I was impressed by how you tackled all the issues that had piled up last week in one go! I always learn a lot by browsing issues or contributing to react-hook-form and your activities. 
This time, I’m going to fix a problem related to the `get` function. The issue was well organized, so it was easy to identify the problem.

### Issue

Error thrown when field name contains single or double quotes

### Cause

The `compact` function cannot handle field names that contain single or double quotes.

### Solution

Replace all single/double quote paths using the `stringToPath` function (the `set` function already uses `stringToPath`).


bluebill1049 (dismissed):
Nice one 👍

src/utils/get.ts
```diff
@@ -12,7 +13,8 @@ export default <T>(
     return defaultValue;
   }
 
-  const result = compact(path.split(/[,[\].]+?/)).reduce(
+  const tempPath = isKey(path) ? [path] : stringToPath(path);
+  const result = tempPath.reduce(
```
bluebill1049: Save an extra variable, can just keep temPath inline.

`(isKey(path) ? [path] : stringToPath(path)).reduce`

bluebill1049 (approved):

PR #12859, opened 2025-05-31
🐞 fix: add type info for callback args in subscribe

### LINKED ISSUE #12856: issue: `name` type missing in `callback` args from `subscribe` util
### Version Number

7.56.4

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/icy-microservice-2gf86r

### Steps to reproduce

1. Use the `subscribe` util within a React component (TS project)
2. Inside the `callback` param, try to access `formValues.name`

You'll see that TS complains about `name` not existing within `formValues`, but it actually exists.

Image

### Expected behaviour

The `formValues.name` property can be accessed within `callback` of the `subscribe` util without type errors.

I think it should be as simple as adding `{name: string }` to the right part in the `callback` arg [in the UseFromSubscribe type here](https://github.com/react-hook-form/react-hook-form/blob/90cdaee9ba148e6c28fe3cb69f885ae1541f7c04/src/types/form.ts#L400).

Chrome

### Relevant log output

```shell

```

### Code of Conduct

candymask0712:
to resolve #12856

Hello, Bill 😁

### Issue
When trying to access the `name` field in the callback arguments of `subscribe`, a type error occurs.

### Cause
No type information is provided for the`name` field in the callback arguments of `subscribe`.

### Solution
Add type information for the `name` field in the callback arguments of `subscribe`.

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -347,6 +347,7 @@ export type FromSubscribe<TFieldValues extends FieldValues> = <TFieldNames exten
     formState?: Partial<ReadFormState>;
     callback: (data: Partial<FormState<TFieldValues>> & {
         values: TFieldValues;
+        name?: InternalFieldName;
     }) => void;
     exact?: boolean;
     reRenderRoot?: boolean;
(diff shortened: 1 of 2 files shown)


bluebill1049:
oh, nice one!

bluebill1049 (approved):
LGTM!

PR #12890, opened 2025-06-13
🧧 default to crypto.randomUUID in generateId

mastermatt:
## Proposed Changes

Use the native crypto module to generate UUIDv4s, when the browser supports it. 

I feel silly opening this PR, but I had a pen-test audit call out that our application was using an insecure UUID generator.

## Type of change
Enhancement?

## Checklist:

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -56,7 +56,7 @@ export default tseslint.config(
       '@typescript-eslint/ban-types': 'off',
       '@typescript-eslint/no-explicit-any': 'off',
       '@typescript-eslint/explicit-module-boundary-types': 'off',
-      "@typescript-eslint/consistent-type-imports": "error",
+      '@typescript-eslint/consistent-type-imports': 'error',
       '@typescript-eslint/no-unused-vars': [
         'warn',
         { ignoreRestSiblings: true },
--- a/src/logic/generateId.ts
+++ b/src/logic/generateId.ts
@@ -1,4 +1,8 @@
 export default () => {
+  if (crypto?.randomUUID) {
+    return crypto.randomUUID();
+  }
+
   const d =
     typeof performance === 'undefined' ? Date.now() : performance.now() * 1000;
 
--- a/src/logic/generateId.ts
+++ b/src/logic/generateId.ts
@@ -1,4 +1,4 @@
-import isUndefined from "../utils/isUndefined";
+import isUndefined from '../utils/isUndefined';
 
 export default () => {
   if (!isUndefined(crypto) && crypto.randomUUID) {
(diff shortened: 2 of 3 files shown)


eslint.config.mjs
```diff
@@ -56,7 +56,7 @@ export default tseslint.config(
       '@typescript-eslint/ban-types': 'off',
       '@typescript-eslint/no-explicit-any': 'off',
       '@typescript-eslint/explicit-module-boundary-types': 'off',
-      "@typescript-eslint/consistent-type-imports": "error",
+      '@typescript-eslint/consistent-type-imports': 'error',
```
mastermatt, line 59: This was done automatically when I ran the test suite.

bluebill1049 (dismissed):

bluebill1049 (approved):

src/logic/generateId.ts
```diff
@@ -1,4 +1,10 @@
+import isUndefined from '../utils/isUndefined';
+
 export default () => {
+  if (!isUndefined(crypto) && crypto.randomUUID) {
+    return crypto.randomUUID();
```
ITenthusiasm, line 5: @bluebill1049 I think this is an awesome change, and I know this already got merged. But I just wanted to point out that [`crypto.randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) will only work in secure contexts. So it might be good to only execute this function [in that scenario](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts#feature_detection).

If this isn't a concern, feel free to ignore. :)

bluebill1049: should be fine, will keep mointor for potential issues

mastermatt: Unless I'm mistaken, in a non-secure context, `crypto ` is still defined globally, but has no properties, so the second half of of the `if` statement is falsey.
I've confirmed on Chrome 137 and Firefox 139 (osx 15.5).

sschickowski: When running vitest with a jsdom environment, tests will fail with `useFieldArray` 
```
ReferenceError: crypto is not defined
 ❯ generateId node_modules/react-hook-form/src/logic/generateId.ts:4:20
```

hartmannco: In React Native with Hermes (I couldn’t test other setups), this also fails with:
`ReferenceError: Property 'crypto' doesn't exist`

ITenthusiasm: Seems a `typeof crypto !== "undefined"` check would be better. Might also be good to double-verify that `crypto.randomUUID` is indeed a function (instead of just a property).

kai-dorschner-twinsity: I'm a bit late to the party, but my 2 cents: could we use the react [useId() hook](https://react.dev/reference/react/useId)?

PR #12896, opened 2025-06-18
fix: watch return type based on defaultValue

### LINKED ISSUE #12705: issue: watch should provide undefined as a possible TypeScript type
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/nklsmk

### Steps to reproduce

The `watch` documentation (https://react-hook-form.com/docs/useform/watch) says that when defaultValues are not provided, the value returned by watch (or useWatch) will be undefined.  This is fine, but the TypeScript support should make sure that undefined is a possible type for the returned value.  Right now it will just use whatever the type is of the form, which often are values that do not have undefined as an option.  This leads to a situation where we cannot trust the TypeScript compiler or our IDE.

1. Go to the codesandbox link: https://codesandbox.io/p/sandbox/nklsmk
1. See the errorImportant parts:

```

type MyType = {
  aString: string;
};

  const { watch } = useForm<MyType>();

  // Hover over aStringValue and see that it is showing only
  //   type 'string' instead of type 'string | undefined'
  const aStringValue = watch("aString");

  // And you will see an error in the preview because I am using
  //   a undefined value with no IDE warning that it could be
  //   undefined.
  return (
    <div className="App">
      Here is aStringValue.toUpperCase(): {aStringValue.toUpperCase()}
    </div>
  );
}

```

### Expected behaviour

The type of `aStringValue` above should be `string | undefined`

_No response_

### Relevant log output

```shell

```

### Code of Conduct

codepunkt:
Fixes #12705 

Attention: `pnpm tsd` logs

```
The test file `dist/index.test-d.ts` or `dist/index.test-d.tsx` does not exist in `/Users/christoph/repos/own/react-hook-form`. Create one and try again.
```


bluebill1049:
Could you please update the api contract?

MTG2000:
Update: Just saw that there was already an issue opened for that.

Hey

I noticed that after the new version, the type of the value returned by `watch` when the `useForm` **has** some default values always include `undefined`

Ex:
```ts
 const form = useForm({
    defaultValues: {
      name: '',
    },
  })

  const nameWatched = form.watch('name')
  //     ^? string | undefined
  ```
  
 Is this correct?
 AFAIK, `nameWatched` will always have a string value.

(Even if I pass a form type `{ name: string }` to useForm generic, it's still the same.

PR #12904, opened 2025-06-20
fix: typo in `UseFormSubscribe` and missing event type for callback data in subscribe

### LINKED ISSUE #12899: feat: allow to subscribe to changes of a particular event type using `subscribe`
When using `watch` you could get an event type in the callback:

```typescript
useEffect(() => {
  const { unsubscribe } = watch((values, { name, type }) => {
    if (type === 'change') {
      console.log(values);
    }
  });

  return () => unsubscribe();
}, [watch]);
```

This way we were able to distinguish between user input and a value being set with `setValue`.

After migrating to `subscribe` I found out that it doesn't provide such functionality.

Maybe add an event type to the callback data.

```typescript
useEffect(() => {
  const callback = subscribe({
    formState: {
      values: true,
    },
    callback: ({ values, type }) => {
      if (type === 'change') {
          console.log(values);
      }
    },
  });

  return () => callback();
}, [subscribe]);
```

pushys:
Closes #12899.

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -761,7 +761,7 @@ export type UseFormReturn<TFieldValues extends FieldValues = FieldValues, TConte
     control: Control<TFieldValues, TContext, TTransformedValues>;
     register: UseFormRegister<TFieldValues>;
     setFocus: UseFormSetFocus<TFieldValues>;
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):
LGTM

PR #12910, opened 2025-06-24
added an example of full usage of react-hook-form with automated building form with validations and all type of inputs

faridsh69:
Here you can find a component named Form that accept these props:

inputs: Form is a component that will build a form based on array of inputs in props.
schema: Form accepts an schema and do live validation based on that schema with statistics.
columns: For each input you can set width of the input based on 12 column grid.
onCahngeInput: You have access to current form data and current changed input via this callback.
uikitMapper: You can override the default uikit components.
Grouping fields: You can manage unlimit group in group inputs with not even 1 line of code.


faridsh69:
Please let me know if I need to move the source code of Form component to examples of react-hook-form here and then someone can share it through codesandbox,

Right now I moved all codes inside codesandbox will full usage example of the component

PR #12914, opened 2025-06-25
feat: support deep equality checking with circular references

n8pjl:
Let deepEqual handle objects with circular references without infinitely recursing.


bluebill1049:
could you include a test case?

n8pjl:
Sorry about the delay, but I think this covers applicable situations well.

bluebill1049 (approved):
LGTM

PR #12921, opened 2025-06-30
docs: remove blue underline from README.md

Marukome0743:
## Summary

Documentation:
- Condense sponsor link markup in [README.md](https://github.com/react-hook-form/react-hook-form/blob/master/README.md) by collapsing `<a>` and `` tags onto single lines to remove blue underlines from sponsor logos.

## Appearance

| Before | After |
|--------|--------|
| before | after |

## Why?

Previously, the images were displayed with line breaks between them, causing blue link lines to appear, which disrupted the clean look of the images. By removing these line breaks, the images now appear without the blue link lines, providing a more cohesive and visually appealing README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,92 +62,38 @@ function App() {
 
 We’re incredibly grateful to these kind and generous sponsors for their support!
 
-<a target="_blank" href="https://www.follower24.de/">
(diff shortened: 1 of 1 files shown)


bluebill1049 (approved):
LGTM! nice i always wondered why

PR #12928, opened 2025-07-03
🚀 chore: use TSTyche for type testing

mrazauskas:
It is admirable to see how well type tests are organised in this repo. I was writing many type tests, but available tooling required numerous workarounds and did not scale.

As an answer to that, I created [TSTyche](https://tstyche.org), a type testing tool I always wanted to have. Perhaps you will like it too?

Here is what TSTyche brings in:

- assertion that compare two types without a need to create intermediate values
- `it()` and `describe()` helpers. The test reads and acts the same as your unit tests. You can use `.only` and `.skip`. Also from command line: `tstyche --only DeepPartial`
- report that lists TypeScript version, used `tsconfig.json` and test files:
   Screenshot 2025-07-03 at 08 06 37
- run the tests agains specific TypeScript version or even a range of versions: `tstyche --target '>=5.6'`
- the install size is only 250kB (to compare, the install size of `tsd` is 38MB)
- (almost forgot, obviously there is no requirement to create empty `.d.ts` or so)

And there is more: watch mode, [checks of errors](https://tstyche.org/reference/config-file#check-suppressed-errors) silenced by `// @ts-expect-error`, [template test](https://tstyche.org/guides/template-test-files) files, etc.

---

Thanks for your time. If it sounds useful, I would be glad to migrate all the tests.


package.json
```diff
@@ -109,6 +109,7 @@
     "rollup": "^4.44.1",
     "rollup-plugin-typescript2": "^0.36.0",
     "tsd": "^0.31.2",
```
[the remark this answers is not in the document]
mrazauskas: Too early. Removing this would break CI. Actually I did not include TSTyche to any CI job. That is simple, I just wanted to keep the PR focused. This is just a demo. It shouldn’t be merged until all tests are migrated.

src/__typetest__/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+  "extends": "../../tsconfig.json",
+  "include": ["**/*"],
+  "exclude": []
```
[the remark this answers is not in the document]
mrazauskas: Can be this is base on TypeScript documentation, but not on real code. The reality is more complex. The logic of the `include` pattern does not include the `node_modules` directories (and paths starting with a dot) by default. Those can only be included if explicitly specified. Not the case here. Therefore `"exclude": []` is correct, it is only meant to override inherited `exclude` patterns.

The logic that turns glob pattern into regex, can be found here: https://github.com/microsoft/TypeScript/blob/833a8d492c728d606454865e8c0fee84842f9f10/src/compiler/utilities.ts#L9543-L9710

---

Here is a quick script to see how the internal logic works:

```js
import process from "node:process";
import ts from "typescript";

// @ts-expect-error 'matchFiles' is internal
const files = ts.matchFiles(
  /* path */ process.cwd(),
  /* extensions */ undefined,
  /* excludes */ [],
  /* includes */ ["scripts/**/*"],
  /* useCaseSensitiveFileNames */ true,
  /* currentDirectory */ process.cwd(),
  /* depth */ undefined,
  // @ts-expect-error 'getAccessibleFileSystemEntries' is internal
  ts.sys.getAccessibleFileSystemEntries,
  ts.sys.realpath,
);

console.log(files);
```

mrazauskas:
@bluebill1049  I found some time to migrate all the tests. Please, take a look. No pressure. Only wanted to show what is possible.

Unfortunately the diff is rather overwhelming. I kept the same file names hoping to keep the diff more readable, that works in som places.

src/__tests__/logic/getCheckboxValue.test.ts
```diff
@@ -15,7 +15,7 @@ describe('getCheckboxValue', () => {
           name: 'bill',
           checked: true,
           value: '3',
-          // @ts-expect-error this is a mock for html input
+          // @ts-expect-error! this is a mock for html input
```
mrazauskas, line 18: TSTyche checks messages of suppressed errors. That is not important in unit tests, adding the `!` disables this check.

src/__tests__/type.test.tsx
```diff
@@ -610,9 +610,9 @@ test('useWatch should correctly select the name from object like param', () => {
       control,
     });
 
-    expectType<{ second: string }>(resultFromObj);
-    expectType<{ second: string }>({ ...resultFromObj });
-    expectType<{ second: string }>(resultFromInline);
+    tstyche.expect(resultFromObj).type.toBe<{ second: string }>();
+    tstyche.expect({ ...resultFromObj }).type.toBe<{ second: string }>();
+    tstyche.expect(resultFromInline).type.toBe<{ second: string }>();
```
mrazauskas, line 618: These assertions is the reason why unit tests got included in the type test run. Currently `tsd` does not check them.

src/__typetest__/__fixtures__/index.ts
```diff
@@ -1,4 +1,4 @@
 export * from './pathString';
 export * from './traversable';
 export * from './tuple';
-export * from './type';
+export * from './zodResolver';
```
mrazauskas, line 4: Perhaps `zodResolver` should be moved into `__fixtures__`?

src/__typetest__/errors.test-d.ts
```diff
@@ -1,21 +1,22 @@
-import { expectType } from 'tsd';
+import { describe, expect, it } from 'tstyche';
 
-import type { FieldError, FieldErrors, GlobalError, Merge } from '../types';
+import type { FieldError, GlobalError, Merge } from '../types';
+import { useForm } from '../useForm';
```
mrazauskas, line 4: The last test case in this file is the reason. It was a struggle to reach nested fields using the `FieldErrors` type. Rewriting these tests to e2e style makes it all work. Perhaps, that reflects a user story better as well?

tstyche.config.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "https://tstyche.org/schemas/config.json",
+  "checkSuppressedErrors": true,
+  "checkSourceFiles": false,
```
mrazauskas, line 4: The source files are all other files in the program that are not test files. They are checked via `tsc --noEmit`. No need to waste time.

mrazauskas (commented):
Few notes are left in the inline comments.

tstyche.config.json
mrazauskas: Hm.. Only problem that `__fixtures__/*` are not checked anymore. I have to think about this.

mrazauskas: This is addressed in the latest `tstyche@4.3.0`, which now has the [`fixtureFileMatch`](https://tstyche.org/reference/config-file#fixture-file-match) configuration option. (The real world inspired usage / requirements are the reasons why I create PRs like this one.)

bluebill1049 (dismissed):

mrazauskas:
Time to give up, I guess.

PR #12932, opened 2025-07-04
test(utils): add unit tests for live function with Ref type

kamja44:
This PR adds unit tests for the live utility function to verify its behavior under different scenarios:

Returns true when the ref is an HTMLElement and is connected to the DOM.

Returns false when the ref is not connected.

Returns false when the ref is not an HTMLElement.

These tests use a mocked version of isHTMLElement to isolate the function logic.
The goal is to ensure the correctness and reliability of the live function used internally in the library.


bluebill1049 (approved):
LGTM

PR #12935, opened 2025-07-04
chore: improve `exclude` patterns

mrazauskas:
This improves `exclude` patterns of 'tsconfig.json'. See inline comments of the details.

Currently `dist/__tests__` and `dist/__typetest__` are emitted. The `postbuild` deletes `dist/__tests__`, but `dist/__typetest__` are shipped to npm registry. This change makes sure no test files are emitted to the `dist` directory. That build time shortens and the shipped packed becomes smaller too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,7 @@
   "types": "dist/index.d.ts",
   "sideEffects": false,
   "files": [
-    "dist",
-    "dist/__tests__"
+    "dist"
   ],
   "exports": {
     "./package.json": "./package.json",
@@ -30,7 +29,7 @@
     "prebuild": "pnpm clean",
     "build": "pnpm build:modern",
     "build:watch": "pnpm build:modern -w",
-    "postbuild": "rimraf dist/__tests__ && node ./scripts/rollup/assert-esm-exports.mjs && node ./scripts/rollup/assert-cjs-exports.cjs",
+    "postbuild": "node ./scripts/rollup/assert-esm-exports.mjs && node ./scripts/rollup/assert-cjs-exports.cjs",
     "build:modern": "rollup --bundleConfigAsCjs -c ./scripts/rollup/rollup.config.js",
     "build:esm": "rollup --bundleConfigAsCjs -c ./scripts/rollup/rollup.esm.config.js",
     "prettier:fix": "prettier --config .prettierrc --write .",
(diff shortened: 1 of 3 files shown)


scripts/rollup/createRollupConfig.js
```diff
@@ -28,9 +28,7 @@ export function createRollupConfig(options, callback) {
     external: Object.keys(pkg.peerDependencies),
     plugins: [
       typescript({
-        tsconfig: options.tsconfig,
```
mrazauskas, line 31: This is always `undefined` in current setup.

scripts/rollup/createRollupConfig.js
```diff
@@ -28,9 +28,7 @@ export function createRollupConfig(options, callback) {
...
         clean: true,
-        exclude: ['**/__tests__', '**/*.test.ts', '**/__typetest__'],
```
mrazauskas, line 33: Duplication.

tsconfig.json
```diff
@@ -18,16 +18,6 @@
...
-    "examples",
-    "cypress",
-    "src/*.test.ts",
-    "src/*.test.tsx",
-    "src/*.test-d.ts",
-    "src/*.test-d.tsx",
-    "src/__mocks__"
-  ]
+  "include": ["src/**/*"],
+  "exclude": ["**/__tests__/**/*", "**/__typetest__/**/*"]
```
mrazauskas, line 22: The `include` pattern specifies that only files from the `src` directory should be included. `exclude` is applied after includes are resolved. The `src` directory does not have any nested directories like `node_modules`, `app`, `examples`, `cypress`. Also `src/__mocks__` does not exist. These can be removed, I think.

The test currently live under `src/__tests_/*` and `src/__typetest_/*`, but these are not excluded by `src/*.test.ts` or `src/*.test-d.ts`. The `**/` is missing in these globs.

The added `**/__tests__/**/*` and `**/__typetest__/**/*` are targeting the test files. This means neither tests nor type tests will be emitted to the `dist` directory. (In contrary current npm package ships with the type tests.)

package.json
```diff
@@ -12,8 +12,7 @@
   "types": "dist/index.d.ts",
   "sideEffects": false,
   "files": [
-    "dist",
-    "dist/__tests__"
```
mrazauskas, line 16: Currently `dist/__tests__` is deleted explicitly by the `postbuild` script.

This change excludes emitting anything to `dist/__tests__`.

bluebill1049 (approved):
LGTM, nice clean up.

PR #12939, opened 2025-07-06
test: add unit tests for unsetEmptyArray behavior

kamja44:
Adds unit tests for the `unsetEmptyArray` utility function, which unsets a given path in an object if the value is an empty array (or falsy after compaction).

- Tests that `unset()` is called when the target value is an empty array
- Tests that `unset()` is **not** called when the array has truthy elements
- Mocks `unset()` using Jest and verifies call behavior

This function is used internally to clean up empty values in form-related objects like `errors` or `dirtyFields`.  
Having test coverage ensures correct conditional cleanup and improves confidence in future refactors.

### DIFF
--- a/src/__tests__/logic/unsetEmptyArray.test.ts
+++ b/src/__tests__/logic/unsetEmptyArray.test.ts
@@ -1,0 +1,31 @@
+import unsetEmptyArray from '../../logic/unsetEmptyArray';
+import unset from '../../utils/unset';
+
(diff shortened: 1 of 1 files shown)


bluebill1049:
could you please combine all your prs into one?

kamja44:
Sure! I will close the existing three PRs(#12939, #12938, #12937) and create a new one that combines all three tasks(#12940).

PR #12942, opened 2025-07-06
fix(build): Remove React wildcard import to resolve ESM build issues

### LINKED ISSUE #12763: issue: Build failure after auto-updating to 7.56.0: Can't import named exports like useEffect/useRef
### Version Number

7.56.0

### Codesandbox/Expo snack

N/A

### Steps to reproduce

1. package.json had "react-hook-form": "^7.54.2".

2. After running npm install or yarn install, the package auto-updated to 7.56.0.

3. Attempted to build the project (e.g., with Webpack).

4. The following errors appeared during the build:
`Can't import the named export 'useEffect' from non EcmaScript module (only default export is available)
Can't import the named export 'useRef' from non EcmaScript module (only default export is available)`

A similar issue was previously reported [here](https://github.com/react-hook-form/react-hook-form/issues/10465).

Following the suggested workaround from that issue (locking the version without a caret ^) resolved the problem.

It seems that the issue was introduced in 7.56.0.

### Expected behaviour

The project should build successfully.

Named exports like useEffect and useRef should be properly imported without errors.

_No response_

### Relevant log output

```shell
[vite] Internal server error: Can't import the named export 'useEffect' from non EcmaScript module (only default export is available)
[vite] Internal server error: Can't import the named export 'useRef' from non EcmaScript module (only default export is available)
```

### Code of Conduct

### LINKED ISSUE #12892: issue: Build failure in 7.56.3: Can't import the named export 'useEffect' from non EcmaScript module
### Version Number

7.56.3

### Steps to reproduce

Build is failing with

```
Can't import the named export 'useEffect' from non EcmaScript module (only default export is available)
Can't import the named export 'useLayoutEffect' from non EcmaScript module (only default export is available)
```

both failures point to index.esm.js line 220, a line that contains

```
const useIsomorphicLayoutEffect =
  typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect;
```

After downgrading to 7.54.2 the issue went away.

Looks similar to #12763 but the failing hooks are different.

### Expected behaviour

The build passes with no issues.

### Code of Conduct

afontcu:
Closes #12892
Closes #12763

This PR removes the `import * as React from 'react'` wildcard import in favor of a default import. This is the only wildcard import in the codebase and is known to cause issues with some build tools (see issues above).

I've been investigating this issue and it seems related to build tooling not supporting `.mjs` files properly. Given that there's only one instance of the wildcard import it seemed like a legit fix.

Here's [the resulting diff](https://www.jstoolset.com/diff/b9d6cabf0a26fdd8) of `index.esm.js`. Note that the only meaningful change is that `import * as React from 'react'` is gone, other changes are just a rename from `React__default` to `React`.

While this is a targeted fix, a more robust solution could be to update the build process or add a lint rule to prevent future wildcard imports. I'm happy to explore this further.

Since it's my first PR in react-hook-form, let me know if I'm missing any important step :) happy to iterate.

thanks!

### DIFF
--- a/src/useIsomorphicLayoutEffect.ts
+++ b/src/useIsomorphicLayoutEffect.ts
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React from 'react';
 
 export const useIsomorphicLayoutEffect =
   typeof window !== 'undefined' ? React.useLayoutEffect : React.useEffect;


bluebill1049:
oh wow! Thanks very much for that. That was a tiny mistake that hard to spot.

bluebill1049 (approved):

afontcu:
> oh wow! Thanks very much for that. That was a tiny mistake that hard to spot.

awesome, my pleasure! 😄 When can I expect the upcoming version to be available?

afontcu:
> awesome, my pleasure! 😄 When can I expect the upcoming version to be available?

sorry for the ping @bluebill1049, do you have a fixed release schedule? was wondering when this fix would become available to install :) tyvm!

Issue #12821, opened 2025-05-16
issue: passed to form.watch in useEffect is being called in a loop


shahbazshueb:
### Version Number

7.56.3

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/still-bird-v2t687?workspaceId=ws_BHZmXkUXGz6U7TXYujBZKY

### Steps to reproduce

So I was submitting the form on change in the following way:

```
useEffect(() => {
    // TypeScript users 
    // const subscription = watch(() => handleSubmit(onSubmit)())
    const subscription = watch(handleSubmit(onSubmit));
    return () => subscription.unsubscribe();
}, [handleSubmit, watch]);
```

This was live and was working fine for me in v7.50.1 but as soon as I update to v7.56.3, it stopped working as expected and now it the watch calls the callback function again and again. The attached sandbox reproduces this problem. I had to downgrade my package version for now to ensure it works properly but would love to know when this is fixed or what I can do to fix this issue.

### Expected behaviour

On each chage in the form, the callback being passed to the watch function should only be called once rather than in a loop.

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

shahbazshueb:
@robinmartensss WIth this approach, the problem still persists, the callback will still be called again and again and cause performance issues?

robinmartensss:
I tested it in your sandbox and this should work, if not I won't be able to help you further, sorry!

```
useEffect(() => {
    const subscription = form.watch((_, { name, type }) => {
      if (type === "change") {
        onSubmit();
      }
    });
    return () => subscription.unsubscribe();
  }, [form, onSubmit]);
```

You can even add a specific name if that's needed

bluebill1049:
please consider to use https://react-hook-form.com/docs/useform/subscribe instead

Kapaak:
> > > [@bluebill1049](https://github.com/bluebill1049)  please consider to use https://react-hook-form.com/docs/useform/subscribe instead
> > 
> > 
> > I wanted to use subscribe instead of watch, but the subscribe callback doesnt provide "name" which I use to check, which field has changed. Is there any way to get the name in the callback?
> > ```
> > const subscription = watch(
> >       ({ ownRules, templateId, rules, advisorId }, { name }) => {
> >         switch (name) {
> >           case 'firstName':
> >               ....
> >             break;
> > 
[...]

PR #12945, opened 2025-07-07
Trigger watch callbacks in response to value changes only

### LINKED ISSUE #12721: issue: `watch` with callback triggers when calling `clearErrors`
### Version Number

7.55.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/react-hook-form-v7-ts-clearerrors-forked-qlspzg

### Steps to reproduce

1. Open attached sandbox
2. Click on 'Clear First Name Errors' or 'Clear All Errors'

### Expected behaviour

The `onChange` callback didn' triggered.

> Note: I think, that when we want clearing form errors we don't expect that any of watch callbacks will triggered (because it's nothing changed). The expected behaviour was on previous version (7.54.2), but in current version (7.55.0) it changed 

_No response_

### Relevant log output

```shell

```

### Code of Conduct

joshkel:
Prior to 7.55.0, `watch` with a callback would trigger that callback only in response to value changes.  (This is the behavior that I expected from [the documentation](https://react-hook-form.com/docs/useform/watch): "Subscribe to field update/change," and the other forms of the `watch` API are specifically for field values.) Some code in our application depended on this behavior; for example:

```ts
useEffect(() => {
  const subscription = watch(async (formData) => {
    if (await trigger()) {
      doSomethingWithValidForm(formData);
    }
  });
  return () => subscription.unsubscribe();
}, [trigger, watch]);
```

React Hook Form 7.55 changed this behavior to trigger the watch callback on every state change, not just value changes.  Specifically, in 7f95b265d16aee8b6b34d2746f4248aebb9d4678, every occurrence of `_subjects.values.next({ values: ... })` was changed to `_subjects.state.next({ values: ... })`, and `watch` now subscribes to `_subjects.state` instead of `_subjects.value`. As a result, the watch callback to be called on every state change.  Note that, in that commit, tests in `watch.test.tsx` had to be updated to reflect the extra calls.

We can get back the pre-7.55 behavior by only triggering the `watch` callback if the `values` property is present in the payload of the state change, since it was present for every call to `_subject.values.next` in pre-7.55.

Note that this change potentially breaks compatibility with 7.55 through 7.60 (since `watch` is no longer called on every state change), but it fixes a break in backward compatibility with pre-7.55.

I'd really appreciate it if this could be merged: the break in backward compatibility relative to 7.54 and below has affected multiple users, and any issues with changing behavior relative to 7.55 and above should be less severe (especially since `watch` is deprecated starting in 7.55 and the easier-to-use `subscribe` API is now available). Thank you.

Fixes #12721 and #12821. See also #12918.

### DIFF
--- a/src/__tests__/useForm/watch.test.tsx
+++ b/src/__tests__/useForm/watch.test.tsx
@@ -613,10 +613,10 @@ describe('watch', () => {
     render(<App />);
 
     fireEvent.click(screen.getByRole('button'));
-    expect(mockedFn).toHaveBeenCalledTimes(2);
+    expect(mockedFn).toHaveBeenCalledTimes(1);
 
     fireEvent.click(screen.getByRole('button'));
-    expect(mockedFn).toHaveBeenCalledTimes(4);
+    expect(mockedFn).toHaveBeenCalledTimes(2);
   });
 
   it('should remain isReady form state for subscription', () => {
--- a/src/__tests__/useWatch.test.tsx
+++ b/src/__tests__/useWatch.test.tsx
@@ -342,6 +342,48 @@ describe('useWatch', () => {
     expect(screen.getByText('345')).toBeVisible();
   });
 
+  it('should avoid triggering extra callbacks', () => {
+    const onChange = jest.fn();
+    type FormInputs = {
+      firstName: string;
+    };
+
+    const App = () => {
+      const {
+        register,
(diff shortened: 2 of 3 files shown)


bluebill1049:
Thanks again for looking into this and fix the root cause.

bluebill1049 (approved):

PR #12968, opened 2025-07-16
fix: ensure each createFormControl.subscribe subscription listens only to the changes it subscribes to

dusan233:
Not sure if its a bug fix or a feature. When you register multiple subscriptions, their formState filters are merged, so the last called subscription will listen on updates for both its own fields and all fields subscribed to by earlier subscriptions.

Example: [https://codesandbox.io/p/sandbox/vwfnsk?file=%2Fsrc%2FApp.tsx%3A5%2C18](https://codesandbox.io/p/sandbox/vwfnsk?file=%2Fsrc%2FApp.tsx%3A5%2C18)
Steps:
1. notice that App component is subscribing only to value changes on field "lastName".
2. type one character into "lastName" field.
3. notice that "lastName" subscription callback is being called both for value change and fields dirty state being changed as if we subscribed like: 
```
subscribe({
      formState: {
        values: true,
        dirtyFields: true,
      },
      name: 'lastName',
      callback: (state) => {
        console.log('Form state changed:', state);
      },
    });
```

This PR ensures that each subscription listens only to the changes it subscribes to.


src/logic/createFormControl.ts
```diff
@@ -160,6 +160,9 @@ export function createFormControl<
     isValid: false,
     errors: false,
   };
+  const _proxyFormState: ReadFormState = {
+    ...defaultProxyFormState,
+  };
   let _proxySubscribeFormState = {
     ..._proxyFormState,
   };
```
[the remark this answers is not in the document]
dusan233: I’m not sure why Copilot recommends removing selected lines. `_proxySubscribeFormState` is still referenced by many internal formControl methods and it needs to be initially defined just like `_proxyFormState`.

src/logic/createFormControl.ts
```diff
@@ -1059,7 +1062,10 @@ export function createFormControl<
     };
     return _subscribe({
       ...props,
-      formState: _proxySubscribeFormState,
+      formState: {
+        ...defaultProxyFormState,
+        ...props.formState,
+      },
```
bluebill1049, line 1072: Do we still need this `defaultProxyFormState`? 

what's the diff if we just do: 

```
..._proxyFormState,
...props.formState,
```

dusan233: I think we do still need `defaultProxyFormState`. The example below also mutates `_proxyFormState`, so we may end up subscribing to changes in formState that we didn't request in the `subscribe` call.

```
const fieldFormState = useFormState({
    name: "field1",
    control,
 });

fieldFormState.touchedFields
```
Do you want me to create test for this?

bluebill1049: > Do you want me to create test for this?

yes plz <3

bluebill1049 (approved):

PR #12971, opened 2025-07-18
fix: prevent onBlur for readOnly fields

Adityapradh:
Prevents onBlur event from firing for readonly input fields to fix unintended validation triggers.


bluebill1049:
thanks for the fix can you attach some test coverage?

bluebill1049 (approved):

Issue #12838, opened 2025-05-22
Feature: return defaultValue from getFieldState in order to have opportunity reset dependent field correctly


evgeniyworkbel:
**Is your feature request related to a problem? Please describe.**
I have a business requirement to make dependent selects. E.g., 2 selects - second is disabled while value is not chosen in the 1st one. When value of 1st has chosen, the 2nd select enables and fetches options for it (for example). Also I have requirement to reset value in the 2nd select after I has resetted value in the 1st one.

This is pretty simple example, but there are 8-10 dependent selects in my production code.

It will be very nice and convenient to make opportunity to get default value of field in order to reset dependent field to its default value correctly.

Now I pass 2 custom props for solving business req: `deps` with `resetOptionsForDeps`. These are arrays. In custom component during `onChange` fn call I iterates through `deps` and then find reset config in `resetOptionsForDeps` by current index. The disadvantage of the solution is that we need to provide `defaultValue` for each dependent field in `resetOptionsForDeps` within `deps` prop always. This causes appearing many duplications of code, need to give unique name for `resetOptionsForDeps`, write typescript typing for it (we move away them from react component to helper file) and etc.

I tried to solve this another way: in custom component I took `defaultValues` from `formState`, then I found default value by key like const `defaultValue = defaultValues[dependantName]` and this works, but unfortunatelly except for fields which were defined via `useFieldArray`: usually for this array-like fields we set default value as empty array (`[]`) in order to user add entity to array if he wants. Most of the cases, we keep objects in array-like fields - they represent some business entity (like identity document or others), so default value is `undefined`.

Here is example:
```
/** helper file*/
const docEntityDefaultValues = {
  dateFrom: '',
  dateTo: ''
};

/** high-level component */
const methods = useForm({
  defaultValues: {
    docs: [],
  }
});
const { fields,append } = useFieldArray({
  name: "docs"
})
...
return (
  <form>
    // fields mapping

    <button onClick={() => append(docEntityDefaultValues)}>
  </form>
);

/** select (custom component)*/
...
const {resetField} = useFormContext();

 <Controller
    name={name}
    render={({ field, formState }) => {
      const { defaultValues } = formState;

      return (
         <select onChange={(e) => {
           deps?.forEach((item, index) => {
             // current solution
             const  options = resetOptionsForDeps ? resetOptionsForDeps[index] : undefined;
             resetField(item, options);
            
             // solution which I try, but the problem with fieldArray occured
[...]

PR #12981, opened 2025-07-24
feat: Add defaultValue to getFieldState

evantrimboli:
Possibly resolves:

https://github.com/react-hook-form/react-hook-form/issues/12838

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -942,13 +942,21 @@ export function createFormControl<
   const getFieldState: UseFormGetFieldState<TFieldValues> = (
     name,
     formState,
-  ) => ({
-    invalid: !!get((formState || _formState).errors, name),
-    isDirty: !!get((formState || _formState).dirtyFields, name),
-    error: get((formState || _formState).errors, name),
-    isValidating: !!get(_formState.validatingFields, name),
-    isTouched: !!get((formState || _formState).touchedFields, name),
-  });
+  ) => {
+    const state = formState || _formState;
+    const error = get(state.errors, name);
+    return {
+      invalid: !!error,
+      isDirty: !!get(state.dirtyFields, name),
+      error,
+      isValidating: !!get(_formState.validatingFields, name),
+      isTouched: !!get(state.touchedFields, name),
+      defaultValue: get(
+        formState ? formState.defaultValues : _defaultValues,
+        name,
+      ),
+    };
+  };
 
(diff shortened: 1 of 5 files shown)


src/logic/createFormControl.ts
```diff
@@ -942,13 +942,21 @@ export function createFormControl<
...
+    const state = formState || _formState;
+    const error = get(state.errors, name);
+    return {
+      invalid: !!error,
+      isDirty: !!get(state.dirtyFields, name),
+      error,
+      isValidating: !!get(_formState.validatingFields, name),
+      isTouched: !!get(state.touchedFields, name),
+      defaultValue: get(
+        formState ? formState.defaultValues : _defaultValues,
```
evantrimboli, line 955: This code seems a little strange, it seems like it should be able to access `state.defaultValues` however `defaultValues` is never assigned to the `_formState` object.

It's assigned in two spots:

https://github.com/react-hook-form/react-hook-form/blob/343ac030cd00029c6aa0d245d01c6bb2646a303d/src/logic/createFormControl.ts#L133

https://github.com/react-hook-form/react-hook-form/blob/343ac030cd00029c6aa0d245d01c6bb2646a303d/src/logic/createFormControl.ts#L1324

Should it also be assigned to `_formState`?

evgeniyworkbel:
@evantrimboli hello! I am the author of #12838 and I notice that your PR doesn't resolve my feature-request fully, so the PR can be merged after code review and approvals, but my topic #12838 shouldn't be closed

My feature-req has the following description:

> I tried to solve this another way: in custom component I took defaultValues from formState, then I found default value by key like const defaultValue = defaultValues[dependantName] and this works, but unfortunatelly except for fields which were defined via useFieldArray: usually for this array-like fields we set default value as empty array ([]) in order to user add entity to array if he wants. Most of the cases, we keep objects in array-like fields - they represent some business entity (like identity document or others), so default value got as undefined.

My feature-req has the following expectation:
> Describe the solution you'd like
getFieldState API could return defaultValue with which the field was registered. Maybe getFieldState is not possible in current implementation (because returned obj from getFieldState is calculated based on defaultValues from formState), so maybe it needs to provide useFieldArray with own default values (for entity), create obj which will keep default values in respect that fieldArray's defaultValues (e.g. exhaustiveDefaultValues), calculate defaultValue and return it from getFieldState.

evantrimboli:
This is likely stale now. Will close it out.

PR #12985, opened 2025-07-25
fix: do not override prototype of `data` in `cloneObject`

Ty3uK:
I've encountered an issue while using `effectTsResolver` and `DateTime` structure in particular: in `handleSubmit` callback inside `formValues` my `dateTime` field always fails validation against `effect/Schema`.

After some digging I realized that the issue is in `cloneObject` util: the base object for `copy` is just a plain `{}` but in `effect` many data types heavily relies on prototypes.

For example the same `DateTime`: https://github.com/Effect-TS/effect/blob/main/packages/effect/src/internal/dateTime.ts#L203-L212

### DIFF
--- a/src/__tests__/utils/cloneObject.test.ts
+++ b/src/__tests__/utils/cloneObject.test.ts
@@ -166,4 +166,13 @@ describe('clone', () => {
       delete Array.prototype.somePolyfill;
     });
   });
+
+  it('should not override prototype of nested object', () => {
+    const UtcProto = {
+      _tag: 'Utc',
+    };
+    const dateTime = Object.create(UtcProto);
+    const copy = cloneObject(dateTime);
+    expect(copy._tag).toBe('Utc');
+  });
 });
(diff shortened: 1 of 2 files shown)


Ty3uK:
@bluebill1049, should I rewrite test to this?
```typescript
it('should not override prototype of nested object', () => {
  const UtcProto = {
    _tag: 'Utc',
  };
  const formValues = {
    dateTime: Object.create(UtcProto),
  };
  const copy = cloneObject(formValues);
  expect(Object.getPrototypeOf(copy.dateTime)).toEqual(UtcProto);
  expect(copy.dateTime._tag).toBe('Utc');
});
```

bluebill1049:
> @bluebill1049, should I rewrite test to this?
> 
> ```ts
> it('should not override prototype of nested object', () => {
>   const UtcProto = {
>     _tag: 'Utc',
>   };
>   const formValues = {
>     dateTime: Object.create(UtcProto),
>   };
>   const copy = cloneObject(formValues);
>   expect(Object.getPrototypeOf(copy.dateTime)).toEqual(UtcProto);
>   expect(copy.dateTime._tag).toBe('Utc');
> });
> ```

yes, please. we will get it in.

Ty3uK:
@bluebill1049, done :slightly_smiling_face:

bluebill1049 (approved):

PR #12986, opened 2025-07-25
🧿 feat: <Watch /> component

tesseractjh:
# Proposal

The `Watch` component is a syntax sugar of the `useWatch` hook.

## AS-IS

```tsx
import { useForm, useWatch } from 'react-hook-form';

const FooSubscriber = ({ control }) => {
  const foo = useWatch({ control, name: 'foo' });
  return <span>{foo}</span>;
};

const App = () => {
  const { register, control } = useForm();

  return (
    <div>
      <form>
        <input {...register('foo')} />
        <input {...register('bar')} />
      </form>
      {/* re-render only when value of `foo` changes */}
      <FooSubscriber control={control} />
    </div>
  );
};
```

When using `useWatch`, the entire component re-renders whenever the subscribed values change. To narrow the re-render scope, component separation was previously necessary, but with `Watch`, you can easily limit the re-render scope without separating components.

## TO-BE

```tsx
import { useForm, Watch } from 'react-hook-form';

const App = () => {
  const { register, control } = useForm();

  return (
    <div>
      <form>
        <input {...register('foo')} />
        <input {...register('bar')} />
      </form>
      {/* re-render only when value of `foo` changes */}
      <Watch
        control={control}
        names={['foo']}
        render={([foo]) => <span>{foo}</span>}
      />
    </div>
  );
};
```

- The `Watch` component subscribes to values of the paths provided through the `names` prop, and the `render` function receives the latest values as arguments.
- The types of `render` function's arguments are accurately inferred.


src/watch.tsx
```diff
@@ -0,0 +1,66 @@
...
+  names: TFieldNames;
+  children?:
+    | ReactNode
+    | ((values: GetValues<TFieldValues, TFieldNames>) => ReactNode);
+}) => {
+  const values = useWatch({ control, name: names });
+
+  return typeof children === 'function'
+    ? children(values as GetValues<TFieldValues, TFieldNames>)
+    : children;
```
bluebill1049: I am not sure we need to support children props here, we can probably stay the same as `src/controller.tsx` what do you think?

bluebill1049 (commented):
Nice one.

src/watch.tsx
tesseractjh: I agree as well. For consistency, I replaced it with a prop named `render`, just like the `Controller`.
[fix: replace children prop to render prop](https://github.com/react-hook-form/react-hook-form/pull/12986/commits/daa649e2b43b7d5089755ae6618a020149cc15b0)

bluebill1049 (approved):

bluebill1049:
getting it release to the next minor version, i will update the doc as well.

PR #13000, opened 2025-07-30
fix: use field name to update isValidating fields

### LINKED ISSUE #12999: issue: validatingFields is not properly tracking nested field validation state
### Version Number

7.61.1

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/2k2644

### Steps to reproduce

1. Create a form with a nested field structure
2. Register a field with async validation
3. Use trigger() to manually trigger validation
4. Check the validatingFields object during validation

When using the trigger() function to validate nested fields, the validatingFields object incorrectly tracks the validation state at the wrong path level. Instead of showing the validation state at the proper nested path (e.g., validatingFields.nested.field), it appears to be tracked at the root level using only the field name (e.g., validatingFields.field).

### Expected behaviour

When calling `trigger()` for validation, a nested field like `"nested.field"` should be in the `validatingFields` object like: 

`validatingFields.nested.field === true`

_No response_

### Relevant log output

```shell

```

### Code of Conduct

GorkemKir:
To resolve #12999

Hi Bill,

### Issue 
Updating validatingFields state is not properly handling nested field names when called via `trigger()`.

### Solution
Using `_f.name` to update `isValidating` state

### DIFF
--- a/src/__tests__/useForm/trigger.test.tsx
+++ b/src/__tests__/useForm/trigger.test.tsx
@@ -9,7 +9,13 @@ import {
 } from '@testing-library/react';
 
 import { VALIDATION_MODE } from '../../constants';
-import type { Control, FieldPath } from '../../types';
+import type {
+  Control,
+  FieldPath,
+  FieldValues,
+  FormState,
+  UseFormGetFieldState,
+} from '../../types';
 import { useController } from '../../useController';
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks for the pr, can you include some test coverage?

bluebill1049 (approved):

PR #13008, opened 2025-08-03
chore: upgrade typescript-eslint

abnud11:
This PR upgrades typescript-eslint from 7 to 8, and a lot of linting errors happened after that.
I turned some of them off since they didn't catch any meaningful bugs in my opinion, but I'd like to hear your opinion on the matter.


src/logic/createFormControl.ts
```diff
@@ -1257,7 +1259,7 @@ export function createFormControl<
           errors: {},
         });
         try {
-          await onValid(fieldValues as TTransformedValues, e);
+          await onValid(fieldValues as unknown as TTransformedValues, e);
```
bluebill1049: what's the reason of needing as unknow now?

bluebill1049 (commented):
Nice work just one comment.

src/logic/createFormControl.ts
abnud11: It was related to `ResolverResult`, where the value was sometimes typed as `object`. I changed its type, and since' unknown' is no longer needed, it's unnecessary.

bluebill1049 (approved):
lgtm!

PR #13041, opened 2025-08-29
fix: unregister previous field when switching conditional Controllers

### LINKED ISSUE #13039: issue: Field values leak across conditional renders when Controllers share the same render position
### Version Number

7.62.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/rk-react-hook-form-issue-forked-gt9w5s

### Steps to reproduce

1. Render a form with two conditional modes (e.g., Step A and Step B).
2. In Step A, enter a value in a field.
3. Switch to Step B, which renders a different field in the same position with a different name.
4. Observe that the value from Step A incorrectly appears in Step B’s field.

### Expected behaviour

When conditionally rendering forms, field values leak across modes if Controllers occupy the same render position. This happens even when the fields have different names and validation rules.

Expected behavior:

Switching between conditional form modes should not carry over values from one mode to another when the field names are different.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

m2na7:
fix #13039

This PR addresses the issue where field values could leak between Controllers rendered conditionally at the same position.

`useController` now tracks the previous field name and automatically unregisters it when the name changes, preventing removed fields from lingering in the form state. Field Array functionality is preserved.

If users prefer handling this with explicit `key` props, this PR can be closed! 🙂

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -1246,4 +1246,61 @@ describe('useController', () => {
 
     expect(renderCounter).toEqual({ test: 3, test_with_suffix: 3 });
   });
+
+  it('should prevent field value leakage when field names change at same position', () => {
+    type FormValues = {
+      type: 'personal' | 'business';
+      personalName: string;
+      businessName: string;
+    };
+
+    const Component = () => {
+      const { control, watch, setValue } = useForm<FormValues>({
+        defaultValues: {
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):

bluebill1049:
Thanks for the contribution and pr.

PR #13056, opened 2025-09-13
fix: only excute trigger fuction when deps has a valid array

### LINKED ISSUE #13040: issue: Having any deps array causes subscribe() to fire error updates on every key stroke
### Version Number

7.62.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/lctvk3

### Steps to reproduce

1. Go to https://codesandbox.io/p/sandbox/lctvk3
2. Cause and clear errors by typing something in "First name" input and then emptying the field
3. Subscription to errors formState fires only when error changes (field goes from empty => not empty OR from not empty => empty)
4. Type anything in "Last name" input
5. Subscription to errors formState fires on every input, regardless of if an error was caused/cleared or not

### Expected behaviour

subscribe() should only fire updates, if the subscribed formState changes

Chrome

### Relevant log output

```shell

```

### Code of Conduct

candymask0712:
To resolve #13040

Hello, Bill 😁
This PR includes a small fix.

### Issue

When `deps` is an empty array, errors keep updating on every key input.

### Cause

An empty array is still evaluated as truthy, which caused trigger to run on every `onChange` event.

### Solution

Updated the logic so that `trigger` function inside `onChange` only executes when the dependency array contains elements.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -858,6 +858,7 @@ export function createFormControl<
 
       if (isFieldValueUpdated) {
         field._f.deps &&
+          (!Array.isArray(field._f.deps) || field._f.deps.length > 0) &&
           trigger(
             field._f.deps as
               | FieldPath<TFieldValues>


bluebill1049 (approved):
LGTM

PR #13063, opened 2025-09-24
fix: preserve Controller's defaultValue

### LINKED ISSUE #12670: issue:  🐛  Breaking Changes in getValues() Behavior Between 7.54.1 and 7.54.2
### Version Number

7.54.2

#### **Describe the bug**  
After upgrading from `7.54.1` to `7.54.2`, the behavior of `getValues()` has changed when using `shouldUnregister: true` in conjunction with `FormProvider` and `Controller`.  

Previously, `getValues()` would return the `defaultValue` defined inside the `Controller`.However, in `7.54.2`, `defaultValue` is now `undefined`.  
___

#### **Steps to reproduce**  
1. Go to: [Codesandbox](https://codesandbox.io/p/sandbox/react-hook-form-v7-controller-ts-forked-2yqvs2?file=%2Fsrc%2Findex.tsx%3A23%2C5)
2. Click on ‘getValues’
3. Check console
4. Change version to `7.54.1`
5. Repeat step 2 and 3

___

#### **Code Sandbox / Reproducible Example**  
```jsx
import React from "react";
import { createRoot } from "react-dom/client";
import {
  useForm,
  Controller,
  useFormContext,
  FormProvider,
} from "react-hook-form";

const Input = () => {
  const { control } = useFormContext();

  return (
    <Controller
      control={control}
      name="name"
      defaultValue="default-value-provided"
      render={({ field: { onChange, onBlur, value, ref } }) => (
        <input onChange={onChange} onBlur={onBlur} value={value} />
      )}
    />
  );
};

function App() {
  const formMethods = useForm({
    shouldUnregister: true,
  });

  return (
    <main>
      <FormProvider {...formMethods}>
        <form>
          <Input />
        </form>

        <br />

        <button onClick={() => console.log(formMethods.getValues())}>
          getValues
        </button>
      </FormProvider>
    </main>
  );
}

const container = document.getElementById("root")!;
const root = createRoot(container);

root.render(
  <React.StrictMode>
    <App />
  </React.StrictMode>
);
```
___

### Expected behaviour

- In version **7.54.1**, calling `getValues()` returns the provided `defaultValue`.  
- In version **7.54.2**, `defaultValue` is now `undefined`, which seems to be an unintended breaking change.  

Chrome

### Relevant log output

```shell

```

### Code of Conduct

### LINKED ISSUE #12798: issue: when shouldUnregister is true, defaultValue in Controller always undefined
### Version Number

7.56.1

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/kind-poincare-fxfhcd

### Steps to reproduce

1. Open the sandbox: https://codesandbox.io/p/sandbox/kind-poincare-fxfhcd
2. Hit submit -> the submitted form does not contains the default value
3. try comment the `shouldUnregister` property
4. Hit submit -> the submitted form contains the default value

### Expected behaviour

when `shouldUnregister` is true, it should not affects the behavior of the the mounted inputs component, `defaultValue` in Controller should register the value

Chrome, Firefox, Safari

### Relevant log output

```shell

```

### Code of Conduct

gynekolog:
When `shouldUnregister` is true, the field restoration logic was only considering useForm's `defaultValues` but ignoring the Controller's defaultValue prop. 
This caused the `defaultValue` to be lost with React.Strict Mode.

close https://github.com/react-hook-form/react-hook-form/issues/12670
close https://github.com/react-hook-form/react-hook-form/issues/12798

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -833,24 +833,22 @@ describe('useController', () => {
       const { isValid } = form.formState;
 
       return (
-        <React.StrictMode>
-          <FormProvider {...form}>
-            <Controller
-              render={({ field }) => (
-                <input value={field.value} onChange={field.onChange} />
-              )}
-              name="name"
-              rules={{
-                required: true,
-              }}
-            />
-            <p>{isValid ? 'valid' : 'not'}</p>
-          </FormProvider>
-        </React.StrictMode>
+        <FormProvider {...form}>
+          <Controller
+            render={({ field }) => (
+              <input value={field.value} onChange={field.onChange} />
+            )}
+            name="name"
+            rules={{
+              required: true,
+            }}
+          />
+          <p>{isValid ? 'valid' : 'not'}</p>
+        </FormProvider>
       );
     };
 
-    render(<App />);
+    render(<App />, { reactStrictMode: true });
 
     await waitFor(() => {
       screen.getByText('not');
@@ -879,18 +877,16 @@ describe('useController', () => {
       } = methods;
 
       return (
-        <React.StrictMode>
-          <FormProvider {...methods}>
-            <form>
-              <Form />
-              {dirtyFields.lastName ? 'dirty' : 'pristine'}
-            </form>
-          </FormProvider>
-        </React.StrictMode>
+        <FormProvider {...methods}>
+          <form>
+            <Form />
+            {dirtyFields.lastName ? 'dirty' : 'pristine'}
+          </form>
+        </FormProvider>
       );
     }
 
-    render(<App />);
+    render(<App />, { reactStrictMode: true });
 
     fireEvent.change(screen.getByRole('textbox'), {
       target: {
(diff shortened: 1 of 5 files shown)


gynekolog:
Oh, and I removed the other React.StrictMode wrappers from the rendered components and instead used the [render options with reactStrictMode](https://testing-library.com/docs/react-testing-library/api/#render-options-reactstrictmode).

gynekolog:
I’m not sure why the issue started in version 7.54.2.

If you switch the repo to v7.54.1 and run the tests, they all pass. However, tests using React.StrictMode don’t behave correctly, because React.StrictMode should be applied in the render function, not inside the component.

You can try it yourself: open `useController.test.tsx` in v7.54.1, find the test called `should restore defaultValues with react strict mode double useEffect`, move `React.StrictMode` from the App component into the render function, and rerun the tests. The test will fail.

With the latest version (7.63.0), though, the same test passes even after moving/fixing React.StrictMode. Why? I don’t know. But if you remove [this changed line from v7.54.2](https://github.com/react-hook-form/react-hook-form/pull/12480/files), the test still pass.

bluebill1049 (approved):

Issue #13009, opened 2025-08-03
Compatibility with Lenses


aspirisen:
Hello 👋

While working on the [**Lenses** project](https://github.com/react-hook-form/lenses), I’ve encountered several challenges that would benefit from improvements in the core `react-hook-form` library. 
---

### 1. **Allow [Control](https://github.com/react-hook-form/react-hook-form/blob/master/src/types/form.ts#L814) to Accept Generic Value Types**

Currently, the [`Control`](https://github.com/react-hook-form/react-hook-form/blob/master/src/types/form.ts#L814) type requires both a value type and a path. This presents a problem when we don't have a specific path, which is often the case when working with generic or path-agnostic components.

#### Problem

We’d like to support simple usage like:

```ts
const { register } = useForm<string>();
return <input {...register()} />;
```

But `Control<T, Path>` requires a valid path type—so passing `string` or `array` causes type errors. To work around this, we created a shim:

```ts
Control<
  HookFormControlShim<{ formValues: { etc: {} } }>,
  '__DO_NOT_USE_HOOK_FORM_CONTROL_SHIM__'
>
```

#### How the Shim Works

* `HookFormControlShim` is a helper type that wraps the actual type you want to work with in a known structure.
* `__DO_NOT_USE_HOOK_FORM_CONTROL_SHIM__` is a constant string literal used as a dummy path. It always matches the single known key inside the shim object.

Here’s a simplified version:

```ts
type HookFormControlShim<T> = { __DO_NOT_USE_HOOK_FORM_CONTROL_SHIM__: T };
```

This allows us to "trick" the type system into accepting any generic type by wrapping it in a known object shape with a fixed path key. Then, immediately, we unwrap it when passing the path to the form methods.

#### Suggestion

Allow the `Control<T>` type to accept arbitrary generic types without constraints and requiring a valid `Path<T>` up front. This would eliminate the need for `HookFormControlShim` and enable straightforward usage in path-agnostic code.

---

### 2. **Support Transformed Controllers**

Lenses uses a `reflect` method that can remap or restructure values, requiring transformations during value access and update.

Currently:

* I created a wrapper around [useFieldArray](https://github.com/react-hook-form/lenses/blob/main/src/rhf/useFieldArray.ts) that performs value transformation.
[...]

PR #13070, opened 2025-09-30
🎥  feat: make `useWatch` and `useController` to react to `name` change

aspirisen:
Make `useController` and `useWatch` to react to `name` and `control` change

Closes: https://github.com/react-hook-form/react-hook-form/issues/12974
Related: https://github.com/react-hook-form/react-hook-form/issues/13009


bluebill1049:
e2e test breaks

aspirisen:
@bluebill1049 this looks like an CI issue
image

bluebill1049:
Have you tried locally? I don't see this issue on other prs.

aspirisen:
Locally they work
image
Looks like on CI the tests are not event started, some problem with cypress installation

bluebill1049:
interesting i will check over the weekend thanks @aspirisen for the pr!

src/useWatch.ts
```diff
@@ -269,56 +269,100 @@ export function useWatch<TFieldValues extends FieldValues>(
...
 
-  return value;
+  // If name or control changed for this render, synchronously reflect the
+  // latest value so callers (like useController) see the correct value
+  // immediately on the same render.
+
+  // Optimize: Check control reference first before expensive deepEqual
+  const controlChanged = _prevControl.current !== control;
+  const nameChanged = !controlChanged && !deepEqual(_prevName.current, name);
+  const shouldReturnImmediate = controlChanged || nameChanged;
```
bluebill1049: I think we can bring this as part of the following `memo`, probably only need the controlChanged to be left out.

src/useWatch.ts
```diff
@@ -269,56 +269,100 @@ export function useWatch<TFieldValues extends FieldValues>(
...
-    [control, name],
-  );
+  const [value, updateValue] = React.useState(() => {
+    const defaultValue = control._getWatch(
+      name as InternalFieldName,
+      _defaultValue.current as DeepPartialSkipArrayKey<TFieldValues>,
+    );
+
+    return _compute.current ? _compute.current(defaultValue) : defaultValue;
+  });
```
bluebill1049, line 284: can we memo this default state?

aspirisen: As I understand the default value should be set only once on initial rendering, with the callback in use state it has been already memorized 

With useMemo we would need to set deps and therefore change memo on that deps change, what it strange since we are using defaultValue only once and we are not going to react and recalculate the default values change

bluebill1049: but it get re-run each render? may as well as skip that.

aspirisen: no, that function will be called only once to get initial state, on subsequent renders the passed function will not be called

bluebill1049: oh yea you are right, it's a callback

bluebill1049 (approved):

Issue #13017, opened 2025-08-07
issue: isDirty form state is not updated without subscription


Song1201:
### Version Number

7.61.1

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/vigilant-faraday-lhxw7f

### Steps to reproduce

1. Enter anything in the input textbox, like 'abc'
2. Click "Show page 1" button to show page 1
3. `isDirty` is `false` in `Page1` component
4. Click "rerender page 1" button, page 1 is rerendered while we already subscribed to `isDIrty`, but `isDirty` remains `false`
5. type anything in the textbox again, `isDirty` become `true`
6. Now try to subscribe to `isDirty` before typing in the textbox, by updating the code to deconstruct `formState: {isDirty}` in `App.tsx`
7. Reload the UI, repeat the above steps and `isDirty` is `true` upon first render of page 1

### Expected behaviour

Page 1 has `isDirty` `true` without needing to subscribe to it in `App.tsx`, since we're not even using `isDirty` there.

I understand there is a sentence in the [form state](https://react-hook-form.com/docs/useform/formstate) doc. "Therefore make sure you invoke or read it before a render in order to enable the state update.", but it's kind of unintuitive to read the value in a component while we're not using it there. Reading a value and never use it, will also be marked as error by some linter rules.

I tried to look into the code, and saw [this](https://github.com/react-hook-form/react-hook-form/blob/1b5a6748a83bbaf3deed4092ec631d6cda927bf7/src/useForm.ts#L145) where we evaluate `_getDirty()` if it's subscribed, so I'm not sure why we won't get `isDirty: true` in `Page1` component. 

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bluebill1049:
if you didn't subscribe isDirty or other formState, RHF will just skip the computation and that's expected to save additional computation.

PR #13072, opened 2025-10-02
fix: resolve late subscription issue for formState properties(isDirty)

shunn2:
This fixes the issue where components subscribing to formState properties (like isDirty, dirtyFields) after form values have changed would receive stale state values.

Problem:
- Issue: [https://github.com/react-hook-form/react-hook-form/issues/13017](https://github.com/react-hook-form/react-hook-form/issues/13017)

Solution:
Always keep _formState.isDirty up-to-date regardless of subscription status in updateTouchAndDirty()

Fixes: #[issue-number]

### DIFF
--- a/src/__tests__/logic/createFormControl.test.ts
+++ b/src/__tests__/logic/createFormControl.test.ts
@@ -62,4 +62,68 @@ describe('createFormControl', () => {
 
     expect(isEmptyObject).toHaveBeenCalledTimes(3);
   });
+
+  describe('isDirty state updates without subscription', () => {
+    it('should keep _formState.isDirty up-to-date even when not subscribed', () => {
+      const { register, control } = createFormControl({
+        defaultValues: {
+          name: '',
+        },
+      });
+
+      expect(control._proxyFormState.isDirty).toBe(false);
+
+      const ref = document.createElement('input');
+      ref.name = 'name';
+      const { onChange } = register('name', {});
+      onChange({ target: { name: 'name', value: '' } });
+      register('name', {}).ref(ref);
+
+      expect(control._formState.isDirty).toBe(false);
+
+      onChange({ target: { name: 'name', value: 'John' } });
+
+      expect(control._formState.isDirty).toBe(true);
+      expect(control._proxyFormState.isDirty).toBe(false);
+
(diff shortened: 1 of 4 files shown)


bluebill1049:
Could you include some unit tests for this change?

shunn2:
> Could you include some unit tests for this change?

07321368bd4f2f6ca28aefe2d7f54dafa8a1491a
I added a test for that change. If you need anything else, please let me know. Thanks. 😊

image

bluebill1049:
thanks, let me discuss this issue on the issue thread.

shunn2:
Hi @bluebill1049 I checked the your comment on the issue.
totally agree that skipping computation when not subscribed is important for performance.
What do you think about syncing `isDirty` internally based on `dirtyFields` instead?  
It wouldn’t call `_getDirty()` or add any extra work, but would keep the internal state consistent so late subscribers still get the correct value without extra cost.

bluebill1049:
> Hi @bluebill1049 I checked the your comment on the issue. totally agree that skipping computation when not subscribed is important for performance. What do you think about syncing `isDirty` internally based on `dirtyFields` instead? It wouldn’t call `_getDirty()` or add any extra work, but would keep the internal state consistent so late subscribers still get the correct value without extra cost.

sounds good 👍

shunn2:
> > Hi @bluebill1049 I checked the your comment on the issue. totally agree that skipping computation when not subscribed is important for performance. What do you think about syncing `isDirty` internally based on `dirtyFields` instead? It wouldn’t call `_getDirty()` or add any extra work, but would keep the internal state consistent so late subscribers still get the correct value without extra cost.
> 
> sounds good 👍

thanks for review 

commit: e7d3fcad4f017416eee2590f28d64ad5aaaad55b

This Commit proposes syncing isDirty internally based on dirtyFields to keep form state consistent for late subscriptions.
It doesn’t call _getDirty() or introduce extra computation when isDirty isn’t subscribed — maintaining the existing performance behavior while ensuring correct state visibility for new subscribers.

shunn2:
@bluebill1049 Hi, let me know if there’s anything else you’d like me to update!

bluebill1049:
it's a risky change atm, going to close for now.

PR #13079, opened 2025-10-06
Fix(getDirtyFields): submit fields with null values when using useForm

### LINKED ISSUE #12815: issue: fields with value `null` are no longer submitted when using `useForm({ values })`
### Version Number

7.56.2

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/usecontroller-forked-cyj47p

### Steps to reproduce

1. Open browser DevTools to see the console
2. Submit the form
3. See submitted data: `{}` (empty object)

### Expected behaviour

Expected submitted data should have been `{ views: null }`

## Additional Info

This seems to be a regression brought by version 7.55.0.

The same sandbox in version 7.54.2 shows the expected behavior: https://codesandbox.io/p/devbox/usecontroller-forked-gnn7m2

It's also interesting to note that the issue only happens when using `useForm({ values: {} })`.
Using `useForm({ defaultValues: {} })` instead does not show the issue.

## Use case context

We are building a `NumberInput` component and would like this component to submit a `null` value in the form when the input is not filled.
Hence we set `defaultValue: null` when calling `useController`.

We can control the parsing of the value by overriding `onChange`, but we need to be able to set the value to `null` by default so that it submits `null` also when the input is empty and untouched.

Firefox

### Relevant log output

```shell

```

### Code of Conduct

jonathanarnault:
## Problem

Fixes #12815

Fields with `null` values are not submitted when using `useForm({ values })` since v7.55.0.

**Before:**
```typescript
const { control, handleSubmit } = useForm({ values: { views: null } });
```
On submit: `{}` (null value is omitted)

**Expected:**
On submit: `{ views: null }`

**Root Cause:**
In `src/logic/getDirtyFields.ts`, the `markFieldsDirty` function used `!isNullOrUndefined(data[key])` which excluded `null` values from being marked as dirty, causing them to be omitted during form submission.

## Solution

Changed the condition to only exclude `undefined` values:

```typescript
} else if (!isUndefined(data[key])) {
  fields[key] = true;
}
```

This allows `null` to be treated as a valid field value while still excluding `undefined` fields.

## Testing

Added 8 tests:
- 2 unit tests in `src/__tests__/logic/getDirtyFields.test.ts`
- 6 integration tests in `src/__tests__/useForm/useFormWithNullValues.test.tsx`

All 968 tests pass.

## Changes

- `src/logic/getDirtyFields.ts`: Changed condition in `markFieldsDirty`
- `src/__tests__/logic/getDirtyFields.test.ts`: Added unit tests
- `src/__tests__/useForm/useFormWithNullValues.test.tsx`: Added integration tests

Backward compatible, no breaking changes.

### DIFF
--- a/src/__tests__/logic/getDirtyFields.test.ts
+++ b/src/__tests__/logic/getDirtyFields.test.ts
@@ -341,4 +341,39 @@ describe('getDirtyFields', () => {
       ],
     });
   });
+
+  it('should include fields with null values', () => {
+    expect(
+      getDirtyFields(
+        {},
+        {
+          views: null,
+          name: 'test',
+          count: 0,
+        },
+      ),
+    ).toEqual({
+      views: true,
+      name: true,
+      count: true,
+    });
+  });
+
+  it('should mark null values as dirty when comparing with defaultValues', () => {
+    expect(
+      getDirtyFields(
+        {
+          views: null,
+          name: 'test',
+        },
+        {
+          views: null,
+          name: 'other',
+        },
+      ),
+    ).toEqual({
+      views: false,
+      name: true,
+    });
+  });
 });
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):
LGTM

PR #13083, opened 2025-10-08
🐞 fix: respect parent-provided useFieldArray rules (#13082)

### LINKED ISSUE #13082: issue: Regression useFieldArray rule attribute not being respected when passed from parent
### Version Number

7.64.0

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/react-hook-form-usefieldarray-rules-forked-9lgp8m?workspaceId=ws_SpFfrNeSXw3k7QXR2peBdn

### Steps to reproduce

 I am upgrading from 7.52.1 to 7.63.0 (also verified regression with 7.64.0) and I have an existing scenario that is now broken:

1. Go to [this codesandbox](https://codesandbox.io/p/devbox/react-hook-form-usefieldarray-rules-forked-9lgp8m?workspaceId=ws_SpFfrNeSXw3k7QXR2peBdn)
2. Either delete all field-rows or append until 4 is reached then delete one to make the form invalid
3. Click submit
4. Console will log "validSubmit" (first param in handleSubmit)

### Expected behaviour

onInvalidSubmit should be called as the form does not pass the rules provided by the parent.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

Han5991:
⏺ Summary

  fixes #13082 

  Problem

  When rules were passed to useFieldArray from a parent component, they were wrapped in React.useMemo without proper dependencies, causing the rules to be registered only once during initial render. If the rules object reference changed (e.g., created inline or from useMemo in parent), the validation would not update
  properly.

  Solution

  - Removed unnecessary React.useMemo wrapper around the control.register call in src/useFieldArray.ts:112-118

  Changes

  - src/useFieldArray.ts: Simplified rules registration by removing useMemo wrapper
  - src/tests/useFieldArray.test.tsx: Added comprehensive test case validating that rules from parent component are properly enforced

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -112,7 +112,7 @@ export function useFieldArray<
   React.useMemo(
     () =>
       rules &&
-      fields.length &&
+      fields.length >=0 &&
       (control as Control<TFieldValues, any, TTransformedValues>).register(
         name as FieldPath<TFieldValues>,
         rules as RegisterOptions<TFieldValues>,
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -111,7 +111,8 @@ export function useFieldArray<
 
   React.useMemo(
     () =>
-      rules && fields.length &&
+      rules &&
+      fields.length &&
       (control as Control<TFieldValues, any, TTransformedValues>).register(
         name as FieldPath<TFieldValues>,
         rules as RegisterOptions<TFieldValues>,
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -112,7 +112,7 @@ export function useFieldArray<
   React.useMemo(
     () =>
       rules &&
-      fields.length >=0 &&
+      fields.length >= 0 &&
       (control as Control<TFieldValues, any, TTransformedValues>).register(
         name as FieldPath<TFieldValues>,
         rules as RegisterOptions<TFieldValues>,
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -111,12 +111,11 @@ export function useFieldArray<
 
   React.useMemo(
     () =>
-      rules &&
+      rules && fields.length &&
       (control as Control<TFieldValues, any, TTransformedValues>).register(
         name as FieldPath<TFieldValues>,
         rules as RegisterOptions<TFieldValues>,
       ),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
     [control, name, fields.length, rules],
   );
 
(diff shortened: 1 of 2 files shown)


src/useFieldArray.ts
```diff
@@ -109,15 +109,11 @@ export function useFieldArray<
...
-        name as FieldPath<TFieldValues>,
-        rules as RegisterOptions<TFieldValues>,
-      ),
-    [control, rules, name],
-  );
+  rules &&
+    (control as Control<TFieldValues, any, TTransformedValues>).register(
+      name as FieldPath<TFieldValues>,
+      rules as RegisterOptions<TFieldValues>,
+    );
```
bluebill1049: could we try to use rules with ref instead? would be good if we can still keep the memo.

```tsx
const rulesRef = useRef(rules)

reulsRef.current = rules
```

bluebill1049 (commented):
thanks for the PR!

src/useFieldArray.ts
Han5991:
```tsx
  const rulesRef = React.useRef(rules);

  rulesRef.current = rules;

  control._names.array.add(name);

  rulesRef.current &&
    (control as Control<TFieldValues, any, TTransformedValues>).register(
      name as FieldPath<TFieldValues>,
      rulesRef.current as RegisterOptions<TFieldValues>,
    );
```
`rulesRef` as suggested. This keeps the latest `rules` value in a ref while preserving memoization. Test passes. Does this look right?

bluebill1049:
```tsx
const rulesRef = React.useRef(rules);

  rulesRef.current = rules;

  const _actioned = React.useRef(false);

  control._names.array.add(name);

  React.useMemo(
    () =>
      rules &&
      (control as Control<TFieldValues, any, TTransformedValues>).register(
        name as FieldPath<TFieldValues>,
        rulesRef.current as RegisterOptions<TFieldValues>,
      ),
    [control, name],
  );
  ```
  
  what about this?

Comment (Han5991) [reply src/useFieldArray.ts]: image

re-ran the test suite, but it still fails.

Comment (Han5991) [reply src/useFieldArray.ts]: image

```tsx
  React.useMemo(
    () =>
      rules &&
      (control as Control<TFieldValues, any, TTransformedValues>).register(
        name as FieldPath<TFieldValues>,
        rules as RegisterOptions<TFieldValues>,
      ),
    // eslint-disable-next-line react-hooks/exhaustive-deps
    [control, name, fields.length, rules],
  );
```
What do you think about this code?

bluebill1049 (approved):
ty very much for the fix

PR #13091, opened 2025-10-10
fix: watch() returning undefined immediately after reset() - Issue #13088

scato3:
Set _state.mount to true in _reset function to ensure watch() returns the correct value immediately after reset() is called.

Previously, when reset() was called without any prior field interaction, _state.mount remained false, causing _getWatch to return _defaultValues instead of the updated _formValues.

- Added _state.mount = true after updating _formValues in _reset
- Added test case to verify watch() returns updated value after reset()
- Updated existing test expectation to match new behavior

Fixes #13088

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1391,12 +1391,6 @@ export function createFormControl<
           : ({} as TFieldValues)
         : (cloneObject(values) as TFieldValues);
 
-      _state.mount =
-        !_proxyFormState.isValid ||
-        !!keepStateOptions.keepIsValid ||
-        !!keepStateOptions.keepDirtyValues ||
-        (!_options.shouldUnregister && !isEmptyObject(values));
-
       _subjects.array.next({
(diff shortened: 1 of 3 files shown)


src/logic/createFormControl.ts
```diff
@@ -1410,11 +1412,6 @@ export function createFormControl<
       focus: '',
     };
 
-    _state.mount =
-      !_proxyFormState.isValid ||
-      !!keepStateOptions.keepIsValid ||
-      !!keepStateOptions.keepDirtyValues;
```
bluebill1049, line 1416: This would probably lead to some regression on user don't want to run

- update dirty values
- valid state

scato3: I've addressed your feedback and made the changes. Could you please take a look?

  The fix now preserves the original 3 conditional mount checks and adds an additional condition specifically for the reset

bluebill1049 (approved):
LGTM

PR #13103, opened 2025-10-17
fix(form-state, reset): programmatic dirty in disabled forms; treat reset({}) as explicit values (#13100, #13099)

ankit1324:
For issue #13100 (programmatic dirty in disabled form)

## Summary

**Problem**: Calling setValue(..., { shouldDirty: true }) while the form is disabled didn’t update fieldState.isDirty or formState.isDirty.

**Root cause**: Both updateTouchAndDirty and _getDirty short-circuited when _options.disabled was true, so programmatic updates couldn’t affect dirty state.

## Fix:

In updateTouchAndDirty, allow programmatic updates to set dirty even when disabled (detect shouldDirty === true).

In _getDirty, remove the disabled short-circuit so form dirty recalculation compares current values vs defaults regardless of disabled state.

Behavior (before → after):

Before: setValue('foo', 'x', { shouldDirty: true }) on a disabled form → isDirty stayed false.

After: Same call → fieldState.isDirty and formState.isDirty become true. User input is still blocked by disabled.

Tests: Added/updated tests to cover disabled + programmatic updates; all existing tests pass.

Backward compatibility: No breaking changes; only affects programmatic updates when shouldDirty: true.

Closes: #13100

For issue #13099 (reset({}) vs reset())

## Summary

**Problem**: reset({}) behaved like reset() and fell back to default values, forcing a “double reset” workaround.

**Root cause**: A truthiness/empty-object check treated undefined (no args) and {} (explicit empty) the same.

## Fix: 

Distinguish no argument from empty object so reset({}) uses {} as the new values.

Behavior (before → after):

Before: reset({}) → values restored from defaults.

After: reset({}) → values cleared to {} (as provided).

Tests: Added tests for reset({}) semantics; updated the one that relied on the old behavior; all pass.

Backward compatibility: No breaking changes; clarifies intended semantics.

Closes: #13099

### DIFF
--- a/src/__tests__/useForm/reset.test.tsx
+++ b/src/__tests__/useForm/reset.test.tsx
@@ -1509,13 +1509,13 @@ describe('reset', () => {
     });
   });
 
-  it('should return defaultValues in useWatch and watch when using calling reset with empty object', async () => {
+  it('should reset to empty values when reset({}) is called, not default values', async () => {
     const defaultValues = {
       something: 'anything',
     };
 
     function App() {
-      const { control, reset, register, watch } = useForm({
+      const { control, reset, register, watch, getValues } = useForm({
         defaultValues,
       });
       const watchValue = watch('something');
@@ -1537,6 +1537,7 @@ describe('reset', () => {
           </button>
           <p>watch: {watchValue}</p>
           <p>useWatch: {useWatchValue}</p>
+          <p>getValues: {JSON.stringify(getValues())}</p>
         </form>
       );
     }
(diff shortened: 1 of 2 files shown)


bluebill1049:
worth to check if latest already fixed this issue, if not please open issue again.

ankit1324:
Hi @bluebill1049 , I checked against the latest `master` (af84a5d0, after v7.65.0) and both issues this PR addressed still reproduce:

**#13099 — `reset({})` falls back to `defaultValues` instead of clearing values**

```tsx
const { result } = renderHook(() =>
  useForm({ defaultValues: { firstName: 'default' } }),
);

act(() => result.current.setValue('firstName', 'changed'));
act(() => result.current.reset({}));

result.current.getValues();
// expected: {}
// actual:   { firstName: 'default' }
```

**#13100 — `setValue(..., { shouldDirty: true })` doesn't mark dirty when the form is disabled**

```tsx
const { result } = renderHook(() =>
  useForm({ defaultValues: { foo: 'bar' }, disabled: true }),
);

act(() => result.current.register('foo'));
act(() => result.current.setValue('foo', 'x', { shouldDirty: true }));

result.current.getFieldState('foo').isDirty; // expected: true, actual: false
result.current.formState.isDirty;            // expected: true, actual: false
```

Since I'm not the author of #13099 or #13100 I can't reopen them myself — could you reopen them, or would you prefer I file fresh issues?

bluebill1049:
Prefer to fire a new issue and I will double check if this is expected behaviour or issue

PR #13108, opened 2025-10-21
fix(Watch): correct render function parameter typing

### LINKED ISSUE #13105: issue: Watch component render function parameter typed as never
### Version Number

7.65.0

### Codesandbox/Expo snack

https://www.typescriptlang.org/play/?ssl=21&ssc=1&pln=2&pc=1#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgIilQ3wG4AoUSWOAbzgHUUY0ALAGjgFcBnJAMWh4AvnGy4CRdDAC0rCBADWMzELLlKAOxhIomdEjiCoIAGooANlyQ865OGIUAuODxhRgmgOYVhGzFyaGMAQmnAAwriQmkjaABQAlHYOaKFudHCp2jgWcKIAvNx8xiAAPCXmVjYAfIkU9nBEMFxQYXENDqUAJsAAbtUdDnClTCysg0OZoe4QFvm0WTMWfpOTmiggNvMA2viqEPgAuiurDkSaXbrzcdv7h0n51cM9-bT7wqUA9C-VJ5OfA0mXx+DQSvnIQA

### Steps to reproduce

1. Use the new Watch component with TypeScript as shown:

```tsx
import { Watch, useForm } from "react-hook-form";

interface FormValues {
  foo: string;
}

function Component() {

  return (
    <div>
      <Watch
        control={control}
        names={["foo"]}
        render={([foo]) => <div>{foo}</div>}
      />
    </div>
  );
}
```

2. Observe that the parameter passed to the render function is typed as `never`, instead of the expected type (supposedly `[string]`).

3. We can notice that the property `names` is typed as `"foo"[]` instead of `["foo"]`. Suffixing with `as const` fixes the issue.

```ts
names={["foo"] as const}
```

### Expected behaviour

The parameter of the render function should be correctly typed based on the `names` prop (e.g., as a tuple of the types of the watched fields), not `never`. This would allow TypeScript to infer the correct types for the watched values.

### Code of Conduct

dusan233:
fixes #13105

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -935,7 +935,7 @@ export type ValidationValueMessage<TValidationValue extends ValidationValue = Va
 };
 
 // @public
-export const Watch: <TFieldNames extends readonly FieldPath<TFieldValues>[], TFieldValues extends FieldValues = FieldValues, TContext = any, TTransformedValues = TFieldValues>({ control, names, render, }: WatchProps<TFieldNames, TFieldValues, TContext, TTransformedValues>) => ReactNode;
+export const Watch: <const TFieldNames extends readonly FieldPath<TFieldValues>[], TFieldValues extends FieldValues = FieldValues, TContext = any, TTransformedValues = TFieldValues>({ control, names, render, }: WatchProps<TFieldNames, TFieldValues, TContext, TTransformedValues>) => ReactNode;
 
 // @public (undocumented)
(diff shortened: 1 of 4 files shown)


bluebill1049:
could you include some tests?

dusan233:
> could you include some tests?

I moved the type assertions into a `.test-d.ts` file. Previously they existed in `watch.test.tsx` but didn’t run because `expectType` was used outside a `.test-d.ts` file.

bluebill1049 (approved):

mamlzy:
@bluebill1049 Could you publish this pr? this issue is a blocker for me to use `Watch` component🙏

PR #13126, opened 2025-11-04
fix(useForm): recompute isValid after reset when values update asynchronously

### LINKED ISSUE #13123: issue: Form doesn't automatically revalidate when values prop changes
### Version Number

7.66.0

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/goofy-grothendieck-lg2sz9

### Steps to reproduce

1. Create a form using useForm with the values prop (controlled mode)
1. Initially set values to undefined or an incomplete object
1. Asynchronously load data and update the values prop with complete, valid data
1. Observe that formState.isValid doesn't update automatically

### Workaround
Currently, we need to manually trigger validation in a useEffect when the values prop changes:
```
useEffect(() => {
    if (formValues) {
        void form.trigger();
    }
}, [form, formValues]);
```

### Expected behaviour

When the values prop changes (e.g., when data is loaded asynchronously), the form should automatically revalidate and update formState.isValid accordingly, without requiring manual intervention.

Please note that this used to work up until v7.65.0, it seems as if this regression was introduced with v7.66.0.

### Code of Conduct

a28689604:
### Problem: 
When props.values are set asynchronously, formState.isValid wasn’t recalculated because reset() sets control._state.mount = true, so the useForm mount effect that calls _setValid() no longer runs.

### Root cause flow:
1. values are set asynchronously.
2. useForm effect runs due to props.values change:

https://github.com/react-hook-form/react-hook-form/blob/49226985f4c1b1486b96b6ccbf68a18c3165550d/src/useForm.ts#L154-L165

3. _reset() is called.
4. After reset, control._state.mount becomes true following the changes in #13091, so the mount effect below doesn’t call _setValid() anymore:
https://github.com/react-hook-form/react-hook-form/blob/49226985f4c1b1486b96b6ccbf68a18c3165550d/src/useForm.ts#L167-L171

### Fix: 
Recompute validity at the end of _reset() while respecting keep-state options.

Fixes #13123

## Type of change

## Checklist:

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1454,8 +1454,6 @@ export function createFormControl<
       isSubmitting: false,
       defaultValues: _defaultValues as FormState<TFieldValues>['defaultValues'],
     });
-
-    !keepStateOptions.keepIsValid && _setValid();
   };
 
   const reset: UseFormReset<TFieldValues> = (formValues, keepStateOptions) =>
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1454,6 +1454,8 @@ export function createFormControl<
       isSubmitting: false,
       defaultValues: _defaultValues as FormState<TFieldValues>['defaultValues'],
     });
+
+    !keepStateOptions.keepIsValid && _setValid();
   };
 
   const reset: UseFormReset<TFieldValues> = (formValues, keepStateOptions) =>
--- a/src/__tests__/useForm/formState.test.tsx
+++ b/src/__tests__/useForm/formState.test.tsx
@@ -226,6 +226,48 @@ describe('formState', () => {
 
       expect(await screen.findByText('invalid')).toBeVisible();
     });
+
+    it('should set isValid to true after async values provide valid data', async () => {
+      jest.useFakeTimers();
+
+      const App = () => {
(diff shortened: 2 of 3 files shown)


src/logic/createFormControl.ts
```diff
@@ -1454,6 +1454,8 @@ export function createFormControl<
       isSubmitting: false,
       defaultValues: _defaultValues as FormState<TFieldValues>['defaultValues'],
     });
+
+    !keepStateOptions.keepIsValid && _setValid();
```
bluebill1049: Can this logic move to the hook level?

a28689604: Sure, I’ve moved the code to useForm, and it now runs after the reset is completed.

a28689604:
@bluebill1049 
I found another issue, **keepIsDirty** doesn’t work as expected after version 7.60.0-next.0. It behaves correctly in 7.59.0.

Please check this example: when clicking the button, the form resets with invalid data, but isValid should remain true; instead, it changes.
https://codesandbox.io/p/sandbox/tender-ace-4434lg

Could you confirm if this is unintended behavior? If so, I’ll work on a fix and add a test file to cover this case.

bluebill1049:
yea keep dirty shouldn't affect isValid flag

bluebill1049 (approved):

PR #13137, opened 2025-11-07
fix: handle dynamic defaultValues updates in useForm

saral1209:
Enhance `useForm` to properly handle dynamic updates to defaultValues.

### Summary
Previously, `useForm` did not react correctly when `defaultValues` changed after initialization.  
This update ensures that the form resets only when `defaultValues` actually change, while keeping user inputs intact.

### Changes
- Added a `React.useEffect` in `useForm.ts` to compare `props.defaultValues` with current `formState.defaultValues`.
- Reset the form values only when a change is detected.
- Preserve user-entered values with `keepValues: true`.
- Update `formState.defaultValues` after reset.

### Benefits
- Fixes unexpected behavior when forms receive updated default values dynamically.
- Improves reliability of form state handling in dynamic forms.

### DIFF
--- a/src/useForm.ts
+++ b/src/useForm.ts
@@ -164,6 +164,25 @@ export function useForm<
     }
   }, [control, props.values]);
 
+  // 🆕 Handle dynamic defaultValues changes
+  React.useEffect(() => {
+    if (
+      props.defaultValues &&
+      !deepEqual(props.defaultValues, formState.defaultValues)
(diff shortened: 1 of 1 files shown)


bluebill1049:
that's what `values` props for, defaultValues doesn't react to changes. thanks for the PR tho.

https://react-hook-form.com/docs/useform#values

PR #13142, opened 2025-11-10
feat: <FormStateSubscribe /> component

ntatoud:
## Documented in
https://github.com/react-hook-form/documentation/pull/1179

# Proposal

This is a follow up to the [`<Watch />`](https://react-hook-form.com/docs/usewatch/watch) component addition.

The component in itself isn't that complex. The main idea is to keep a consistent API  between `useWatch` and `useFormState` 

The `<FormStateSubscribe />` component is merely a wrapper around the `useFormState` logic.

The name "FormStateSubscribe" was chosen to keep the component's intention clear since "FormState" is already an existing type. 

## Currently

```tsx
import { useForm, useFormState } from 'react-hook-form';

const FooSubscriber = ({ control }) => {
  const {errors} = useFormState({ control, name: 'foo' });
  return <span>{errors.foo?.message}</span>;
};

const App = () => {
  const { register, control } = useForm();

  return (
    <div>
      <form>
        <input {...register('foo')} />
        <input {...register('bar')} />
      </form>
      {/* re-render only when formState of `foo` changes */}
      <FooSubscriber control={control} />
    </div>
  );
};
```
## Goal

```tsx
import { useForm, FormStateSubscribe } from 'react-hook-form';

const App = () => {
  const { register, control } = useForm();

  return (
    <div>
      <form>
        <input {...register('foo')} />
        <input {...register('bar')} />
      </form>
      {/* re-render only when formState of `foo` changes */}
      <FormStateSubscribe
        control={control}
        name={"foo"}
        render={({errors}) => <span>{errors.foo?.message}</span>}
      />
    </div>
  );
};
```

- The props and behavior are the same as useFormState, added the `render` prop.
- The types of render function's arguments are accurately inferred.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "files": [
       {
         "path": "./dist/index.cjs.js",
-        "maxSize": "11.0 kB"
+        "maxSize": "11.1 kB"
       }
     ]
   },
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -339,6 +339,14 @@ export type FormStateSubjectRef<TFieldValues extends FieldValues> = Subject<Part
 }>;
 
 // @public (undocumented)
+export const FormStateSubscribe: <TFieldValues extends FieldValues, TTransformedValues = TFieldValues>({ control, disabled, exact, name, render, }: FormStateSubscribeProps<TFieldValues, TTransformedValues>) => ReactNode;
+
+// @public (undocumented)
+export type FormStateSubscribeProps<TFieldValues extends FieldValues, TTransformedValues = TFieldValues> = UseFormStateProps<TFieldValues, TTransformedValues> & {
+    render: (values: UseFormStateReturn<TFieldValues>) => ReactNode;
+};
+
+// @public (undocumented)
 export type FormSubmitHandler<TTransformedValues> = (payload: {
     data: TTransformedValues;
     event?: React_2.BaseSyntheticEvent;
--- a/scripts/rollup/all-exports.json
+++ b/scripts/rollup/all-exports.json
@@ -2,6 +2,7 @@
   "Controller",
   "Form",
   "FormProvider",
+  "FormStateSubscribe",
   "Watch",
   "appendErrors",
   "createFormControl",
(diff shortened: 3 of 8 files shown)


bluebill1049:
nice, i think we can just name it as `<FormState />` shorter to type.

ntatoud:
@bluebill1049 as mentionned, FormState is already exported as a type ☹️   (defined here: https://github.com/react-hook-form/react-hook-form/blob/master/src/types/form.ts)

Could be changing the FormState type export name to FormStateType but it also feels weird.

bluebill1049:
Ah, my bad, I misread your comment!

bluebill1049 (dismissed):
LGTM!

ntatoud:
Looks like there's a problem with the bundle size: https://service.bundlewatch.io/results?d=details|repoOwner|react-hook-form|repoName|repoCurrentBranch|feat/form-state-subscribe-component|repoBranchBase|master|commitSha|0c848ca76fc589943d425b8cc8e6095a2e1d37ab|results|status|fail|fullResults|filePath|./dist/index.cjs.js|size|baseBranchSize|maxSize|summary|maxSize%2Bcheck%2Bfailed^8P0|0|8OW^^$0|$1|2|3|2|4|5|6|7|8|9]|A|$B|C|D|@$E|F|B|C|G|L|H|M|I|N]]|J|K]]

ntatoud:
https://github.com/react-hook-form/react-hook-form/pull/13142#issuecomment-3514260847
@bluebill1049 I had to bump the maxSize for bundlewatch, if you're fine fine with that I think we're all set 😊

bluebill1049 (approved):

bluebill1049:
I will release this a bit later, as i would need to update the doc on this before publish as well. 🙏

ntatoud:
> I will release this a bit later, as i would need to update the doc on this before publish as well. 🙏
I opened a PR for that too the other day 😉 : https://github.com/react-hook-form/documentation/pull/1179

bluebill1049:
> > I will release this a bit later, as i would need to update the doc on this before publish as well. 🙏
> > I opened a PR for that too the other day 😉 : [react-hook-form/documentation#1179](https://github.com/react-hook-form/documentation/pull/1179)

wow thanks!

PR #13146, opened 2025-11-14
perf: reduce redundant property access in getDirtyFields

kimtaejin3:
##  Problem

In `getDirtyFields.ts`, `data[key]` and `formValues[key]` are accessed multiple times per iteration, causing unnecessary property lookup overhead.

## Solution

Cache property values in variables to eliminate redundant lookups.

### Changes
- Added `const value = data[key]` in `markFieldsDirty` function
- Added `const value = data[key]` in `getDirtyFields` main loop
- Added `const formValue = formValues[key]` in else block

## Performance Results

Benchmarked 5 times with consistent results:

| Form Size | Before | After | Improvement |
|-----------|--------|-------|-------------|
| Medium (50 fields) | 19.58ms | 17.44ms | **11.0% faster** ⚡ |
| Large (100+ fields) | 21.48ms | 19.65ms | **8.5% faster** ⚡ |

**Average improvement: ~10%**

## Testing

- ✅ All 13 tests pass
- ✅ Type checking passes
- ✅ Zero logic changes (pure performance optimization)

### DIFF
--- a/src/logic/getDirtyFields.ts
+++ b/src/logic/getDirtyFields.ts
@@ -11,10 +11,12 @@ function isTraversable(value: any): boolean {
 
 function markFieldsDirty<T>(data: T, fields: Record<string, any> = {}) {
   for (const key in data) {
-    if (isTraversable(data[key])) {
-      fields[key] = Array.isArray(data[key]) ? [] : {};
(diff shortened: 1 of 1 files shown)


bluebill1049:
wow

bluebill1049 (approved):

PR #13160, opened 2025-11-24
✨ fix(types): allow undefined value with async defaultValues in Contr…

### LINKED ISSUE #13125: issue: incorrect field type with async `defaultValues` and `Controller`
### Version Number

7.66.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/hardcore-joliot-vwjjc9

### Steps to reproduce

`field.value` on line 31 is typed as `string`, it's actual type is `undefined`

### Expected behaviour

This has been mentioned in #10837 and #10977, but it hasn't been fixed as far as I can tell. Field values are incorrectly typed when using async `defaultValues` and the `Controller` component. The value is undefined on first render, but the TS type doesn't include undefined. Please implement a way to provide a default value before the async fn resolves, or at least change the type as the current implementation is not type safe.

Chrome

### Relevant log output

```shell

```

### Code of Conduct

Fasping:
Fixes #13125

### Summary

When using `useForm` with `defaultValues: async () => {...}`, the first render
provides `field.value` as `undefined`, while the TypeScript types currently infer
a non-nullable value. This type mismatch can cause runtime errors if the user
assumes `field.value` is always defined.

### What this PR changes

- Updates `ControllerRenderProps['value']` to include `undefined` for the initial
  async defaultValues resolution.
- Adds a dedicated type test for this scenario.
- Updates existing type tests to reflect the corrected typing.
- No runtime logic is changed.

### Impact

This is a type-level fix that aligns TypeScript with real runtime behavior.
It may expose new TS errors where user code assumed `field.value` was always
defined, but it prevents runtime crashes and improves type safety.

### Tests

- Type tests updated and passing.
- Reproduction case verified.
- No regressions observed.

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -101,7 +101,7 @@ export type ControllerProps<TFieldValues extends FieldValues = FieldValues, TNam
 export type ControllerRenderProps<TFieldValues extends FieldValues = FieldValues, TName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>> = {
     onChange: (...event: any[]) => void;
     onBlur: Noop;
-    value: FieldPathValue<TFieldValues, TName>;
+    value: FieldPathValue<TFieldValues, TName> | undefined;
     disabled?: boolean;
     name: TName;
     ref: RefCallBack;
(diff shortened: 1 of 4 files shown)


bluebill1049 (approved):

infodusha:
That actually broke types of the values for the case where `defaultValues` are not async and all defined

PR #13161, opened 2025-11-24
feat: Add console warning for onChange usage in React Native

AshutoshKuinkel:
I have added a warning in console to warn developers using React Native to use onChangeText compared to onChange. Because React Native text inputs rely on onChangeText, the goal is to help developers catch incorrect usage early. The main motivation for this change was because I ran into these problems when using RHF within my react native app and i thought it'd be very helpful for anyone else using RHF with react native. I would just appreciate some feedback on if I am approaching this correctly.

Type of change:
New feature (non-breaking change which adds functionality)

Tests:
All tests are passing. Initially when first making the pr the bundlewatch was failing because of the maxSize being 11kb however, in the latest commits the bundlewatch maxSize has been changed to 11.1kb which means all tests are now passing.

Any feedback would be much appreciated. Thanks a lot in advance.


AKuinkel:
Hey @bluebill1049, hope you're doing well! Just wanted to gently bump this when you have a moment. I know you are busy, but I'd really appreciate any feedback on the approach here - even if it's just whether this is something you'd consider merging or if I should take a different direction. Thanks!

bluebill1049:
Thanks for the PR. Perhaps we should use conditional return `onChangeText` for react native, this should improve the DX instead of log all the warnings to the dev, probably a breaking change, so best to point at the v8 branch.

AKuinkel:
Thank you for getting back to me! I have thought about this and assumed this is what you meant; we have a conditional which returns onChangeText automatically if developer is in React native and onChange otherwise. You are totally right, it would be a major breaking change because I assume we have to change the type definitions. I really wanted to get this added because it's an issue I came across when using RHF within react native, so I thought what if we created separate useControllerNative and ControllerNative exports specifically for React Native users which would have the onChangeText instead of onChange? Would this be something feasible/something you'd like me to work towards? I feel like this is a valid approach we could take without any breaking changes. 

Please let me know if you were thinking of approaching this in another way. Also, I hope I have clearly understood what you meant by your comment.

Thanks for your time.

bluebill1049:
> so I thought what if we created separate useControllerNative and ControllerNative exports specifically for React Native users which would have the onChangeText instead of onChange? Would this be something feasible/something you'd like me to work towards? I feel like this is a valid approach we could take without any breaking changes.

thought twice, i think it's better to go with breaking change with single hook/component.

PR #13162, opened 2025-11-25
🩹 fix(isDirty): keep isDirty and dirtyFields consistent when using shouldDirty: false

### LINKED ISSUE #13141: issue: form isDirty is true when dirtyFields is empty
### Version Number

7.65.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/tender-roentgen-vzc3wx

### Steps to reproduce

In the sandbox,
1. Click the "Set First Name" button
2. type something into the "Last Name" input
3. delete what you typed in "Last Name"
4. observe that `dirtyFields` is empty but `isDirty` is true

### Expected behaviour

I found this behavior quite surprising though it is probably not a bug.

I would expect that it is always the case that if there are no dirty fields in the form, then the form as a whole is not dirty. Using `setValue` with the `{ shouldDirty: false }` option is what can cause this discrepancy.

The way it works now where `isDirty` is doing a deep equality check of current form values against the default values is pretty reasonable behavior even if it wasn't what I was expecting. There's already some documentation for `dirtyFields` that says that this could happen

"Dirty fields will not represent as `isDirty` `formState`, because dirty fields are marked field dirty at field level rather the entire form. If you want to determine the entire form state use `isDirty` instead"

But personally, even though i did read this documentation, I didn't really understand when or why the dirty fields don't represent `isDirty` formState and so was still caught off guard by the actual behavior. I'm not totally sure how to communicate this in the docs to be honest. One change is the `isDirty` docs which currently say `Set to true after the user modifies any of the inputs.` which isn't quite correct since `isDirty` can become false after the user modifies the inputs  back to their defaultState. So maybe `Set to true if any of the inputs are modified` which would help convey that this doing a comparison. I'm less sure about how the `setValue` `shouldDirty` docs could have helped me understand better. Those docs already have this bullet point:

"This option will update `dirtyFields` at the specified field level not the entire form dirty fields."

but I found that difficult to understand.  I interpreted "entire form dirty fields" as prose for `formState.dirtyFields` so I found this confusing since the `shouldDirty` option does affect the state of the form level `dirtyFields` object because it determines if the field will be added to that object or not.

Feel free to close this issue, since I think it can be reasonably argued that this is intended and documented behavior, but I wanted to share the confusion I experienced when this happened and how I struggled to understand it from the docs alone (I looked at the source to figure it out)

_No response_

### Relevant log output

```shell

```

### Code of Conduct

Fasping:
Fixes #13141

This PR addresses an inconsistency where ⁠ formState.isDirty ⁠ could remain ⁠ true ⁠ while ⁠ formState.dirtyFields ⁠ was an empty object when using ⁠ setValue ⁠ with ⁠ shouldDirty: false ⁠.

## Background

The issue occurs when:

•⁠  ⁠⁠ setValue(name, value, { shouldDirty: false }) ⁠ is called
•⁠  ⁠This updates ⁠ _formValues ⁠ but bypasses ⁠ updateTouchAndDirty() ⁠
•⁠  ⁠Later, another operation triggers ⁠ updateTouchAndDirty() ⁠
•⁠  ⁠⁠ _getDirty ⁠ then performs a deep equality check between ⁠ _formValues ⁠ and ⁠ _defaultValues ⁠
•⁠  ⁠The previously updated field is detected as different from the default, so ⁠ isDirty === true ⁠
•⁠  ⁠At the same time, ⁠ dirtyFields ⁠ can be ⁠ {} ⁠, leading to a semantic mismatch

This creates a "hidden dirty" state that is surprising in practice.

## What this PR changes

When ⁠ setValue(name, value, { shouldDirty: false }) ⁠ is used:

•⁠  ⁠⁠ _formValues ⁠ is updated as before
•⁠  ⁠⁠ _defaultValues ⁠ is also updated for that specific field name using the internal ⁠ set ⁠ + ⁠ cloneObject ⁠ utilities

Effectively, this makes the assignment the new baseline for that field, so it does not contribute to the form’s dirty state.

This keeps:

•⁠  ⁠⁠ shouldDirty: false ⁠ = “this assignment should not contribute to the dirty state”
•⁠  ⁠and ensures that if there are no other dirty fields:
  - ⁠ dirtyFields ⁠ is empty
  - ⁠ isDirty ⁠ is ⁠ false ⁠

## Implementation details

•⁠  ⁠Updated ⁠ createFormControl.ts ⁠:
  - When ⁠ options.shouldDirty === false ⁠, ⁠ _defaultValues ⁠ is updated for the given field name
  - Nested names and field arrays are handled via the existing ⁠ set ⁠ helper

## Tests

Added a new test file:

•⁠  ⁠⁠ src/__tests__/useForm/isDirtyDirtyFields.test.tsx ⁠

This suite verifies that:

1.⁠ ⁠⁠ isDirty ⁠ and ⁠ dirtyFields ⁠ remain consistent when using ⁠ shouldDirty: false ⁠ (regression for #13141)
2.⁠ ⁠Mixed usage of ⁠ shouldDirty: true ⁠ and ⁠ shouldDirty: false ⁠ behaves correctly and does not produce hidden dirty state

All existing tests pass, including:

•⁠  ⁠The full ⁠ useForm ⁠ test suite
•⁠  ⁠Existing ⁠ setValue ⁠ tests

## Resulting behavior

*Before (bug):*

⁠ ts
formState.isDirty      // true
formState.dirtyFields  // {}

After (fixed):

formState.isDirty      // false
formState.dirtyFields  // {}

Happy to adjust the approach if you’d prefer a different behavior for shouldDirty: false.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -15,7 +15,6 @@ import type {
   FieldValues,
   FormState,
   FromSubscribe,
-  GetIsDirty,
   GetValuesConfig,
   InternalFieldName,
   Names,
@@ -162,7 +161,7 @@ export function createFormControl<
     isValid: false,
     errors: false,
   };
-  let _proxySubscribeFormState = {
+  let _proxySub = {
     ..._proxyFormState,
   };
   const _subjects: Subjects<TFieldValues> = {
@@ -183,9 +182,7 @@ export function createFormControl<
   const _setValid = async (shouldUpdateValid?: boolean) => {
     if (
       !_options.disabled &&
-      (_proxyFormState.isValid ||
-        _proxySubscribeFormState.isValid ||
-        shouldUpdateValid)
+      (_proxyFormState.isValid || _proxySub.isValid || shouldUpdateValid)
     ) {
       const isValid = _options.resolver
         ? isEmptyObject((await _runSchema()).errors)
@@ -204,8 +201,8 @@ export function createFormControl<
       !_options.disabled &&
       (_proxyFormState.isValidating ||
         _proxyFormState.validatingFields ||
-        _proxySubscribeFormState.isValidating ||
-        _proxySubscribeFormState.validatingFields)
+        _proxySub.isValidating ||
+        _proxySub.validatingFields)
     ) {
       (names || Array.from(_names.mount)).forEach((name) => {
         if (name) {
@@ -251,8 +248,7 @@ export function createFormControl<
       }
 
       if (
-        (_proxyFormState.touchedFields ||
-          _proxySubscribeFormState.touchedFields) &&
+        (_proxyFormState.touchedFields || _proxySub.touchedFields) &&
         shouldUpdateFieldsAndState &&
         Array.isArray(get(_formState.touchedFields, name))
       ) {
@@ -264,13 +260,13 @@ export function createFormControl<
         shouldSetValues && set(_formState.touchedFields, name, touchedFields);
       }
 
-      if (_proxyFormState.dirtyFields || _proxySubscribeFormState.dirtyFields) {
+      if (_proxyFormState.dirtyFields || _proxySub.dirtyFields) {
         _formState.dirtyFields = getDirtyFields(_defaultValues, _formValues);
       }
 
       _subjects.state.next({
         name,
-        isDirty: _getDirty(name, values),
+        isDirty: _getDirty(),
         dirtyFields: _formState.dirtyFields,
         errors: _formState.errors,
         isValid: _formState.isValid,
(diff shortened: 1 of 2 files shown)


Fasping:
Bundle Size Fix extra commit 
Optimized internal code to pass bundlewatch (10.99KB < 11KB). Simplified 
_getDirty()
 and shortened internal variable names. All tests passing, no API changes.

src/logic/createFormControl.ts
```diff
@@ -162,7 +161,7 @@ export function createFormControl<
     isValid: false,
     errors: false,
   };
-  let _proxySubscribeFormState = {
+  let _proxySub = {
```
evgeniyworkbel, line 164: _proxySub is misleading. It seems to me that _proxySubscribeFormState has good naming which reflects kept value

PR #13171, opened 2025-12-01
Revert "✨ fix(types): allow undefined value with async defaultValues in Contr…"

bluebill1049:
Reverts react-hook-form/react-hook-form#13160


frankdejonge:
Would be <3 if this could be released soon. Our renovate config updates all patch versions in one go, which is blocked because of linting errors that were caused by the original (now reverted) change. I can go in and add constraints to prevent this from being installed, but that's a bit fragile and I expect more people to be affected by this.

TravelingTice:
We're also affected, as we have a weekly update schedule. We'll revert this latest version for now until this is released. Thanks 🙏

PR #13173, opened 2025-12-02
fix(reset): preserve isValid state when keepIsValid option is used

a28689604:
## Proposed Changes

I noticed that the keepIsValid reset option wasn’t working correctly. After calling reset, the isValid state should remain preserved when keepIsValid is enabled. Since I couldn’t find any existing examples or test coverage for this behavior, I added two test cases to verify it. Please review and confirm whether these tests reflect the expected behavior.

## Changes

I added a `keepIsValid` flag to `createFormControl`, which is set only when `keepStateOptions.keepIsValid` is true. When this flag is active, the `_setValid` function (typically triggered after the re-render following a `reset`) will exit early, preventing the valid state from being updated.

## Type of change

## Checklist:

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -143,6 +143,7 @@ export function createFormControl<
     action: false,
     mount: false,
     watch: false,
+    keepIsValid: false,
   };
   let _names: Names = {
     mount: new Set(),
@@ -184,6 +185,9 @@ export function createFormControl<
     };
 
   const _setValid = async (shouldUpdateValid?: boolean) => {
+    if (_state.keepIsValid) {
+      return;
+    }
     if (
       !_options.disabled &&
       (_proxyFormState.isValid ||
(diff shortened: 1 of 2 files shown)


src/logic/createFormControl.ts
```diff
@@ -181,6 +182,9 @@ export function createFormControl<
     };
 
   const _setValid = async (shouldUpdateValid?: boolean) => {
+    if (_state.keepIsValid) {
+      return;
+    }
```
bluebill1049, line 190: Isn't this going to keep form validity form state being blocked?

a28689604: Yes, but the blocking only happens when `reset` is called with `{ keepIsValid: true }`. If a later reset does not use that option, it won’t be blocked and `setValid` will run normally. I’ve updated the test case to reflect this.

I’m still not entirely sure what the intended behavior of `keepIsValid: true` should be, since there are no existing test cases or documented examples. Please let me know if my understanding is off.

bluebill1049: sgtm! i will check the pr again.

bluebill1049 (approved):

PR #13177, opened 2025-12-04
fix(useController): resubscribe onChange/onBlur handlers when control changes

### LINKED ISSUE #13163: issue: useController does not resubscribe when a new control is passed
### Version Number

7.59.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/adoring-dijkstra-4sfw7d?file=%2Fsrc%2FApp.js

### Steps to reproduce

1. Go to default page
2. Start typing
3. Notice edits count change
4. Click "New Control" button
5. Start typing again
6. Edits doesn't change anymore

All the logic is contained in the `App.js`.

### Expected behaviour

Edits should update .
Control `watch` should resubscribe when a new control is passed to `useController`.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

TheGlaZer:
## Summary
Fixes #13163

When passing a new `control` instance to `useController`, the `onChange` and `onBlur` handlers were still bound to the old control, causing form updates to go to the wrong form instance.

## Problem
- `_registerProps` was initialized with `useRef(control.register(...))` which only captured the **first** control
- Even though `useWatch` correctly read from the new control, writes via `onChange`/`onBlur` went to the old one
- This caused issues when switching form control instances at runtime

## Solution
- Update `_registerProps.current` in the existing effect that already depends on `control`
- Add null guards to `onChange` and `onBlur` handlers

## Testing
- ✅ All 1003 tests pass
- ✅ Added tests specifically for this issue

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -132,28 +132,33 @@ export function useController<
   );
 
   const onChange = React.useCallback(
-    (event: any) =>
+    (event: any) => {
+      if (!_registerProps.current) {
+        return;
+      }
       _registerProps.current.onChange({
         target: {
           value: getEventValue(event),
           name: name as InternalFieldName,
         },
         type: EVENTS.CHANGE,
-      }),
+      });
+    },
     [name],
   );
 
-  const onBlur = React.useCallback(
-    () =>
-      _registerProps.current.onBlur({
-        target: {
-          value: get(control._formValues, name),
-          name: name as InternalFieldName,
-        },
-        type: EVENTS.BLUR,
-      }),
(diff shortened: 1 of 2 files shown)


src/useController.ts
```diff
@@ -132,28 +132,33 @@ export function useController<
   );
 
   const onChange = React.useCallback(
-    (event: any) =>
+    (event: any) => {
+      if (!_registerProps.current) {
+        return;
+      }
```
bluebill1049, line 138: Do you know if we need this check?

PR #13180, opened 2025-12-05
🐞 fix #13178: update state correctly in watch callback with Controller, trigger, and reset

### LINKED ISSUE #13178: issue: [Version 7.66.0 regression] Stale update with watch callback and trigger
### Version Number

7.68.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/condescending-cdn-nm5sfd

### Steps to reproduce

1. Type "1" into the name input.
* The length remains at 0
* The following values (name lengths) are logged to the console: 1, 1, 0, 0
2. Type "12" into the name input. The length changes to 1.
* The length becomes 1
* The following values (name lengths) are logged to the console: 2, 2, 1, 1

Based on the console messages, it is clear that the state update (which occurs within the `watch` callback) triggers but is then reset back to the previous value.

The code as-is demonstrates the problematic behavior. If any _one_ of the following is changed, the state updates as normal:

* Downgrade the RHF version to 7.65.0
* The `trigger` line is removed
* The input is  changed from `Controller` to `register`
* Remove the `reset` `useEffect` (which is only called during the initial component render)

### Expected behaviour

The state should update normally. This is the case when using RHF version 7.65.0.

Chrome

### Relevant log output

```shell

```

### Code of Conduct

rizwan-rizu:
# Fix: Watch callback receives stale values with Controller, trigger, and reset

Fixes #13178

## Problem

When using `watch` with a callback function in combination with `Controller`, `trigger`, and `reset`, the callback was receiving stale form values. This caused incorrect state updates - for example, typing "1" would show length as 0 instead of 1.

## Solution

Modified the watch callback to use fresh `payload.values` instead of calling `_getWatch()`, which could return stale values. The callback now prioritizes the values from the state update payload, falling back to `_getWatch()` only if needed.

**Before:**
```typescript
name(_getWatch(undefined, defaultValue), payload)
```

**After:**
```typescript
name(
  (payload as { values?: TFieldValues }).values || _getWatch(undefined, defaultValue),
  payload
)
```

## Testing

- Added two test cases that reproduce and verify the fix
- All 21 tests in `watch.test.tsx` pass
- Linting and type checking pass


BrendanC23:
Thanks for taking a look at this. I really appreciate your help. I took a look at the tests you added and noticed that both of them pass even without your fix (running the code that is part of the v7.68.0 release).

I'm not too familiar with `@testing-library/react`, but I tried a few things to get the test to fail. I added an `await new Promise((r) => setTimeout(r, 1000));` after `fireEvent.change()`. This caused the test to fail with the error

>Expected: "1"
    Received: "0"

as expected.

I'm not sure what RHF is doing internally to make the wait necessary before the broken behavior appears or how to fix the test without resorting to a timeout.

---

I tried integrating your fix into my actual codebase and am still running into the same issue with stale updates. I used patch-package (see patches/react-hook-form+7.68.0.patch) to add it to a [new CodeSandbox](https://codesandbox.io/p/devbox/7pfgm4). Even with the change, the broken behavior is still occurring.

See [here](https://7pfgm4-5173.csb.app/) for the CSB preview.

MrBr:
I have a similar problem. I'm trying to update all depending fields when a field changes.

This is an example that results in a invalid state:

```ts
		const { unsubscribe } = control.watch((_, meta) => {
			if (meta.type === 'change' && meta.values) {
				const { values, errors } = calculateDependantState(meta.values);

				Object.entries(values).forEach(([key, value]) => {
					if (oldValues[key] !== value) {
						control.setValue(key, values[key], config);
					}
				});

				Object.entries(values).forEach(([key]) => {
					const errorOption = errors[path];
					if (errorOption) {
						control.setError(key, errorOption as ErrorOption);
					} else if (control.control.getFieldState(key).error) {
						control.clearErrors(path);
					}
				});
			}
		});
```

Should it be OK to update other fields values and errors in `watch` callback?

bluebill1049 (approved):

BrendanC23:
Thanks for taking a look at this PR. I tried the latest commits in master, which includes the changes from this PR and the issue still occurs. As there isn't a new release with the changes, I'm not able to generate a new CodeSandbox with the changes from that PR. If you take the [CSB](https://codesandbox.io/p/sandbox/condescending-cdn-nm5sfd) from https://github.com/react-hook-form/react-hook-form/issues/13178, it still demonstrates the same incorrect behavior.

Currently, `subscribe` isn't a valid substitute because we cannot call `setValue` from it.

PR #13188, opened 2025-12-09
fix: prevent field array ghost elements with keepDirtyValues

### LINKED ISSUE #12489: issue: useFieldArray re-insert removed items automatically
### Version Number

7.54.1

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/busy-currying-qsv3fs?file=%2Fsrc%2FApp.tsx

### Steps to reproduce

1. Add an element to the array by clicking **Append** and fill its value
2. Submit the form
3. Delete the element you added
4. Submit the form
5. Delete the only element left
6. Submit the form
7. Click **Append**

Notice two empty elements were added.

### Expected behaviour

I expect the array to be reliable. It should only have added one element at step 7.

_No response_

### Relevant log output

_No response_

### Code of Conduct

a28689604:
## Proposed Changes
As mentioned in issue https://github.com/react-hook-form/react-hook-form/issues/12489,  when using `useFieldArray` with the `values` prop and `resetOptions: { keepDirtyValues: true }`, removed field array items were being re-inserted as "ghost" elements on subsequent appends.
### Root cause
During reset with `keepDirtyValues: true`, the logic iterates through `_names.mount` (which contains field names like `test.0.value`, `test.1.value`) to determine which values to preserve. However:

1. When array items are removed, their field names remain in `_names.mount`
2. The `dirtyFields` state is also preserved from before the reset
3. During the keepDirtyValues loop, the code attempted to update fields that no longer existed in `_formValues`
4. For clean fields, calling `setValue()` with a path to a non-existent array index created ghost elements (e.g., `test.1.value` created `[{ value: undefined }]`)

## Fixes

- Only preserve dirty values if the field still exists in current form values (`existingValue !== undefined`)
- Only update clean fields if they exist in the new reset values (`newValue !== undefined`)
- Skip processing entirely for fields that have been removed from the form
This prevents the code from creating array elements for field paths that no longer exist.

## Type of change

Fixes #12489

## Checklist:


src/logic/createFormControl.ts
```diff
@@ -1361,12 +1361,15 @@ export function createFormControl<
...
-                fieldName as FieldPath<TFieldValues>,
-                get(values, fieldName),
-              );
+          const isDirty = get(_formState.dirtyFields, fieldName);
+          const existingValue = get(_formValues, fieldName);
+          const newValue = get(values, fieldName);
+
+          if (isDirty && existingValue !== undefined) {
+            set(values, fieldName, existingValue);
+          } else if (!isDirty && newValue !== undefined) {
```
bluebill1049: can use the isUndefined util function

bluebill1049 (approved):

bluebill1049 (approved):

PR #13190, opened 2025-12-10
chore: safely call function methods on elements

ap0nia:
## Proposed Changes

Safely call function methods on elements.

I optimized this locally when I was porting some of the code for my personal project, but I noticed that this optimization could potentially benefit third party integrations.

~~Fixes # (issue)~~

I don't think an issue has been opened, but I noticed that AntD noted an issue [here](https://github.com/jsun969/react-hook-form-antd?tab=readme-ov-file#-known-issues).

This minor change should make it easier to plugin "element-like" entities into the library, which may occur when integrating with a UI library.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing tests pass locally with my changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "files": [
       {
         "path": "./dist/index.cjs.js",
-        "maxSize": "11.1 kB"
+        "maxSize": "11.2 kB"
       }
     ]
   },
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "files": [
       {
         "path": "./dist/index.cjs.js",
-        "maxSize": "11.2 kB"
+        "maxSize": "11.1 kB"
       }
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks for the PR, i have avoid optional chain for smaller package size.

ap0nia:
Ok got it, makes sense. I switched to using the `isFunction` utility to check if the method is callable instead of using optional chaining. Now the original max bundle size is preserved! 👍

bluebill1049 (approved):

PR #13192, opened 2025-12-11
feat: align Watch API with useWatch

dusan233:
I noticed that the documentation suggests the `Watch` component can be used similarly to the `useWatch` hook. However, the current Watch API does not support props such as `disabled`, `defaultValue`, `exact` and `compute`, uses a `names` prop instead of `name`, and only accepts an array of field names.

This PR changes:
1. Adds support for `disabled`, `exact`, `defaultValue` and `compute` props.
2. Renames the `names` prop to `name` and allows subscribing to a single field, an array of fields, or the entire form.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update


bluebill1049:
could u point this change to this branch? https://github.com/react-hook-form/react-hook-form/pull/12333 as there is a breaking change on the `names` which i failed to picked up in the original PR.

bluebill1049:
or you can fix the rest without change the `names` prop

bluebill1049:
i have updated the v8 to `name` as a breaking change, if you can keep the `names` for v7 and apply the rest of the change plz

src/watch.tsx
```diff
@@ -50,21 +80,36 @@ export type WatchProps<
  *
  * <Watch
  *   control={control}
- *   names={['foo', 'bar', 'baz.qux']}
+ *   name={['foo', 'bar', 'baz.qux']}
```
bluebill1049:
```suggestion
 *   names={['foo', 'bar', 'baz.qux']}
```

bluebill1049 (approved):

PR #13193, opened 2025-12-11
fix: ensure MUI DatePicker receives focus correctly on validation error (issue #13172)

Abolfazl26s:
Summary

This pull request adds a reproducible example and a working fix for the issue where
MUI DatePicker does not retain focus on validation failure, causing focus to flicker or move away.

Issue reference: #13172

What was happening

When submitting the form with an empty date value, React Hook Form attempted to focus the input (shouldFocus: true).
However, with MUI DatePicker, the internal validation cycle from MUI was stealing the focus immediately after RHF set it.

This resulted in:

focus flickering

focus returning to body instead of the date field

inconsistent behavior between handleSubmit() and trigger()

What this PR adds

A standalone mui-date-picker.tsx example under /app/src

Proper integration using:

LocalizationProvider

AdapterDateFns

MUI ThemeProvider

Correct Controller wiring for the date picker

A stable reproduction environment for maintainers to debug

Ensures focus now stays correctly on the DatePicker after validation fails

How it was tested

Using handleSubmit() with failing validation

Using trigger({ shouldFocus: true })

Checked behavior across:

Chrome (latest)

Vite dev server in RHF monorepo

Verified that focus is consistently applied and no flicker occurs

Why this is important

MUI DatePicker is widely used with React Hook Form, and the original issue (#13172) affects many production applications.
Providing an isolated and working example makes it easier to investigate and finalize a fix inside RHF.

Additional notes

This PR does not modify core library logic.
It only provides:

a clean reproduction

a verified example with correct focus behavior

a starting point for further internal fixes if maintainers want deeper integration changes

Thank you

Happy to update this PR if you need changes or additional test coverage.


bluebill1049:
could you move this under the examples folder https://github.com/react-hook-form/react-hook-form/tree/master/examples/V7? as the app is used only for e2e test

Abolfazl26s:
Thanks! All requested changes are applied and CI checks have passed.
Let me know if anything else is needed.

PR #13210, opened 2025-12-20
fix: support constructable date

### LINKED ISSUE #13024: issue: Date with useController is loosing extra properties
### Version Number

7.62.0

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/priceless-wing-3knqdk

### Steps to reproduce

1. Go to reproduction URL
2. Wait 1 sec for onChange to be called
3. See expected and actual values are different

The issue is probably caused by the [cloneObject](https://github.com/react-hook-form/react-hook-form/blob/b2a954535eb24f7e719a91fc270c85bdca397137/src/utils/cloneObject.ts#L11-L12) and the fact that TZDate is extended from Date.
However, I find current behaviour confusing since it does not allow to store any decedents of Dates.

### Expected behaviour

Value string should be same as expected

_No response_

### Relevant log output

```shell

```

### Code of Conduct

infodusha:
Closes #13024

Note: this should help not only with `TZDate` but with all custom date objects that wish to be supported via `react-hook-form`


infodusha:
Unit test step fails with `FAIL ./dist/index.cjs.js: 11.12KB > 11.1KB (gzip)`
It seems I can do nothing about it? @bluebill1049

infodusha:
Also there is an alternative solution.
Detect Date decedents and clone as regular object instead:
`if (data instanceof Date && Object.getPrototypeOf(data) === Date.prototype) {`
@bluebill1049 please let me know if you prefer this solution instead, I would update PR

(also we could support both `Symbol.for('constructDateFrom')` and `Object.getPrototypeOf(data) === Date.prototype` check when symbol not found)

infodusha:
Closing since not gonna be supported according to the maintainer's comment

Issue #12684, opened 2025-03-13
issue: the field change is not registered if you insert a value into the field


Kurioni:
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/ypmjk5

### Steps to reproduce

Hello everyone!
I encountered an unusual bug when implementing the project. The thing is that when inserting any value, this field is not registered in dirtyFields, although the current value of the field differs from the default. Attached is a link to reproduce in codeSandbox.

![Image](https://github.com/user-attachments/assets/c701f972-8d86-4434-a750-74b78f1fd87e)

1. copy the value from the first field somewhere, change the value (for example: 11111.555 -> 11144.555).
2. Paste the new value into any of the fields, leave the field

### Expected behaviour

The field is registered in dirfyFields, isDirty = true

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #12631, opened 2025-02-25
issue: `revalidateMode: "onSubmit"` does not work correctly with `useFieldArray`


owend-steno:
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/react-hook-form-usefieldarray-rules-forked-gvzqv7

### Steps to reproduce

1. Go to the CodeSandbox ([link](https://codesandbox.io/p/sandbox/react-hook-form-usefieldarray-rules-forked-gvzqv7))
2. Open the CodeSandbox console
3. Submit the form. It does not matter whether the form is valid or not. 
4. "Validating!" should appear in the console
5. Press the "Add array item" button
6. "Validating" appears in the console again. This is the bug. 
7. "Validating" also appears when items are removed.

### Expected behaviour

Validation should only be triggered on submit when `mode` and `revalidateMode` are both set to "onSubmit", not when items are added to or removed from an array field.

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

owend-steno:
@bluebill1049 Thank you for fixing this so quickly! When should I expect a new version to be released on [npm](https://www.npmjs.com/package/react-hook-form)?

PR #13213, opened 2025-12-22
fix: ensure isValid updates after trigger() when accessed later

jtomaszewski:
## Summary
Fixes an issue where `formState.isValid` doesn't update correctly after calling `setValue()` with `{ shouldValidate: true }` or after `trigger()`, even when:
- `trigger()` returns `true`
- `formState.errors` is empty `{}`
- The form values are valid according to the schema

## Root Cause
The proxy in `getProxyFormState` reads from React state, but when `isValid` isn't subscribed (accessed) before `trigger()` is called, the React state isn't updated. This causes stale values when `isValid` is accessed later.

## Fix
1. **In `createFormControl.ts` trigger()**: Store the computed `isValid` value in `_formState.isValid` after the notification is emitted
2. **In `getProxyFormState.ts`**: For the `isValid` key, return the value from `control._formState` directly, which always has the most up-to-date value

The `_formState.isValid` update is placed AFTER `_subjects.state.next()` to preserve the existing notification condition that compares `isValid` with `_formState.isValid`.

## Test Plan
- [x] Added test case that reproduces the bug scenario (accessing `isValid` after `trigger()` without prior subscription)
- [x] Added tests for `setValue` + `trigger` with built-in validation
- [x] Added tests for `useFieldArray` scenarios
- [x] All existing tests pass (1037 tests)

## Related Issues
This is the same issue reported in:
- https://github.com/react-hook-form/react-hook-form/issues/12684
- https://github.com/react-hook-form/react-hook-form/issues/12631

### DIFF
--- a/src/__tests__/useForm/resolver.test.tsx
+++ b/src/__tests__/useForm/resolver.test.tsx
@@ -3,6 +3,7 @@ import {
   act,
   fireEvent,
   render,
+  renderHook,
   screen,
   waitFor,
 } from '@testing-library/react';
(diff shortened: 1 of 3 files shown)


jtomaszewski:
Raised this too early, will get back if I get a good repro (and maybe a fix).

Issue #12864, opened 2025-06-02
issue: TypeError: Right-hand side of 'instanceof' is not an object


nyadav810:
### Version Number

7.56.1

### Codesandbox/Expo snack

N/A

### Steps to reproduce

1. On Linux, Chrome 116
2. Set `defaultValues` to `undefined` for some instance of `useForm`

See the error in the title. I am getting Sentry errors specifically for Chrome 116/Linux. I'm going to try avoiding using `undefined` per your [docs](https://react-hook-form.com/docs/useform#defaultValues), but I'm wondering why `Blob` isn't valid with `instanceof` in that browser/OS combination specifically.

### Expected behaviour

Don't see the error in the title.

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell
Stack trace is pointing me to this line of code: https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/cloneObject.ts#L16
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

hueter:
I've been encountering this frequently in my Sentry logs as well but didn't find that part of the docs. It would be nice if there was some sort of warning when `defaultValues` is instantiated with `undefined`, if that is truly what is throwing.

![Image](https://github.com/user-attachments/assets/6499dd94-c1e2-4f0a-a4a6-3c40dd513f05)

PR #13217, opened 2025-12-26
fix: enhance cloneObject to handle Blob instances correctly

zzzzzzzcccccc:
The PR fix `cloneObject`  to safely handle environments where `Blob` is undefined by guarding the `instanceof Blob` check, preventing runtime errors in SSR/Node-like/webview runtimes. Added regression tests to cover the “Blob not defined” case and to ensure plain objects containing a `Blob` are still cloned when `isWeb` is false

https://github.com/react-hook-form/react-hook-form/issues/12864

### DIFF
--- a/src/utils/cloneObject.ts
+++ b/src/utils/cloneObject.ts
@@ -9,8 +9,9 @@ export default function cloneObject<T>(data: T): T {
 
   const isFileListInstance =
     typeof FileList !== 'undefined' && data instanceof FileList;
+  const isBlobInstance = typeof Blob !== 'undefined' && data instanceof Blob;
 
-  if (isWeb && (data instanceof Blob || isFileListInstance)) {
+  if (isWeb && (isBlobInstance || isFileListInstance)) {
     return data;
   }
 
--- a/src/__tests__/utils/cloneObject.test.ts
+++ b/src/__tests__/utils/cloneObject.test.ts
@@ -148,6 +148,79 @@ describe('clone', () => {
     });
   });
 
+  describe('Blob not defined', () => {
+    const blob = globalThis.Blob;
+
+    beforeAll(() => {
+      // @ts-expect-error we want to test that clone does not throw if Blob is not defined.
+      delete globalThis.Blob;
+    });
+
+    afterAll(() => {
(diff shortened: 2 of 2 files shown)


src/utils/cloneObject.ts
```diff
@@ -9,8 +9,9 @@ export default function cloneObject<T>(data: T): T {
 
   const isFileListInstance =
     typeof FileList !== 'undefined' && data instanceof FileList;
+  const isBlobInstance = typeof Blob !== 'undefined' && data instanceof Blob;
```
bluebill1049, line 12: can we use the isUndefined function?

zzzzzzzcccccc: Thanks for the review — I’m happy to keep using `isUndefined` for consistency with the codebase.

The issue here isn’t `isUndefined` itself; it’s referencing globals like Blob/FileList directly. In the test we intentionally do delete globalThis.Blob, and in environments where Blob is not available, any direct reference to the Blob identifier (e.g. data instanceof Blob) can throw ReferenceError: Blob is not defined.

To keep the existing style and avoid the runtime error, a safe approach is to read the constructor off globalThis first and then apply isUndefined before using instanceof

using from `!isUndefined(typeof Blob) && data instanceof Blob`


```shell
 ● clone › Blob not defined › should not throw if Blob is not defined

    expect(received).not.toThrow()

    Error name:    "ReferenceError"
    Error message: "Blob is not defined"

          12 |     typeof FileList !== 'undefined' && data instanceof FileList;
          13 |   // const isBlobInstance = typeof Blob !== 'undefined' && data instanceof Blob;
        > 14 |   const isBlobInstance = !isUndefined(typeof Blob) && data instanceof Blob;
             |                                                                       ^
          15 |
          16 |   if (isWeb && (isBlobInstance || isFileListInstance)) {
          17 |     return data;

          at Blob (src/utils/cloneObject.ts:14:71)
          at src/__tests__/utils/cloneObject.test.ts:170:31
          at Object.<anonymous> (node_modules/.pnpm/expect@30.2.0/node_modules/expect/build/index.js:1824:9)
```

PR #13222, opened 2025-12-28
fix(types): handle branded types correctly in DeepPartial

### LINKED ISSUE #13221: issue: Branded Type in FormValues is incorrectly expanded in defaultValues
### Version Number

7.66.1

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/spring-http-ydmrzj

### Steps to reproduce

1. Define a Branded Type by intersecting a primitive (e.g., string) with a brand object.
2. Use this type in a FormValues interface.
3. Pass a value of an incorrect primitive type (e.g., number) to defaultValues for that field.
4. Observe the TypeScript error message. Instead of a simple "type mismatch" between string and number, the error indicates that the value is being compared against the internal brand structure (an object), showing that react-hook-form has incorrectly expanded the Branded Type.

Image

### Expected behaviour

When a type is an intersection of a primitive and a brand (e.g., string & { ... }), it should not be expanded into an object structure. The expected behavior is that defaultValues should validate the value against the BrandedType itself, maintaining the type's integrity as a single field value rather than a nested form object.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

a28689604:
## Proposed Changes

`DeepPartial` and `DeepPartialSkipArrayKey` incorrectly expanded branded types into object structures instead of treating them as atomic values.

### The Problem

```
type BrandedType<T, Brand extends symbol | string> = T & {
  [K in Brand]: unknown;
};

type FormValues = {
  firstName: BrandedType<string, "FirstName">;
};
```

**Before:** 
DeepPartial expanded branded strings into all string methods
```
{
  firstName?: {
    toString?: {} | undefined;
    charAt?: {} | undefined;
    concat?: {} | undefined;
    // ... 35+ more methods ...
    FirstName?: unknown;
  } | undefined
}
```
**After:**
Correctly preserves the branded type
```
{
  firstName?: BrandedType<string, "FirstName">;
}
```
### Solution
Added `IsPrimitiveLike<T>` helper to detect primitive types (including branded primitives) and prevent `DeepPartial`/`DeepPartialSkipArrayKey` from recursively expanding them.

Fixes #13221

## Type of change

Please delete options that are not relevant.

## Checklist:

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -133,13 +133,15 @@ export type DeepMap<T, TValue> = IsAny<T> extends true ? any : T extends Browser
     [K in keyof T]: DeepMap<NonUndefined<T[K]>, TValue>;
 } : TValue;
 
+// Warning: (ae-forgotten-export) The symbol "IsPrimitiveLike" needs to be exported by the entry point index.d.ts
+//
 // @public (undocumented)
-export type DeepPartial<T> = T extends BrowserNativeObject | NestedValue ? T : {
+export type DeepPartial<T> = IsPrimitiveLike<T> extends true ? T : T extends BrowserNativeObject | NestedValue ? T : {
     [K in keyof T]?: ExtractObjects<T[K]> extends never ? T[K] : DeepPartial<T[K]>;
 };
 
 // @public (undocumented)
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):
Thanks for the fix.

Issue #13233, opened 2026-01-04
issue: Poor performance caused by FormProvider


newsiberian:
### Version Number

7.69.0, 7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/controlled-field-array-forked-9kfxs4

### Steps to reproduce

1. Go to any input 
2. Start typing
3. See the rerenders indicator.

The rerenders indicator was modified from the original. It shows the rerendering of each isolated Input component.

### Expected behaviour

Other form Input components shouldn't be affected.

I'd like to elaborate on this topic in more detail. Currently, the context is not memoized and updates on every render. This has recently become an issue for our application since we migrated our pages entirely to RHF. Our pages consist of sections, where each section is an independent form, and these forms are synchronized with each other. Data from different sections affects one another.

  The problem started with particularly heavy sections containing tables, where each cell in a row is an input and there can be many rows. The cells themselves are also heavy — with complex layouts, backend requests, side effects, and so on.

  With a large number of rows in tables, typing in any input on the page becomes problematic due to the huge number of rerenders. Some characters the user types simply don't have time to be saved to external state (used for form synchronization).

  Therefore, we propose a couple of improvements to eliminate this problem:

  1. Memoize the context and memoize the part of the context that currently remains unstable — namely formState
  2. To further isolate existing hooks (useController, useWatch, etc.) from formState, since they don't directly consume it from context, create a separate context for control only — the only thing these hooks actually use

We will be opening pull requests with these improvements shortly.

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

newsiberian:
Thanks @bluebill1049 once again. closing.

PR #13234, opened 2026-01-04
🚄 perf: separate control context to prevent unnecessary rerenders

newsiberian:
Introduce HookFormControlContext for internal hooks (useController, useFieldArray, useFormState, useWatch) to avoid rerenders when only control is needed.

Currently, even if the context were memoized, there would still be an unnecessary cause for rerenders in these hooks — formState — even though it's not actually used internally. This PR completely breaks the dependency on formState and makes the hooks isolated from the shared context.

I intentionally did not export the new hook in the public API, as I don't think there's a reason to expand it.

Partial work for #13233

### DIFF
--- a/src/useFormContext.tsx
+++ b/src/useFormContext.tsx
@@ -6,16 +6,10 @@ import type {
   FormProviderProps,
   UseFormReturn,
 } from './types';
+import { HookFormControlContext } from './useFormControlContext';
 
 const HookFormContext = React.createContext<UseFormReturn | null>(null);
 HookFormContext.displayName = 'HookFormContext';
-
-/**
- * Separate context for `control` to prevent unnecessary rerenders.
- * Internal hooks that only need control use this instead of full form context.
- */
-const HookFormControlContext = React.createContext<Control | null>(null);
-HookFormControlContext.displayName = 'HookFormControlContext';
 
 /**
  * This custom hook allows you to access the form context. useFormContext is intended to be used in deeply nested structures, where it would become inconvenient to pass the context as a prop. To be used with {@link FormProvider}.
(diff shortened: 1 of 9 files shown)


src/useFormContext.tsx
```diff
@@ -46,6 +58,20 @@ export const useFormContext = <
...
+  TFieldValues extends FieldValues,
+  TContext = any,
+  TTransformedValues = TFieldValues,
+>(): Control<TFieldValues, TContext, TTransformedValues> =>
+  React.useContext(HookFormControlContext) as Control<
+    TFieldValues,
+    TContext,
+    TTransformedValues
+  >;
+
```
bluebill1049: we could just move this into a diff file.

bluebill1049 (commented):
could u increase the bundle size config at the package.json? nice work

src/useFormContext.tsx
newsiberian: do you mean you'd prefer to see this hook within a separate `src/useFormControlContext.ts`?

bluebill1049: yes plz. 🙏

newsiberian: done. Thank you for you hard work on this project. 🙏  I've increased the bundle watch limit w/ small reserve for the future changes

bluebill1049 (approved):

mattsputnikdigital:
Quick question about this change.

Previously I was passing in an additional function to the <FormProvider> which I could get back from useFormContext(). This change breaks that. Would there be a way to still allow this?

newsiberian:
> Quick question about this change.
> 
> Previously I was passing in an additional function to the which I could get back from useFormContext(). This change breaks that. Would there be a way to still allow this?

Could you consider using one more custom context for your function? You can create your own context as a one more wrapper around forms.

PR #13238, opened 2026-01-07
🧶 feat: add `isFocused` function to check if field has focus

chrisivo:
# feat: add `isFocused` function to check if field has focus

## Summary
- Adds new `isFocused(name)` function to `useForm` return value that checks whether a registered field currently has focus
- Returns `true` if the field is focused, `false` otherwise (including for unregistered fields or SSR)
- Supports all field types including radio button groups and nested field names

## Motivation
This provides a way to programmatically check focus state without needing to wire up separate `onFocus`/`onBlur` handlers and manage state manually. Useful for conditional logic, analytics, or UI feedback based on focus state.

## Example usage
```tsx
const { register, isFocused } = useForm();

// Check if field has focus
if (isFocused('email')) {
  // field is currently focused
}
```

## Test plan
- [x] Returns `true` when field has focus, `false` when blurred
- [x] Returns `false` for unregistered fields
- [x] Works with radio button groups (checks first ref)
- [x] Works with nested field names (e.g., `user.name`)


bluebill1049:
thanks for the PR, we have no plan to include this feature atm. 🙏

PR #13240, opened 2026-01-08
fix: controller onChange stops working after reset when accessing for…

### LINKED ISSUE #12337: issue: when some properties on formstate from a controller is accessed, it makes other controllers onChange stop working
**Version Number:**
7.53.0

**Codesandbox Link:**
https://codesandbox.io/p/sandbox/pedantic-monad-94vg9g?workspaceId=1928a77f-2690-44ab-88fd-f1021e363df6

**Steps to Reproduce:**
1. Open the provided [Codesandbox link](https://codesandbox.io/p/sandbox/pedantic-monad-94vg9g?workspaceId=1928a77f-2690-44ab-88fd-f1021e363df6).
2. Open the terminal for the preview.
3. Click the "Change Name" button.
4. Observe that the form is submitted with the default values, not the updated name.

**Description:**
I tested multiple versions and determined the bug was introduced in version 7.51.0. The issue occurs when attempting to submit a form after updating the field values.

I noticed the form submits with the default values instead of the updated values, specifically with the name field in this example. While I have not tested all properties on formState, I have observed this bug with several properties, although it does not affect all of them.

Additionally, I attempted to use the <Controller> component, but the issue persists when accessing properties on formState through this component as well.

ca054a3a-4ea2-4b93-9c35-50239c8ec66f

**Expected Behaviour:**
The form should submit with the updated name value after clicking the "Change Name" button.

**Browsers Tested:**
Chrome

**Relevant Log Output:**
No response.

### Code of Conduct

masoombaba16:
## What
Fixes a regression where Controller fields stop responding to onChange
after reset when formState properties are accessed.

## Why
Reset did not emit required formState flags, causing shouldRenderFormState
to return false and preventing Controller re-registration.

## How
Ensures reset emits required formState flags and adds a regression test.

Fixes #12337


masoombaba16:
I’ve opened a PR that fixes this regression and adds a regression test to
prevent it from reoccurring. Happy to iterate if needed.

PR #13241, opened 2026-01-09
Fix : ensure the isValidating is set when trigger is called programma…

### LINKED ISSUE #13218: issue: methods.trigger() wont toggle "isValidating" to true when called programaticaly
### Version Number

7.44.3,  7.69.0

### Codesandbox/Expo snack

.

### Steps to reproduce

tested with validation from yup resolver

```
<div onClick={() => formMethods.trigger()}>
    THIS WORKS OK meaning isValidating is toggled  false -> true -> false in the nested components
</div>

<div
   onClick={() => {
       setTimeout(() => formMethods.trigger(), 1000);
    }}>  THIS DOES NOT WORK, meaning validation is triggered, errors are calculated properly but the "isValidating" is called only once with "false" value at the end of the process.</div>
```

### Expected behaviour

```
<div onClick={() => formMethods.trigger()}>
    THIS WORKS OK meaning isValidating is toggled  false -> true -> false in the nested components
</div>

<div
   onClick={() => {
       setTimeout(() => formMethods.trigger(), 1000);
    }}>  THIS DOES NOT WORK, meaning validation is triggered, errors are calculated properly but the "isValidating" is called only once with "false" value at the end of the process.</div>
```

_No response_

### Relevant log output

```shell

```

### Code of Conduct

Karan7687:
Fixes an issue where calling trigger() asynchronously (e.g. via setTimeout or useEffect)
does not toggle isValidating to true.

This ensures isValidating is set when trigger() starts and reliably reset after validation.

Closes #13218

### DIFF
--- a/src/__tests__/useForm/formState.test.tsx
+++ b/src/__tests__/useForm/formState.test.tsx
@@ -270,6 +270,45 @@ describe('formState', () => {
     });
   });
 
+  //added
+  it('should toggle isValidating when trigger is called asynchronously', async () => {
+    jest.useFakeTimers();
+
+    const resolver = jest.fn(async () => ({
+      values: {},
+      errors: {},
+    }));
+
+    const { result } = renderHook(() =>
+      useForm({
+        resolver,
+      }),
+    );
+
+    expect(result.current.formState.isValidating).toBe(false);
+
+    act(() => {
+      setTimeout(() => {
+        result.current.trigger();
+      }, 100);
+    });
+
+    act(() => {
+      jest.advanceTimersByTime(100);
+    });
+
+    // wait until validation completes
+    await waitFor(() =>
+      expect(result.current.formState.isValidating).toBe(false),
+    );
+
+    // resolver must have been called
+    expect(resolver).toHaveBeenCalled();
+
+    jest.useRealTimers();
+  });
+
+  //here
   it('should be a proxy object that returns undefined for unknown properties', () => {
     const { result } = renderHook(() => useForm());
 
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -900,50 +900,58 @@ export function createFormControl<
     let validationResult;
     const fieldNames = convertToArrayPayload(name) as InternalFieldName[];
 
-    if (_options.resolver) {
-      const errors = await executeSchemaAndUpdateState(
-        isUndefined(name) ? name : fieldNames,
-      );
+    // works for async triggr
+    _updateIsValidating(fieldNames, true);
 
-      isValid = isEmptyObject(errors);
-      validationResult = name
-        ? !fieldNames.some((name) => get(errors, name))
-        : isValid;
-    } else if (name) {
-      validationResult = (
-        await Promise.all(
-          fieldNames.map(async (fieldName) => {
-            const field = get(_fields, fieldName);
-            return await executeBuiltInValidation(
-              field && field._f ? { [fieldName]: field } : field,
-            );
-          }),
-        )
-      ).every(Boolean);
-      !(!validationResult && !_formState.isValid) && _setValid();
-    } else {
-      validationResult = isValid = await executeBuiltInValidation(_fields);
+    try {
+      if (_options.resolver) {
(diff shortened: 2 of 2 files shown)


bluebill1049:
could u include some tests for this?

Karan7687:
Updated the Test to avoid some timing issues and still covering the async trigger behavior.
Tests are passing now, thanks for the suggestion!

src/__tests__/useForm/formState.test.tsx
```diff
@@ -270,6 +270,45 @@ describe('formState', () => {
     });
   });
 
+  //added
```
bluebill1049, line 273:
```suggestion
```

src/__tests__/useForm/formState.test.tsx
```diff
@@ -270,6 +270,45 @@ describe('formState', () => {
...
+      expect(result.current.formState.isValidating).toBe(false),
+    );
+
+    // resolver must have been called
+    expect(resolver).toHaveBeenCalled();
+
+    jest.useRealTimers();
+  });
+
+  //here
```
bluebill1049, line 311:
```suggestion
```

src/logic/createFormControl.ts
```diff
@@ -900,50 +900,58 @@ export function createFormControl<
...
+        iterateFieldsByAction(
+          _fields,
+          _focusInput,
+          name ? fieldNames : _names.mount,
+        );
+
+      return validationResult;
+    } finally {
+      _updateIsValidating(fieldNames);
+    }
```
bluebill1049, line 954: i don't think we will need the try and catch, because rhf shouldn't shallow errors and user land should sort it out.

Karan7687: I have taken out the try/catch and allowed errors to propagate up,to be dealt with on the user level. The isValidating flag is still being flipped as it should, via the resolver lifecycle methods. Take a look and see if that looks okay

Karan7687:
I have addressed the feedback,added the coverage for the async trigger case and merged the latest master.

tswirski:
Because what we need is to have valid isValidating flag after *trigger* called programmatically for entire form - We've work-around that in following way.

` const formStateProxyFactory = useRef(() => {
    return {
      get(obj, prop) {
        if (prop === 'isValidating') {
          return selfIsValidating.current;
        }
        return obj[prop];
      },
    };
  });

  const formMethodsProxyFactory = useRef(() => {
    let triggerFn;

    const triggerProxy = (name: string, options: any): Promise<any> => {
      if (name) {
        return triggerFn?.(name, options);
      }
      return new Promise(resolve => {
        selfIsValidating.current = true;
        requestAnimationFrame(() =>
          flushSync(async () => {
            const result = await triggerFn(name, options);
            selfIsValidating.current = false;
            resolve(result);
          })
        );
      });
    };

    return {
      get(obj, prop) {
        if (prop === 'trigger') {
          triggerFn = obj[prop];
          return triggerProxy;
        }

        if (prop === 'formState') {
          if (!formStateProxy.current) {
            formStateProxy.current = new Proxy(obj[prop], formStateProxyFactory.current());
          }
          return formStateProxy.current;
        }

        return obj[prop];
      },
    };
  });

  useEffect(() => {
    selfIsValidating.current = methods.formState.isValidating;
  }, [methods.formState.isValidating]);

  useEffect(() => {
    formStateProxy.current = null;
  }, [methods.formState]);

  useEffect(() => {
    setFormMethodsProxy(new Proxy(methods, formMethodsProxyFactory.current()));
  }, [methods]);
`

It doesn't throw error for flushSync, and it works as we expect it to work (so far).

Karan7687:
Thanks for sharing this, that is a clever workaround.

my intention with this change was to just make trigger() work consistently at the core level.Consumers should not have to rely on Proxies,flushSync or RAF to see the correct isValidating state.

Since trigger() already returns a promise and starts async validation internally settings isValidating at the beginning of that process seemed like a reasonable and predictable default.The test i added just confirms that behavior.

i am open to changing it if you think this should be outside the core i am happy to work together.

PR #13245, opened 2026-01-10
docs: clarify native HTML validation behavior

ZakirAliyev:
Clarifies that native HTML validation is disabled by default
and must be explicitly enabled via shouldUseNativeValidation.

Docs-only change.


bluebill1049:
thanks, more details on the api have moved to doc instead.

PR #13258, opened 2026-01-18
fix: consistent error path for field array validation errors

NidhiPednekar:
## Issue
Error paths were inconsistent between `trigger()` and `handleSubmit()` when validating field arrays:
- `trigger('items')` returned `error.message`
- `handleSubmit()` returned `error.root.message`

## Root Cause
The `schemaErrorLookup.ts` function was appending `.root` suffix to field array error paths, causing inconsistency.

## Solution
Modified `schemaErrorLookup.ts` to return consistent field names without the `.root` suffix for both validation methods.

## Changes
- Fixed error path lookup in `src/logic/schemaErrorLookup.ts`
- Updated test expectations in `src/__tests__/logic/schemaErrorLookup.test.ts`

## Testing
- All 1033 tests pass 
- Related tests verified and updated
- No breaking changes

### DIFF
--- a/src/__tests__/logic/schemaErrorLookup.test.ts
+++ b/src/__tests__/logic/schemaErrorLookup.test.ts
@@ -265,7 +265,7 @@ describe('errorsLookup', () => {
       ),
     ).toEqual({
       error: { message: 'correct-root', type: 'root' },
-      name: 'test.0.nested.root',
+      name: 'test.0.nested',
     });
 
     expect(
--- a/src/__tests__/logic/schemaErrorLookup.test.ts
+++ b/src/__tests__/logic/schemaErrorLookup.test.ts
@@ -279,4 +279,104 @@ describe('errorsLookup', () => {
       name: 'test.0.nested.0.deepNested',
     });
   });
+
+  it('should return consistent error path for field array validation (issue #13258)', () => {
+    const fieldArrayError = {
+      type: 'custom',
+      message: 'at_least_1_item',
+      ref: {},
+    };
+
+    expect(
+      schemaErrorLookup(
+        {
+          items: fieldArrayError as any,
+        },
+        {
+          items: [],
+        },
+        'items',
+      ),
+    ).toEqual({
+      error: fieldArrayError,
+      name: 'items',
+    });
+
+    expect(
+      schemaErrorLookup(
+        {
+          items: fieldArrayError as any,
+        },
+        {
+          items: [],
+        },
+        'items.field',
+      ),
+    ).toEqual({
+      error: fieldArrayError,
+      name: 'items',
+    });
+
+    expect(
+      schemaErrorLookup(
+        {
+          items: fieldArrayError as any,
+        },
+        {
+          items: [],
+        },
+        'items.0.field',
+      ),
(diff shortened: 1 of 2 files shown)


bluebill1049:
can you share a codesandbox what's the issue on this? and if can add some unit tests coverage would be great thanks.

NidhiPednekar:
> can you share a codesandbox what's the issue on this? and if can add some unit tests coverage would be great thanks.

Thanks for the review!!

## Problem Demonstration
The issue is demonstrated in the CodePen from the original issue:
https://codepen.io/ferhatavdic/pen/KwVZPNq

When items array is empty:
- `trigger('items')` returns: `error.message: "at_least_1_item"` 
- `handleSubmit()` returns: `error.root.message: "at_least_1_item"`  -->INCONSISTENT

## Solution
Modified `src/logic/schemaErrorLookup.ts` to return consistent error paths. Now both methods return `error.message`.

## Test Coverage
Added 3 comprehensive unit tests in `src/__tests__/logic/schemaErrorLookup.test.ts`:
1. `should return consistent error path for field array validation (issue #13258)`
2. `should handle field array errors consistently with trigger() and handleSubmit()`
3. `should not wrap field array errors in root property`

All 7 tests pass 

## Verification
Run: `npm test -- src/__tests__/logic/schemaErrorLookup.test.ts`
Result: ✅ 7 passed, 0 failed

The test cases prove the fix works correctly and prevents regression.

src/__tests__/logic/schemaErrorLookup.test.ts
```diff
@@ -265,7 +265,7 @@ describe('errorsLookup', () => {
       ),
     ).toEqual({
       error: { message: 'correct-root', type: 'root' },
-      name: 'test.0.nested.root',
+      name: 'test.0.nested',
     });
```
bluebill1049, line 269: thanks for the added tests and codesandbox, if the error type is looking for root error, then the test case reflect correct?

NidhiPednekar: yes, the test cases are correct . The original test checks behavior when a root error is present, while my tests focus on field array errors without a root error

bluebill1049: so why we have to change the test to get it pass? I think this would cause a regression.

NidhiPednekar: yep, that test change is the fix. I removed the .root suffix so the test was updated to match how it works now

bluebill1049: Sorry for not being clear earlier. In

`it('should find the nearest root error', () => {`

the original test correctly identified the root error. The change to test.0.nested instead points to its parent, which is the regression I’m referring to here (see screenshot).

Since this alters the resolved error path, it would be considered a breaking change.

image

NidhiPednekar: oh got it. thanks for correcting me. now I have updated it to skip .root only for field arrays, while keeping it for real root errors. Could you please take a look and confirm if this works correctly now?

PR #13279, opened 2026-02-08
fix: emit field name in clearErrors to enable shouldSubscribeByName filtering

### LINKED ISSUE #13277: issue: clearErrors(name) does not emit signal name -> breaks shouldSubscribeByName filtering, causes broad rerenders
### Version Number

7.71.1

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/reverent-dawn-27ln9h

### Steps to reproduce

1. Create a form with at least two fields (e.g. `a` and `b`) using `useForm`.
2. Render two child components which each call `useFormState({ control, name: '<fieldName>', exact: true })` and access `formState.errors` (e.g. to display / compute a field error).
3. Trigger an error on field `a` (e.g. required / minLength) so `errors.a` exists.
4. Call `clearErrors('a')` (or via a button).
5. Observe: both components re-render / receive the update, not just the one for `a`.

### Expected behaviour

Only subscribers for the cleared field(s) should be notified. In other words, `clearErrors('a')` should not cause components subscribed to other field names (e.g. `b`) to re-render, when `exact: true` is used.

_No response_

### Relevant log output

```shell
`clearErrors('a')` emits a formState update without a `name` signal, so `shouldSubscribeByName(...)` cannot filter and all `errors` subscribers are notified, causing broad re-renders.
```

### Code of Conduct

veeceey:
## Summary
Fixes #13277

When `clearErrors(name)` is called with a single field name, it now includes the `name` in the state update notification. This allows `shouldSubscribeByName` to properly filter subscribers, preventing unnecessary re-renders for components that subscribe to other fields with `exact: true`.

## Changes
- Modified `clearErrors` function in `src/logic/createFormControl.ts` to emit the field name when clearing a single error
- When clearing multiple errors or all errors, continues to emit without a name (maintains backward compatibility)

## Behavior
**Before:**
```typescript
// Component subscribed to field 'b' with exact: true
clearErrors('a') // Both 'a' and 'b' components re-render
```

**After:**
```typescript
// Component subscribed to field 'b' with exact: true  
clearErrors('a') // Only 'a' component re-renders
```

## Testing
- All existing tests pass (1036 tests)
- TypeScript type check passes
- No breaking changes, as the function continues to work for all existing use cases

## Note
This change makes `clearErrors` consistent with `setError`, which already includes the name in its state notification for single-field operations.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -982,14 +982,30 @@ export function createFormControl<
   });
 
   const clearErrors: UseFormClearErrors<TFieldValues> = (name) => {
-    name &&
-      convertToArrayPayload(name).forEach((inputName) =>
-        unset(_formState.errors, inputName),
-      );
+    if (name) {
+      const names = convertToArrayPayload(name);
+      names.forEach((inputName) => {
+        unset(_formState.errors, inputName);
+      });
 
-    _subjects.state.next({
-      errors: name ? _formState.errors : {},
-    });
+      // Emit notification for each cleared field when array is provided,
+      // or single notification when a single field name is provided
+      if (names.length === 1) {
+        _subjects.state.next({
+          name: names[0],
+          errors: _formState.errors,
+        });
+      } else {
+        // For multiple names, emit without name to notify all subscribers
+        _subjects.state.next({
+          errors: _formState.errors,
+        });
+      }
+    } else {
+      _subjects.state.next({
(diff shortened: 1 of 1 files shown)


veeceey:
Closing this PR as the fix has already been merged via #13280. Thanks!

bluebill1049:
Appreciate the effort on the fix!

PR #13280, opened 2026-02-09
fix(clearErrors): emit name signal for targeted field updates

### LINKED ISSUE #13277: issue: clearErrors(name) does not emit signal name -> breaks shouldSubscribeByName filtering, causes broad rerenders

kaigritun:
## Problem

When `clearErrors(name)` is called with a specific field name, the state emission does not include the `name` property. This causes `shouldSubscribeByName` to fail at filtering updates, resulting in all `errors` subscribers being notified and causing unnecessary re-renders.

Fixes #13277

## Solution

When clearing specific field error(s), emit the state update with the `name` property for each cleared field. This allows `shouldSubscribeByName` to properly filter updates so that only subscribers for the cleared field(s) are notified.

**Before:**
```typescript
_subjects.state.next({
  errors: name ? _formState.errors : {},
});
```

**After:**
```typescript
if (names) {
  // Emit for each cleared field with the field name
  names.forEach((inputName) => {
    _subjects.state.next({
      name: inputName,
      errors: _formState.errors,
    });
  });
} else {
  // Clear all errors - emit without name to notify all subscribers
  _subjects.state.next({
    errors: {},
  });
}
```

## Changes

1. Updated `clearErrors` in `createFormControl.ts` to emit with `name` property for targeted clears
2. Added test case verifying that `clearErrors('a')` doesn't cause components subscribed to field `'b'` with `exact: true` to re-render

## Testing

- All existing tests pass (1037 tests)
- New test case specifically validates the fix for #13277

### DIFF
--- a/src/__tests__/useForm/clearErrors.test.tsx
+++ b/src/__tests__/useForm/clearErrors.test.tsx
@@ -8,7 +8,9 @@ import {
   waitFor,
 } from '@testing-library/react';
 
+import type { Control } from '../../types';
 import { useForm } from '../../useForm';
+import { useFormState } from '../../useFormState';
 
 describe('clearErrors', () => {
   it('should remove error', () => {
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):

MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

bluebill1049:
Thanks for the heads-up! Much appreciated.

PR #13289, opened 2026-02-24
🍖 enhance getEventValue to handle file inputs

ZUR1C4T0:


bluebill1049:
what's the use case try to improve here?

ZUR1C4T0:
> what's the use case try to improve here?

## Context

I am migrating my project from native HTML elements to :contentReference[oaicite:0]{index=0} components. Because of this, I changed from using the `register` method of `useForm()` to using the `Controller` component.

When I migrated the file input field, I found that its value was not updated correctly in the form. This caused an error in my zod schema, which was working fine before. Previously, the input returned a `FileList` object when using `register`.

## Problem

While debugging, I followed the execution flow from the `onChange` method and found a function that extracts the value from the event.

This function works correctly for checkbox inputs (using `event.target.checked`), but not for file inputs.

For file inputs, the correct value is in `event.target.files`, but since the input is not a checkbox, the function returns `event.target.value` by default.

## Workaround in my code

To fix this in my project, I had to manually pass the correct value (`event.target.files`) to the `onChange` handler:

```tsx
<Controller
  name="logo"
  control={form.control}
  render={({ field: { value, onChange, ...field }, fieldState }) => (
    <Field data-invalid={fieldState.invalid}>
      <FieldLabel htmlFor="logo">
        Logo de la feria <span className="text-destructive">*</span>
      </FieldLabel>
      <Input
        {...field}
        type="file"
        id="logo"
        accept="image/png, image/jpeg, image/jpg"
        aria-invalid={fieldState.invalid}
        onChange={(e) => onChange(e.target.files)} // <-- HERE
        required
      />
      {!fieldState.invalid ? (
        <FieldDescription>Tamaño máximo de 5MB.</FieldDescription>
      ) : (
        <FieldError errors={[fieldState.error]} />
      )}
    </Field>
  )}
/>
```

## Proposal

This pull request updates the value extraction logic to correctly handle file inputs by using `event.target.files` instead of `event.target.value`.

## Notes

I am not fully sure if this change could introduce conflicts in other cases, so this is just a proposal.

The change passed the existing tests, but I am not sure if we should add new tests to cover scenarios like file inputs with Controller.

bluebill1049 (approved):

PR #13305, opened 2026-03-13
test(utils): add unit tests for isDateObject

### LINKED ISSUE #13304: Add unit tests for isDateObject utility
### Is your feature request related to a problem? Please describe.

I noticed that `src/utils/isDateObject.ts` does not have unit test coverage, while many other utility functions in the repository include dedicated tests.

This utility checks whether a given value is a Date object, and adding tests would help ensure consistent behavior and improve overall test coverage.

### Describe the solution you'd like

Add a new test file:

src/__tests__/utils/isDateObject.test.ts

The tests will verify:

- Returns `true` for Date objects
- Returns `false` for non-Date values such as:
  - string
  - number
  - object
  - null
  - undefined
  - array
  - function

Example:

```ts
import isDateObject from '../../utils/isDateObject';

describe('isDateObject', () => {
  it('should return true for Date objects', () => {
    expect(isDateObject(new Date())).toBe(true);
  });

  it('should return false for non-Date values', () => {
    expect(isDateObject('2021-01-01')).toBe(false);
  });
});

Venkat-Entropik:
Adds unit tests for the `isDateObject` utility.

The tests verify that the function:
- Returns `true` for Date objects
- Returns `false` for non-Date values (string, number, object, null, undefined, array, function)

This improves test coverage and ensures consistent behavior of the utility.

Closes #13304

### DIFF
--- a/src/__tests__/utils/isDateObject.test.ts
+++ b/src/__tests__/utils/isDateObject.test.ts
@@ -1,0 +1,17 @@
+import isDateObject from '../../utils/isDateObject';
+
+describe('isDateObject', () => {
+  it('should return true when value is a Date object', () => {
+    expect(isDateObject(new Date())).toBe(true);
+  });
+
+  it('should return false when value is not a Date object', () => {
+    expect(isDateObject('2021-01-01')).toBe(false);
+    expect(isDateObject(1609459200000)).toBe(false);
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks for the test, i don't think we need it for now. 🙏

PR #13312, opened 2026-03-15
fix: memoize control in HookFormControlContext to prevent render conflicts (#13272)

olagokemills:
## Proposed Changes

This PR resolves a race condition introduced in v7.71.0 where [Controller](cci:1://file:///Users/djunehor/Sites/react-hook-form/src/useController.ts:25:0-260:1) components could receive mismatched context values mid-render, leading to the error: `'Cannot update a component (Controller) while rendering a different component (Controller)'`.

The issue occurred when swapping between different [FormProvider](cci:1://file:///Users/djunehor/Sites/react-hook-form/src/useFormContext.tsx:54:0-157:2) instances (e.g., conditionally rendering between 'Submit' mode and 'Review' mode views). Previously, `HookFormControlContext` was receiving the `control` object separately from the main `HookFormContext` memoization, causing a timing mismatch during React reconciliation.

The fix ensures atomic context updates by deriving the `control` value from the same memoized object used for the main `HookFormContext`. I've also rewritten the regression tests to accurately simulate the reported bug by conditionally rendering separate [FormProvider](cci:1://file:///Users/djunehor/Sites/react-hook-form/src/useFormContext.tsx:54:0-157:2) instances.

Fixes #13272

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:


bluebill1049:
can you check out the comments?

olagokemills:
Hi @bluebill1049, I've addressed the review comments in the latest commit:

- **Atomicity of Context Updates**: Refactored [FormProvider](cci:1://file:///Users/djunehor/Sites/react-hook-form/src/useFormContext.tsx:54:0-157:2) to derive the `control` context value from the same `useMemo` block used for the main form context. This ensures both contexts update atomically during the same render, correctly addressing the timing mismatch without using a no-op `useMemo`.
- **Updated Regression Test**: Rewrote the test to conditionally render two separate [FormProvider](cci:1://.src/useFormContext.tsx:54:0-157:2) instances. This correctly simulates the reconciliation path reported in #13272, which was previously missed by just swapping props on a single instance.
- **Validation Complete**: Verified locally that all tests pass and the "Cannot update a component while rendering" warnings are no longer thrown in both the conditional-rendering and prop-swapping scenarios.

bluebill1049:
Thanks, i will check.

bluebill1049 (approved):

idrisakmal:
thank you @bluebill1049 @olagokemills for this PR. Looking forward for the next release

PR #13318, opened 2026-03-19
🐞 fix: isNameInFieldArray should check all ancestor paths for nested field arrays

shahmir-oscilar:
## Summary

Fixes #13317

`isNameInFieldArray` only checks the first `.\d+` segment via `getNodeParentName`, which returns the wrong parent for field arrays nested under numeric path segments.

For a Controller at `steps.0.items.2.name`, `getNodeParentName` returns `steps` (first `.\d+` match) instead of `steps.0.items`. Since `steps` is not in the field array name set, `isArrayField` is `false`, and the `useController` name-change cleanup from #13041 incorrectly unregisters the shifted field path — wiping the form store values for that item.

Top-level field arrays (e.g., `items.0.name` → parent `items`) are unaffected because the first `.\d+` segment happens to be the correct one.

## Fix

Changed `isNameInFieldArray` to iterate over all `.\d+` segments in the path and check each prefix against the field array name set, instead of only checking the first.

## Reproduction & verification

**Bug repro (official 7.63.0+):** https://codesandbox.io/p/sandbox/fjl9t6

**Fix verified:** https://codesandbox.io/p/sandbox/react-hook-form-data-loss-bug-fixed-3x2g3d

---

*Co-authored with [Claude Code](https://claude.ai/code)*


shahmir-oscilar:
@bluebill1049 Could you take a look at this when you get a chance?

This issue affected our code when we upgraded from 7.51.2 to latest (the repro is based on our actual usage), and we had to downgrade to 7.62.0 to resolve it.

Thank you so much!

src/logic/isNameInFieldArray.ts
```diff
@@ -1,6 +1,18 @@
...
+  const pattern = new RegExp(FIELD_ARRAY_INDEX_PATTERN.source, 'g');
+  let match;
+  while ((match = pattern.exec(name)) !== null) {
+    const parentName = name.substring(0, match.index);
+    if (names.has(parentName)) {
+      return true;
+    }
+  }
+  return false;
+};
```
bluebill1049:
```suggestion
export default (names: Set<InternalFieldName>, name: InternalFieldName) =>
  name
    .split('.')
    .some(
      (part, i, arr) =>
        !isNaN(Number(part)) && names.has(arr.slice(0, i).join('.')),
    );
```

src/logic/getNodeParentName.ts
```diff
@@ -1,2 +1,4 @@
+export const FIELD_ARRAY_INDEX_PATTERN = /\.\d+(\.|$)/;
+
 export default (name: string) =>
-  name.substring(0, name.search(/\.\d+(\.|$)/)) || name;
+  name.substring(0, name.search(FIELD_ARRAY_INDEX_PATTERN)) || name;
```
bluebill1049: we can revert this change.

src/logic/isNameInFieldArray.ts
shahmir-oscilar: Done!

src/logic/getNodeParentName.ts
shahmir-oscilar: Done

bluebill1049 (approved):

PR #13322, opened 2026-03-21
🐞 fix: #13320 formState.isValid incorrect on Controller re-mount

candymask0712:
to resolve #13320

Hello. I have created this PR to fix the issue where formState.isValid stays incorrect when a Controller
with validation rules re-mounts. 😁

### Issue

`formState.isValid` remains true when a Controller with a required rule unmounts and re-mounts via
conditional rendering (e.g., checkbox toggle).

### Cause

The register function's if (field) branch (re-registration path) only calls `_setDisabledField()` — it never
calls `_setValid()`. So after re-mount, the form's validity is not re-evaluated.

### Solution

Added `_setValid()` call in the register function when a previously unmounted field is re-registered. A
`_registerFromRef` flag prevents false positives from the ref callback cycle during normal re-renders.

### Changes

- Added re-mount detection and `_setValid()` call in register function (`createFormControl.ts`)
- Added `_registerFromRef` flag with try-finally to distinguish ref callback calls
- Added test case reproducing the exact issue scenario

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -1514,4 +1514,78 @@ describe('useController', () => {
     rerender({ control: form1Result.current.control });
     expect(result.current.field.value).toBe('form1-value');
   });
+
+  it('should update isValid when Controller with required rule re-mounts via checkbox toggle', async () => {
+    type FormValues = {
+      items: { checked: boolean; input: string }[];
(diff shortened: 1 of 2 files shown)


bluebill1049:
huge effort on finding out the root cause and giving a fix! Thank you so much @candymask0712

PR #13299, opened 2026-03-10
🐞 fix: prevent useFieldArray from marking unrelated fields as dirty

6810779s:
## Proposed Changes

**Problem**

When `useFieldArray` methods (`append`, `remove`, etc.) are called, unrelated fields (`name`, `age`, etc.) incorrectly appear in `dirtyFields`, even though they were never modified by the user.

This happens because `_setFieldArray` replaces the **entire** `_formState.dirtyFields` object with the result of `getDirtyFields(_defaultValues, _formValues)`, which includes `false` entries for all non-dirty fields.

```js
// Before: after append() — name and age were never touched
dirtyFields = { name: false, age: false, items: [{ value: true }] }
```

**Solution**
Instead of replacing the entire `dirtyFields` object, only update the root field array portion using `set()` and `get()`. For nested field arrays (e.g., `test.0.keyValue`), the root name (`test`) is extracted to ensure sibling fields within the same array item are properly included in `dirtyFields`, while unrelated top-level fields remain unaffected.

```js
// After: only the field array portion is updated
dirtyFields = { items: [{ value: true }] }
```

**Changes**

1. `src/logic/createFormControl.ts` — In `_setFieldArray`, compute full dirty fields but only apply the root field array path to `_formState.dirtyFields`
2. `src/__tests__/useFieldArray/append.test.tsx` — Add test for append with unrelated fields
3. `src/__tests__/useFieldArray/remove.test.tsx` — Add test for remove with unrelated fields
4. `src/__tests__/useFieldArray/dirtyFields.test.tsx` — Add 10 comprehensive test scenarios covering single/multiple field arrays, nested paths, large forms, and mixed field modifications

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Fixes

Fixes #13054

Also addresses #11402, #7197

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing tests pass locally with my changes

PR #13326, opened 2026-03-23
fix: prevent setValue with shouldDirty from polluting unrelated dirty fields

6810779s:
## Summary
- When using `setValue` with `shouldDirty: true` on a field array, unrelated fields were incorrectly marked as dirty
- Root cause: `setValue`'s field array branch was emitting the full `getDirtyFields()` result, which overwrites `_formState.dirtyFields` entirely
- Applied the same fix already used in `_setFieldArray` — selectively update only the relevant root branch using `getNodeParentName()`

## Related
- Follow-up to #13054
- Reported in https://github.com/react-hook-form/react-hook-form/pull/13299#issuecomment-4110254692

## Test plan
- [x] Added test: `setValue` with `shouldDirty` on field array should not mark unrelated fields dirty
- [x] Added test: `setValue` with `shouldDirty` should preserve existing dirty fields
- [x] All existing tests pass (1053/1053)

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -806,9 +806,13 @@ export function createFormControl<
           _proxySubscribeFormState.dirtyFields) &&
         options.shouldDirty
       ) {
+        const fullDirtyFields = getDirtyFields(_defaultValues, _formValues);
+        const rootName = getNodeParentName(name);
+        set(_formState.dirtyFields, rootName, get(fullDirtyFields, rootName));
+
         _subjects.state.next({
           name,
-          dirtyFields: getDirtyFields(_defaultValues, _formValues),
+          dirtyFields: _formState.dirtyFields,
           isDirty: _getDirty(name, cloneValue),
         });
       }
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
Thank you again for the fix!

src/logic/createFormControl.ts
```diff
@@ -806,9 +806,13 @@ export function createFormControl<
           _proxySubscribeFormState.dirtyFields) &&
         options.shouldDirty
       ) {
+        const fullDirtyFields = getDirtyFields(_defaultValues, _formValues);
+        const rootName = getNodeParentName(name);
+        set(_formState.dirtyFields, rootName, get(fullDirtyFields, rootName));
+
```
bluebill1049, line 812: one minor improvement, i wonder if can turn this into a function within createFormContorl, then it can be reused.

bluebill1049: I will take care of that.

6810779s: Thank you! 🙏

PR #13338, opened 2026-03-27
fix: allow revalidation on fields with existing errors after trigger()

restareaByWeezy:
Closes #13276
Related: https://github.com/orgs/react-hook-form/discussions/11422

When `trigger()` is called manually to validate fields, it shows errors but doesn't set `isSubmitted` to `true`. After that, typing in the field doesn't clear the error because `skipValidation` uses the pre-submit `mode` (e.g. `onSubmit`) which skips onChange validation entirely.

The core problem is that `skipValidation` decides whether to revalidate based on `isSubmitted`, but `trigger()` doesn't set that flag (and it shouldn't — it's not a submission). So fields with errors from `trigger()` get stuck.

The fix: when a field already has an error AND the form hasn't been submitted yet, check if the current event matches `reValidateMode` (default: `onChange`). If it does, don't skip — let the validation run so the error can be cleared.

This only kicks in when:
1. The field has an existing error (e.g. from `trigger()`)
2. `isSubmitted` is `false` (after actual submission, the existing logic handles it)
3. The event type matches `reValidateMode`

So normal form behavior is unchanged — this just unblocks the specific case where `trigger()` errors are stuck.

### DIFF
--- a/src/__tests__/useForm/trigger.test.tsx
+++ b/src/__tests__/useForm/trigger.test.tsx
@@ -940,6 +940,90 @@ describe('trigger', () => {
     expect(count).toEqual(2);
   });
 
+  it('should clear error on user input after trigger() sets it when form is not submitted', async () => {
+    const App = () => {
+      const {
+        register,
+        trigger,
+        formState: { errors },
+      } = useForm<{ test: string }>({
+        mode: 'onSubmit',
+        reValidateMode: 'onChange',
+      });
+
+      return (
+        <div>
+          <input
+            {...register('test', { required: 'This field is required' })}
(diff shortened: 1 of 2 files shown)


bluebill1049:
could you include a unit test for this thanks?

restareaByWeezy:
@bluebill1049 Done! Added tests for both onChange and onBlur reValidateMode — covers the trigger() → user input → error cleared flow. All tests green 👍

bluebill1049:
thank you @restareaByWeezy! I am currently discussing this issue in more detail on the original thread.

PR #13343, opened 2026-03-31
test: add test coverage for isDirty behavior when removing the only field array item (fixes #13334)

yuvinraja:
### Summary
Adds test coverage and example demonstrating correct `isDirty` behavior when removing the only item from a field array using `useFieldArray.remove()`.

### Related Issue
Fixes #13334

### Changes Made
1. **Added test case** in remove.test.tsx:
   - Tests that `isDirty` correctly updates to `true` when removing the only default item from a field array
   - Verifies the behavior described in issue #13334 works as expected

2. **Created example component** fieldArrayIsDirty.tsx:
   - Demonstrates proper usage of `useFieldArray` with `isDirty` tracking
   - Documents common pitfalls (e.g., calling `reset()` in `useEffect` interferes with dirty tracking)
   - Provides interactive example showing when `isDirty` changes state

3. **Added route** to example app for easy testing

### Testing
- ✅ All existing tests pass
- ✅ New test case passes and verifies expected behavior
- ✅ Linting passes
- ✅ Example component works correctly in demo app

### Notes
The test confirms that `react-hook-form` correctly handles `isDirty` when removing items from field arrays. The issue reported in #13334 appears to be caused by improper usage patterns rather than a library bug. This PR adds documentation and test coverage to prevent similar confusion in the future.

### DIFF
--- a/app/src/app.tsx
+++ b/app/src/app.tsx
@@ -28,6 +28,7 @@ import UseFieldArray from './useFieldArray';
 import UseFieldArrayNested from './useFieldArrayNested';
 import CustomSchemaValidation from './customSchemaValidation';
 import WatchFieldArray from './watchUseFieldArray';
+import FieldArrayIsDirty from './fieldArrayIsDirty';
 import UseWatch from './useWatch';
 import FormStateWithNestedFields from './formStateWithNestedFields';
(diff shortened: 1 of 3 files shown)


bluebill1049 (approved):

bluebill1049:
thanks for the PR, the original issue has been fixed.

PR #13348, opened 2026-04-01
fix: handle nested field when parent defaultValue is null

### LINKED ISSUE #13345: issue: `register` sets nested field values to `null` when ancestor in `defaultValues` is `null`
### Version Number

7.72.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/react-hook-form-get-started-ts-forked-vm7rnp

### Steps to reproduce

1. Open the devtools console
2. Seeing two console.logs:
```
{ example: null }
```
Then
```
{ example: { inner: null } }
```
3. (Without touching the input) Click "Submit", we get `{"example":{"inner":null}}`

### Expected behaviour

`getValues()` and the submitted value should be `{ example: { inner: undefined } }` (or `inner` key absent)

_No response_

### Relevant log output

```shell

```

### Code of Conduct

Prasadzoman:
Fixes #13345

###  Problem
When a parent field in `defaultValues` is `null`, registering a nested field (e.g. `example.inner`) results in incorrect behavior.

The nested value is not properly initialized.

### Solution
- Preserved existing traversal logic
- Fixed handling of `null` parent values
- Ensures nested field can be registered correctly

###  Tests
- Added test to reproduce issue
- Verified correct output:
  { example: { inner: "" } }

###  Notes
- Matches browser behavior for empty inputs
- No regressions (all tests passing)

### DIFF
--- a/src/utils/get.ts
+++ b/src/utils/get.ts
@@ -13,13 +13,16 @@ export default <T>(
     return defaultValue;
   }
 
-  const result = (isKey(path) ? [path] : stringToPath(path)).reduce(
-    (result, key) =>
-      isNullOrUndefined(result) ? result : result[key as keyof T & object],
-    object,
-  );
+  const paths = isKey(path) ? [path] : stringToPath(path);
 
-  return isUndefined(result) || result === object || result === null
+  const result = paths.reduce<any>((result, key, index) => {
+    if (result === null && index < paths.length) {
+      return undefined;
+    }
+    return isNullOrUndefined(result) ? result : result[key];
+  }, object);
+
+  return isUndefined(result) || result === object
     ? isUndefined(object[path as keyof T])
       ? defaultValue
       : object[path as keyof T]
--- a/src/utils/get.ts
+++ b/src/utils/get.ts
@@ -19,7 +19,7 @@ export default <T>(
     object,
   );
 
-  return isUndefined(result) || result === object
+  return isUndefined(result) || result === object || result === null
     ? isUndefined(object[path as keyof T])
       ? defaultValue
       : object[path as keyof T]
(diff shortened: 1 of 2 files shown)


Prasadzoman:
Hi, I’ve pushed updates addressing the review feedback. Could you please take another look when you get a chance?

src/utils/get.ts
```diff
@@ -13,11 +13,14 @@ export default <T>(
...
-      isNullOrUndefined(result) ? result : result[key as keyof T & object],
-    object,
-  );
+  const paths = isKey(path) ? [path] : stringToPath(path);
+
+  const result = paths.reduce<any>((result, key, index) => {
+    if (result === null && index < paths.length) {
+      return undefined;
+    }
+    return isNullOrUndefined(result) ? result : result[key];
```
Yihao-G: Thank you very much for the PR to fix the issue.

Just out of curiosity, In what case would `index < paths.length` be false? I don't think it will ever be false. Therefore, this callback becomes:
```typescript
    if (result === null) {
      return undefined;
    }
    return isNullOrUndefined(result) ? result : result[key];
```

Then that means `isNullOrUndefined(result)` only checks if `result` is `undefined` or not. If it's `undefined`, return the `result` which is `undefined`.

Therefore, this whole callback can just become
```typescript
      isNullOrUndefined(result) ? undefined : result[key]
```
same as [the suggestion mentioned in my original issue](https://github.com/react-hook-form/react-hook-form/issues/13345#issuecomment-4166778431).

Prasadzoman:
Good catch  you're absolutely right here.

index < paths.length will always evaluate to true inside a reduce callback, so that condition is effectively dead code and can be removed.

Following that through:

The null check combined with that condition simplifies to returning undefined for null
Which means both null and undefined should short-circuit to undefined

So the callback can be simplified to:
isNullOrUndefined(result) ? undefined : result[key]

bluebill1049 (approved):

PR #13352, opened 2026-04-03
fix: skip re-render in setValue when value is unchanged

### LINKED ISSUE #13351: issue: Calling setValue with existing value triggers re-render
### Version Number

7.72.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/happy-rgb-kps3dv

### Steps to reproduce

Click "set firstName to empty string". This changes the `firstName` field from `""` to `""`. Though the value is still the same, the component will re-render because of `useWatch`.

### Expected behaviour

The form values have not actually changed. `useWatch` should not trigger a re-render. This also occurs with `watch`,  a `watch` callback function, and `form.subscribe`.

Chrome

### Relevant log output

```shell

```

### Code of Conduct

sujiu222:
## Summary

Fixes #13351

## Problem

When `setValue` is called with a value identical to the current field value, subscribers (`useWatch`, `watch`, `form.subscribe`) are still notified, causing unnecessary re-renders even though the value hasn't actually changed.

```tsx
// This should NOT trigger re-render if firstName is already ''
setValue('firstName', '')
```

## Solution

Added a `deepEqual` comparison in `createFormControl.ts` before notifying value subscribers. If the new value is identical to the current value and no side effects (`shouldDirty`, `shouldTouch`, `shouldValidate`) are requested, the `_subjects.state.next()` notification is skipped.

## Testing

- All existing tests pass (53 setValue tests ✅, 47 useWatch tests ✅)
- Added new test case: `should not notify subscribers when setValue is called with an unchanged value`

## Checklist

- [x] Existing tests pass
- [x] New test added
- [x] No breaking changes

### DIFF
--- a/src/__tests__/useForm/setValue.test.tsx
+++ b/src/__tests__/useForm/setValue.test.tsx
@@ -1532,4 +1532,32 @@ describe('setValue', () => {
 
     expect(nextSpy).toHaveBeenCalledTimes(1);
   });
+
+  it('should not notify subscribers when setValue is called with an unchanged value', async () => {
+    const { result } = renderHook(() =>
+      useForm<{ test: string }>({ defaultValues: { test: 'initial' } }),
+    );
+    const control = result.current.control as any;
+
+    // Set the initial value first
+    await act(async () => {
+      result.current.setValue('test', 'initial');
+    });
+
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
LGTM

PR #13365, opened 2026-04-04
fix: safely access field._f during register 

### LINKED ISSUE #13364: issue: registering field array property then field array root throws error
### Version Number

7.72.1

### Codesandbox/Expo snack

N/A

### Steps to reproduce

1. Register a field array element property. e.g. if `items` is a field array, then register `items.0.id`.
2. Register the field array root. e.g. `items`.

See broken test case below.

```tsx
it('can register field array property before field array root', () => {
  const Component = () => {
    const { control } = useForm<{
      test: string;
      test1: { test: string }[];
    }>();

    useController({
      name: 'test1.0.test',
      control,
    });

    useController({
      name: 'test1',
      control,
    });

    return null;
  };

  render(<Component />);
});
```

### Expected behaviour

It works.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

ap0nia:
Resolves #13364 .

The `register` does not null-check `field._f`, which throws an error if `field` is an array.

```ts
useController({ name: 'test1.0.test' })

// Current value of _fields after first register
const _fields = {
  test1: [
    {
      test: { _f }
    }
  ]
}

useController({ name: 'test1' })

// Logic in `register` causes error.
const field = _fields.test1
const shouldRevalidateRemount = !_names.registerName.has(name) && field && !field._f.mount;

// Fix with one extra check.
const shouldRevalidateRemount = !_names.registerName.has(name) && field && field._f && !field._f.mount;
```

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -1588,4 +1588,27 @@ describe('useController', () => {
       expect(screen.getByText('invalid')).toBeVisible();
     });
   });
+
+  it('can register field array property before field array root', () => {
+    const Component = () => {
+      const { control } = useForm<{
+        test: string;
+        test1: { test: string }[];
+      }>();
+
+      useController({
+        name: 'test1.0.test',
+        control,
+      });
+
+      useController({
+        name: 'test1',
+        control,
+      });
+
+      return null;
+    };
+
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
Nice one.

PR #13368, opened 2026-04-05
feat: enhance submit

ap0nia:
Enhance submission.

- `control.handleSubmit` returns the strongly typed output of `onValid` callback.
- New utilities to handle JSON and `FormData`.
- `Form` isolates phases of `handleSubmit` flow that can be memoized later.

Notes

- Bundle size increased from `11.79kb` -> `11.80kb` because of additional handling for function-based action.


src/form.tsx
```diff
@@ -45,84 +47,85 @@ function Form<
...
+              (encType && encType.includes('json'));
+
+            const response = await fetch(action, {
+              method,
+              headers: {
+                ...headers,
+                ...(encType &&
+                  encType !== 'multipart/form-data' && {
+                    'Content-Type': encType,
+                  }),
```
ap0nia: It's ok to spread primitives (boolean), it just might be a code smell.

```ts
const merged = { ...(condition ? { prop: value } : {}) }

// If condition is false, then spreading false is a noop.
const merged = { ...(condition && { prop: value }) } // {}
const merged = { ...false } // {}
```

src/form.tsx
```diff
@@ -45,96 +47,92 @@ function Form<
     onError,
     render,
     onSuccess,
-    validateStatus,
+    validateStatus = (status) => status >= 200 && status < 300,
```
ap0nia: Inline default function initializer actually reduced bundle size rather than declaring it at the module scope.

src/form.tsx
```diff
@@ -45,96 +47,92 @@ function Form<
...
+      encType,
+      onError,
+      onSuccess,
+      method,
+    ],
+  );
+
+  const submit = React.useCallback(
+    async (event?: React.BaseSyntheticEvent) => {
+      const hasError = await handleSubmit(event);
```
ap0nia: `handleSubmit` returns `void` if successful. Otherwise it returns an error object.

bluebill1049:
can you add a seperate PR just to memo the submit callback since you discover the missing memo?


` const submit = useCallback...`

src/form.tsx
ap0nia: Gonna opt into module scope so the reference remains stable when it's used in dependency arrays.

bluebill1049:
do you send to send a pr on this ` 🔧 add submit memo` we can release it this weekend? 

```
const handleSubmit = React.useMemo(
```

bluebill1049:
@copilot resolve the merge conflicts in this pull request

src/form.tsx
```diff
@@ -31,110 +38,90 @@ const POST_REQUEST = 'post';
...
 
-  const submit = async (event?: React.BaseSyntheticEvent) => {
-    let hasError = false;
-    let type = '';
-
-    await control.handleSubmit(async (data) => {
-      const formData = new FormData();
-      let formDataJson = '';
+  const handleSubmit = control.handleSubmit(async (data, event) => {
+    const formData = jsonToFormData(data);
```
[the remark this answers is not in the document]
ap0nia: I confirmed that this change does not affect any existing test cases. And if a resolver **did** change the output values, I wasn't sure why these changes would be ignored in the form data request? I assumed that data is the source of truth, and I can revert as needed.

src/form.tsx
```diff
@@ -31,110 +38,90 @@ const POST_REQUEST = 'post';
...
-      } catch {}
-
-      const flattenFormValues = flatten(control._formValues);
-
-      for (const key in flattenFormValues) {
-        formData.append(key, flattenFormValues[key]);
-      }
+        const shouldStringifySubmissionData =
+          (headers && headers['Content-Type'].includes('json')) ||
+          (encType && encType.includes('json'));
```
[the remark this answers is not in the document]
ap0nia: Adding another inline falsy check is sufficient. Optional chaining is undesirable because it adds bundle size.

PR #13378, opened 2026-04-08
perf: ⚡ memoize submit

ap0nia:
Memoize the submit function. Adds about `+0.2kb` bundle size, `11.63kb` to `11.65kb`.

Also addressed the inconsistencies below. Let me know if any of these were intentional and should be rolled back.
- Create the `FormData` from `data` instead of `control._formValues`.
- Use the resolved `control` instead of `props.control`.
- ~~Explicitly check if `action` is a string. It can be a function that would be called with the `FormData`.~~

### DIFF
--- a/src/form.tsx
+++ b/src/form.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 
 import { flatten } from './utils/flatten';
-import isString from './utils/isString';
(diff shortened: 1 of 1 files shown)


bluebill1049:
thanks can we leave the string check to that other pr?

bluebill1049 (approved):

PR #13382, opened 2026-04-12
fix: sync Controller value from _formValues on render to handle Activ…

### LINKED ISSUE #13381: issue: Controller does not update value as child of Activity
### Version Number

7.72.1

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/controller-activity-bug-6tpzdl

### Steps to reproduce

1. Open Codesandbox
2. Enter text in First Name input on Tab 1
3. Switch to Tab 2
4. Controller still renders old value (empty string)
5. Update text in First Name input on Tab 2 to update Controller value

### Expected behaviour

The Controller should always update the value when used as child of Activity

_No response_

### Relevant log output

```shell

```

### Code of Conduct

ISHANTPATIL2005:
Fixes #13381

## Problem

When `Controller` is inside a React Native `Activity` (or any tab layout 
that keeps components mounted but hidden), it shows a stale value when 
the tab becomes visible again.

Steps to reproduce are in the issue with a CodeSandbox link:
https://codesandbox.io/p/sandbox/controller-activity-bug-6tpzdl

## Why

`useController` uses `useWatch` internally. When a component is hidden 
inside an `Activity`, React skips state updates to that subtree. So 
`useWatch` misses the update and holds the old value.

## Fix

`control._formValues` always has the latest value regardless of React's 
render cycle. On each render, if `_formValues` and `useWatch` disagree, 
we use `_formValues` instead. No extra re-renders.

## Changes

- `src/useController.ts` — sync value from `_formValues` on render
- `src/__tests__/useController.test.tsx` — added regression test

## Code of Conduct

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -1514,6 +1514,53 @@ describe('useController', () => {
     rerender({ control: form1Result.current.control });
     expect(result.current.field.value).toBe('form1-value');
   });
+  it('should reflect updated value when parent visibility is toggled (Activity pattern)', async () => {
+    const Component = () => {
+      const { control, setValue } = useForm({
+        defaultValues: { test: '' },
+      });
+      const [activeTab, setActiveTab] = React.useState(0);
+
+      return (
+        <div>
+          {/* Tab 1: visible initially */}
+          <div style={{ display: activeTab === 0 ? 'block' : 'none' }}>
+            <Controller
+              name="test"
+              control={control}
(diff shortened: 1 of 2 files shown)


iankberry:
I was wondering when this PR was going to be merged? I am currently experiencing an issue where my form fields are not updating properly and I can confirm that this PR fixes that issue. I was able to create a wrapper for the `Controller` component to workaround the issue but it's pretty hacky.

PR #13385, opened 2026-04-16
fix: preserve watch updates on field array unmount fixes #13375

### LINKED ISSUE #13375: issue: List not updated when appending component is unmounted directly afterwards
### Version Number

7.72.1

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/relaxed-leftpad-p7trfw

### Steps to reproduce

1. One component to display the list, one component that appends to it, e.g. a dialog, and a parent that contains both.
2. Show the dialog only conditionally.
3. Close the dialog after appending.

The list doesn't update. Only after opening the dialog again, the update occurs. Explicitly adding `shouldUnregister: false` to `useFieldsArray` in the dialog doesn't change anything (I assume that's the default? It's not documented). Changing it to `true`, crashes the app, because `fields` is undefined.

### Expected behaviour

Appending to a list and then unmounting the component should update the list immediately.

Firefox

### Relevant log output

```shell

```

### Code of Conduct

gkarabelos:
Fixes #13375

## Problem

When `append` is triggered from a conditionally mounted component (e.g. a dialog) and that component unmounts immediately within the same event, `useWatch` consumers for that field array may miss the update.

In the reproduction case, clicking **append + close** does not immediately display the appended item. The item only appears after the component is remounted.

Steps to reproduce are in the issue with a CodeSandbox link:
https://codesandbox.io/p/sandbox/relaxed-leftpad-p7trfw

## Why

`useFieldArray` mutates the form values, but in this specific *append-and-immediate-unmount* sequence, the hook instance that performed the mutation is torn down before watchers receive a final state notification.

As a result, `useWatch` subscribers may not reflect the latest field array values.

## Fix

During `useFieldArray` cleanup, emit one final state update with the latest `formValues`, **only when**:

- An array action occurred in that hook instance (`_actioned.current`), and  
- Values are retained (`shouldUnregister` is `false` at both form and hook level)

This ensures proper synchronization with `useWatch` without altering existing unregister behavior.

### `src/useFieldArray.ts`
- Added `shouldKeepFieldArrayValues` guard in cleanup
- Emit `control._subjects.state.next({ name, values })` during cleanup for the *actioned + retained values* case
- Preserved existing unregister and mount behavior

### `src/__tests__/useFieldArray.test.tsx`
- Added regression test:
  - Verifies that **append + immediate unmount in a single event** updates watched sibling list immediately

## Code of Conduct

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -424,6 +424,9 @@ export function useFieldArray<
     !get(control._formValues, name) && control._setFieldArray(name);
 
     return () => {
+      const shouldKeepFieldArrayValues = !(
+        control._options.shouldUnregister || shouldUnregister
+      );
       const updateMounted = (name: InternalFieldName, value: boolean) => {
         const field: Field = get(control._fields, name);
         if (field && field._f) {
@@ -431,7 +434,14 @@ export function useFieldArray<
         }
       };
 
-      control._options.shouldUnregister || shouldUnregister
+      if (_actioned.current && shouldKeepFieldArrayValues) {
+        control._subjects.state.next({
+          name,
+          values: cloneObject(control._formValues) as TFieldValues,
+        });
+      }
+
+      !shouldKeepFieldArrayValues
         ? control.unregister(name as FieldPath<TFieldValues>)
         : updateMounted(name, false);
     };
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -441,9 +441,9 @@ export function useFieldArray<
         });
       }
 
-      !shouldKeepFieldArrayValues
-        ? control.unregister(name as FieldPath<TFieldValues>)
-        : updateMounted(name, false);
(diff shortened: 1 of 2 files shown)


src/useFieldArray.ts
```diff
@@ -424,14 +424,24 @@ export function useFieldArray<
...
+      if (_actioned.current && shouldKeepFieldArrayValues) {
+        control._subjects.state.next({
+          name,
+          values: cloneObject(control._formValues) as TFieldValues,
+        });
+      }
+
+      !shouldKeepFieldArrayValues
         ? control.unregister(name as FieldPath<TFieldValues>)
         : updateMounted(name, false);
```
bluebill1049:
```suggestion
      shouldKeepFieldArrayValues
? updateMounted(name, false) : control.unregister(name as FieldPath<TFieldValues>);
```

bluebill1049 (approved):

PR #13395, opened 2026-04-21
fix: preserve previous field value when `useController` name changes

Yihao-G:
#13041 introduced a potentially unwanted behaviour that when the `useController`'s `name` is changed, the field value for the previous `name` is lost, because it calls `unregister`. This behaviour isn't documented and cannot be toggled off.

The original issue #13039 that #13041 was fixing has been fixed again by #13070. Thus making the changes in #13041 redundant.

This PR reverts the `useController.ts` changes done in #13041, and solely relies on the changes done in #13070 to fix the original #13039 issue.


bluebill1049 (approved):
thanks for the fixes. LGTM

PR #13398, opened 2026-04-24
🐞 fix: prevent useWatch re-render when unrelated field validation is …

### LINKED ISSUE #12925: issue: useWatch is triggered when unrelated field validation is triggered
### Version Number

7.59.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/awesome-bush-v3snlv

### Steps to reproduce

1. Go to codesandbox
2. Open browser dev console
3. Type into form's input field
4. See how _ComponentWatchingOrganization rendering_ is printed even though that component is using useWatch to watch `organization` which is never updated. The `MyInput` component only updates `firstName` and triggers validation for `secondName` using the `rules`.

### Expected behaviour

I expect ComponentWatchingOrganization not to render because it uses useWatch to watch totally unrelated field in the form. This issue does not exist in v7.54.2.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

cyky:
### Description

Fixes #12925

`useWatch` subscribers re-render when unrelated field validation is triggered via `rules.deps`. This is a regression since v7.55.0 when `_subjects.values` and `_subjects.state` were merged.

### Root cause

In `shouldRenderFormState`, the condition `Object.keys(formState).length >= Object.keys(_proxyFormState).length` acts as a short-circuit that bypasses key-matching. When `trigger()` emits `{ errors: {} }` (1 key) and `useWatch` subscribes with `{ values: true }` (1 key), `1 >= 1` passes despite zero key overlap.

### Changes

- **`shouldRenderFormState.ts`**: Guard the size-based condition with `isRoot` so it only applies to the root `useForm` subscriber. Non-root subscribers now rely on actual key overlap matching.
- **`getProxyFormState.ts`**: `defaultValues` was set as a plain property instead of a tracked getter, so `useFormState` accessing `defaultValues` never registered a subscription. This was masked by the size-based condition but is now properly tracked.
- **Test updates**: Adjusted render count assertions that included spurious re-renders caused by the bug

### DIFF
--- a/src/logic/getProxyFormState.ts
+++ b/src/logic/getProxyFormState.ts
@@ -11,9 +11,7 @@ export default <
   localProxyFormState?: ReadFormState,
   isRoot = true,
 ) => {
-  const result = {
-    defaultValues: control._defaultValues,
-  } as typeof formState;
+  const result = {} as typeof formState;
 
   for (const key in formState) {
     Object.defineProperty(result, key, {
@@ -26,6 +24,22 @@ export default <
 
         localProxyFormState && (localProxyFormState[_key] = true);
         return formState[_key];
+      },
+    });
+  }
+
+  if (!('defaultValues' in formState)) {
+    Object.defineProperty(result, 'defaultValues', {
+      get: () => {
+        const _key = 'defaultValues' as keyof FormState<TFieldValues> &
+          keyof ReadFormState;
+
+        if (control._proxyFormState[_key] !== VALIDATION_MODE.all) {
+          control._proxyFormState[_key] = !isRoot || VALIDATION_MODE.all;
+        }
+
+        localProxyFormState && (localProxyFormState[_key] = true);
+        return control._defaultValues;
       },
     });
   }
--- a/src/logic/getProxyFormState.ts
+++ b/src/logic/getProxyFormState.ts
@@ -28,19 +28,5 @@ export default <
     });
   }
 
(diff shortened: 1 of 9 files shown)


src/logic/getProxyFormState.ts
```diff
@@ -30,5 +28,19 @@ export default <
...
+        if (control._proxyFormState.defaultValues !== VALIDATION_MODE.all) {
+          control._proxyFormState.defaultValues =
+            !isRoot || VALIDATION_MODE.all;
+        }
+
+        localProxyFormState && (localProxyFormState.defaultValues = true);
+        return control._defaultValues;
+      },
+    });
+  }
```
bluebill1049: what's this for?

cyky: The `defaultValues` isn't being passed with the `formState`object here. Without this the useFormState cant subscribe to defaultValues changes. This was being previously masked by the size-based short-circuit in the `shouldRenderFormState`.

There is existing test `formState.defaultValues and useFormState().defaultValues should be equal after reset` that fails without this change.

We could probably include the default values with the formState but I'm not sure if that is better approach?

cyky: Actually I went ahead and removed this and refactored the `useFormState` to include the `defaultValues`.

Let me know if this is better approach.

bluebill1049 (approved):

PR #13409, opened 2026-05-03
fix(Watch): restore TypeScript 4 compatibility

### LINKED ISSUE #13124: Breaking changes for TypeScript 4 projects
We have an application that is using TypeScript 4.7.4. Latest version 7.66.0 no longer compiles due to this change which introduces a TypeScript 5 feature.

> This is causing build errors on our Typescript 4 application, I think this should have been marked as a major version release. 

 _Originally posted by @john-adams-sigparser in [384ad49](https://github.com/react-hook-form/react-hook-form/commit/384ad49f0d582a485b595d4571f7d593470ca3c8#r169522515)_

dfedoryshchev:
The `const TFieldName` modifier added in #13108 is a TypeScript 5.0 language feature, so the published `.d.ts` no longer parses on TypeScript 4.x.

Drops the modifier and adds a variadic-tuple branch (TypeScript 4.0+) to the `TFieldName` constraint. On TypeScript 5+, TypeScript picks the tuple branch for inline literal arrays, preserving the inference that `const TFieldName` provided. The tsd test in `src/__typetest__/watch.test-d.ts` continues to pass.

On TypeScript 4 the type parses, but the `render` callback receives a widened type (`string[]` rather than the tuple) because TS 4 does not preserve literal positions through this constraint shape.

Closes #13124


bluebill1049 (approved):
LGTM. Thanks for the fix.

PR #13416, opened 2026-05-08
fix: useFieldArray re-renders when setValue targets nested array path (#13260)

EduardF1:
## Proposed Changes

`setValue('items.1.value', x)` writes to `_formValues` correctly but only emits on `_subjects.state`. `useFieldArray` subscribes to `_subjects.array`, so its rendered `fields[i]` snapshot stays stale until an `append`/`update`/`replace` forces a remount, contradicting the docs which explicitly recommend `setValue` for in-place updates without remount.

This change detects when a `setValue` target sits inside a registered field array (via the existing `_names.array` set + `isNameInFieldArray` helper) and additionally emits on `_subjects.array` with the array root path, mirroring the existing field-array-root branch. `useFieldArray` then re-renders in place.

Fixes #13260

## Type of change

## Checklist

- [x] I have added a test that proves the fix works
- [x] New and existing tests pass locally with my changes (`useForm` + `useFieldArray` suites: 439 + 205 passing)

## Test plan

- [x] New regression test in `src/__tests__/useForm/setValue.test.tsx` (\"issue #13260\") asserts that `useFieldArray` re-renders the affected item after `setValue('items.1.value', 42)`
- [x] All 11 `useFieldArray` test suites still pass (205/205)
- [x] All 25 `useForm` test suites still pass (439/439)

### DIFF
--- a/src/__tests__/useForm/setValue.test.tsx
+++ b/src/__tests__/useForm/setValue.test.tsx
@@ -1589,4 +1589,45 @@ describe('setValue', () => {
       data: [{ id: true, name: true }],
     });
   });
+
+  it('should re-render useFieldArray fields when setValue targets a nested array path - issue #13260', () => {
+    type FormValues = {
+      items: { value: number }[];
+    };
+
+    const App = () => {
+      const { control, setValue } = useForm<FormValues>({
+        defaultValues: {
+          items: [{ value: 0 }, { value: 0 }, { value: 0 }],
+        },
+      });
+      const { fields } = useFieldArray({ control, name: 'items' });
+
+      return (
+        <div>
+          {fields.map((field, index) => (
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks again!

EduardF1:
Thanks @bluebill1049, glad #13420 landed. If commit attribution allows, would you consider listing me as co-author on the merge given the original repro came from this PR? Appreciate the fast turnaround either way.

bluebill1049:
> Thanks @bluebill1049, glad #13420 landed. If commit attribution allows, would you consider listing me as co-author on the merge given the original repro came from this PR? Appreciate the fast turnaround either way.

for sure, i will add you to the release log as well.

PR #13418, opened 2026-05-08
fix: treat React elements as opaque leaves in deepEqual

### LINKED ISSUE #13415: issue: deepEqual traverses React element \_owner` Fiber → triggers Next.js sync-dynamic-api warnings on `params`/`searchParams` Proxies`
### Version Number

7.72.1

### Codesandbox/Expo snack

(I'll attach a minimal Next.js 16 + RHF 7.72.1 reproduction — let me know if you'd prefer a smaller plain-React repro using a hand-rolled Proxy.)

### Steps to reproduce

1. Use Next.js 16 App Router with a client component page (`'use client'`).
2. Put a React element into a form value's nested field, e.g. an
   `{ label: <ReactNode>, value: string }` option used by a select/autocomplete.
3. Subscribe to `formState.isDirty` (e.g. read it in render, or use `mode: 'all'`).
4. Open the page. The Next.js dev console fills up with:
   - `params are being enumerated. \`params\` is a Promise and must be unwrapped with React.use()`
   - `The keys of \`searchParams\` were accessed directly`

In some configurations React also throws "Maximum update depth exceeded".

### Cause (what I traced)

`useForm`'s isDirty effect calls `_getDirty()` → `deepEqual(getValues(), _defaultValues)`.

`cloneObject` returns non-plain objects by reference, so the React element
inside the form value is the **same reference** in both trees. In 7.72.1
`deepEqual` had no `object1 === object2` early-out, so it enumerated the
element's own keys including `_owner`. On a client render `_owner` points
at the currently-rendering Fiber. `deepEqual` then recurses into the Fiber,
walks `memoizedProps`, and reaches the page component's `params` —
which in Next.js 16 is a Proxy whose `ownKeys` trap fires
`warnForSyncAccess`.

Stack (abridged):

\`\`\`
deepEqual.ts:18
deepEqual.ts:45  (×N recursive)
createFormControl.ts:655  // _getDirty: !deepEqual(getValues(), _defaultValues)
useForm.ts:145             // useEffect on formSta
commitHookEffectListMount
\`\`\`

### Why 7.73.1 hides this

PR #13362 added `if (object1 === object2) return t
`deepEqual`. Since `cloneObject` preserves the React element's identity,
the early-out short-circuits before reaching `_own

This works today because the early-out happens to
reference, but it's load-bearing on a perf optimization rather than an
intentional guard. Anything that triggers a deeper
element (e.g. a future `deepEqual` refactor, a devtool snapshot,
`structuredClone`, a redux serializable-check midd
the same path.

### Expected behaviour

`deepEqual` (and `cloneObject`) should treat React elements as opaque
leaves rather than recursing into their internals. A `$$typeof ===
Symbol.for('react.element')` (or `react.transitional.element`) check
would make this an explicit contract instead of an accidental side
effect of identity equality.

Alternatively: document that form values must be JSON-serializable plain
data, and surface a dev-mode warning when a non-plain non-Date/Blob/File
value is set.

Chrome

### Relevant log output

```shell

```

### Code of Conduct

mixelburg:
Fixes #13415

When form values contain React elements (e.g. option objects with JSX labels), `deepEqual` recurses into internal React properties like `_owner`, which points at the current Fiber on client renders. In Next.js 16, this walks into `params`/`searchParams` Proxies and triggers `warnForSyncAccess` dev warnings.

**Changes:**

- Added `object1 === object2` reference equality check at the top of `deepEqual` — this is both a performance optimization and prevents any traversal of identical references
- Added `isReactElement` check using `$$typeof === Symbol.for('react.element')` to treat React elements as opaque leaves. When either argument is a React element, comparison falls back to reference equality (same ref = equal, different ref = not equal)

This prevents `deepEqual` from recursing into `_owner` Fiber trees and also avoids issues with any future changes that might break the accidental identity-equality protection.

**Test:** Added test case verifying React elements with different internal structures are treated as opaque — different elements return `false`, same reference returns `true`, and nested elements in form values don't cause recursion into `_owner`.

### DIFF
--- a/src/__tests__/utils/deepEqual.test.ts
+++ b/src/__tests__/utils/deepEqual.test.ts
@@ -157,7 +157,9 @@ describe('deepEqual', () => {
   });
 
   it('should treat React elements as opaque and not recurse into internals', () => {
-    const REACT_ELEMENT_TYPE = Symbol.for ? Symbol.for('react.element') : 0xeac7;
+    const REACT_ELEMENT_TYPE = Symbol.for
+      ? Symbol.for('react.element')
+      : 0xeac7;
 
     const element1 = {
       $$typeof: REACT_ELEMENT_TYPE,
--- a/src/__tests__/utils/deepEqual.test.ts
+++ b/src/__tests__/utils/deepEqual.test.ts
@@ -155,4 +155,33 @@ describe('deepEqual', () => {
     expect(deepEqual({ value: NaN }, { value: 'NaN' })).toBeFalsy();
     expect(deepEqual([NaN], [0])).toBeFalsy();
   });
+
+  it('should treat React elements as opaque and not recurse into internals', () => {
+    const REACT_ELEMENT_TYPE = Symbol.for ? Symbol.for('react.element') : 0xeac7;
+
+    const element1 = {
+      $$typeof: REACT_ELEMENT_TYPE,
+      type: 'div',
+      key: null,
+      ref: null,
+      props: { children: 'Hello' },
+      _owner: { memoizedProps: { params: new Proxy({}, {}) } },
+    };
+
+    const element2 = {
+      $$typeof: REACT_ELEMENT_TYPE,
+      type: 'div',
+      key: null,
+      ref: null,
+      props: { children: 'World' },
+      _owner: { memoizedProps: { params: new Proxy({}, {}) } },
+    };
+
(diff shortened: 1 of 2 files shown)


bluebill1049:
thanks for the PR, how come react node become the part of the diff. I don't think rhf supports that.

Issue #13436, opened 2026-05-13
issue: Clarify expected behavior when using form level validation


johnstrand:
### Version Number

7.75.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/suspicious-pasteur-ryrsqq

### Steps to reproduce

Click on submit, observe the error object with the property `firstName` set, but with an empty message despite a message being set inside the `validate` callback.

Mostly asking for clarification of expected behavior; in `validateForm` (starting at line 563 in `src/logic/createFormControl.ts`), there's a check if `props.validate` is set, and then the following block is run (some code omitted for brevity's sake):

```typescript
const result = await props.validate({
	formValues: _formValues,
	formState: _formState,
	name,
	eventType,
	});

if (isObject(result)) {
	for (const key in result) {
		const error = result[key];

		if (error) {
			setError(`${FORM_ERROR_TYPE}.${key}`, {
				message: isString(result.message) ? result.message : '',
				type: INPUT_VALIDATION_RULES.validate,
			});
		}
	}
}
```

My question is, on line 584, should
`message: isString(result.message) ? result.message : '',`
actually be
`message: isString(error.message) ? error.message : '',`

I tried making that change on a local copy of the source code, and that caused `message` to be properly set on `formState.errors.form.firstName`

### Expected behaviour

`formState.errors.form.firstName.message` should be set if returned from the `validate`-function, probably

### What browsers are you seeing the problem on?

Chrome

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bluebill1049:
nice find that's a bug, do you want to send a pr since you discover it 👍

johnstrand:
> nice find that's a bug, do you want to send a pr since you discover it 👍

Absolutely, I've created a PR

PR #13437, opened 2026-05-14
Implemented fix, added tests

johnstrand:
Fixes issue #13436, ensures that when the validator returns an object, the message is preserved. Also, since the user is required to return a type on the object, made sure that it also is preserved (if specified, otherwise fallback to default)


bluebill1049 (approved):
LGTM

PR #13447, opened 2026-05-17
🐞 fix(useFieldArray): preserve managed ids on descendant setValue

maxkostow:
### Follow-up to #13446

#13446 stopped the array subscriber from clobbering mutation-method ids by gating regeneration on `_actioned.current`. That covers the `append`/`remove`/`swap`/`move`/… path, but **not an external nested `setValue`**.

### The remaining bug

`setValue('rows.3.isLoading', true)` goes through neither path:

- it's not a mutation method, so `_actioned.current` is `false`;
- it's a descendant write, so the subscriber still ran `ids.current = fieldValues.map(generateId)` and reminted **every** `field.id`.

With the documented `key={field.id}` pattern this remounts every row on every nested write. If a row has a mount effect (e.g. fetch-on-mount that calls `setValue` again), it becomes an infinite remount/refetch loop.

### Fix

A descendant write provably cannot change the array's length or order — only a leaf primitive is `set()`. So those ids must survive. The `createFormControl` emission that fans a non-field-array write out to its field-array parents is tagged `isDescendantUpdate`, and the subscriber skips id regeneration for it, alongside the existing `_actioned` guard.

Untagged emissions are unchanged: whole-array `setValue('rows', …)` and `reset()` still regenerate ids (no element-identity guarantee there), so upstream behavior is preserved for every ambiguous case — the change only affects the case where positional identity is *guaranteed*.

### Tests

Added a regression test in `useFieldArray.test.tsx` asserting both halves of the contract: descendant `setValue` keeps `field.id` stable; whole-array `setValue` still regenerates. Full `useFieldArray` suite 210/210, `setValue`/`setValues` 64/64.

### DIFF
--- a/src/__tests__/useFieldArray.test.tsx
+++ b/src/__tests__/useFieldArray.test.tsx
@@ -1372,6 +1372,48 @@ describe('useFieldArray', () => {
       },
     );
 
+    it('should preserve field ids when a descendant value is set, but regenerate them on a whole-array set', async () => {
+      let setValue: any;
+      let ids: string[] = [];
+      const Component = () => {
+        const { register, control, setValue: tempSetValue } = useForm({
+          defaultValues: {
+            test: [{ name: 'a' }, { name: 'b' }, { name: 'c' }],
+          },
+        });
+        const { fields } = useFieldArray({ name: 'test', control });
+
+        setValue = tempSetValue;
+        ids = fields.map((field) => field.id);
+
+        return (
+          <form>
+            {fields.map((field, i) => (
+              <input key={field.id} {...register(`test.${i}.name` as const)} />
+            ))}
+          </form>
+        );
+      };
+
+      render(<Component />);
+
+      const initialIds = [...ids];
+
+      // Descendant write: array length/order provably unchanged -> ids stable
+      // (regression for the field-array remount loop, #12665-style).
+      await act(async () => {
+        setValue('test.1.name', 'b-changed');
+      });
+      expect(ids).toEqual(initialIds);
+
+      // Whole-array set: no element-identity guarantee -> ids regenerate,
+      // preserving upstream behavior.
+      await act(async () => {
+        setValue('test', [{ name: 'x' }, { name: 'y' }, { name: 'z' }]);
+      });
+      expect(ids).not.toEqual(initialIds);
+    });
+
     it.each(['dirtyFields'])(
       'should unset name from dirtyFieldRef if array field values are not different with default value when formState.%s is defined',
       (property) => {
--- a/src/__tests__/useFieldArray.test.tsx
+++ b/src/__tests__/useFieldArray.test.tsx
@@ -1376,7 +1376,11 @@ describe('useFieldArray', () => {
       let setValue: any;
       let ids: string[] = [];
       const Component = () => {
-        const { register, control, setValue: tempSetValue } = useForm({
+        const {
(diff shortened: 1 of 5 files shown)


src/__tests__/useFieldArray.test.tsx
```diff
@@ -1372,6 +1372,52 @@ describe('useFieldArray', () => {
...
     );
 
+    it('should preserve field ids when a descendant value is set, but regenerate them on a whole-array set', async () => {
+      let setValue: any;
+      let ids: string[] = [];
+      const Component = () => {
+        const {
+          register,
+          control,
+          setValue: tempSetValue,
```
bluebill1049, line 1382: can you fix the lint?

bluebill1049 (changes requested):

src/__tests__/useFieldArray.test.tsx
maxkostow: Thanks for merging these in and sorry for the multiple PRs. I was testing the changes on my code after opening. Will test before opening next time 👯‍♂️

maxkostow:
@bluebill1049 I think this PR may be running into a react performance regression from #13420. Isn't the whole point of `useFieldArray` to decouple the rendering of the array from the elements? If you want other updates on the field, you can call `useWatch`.

Basically we expect the `useFieldArray` rendering to be stable with calls like
```js
setValue('item.0.a', ...)
setValue('item.0.b', ...)
setValue('item.0.c', ...)
```

Is that a no longer the expectation?

From the [docs](https://github.com/react-hook-form/documentation/blob/1237ac91ed1ece3b851914743ca84c264ed2ab01/src/content/docs/usefieldarray.mdx#return):
> fields | object & { id: string } | This object contains the defaultValue and key for your component.
> -- | -- | --

bluebill1049:
> @bluebill1049 I think this PR may be running into a react performance regression from #13420. Isn't the whole point of `useFieldArray` to decouple the rendering of the array from the elements? If you want other updates on the field, you can call `useWatch`.
> 
> Basically we expect the `useFieldArray` rendering to be stable with calls like
> 
> ```js
> setValue('item.0.a', ...)
> setValue('item.0.b', ...)
> setValue('item.0.c', ...)
> ```
> 
> Is that a no longer the expectation?
> 
> From the [docs](https://github.com/react-hook-form/documentation/blob/1237ac91ed1ece3b851914743ca84c264ed2ab01/src/content/docs/usefieldarray.mdx#return):
> 
> > fields
> > object & { id: string }
> > This object contains the defaultValue and key for your component.

should be sable, yes if that's cause the issue with that PR. I can revert that pr.

bluebill1049:
does that means all the fix you have done, all come down to the same root cause https://github.com/react-hook-form/react-hook-form/pull/13420#issuecomment-4472139263

maxkostow:
#13446 and #13447 yes but not the `setValues` optimizations/fixes (#13445 #13448 #13450)

bluebill1049:
> #13446 and #13447 yes but not the `setValues` optimizations/fixes (#13445 #13448 #13450)

thanks i will revert.

bluebill1049:
all done, i will try to get a patch release this week.

maxkostow:
Thanks! Appreciate any urgency you can put behind it because we are in desperate need of the performance benefits from the batch `setValues` but need this fixed in order to release.

PR #13454, opened 2026-05-18
🐞 fix(setValues): don't re-grow field array when removing an element

maxkostow:
## Summary

Calling the `setValues` batch API with a shortened field-array (removing an element) left a phantom `{ ...: undefined }` entry behind, re-growing the array back to its original length.

### Root cause

`setValues` merges the shrunk array into `_formValues` correctly, then loops over `_names.mount`. The removed row's leaf path (e.g. `test.2.name`) is still in `_names.mount` because React hasn't unmounted/re-rendered yet, so it gets pushed through `_setValue`. `setFieldValue`'s unconditional `set(_formValues, name, …)` then re-creates the path, re-growing the array.

### Fix

The loop now walks the merged tree by own keys:
- A path that no longer resolves (a removed array index) is **skipped**.
- A present-but-`undefined` leaf is **still synced**, since `undefined` is a valid field value (e.g. `setValues({ a: undefined })` must still clear a mounted input).

It also reads the merged-authoritative value instead of the raw partial, so unspecified top-level keys are no longer clobbered.

## Tests

Added to `src/__tests__/useForm/setValues.test.tsx`:
1. `should not leave a stale element behind when setValues removes a field-array item` — regression repro (red before the fix).
2. `should still propagate an explicitly set undefined value to a mounted field` — guards against the structural check over-skipping legitimate `undefined` clears.

Full suite: **1116/1116 passing**.

### DIFF
--- a/src/__tests__/useForm/setValues.test.tsx
+++ b/src/__tests__/useForm/setValues.test.tsx
@@ -8,6 +8,7 @@ import {
 } from '@testing-library/react';
 
 import { Controller } from '../../controller';
+import { useFieldArray } from '../../useFieldArray';
 import { useForm } from '../../useForm';
 
 describe('setValues', () => {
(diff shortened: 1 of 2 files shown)


maxkostow:
Fixed by https://github.com/react-hook-form/react-hook-form/pull/13576

PR #13478, opened 2026-05-25
fix: prevent prototype pollution read in get() utility

### LINKED ISSUE #13477: [Security/Medium]: Prototype Pollution Read Vulnerability in get() utility
## Description

The `get()` utility function in `src/utils/get.ts` allows reading properties from `Object.prototype` via the `__proto__` key. While `set.ts` has protection against prototype pollution (line 31: `if (key === '__proto__' || key === 'constructor' || key === 'prototype')`), the `get.ts` function lacks the same protection.

This creates a prototype pollution read vulnerability that can expose sensitive data if `Object.prototype` is polluted.

## Steps to Reproduce

```javascript
// Pollute Object.prototype (simulating an attack from another library)
Object.prototype.polluted = 'SECRET_DATA';

const formValues = { name: 'John' };

// Using react-hook-form's get utility
import get from './src/utils/get';

// Attack 1: Read via dot notation
const leaked1 = get(formValues, '__proto__.polluted'); // Returns 'SECRET_DATA'

// Attack 2: Read via array notation
const leaked2 = get(formValues, '__proto__[polluted]'); // Returns 'SECRET_DATA'

console.log(leaked1, leaked2); // 'SECRET_DATA' 'SECRET_DATA'
```

## Impact

1. **Data Leakage**: If an attacker can pollute `Object.prototype` (via a vulnerable dependency, JSON merge, or other means), they can exfiltrate sensitive data using react-hook-form's `get()` function.

2. **Widespread Usage**: The `get()` utility is used extensively throughout react-hook-form:
   - `watch()` function (line 14, 21 in `generateWatchOutput.ts`)
   - `useController` (multiple locations)
   - `validateField` (line 55)
   - `createFormControl` (multiple locations)

3. **User-Facing Exposure**: End users can indirectly trigger this via `watch('__proto__.polluted')` or similar API calls.

## Root Cause

The `stringToPath` utility (`src/utils/stringToPath.ts`) splits paths by brackets:
```typescript
export default (input: string): string[] =>
  input.split(/[.[\]'"]/g).filter(Boolean);
```

This correctly parses `__proto__[polluted]` into `['__proto__', 'polluted']`. However, `get.ts` does not validate these keys before accessing the object, unlike `set.ts`.

## Suggested Fix

Add the same prototype protection found in `set.ts` to the `get()` function:

```typescript
export default <T>(
  object: T,
  path?: string | null,
  defaultValue?: unknown,
): any => {
  if (!path || !isObject(object)) {
    return defaultValue;
  }

  const paths = isKey(path) ? [path] : stringToPath(path);

  // Block prototype access
  if (paths.some(key => key === '__proto__' || key === 'constructor' || key === 'prototype')) {
    return defaultValue;
  }

  const result = paths.reduce<any>((result, key) => {
    return isNullOrUndefined(result) ? undefined : result[key];
  }, object);

  return isUndefined(result) || result === object
    ? isUndefined(object[path as keyof T])
      ? defaultValue
      : object[path as keyof T]
    : result;
}
```

## Environment

- Version: 7.76.1
- OS: macOS
- Node.js: v18+

## Additional Context

This is a read-only variant of the prototype pollution issue. The write variant is already protected in `set.ts` (line 31). Applying the same protection to `get.ts` ensures defense-in-depth.

References:
- https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/set.ts#L31
- https://github.com/react-hook-form/react-hook-form/blob/master/src/utils/get.ts
- https://cheatsheetseries.owasp.org/cheatsheets/JavaScript_Security_Cheat_Sheet.html#prototype-pollution

xxiaoxiong:
Fixes #13477

Why
The `get()` utility function in `src/utils/get.ts` allows reading properties from `Object.prototype` via the `__proto__` key, while `set.ts` already has protection against prototype pollution on line 31 (`if (key === '__proto__' || key === 'constructor' || key === 'prototype')`).

Changes
Added the same check to the `reduce` callback in `get.ts` to guard against `__proto__`, `constructor`, and `prototype` keys, matching the existing pattern in `set.ts`.

Updated component: `src/utils/get.ts`

Testing
- `get(formValues, '__proto__.polluted')` now returns `undefined` instead of the prototype value.
- `get(formValues, 'constructor')` now returns `undefined`.
- `get(formValues, 'prototype')` now returns `undefined`.
- Existing `get()` calls with normal paths continue to work unchanged.

Surface area
- Security
- Bug fix


bluebill1049:
This PR is too big, could you send another pr with relevant and isolated fix.

PR #13485, opened 2026-05-29
fix: use reactive values prop over defaultValues when shouldUnregister is true

### LINKED ISSUE #12697: issue: When configuring `values` and `defaultValues` in `useForm`, the `values` data is not set correctly in `formValues` when `shouldUnregister` is set to `true`.
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/serene-wildflower-3lzwc5?file=%2Fsrc%2FApp.tsx%3A13%2C5-13%2C21&workspaceId=ws_DUpBMQkanMVMH4sxTkz7fG

### Steps to reproduce

1. Configure `useForm` with `defaultValues: {firstName: '1'}` and `values: {firstName: '2'}`
2. The input field displays '2' as expected when `shouldUnregister` is `false`
3. Change `shouldUnregister` to `true`
4. Observe the input field value changing to '1'

```
import "./styles.css";
import { Controller, useForm } from "react-hook-form";
import { useState } from "react";

interface FormData {
  firstName: string;
}

  const [submitted, setSubmitted] = useState<FormData | undefined>();
  const { control, handleSubmit } = useForm<FormData>({
    shouldUnregister: true,
    defaultValues: {
      firstName: "1",
    },
    values: {
      firstName: "2",
    },
  });

  return (
    <div className="App">
      <h1>Form bug</h1>
      <form
        className="form"
        onSubmit={handleSubmit((data) => {
          console.log(JSON.stringify(data, undefined, " "));
          setSubmitted(data);
        })}
      >
        <Controller
          name={"firstName"}
          control={control}
          rules={{ required: "Required field" }}
          render={({ field: { onChange, name, value }, fieldState }) => (
            <input name={name} value={value || ""} onChange={onChange} />
          )}
        />
        <button>Submit!</button>
      </form>
      {submitted && <pre>{JSON.stringify(submitted, undefined, " '")}</pre>}
    </div>
  );
}
```

### Expected behaviour

The value of the input field should consistently be '2', regardless of the `shouldUnregister` setting.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

Dharya-dev:
## Summary
Hey! I noticed that when using `useForm` with both reactive `values` and static `defaultValues`, setting `shouldUnregister: true` causes the input fields to incorrectly display the `defaultValues` instead of the reactive `values`.

This PR fixes that by ensuring `useController` checks the reactive `values` prop first before falling back to `defaultValues` during the field re-registration cycle.

Resolves #12697

## Root Cause
In `useController.ts`, when `shouldUnregister` is true, the `useEffect` re-populates `_formValues` from `_defaultValues` on field remount. But it never checks if a reactive `values` prop exists — so `defaultValues: { firstName: '1' }` always wins over `values: { firstName: '2' }`.

## Changes
- **`src/useController.ts`** — Changed the value lookup from `control._defaultValues` to `control._options.values || control._defaultValues`, so the reactive `values` prop takes priority when present.
- **`src/__tests__/useForm.test.tsx`** — Added a regression test that reproduces the exact scenario from #12697 (Controller + shouldUnregister + values + defaultValues).

## Test Results
```
PASS src/__tests__/useForm.test.tsx
PASS src/__tests__/controller.test.tsx
PASS src/__tests__/useController.test.tsx

Test Suites: 3 passed, 3 total
Tests:       157 passed, 157 total
```

Let me know if you need any changes!

### DIFF
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -214,7 +214,9 @@ export function useController<
     if (_shouldUnregisterField) {
       const value = cloneObject(
         get(
-          control._options.values || control._defaultValues,
+          shouldUnregister
+            ? control._defaultValues
+            : control._options.values || control._defaultValues,
           name,
           get(
             control._options.defaultValues,
--- a/src/useController.ts
+++ b/src/useController.ts
@@ -214,7 +214,7 @@ export function useController<
     if (_shouldUnregisterField) {
       const value = cloneObject(
         get(
-          control._defaultValues,
+          control._options.values || control._defaultValues,
           name,
           get(
             control._options.defaultValues,
--- a/src/__tests__/useForm.test.tsx
+++ b/src/__tests__/useForm.test.tsx
@@ -2314,6 +2314,39 @@ describe('useForm', () => {
     expect(onInvalid).not.toHaveBeenCalled();
   });
 
+  it('should use values prop over defaultValues with shouldUnregister and Controller (#12697)', async () => {
+    const App = () => {
+      const { control, handleSubmit } = useForm<{
+        firstName: string;
+      }>({
+        shouldUnregister: true,
+        defaultValues: {
+          firstName: '1',
+        },
+        values: {
+          firstName: '2',
+        },
+      });
+
+      return (
+        <form onSubmit={handleSubmit(noop)}>
+          <Controller
+            name="firstName"
+            control={control}
+            render={({ field }) => <input {...field} />}
+          />
+          <button>submit</button>
+        </form>
+      );
+    };
+
+    render(<App />);
+
+    await waitFor(() => {
+      expect(screen.getByRole('textbox')).toHaveValue('2');
+    });
+  });
+
   it('should only update async form values which are not interacted', async () => {
     type FormValues = {
       test: string;


src/useController.ts
```diff
@@ -214,7 +214,7 @@ export function useController<
     if (_shouldUnregisterField) {
       const value = cloneObject(
         get(
-          control._defaultValues,
+          control._options.values || control._defaultValues,
```
bluebill1049: can we do a flag check here with shouldUnregister?

`shouldUnregister ? control._defaultValues :  control._options.values ||  control._defaultValues,`

bluebill1049 (changes requested):

Dharya-dev:
Hey @bluebill1049, I've pushed the change as requested! Added the `shouldUnregister` flag check so that field-level unregister correctly falls back to `defaultValues`, while form-level unregister uses the reactive `values` prop. I rebased on top of the regression tests you added, and everything is passing locally. Let me know if anything else is needed!

bluebill1049 (approved):

PR #13509, opened 2026-06-07
fix: surface resolver errors for related fields

louzhedong:
This updates the resolver-onChange path so errors returned for related fields are not dropped when the current field lookup misses.

Validation:
- pnpm.cmd test -- src/__tests__/useForm.test.tsx src/__tests__/controller.test.tsx src/__tests__/useFormState.test.tsx --runInBand --watch=false
- pnpm.cmd type

### DIFF
--- a/src/__tests__/useForm.test.tsx
+++ b/src/__tests__/useForm.test.tsx
@@ -1068,7 +1068,7 @@ describe('useForm', () => {
         expect(methods.formState.isValid).toBeFalsy();
       });
 
-      it('should update errors when the resolver returns an error for a different field', async () => {
+      it('should update errors for dependent fields returned by the resolver', async () => {
         const resolver = jest.fn((values: any) => {
           const errors =
             values.brand === 'adidas' && values.country === 'usa'
@@ -1100,7 +1100,10 @@ describe('useForm', () => {
 
           return (
             <form>
-              <input {...register('brand')} placeholder="brand" />
+              <input
+                {...register('brand', { deps: ['country'] })}
+                placeholder="brand"
+              />
               <input {...register('country')} placeholder="country" />
               <p>{errors.country?.message}</p>
             </form>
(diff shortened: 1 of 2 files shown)


louzhedong:
I’ve kept this branch validated locally and the resolver fallback is still behaving as intended on the targeted regression tests. If you’d prefer a narrower fallback strategy, I can revise the lookup path.

louzhedong:
I re-checked the failing Cypress job. The failures are concentrated in asicSchemaValidation.cy.ts and customSchemaValidation.cy.ts (the onSubmit / onBlur / onChange cases), while build and type jobs are green.

That makes this look like a regression in the schema-validation Cypress coverage rather than a broader packaging/build issue. I can dig deeper into the fallback path or trim the regression if you want me to focus on one of those suites first.

louzhedong:
I narrowed the resolver fallback so it only applies to fields declared in the current field's deps. That keeps dependent-field errors visible when they are actually related, while avoiding the broader fallback that was surfacing unrelated resolver errors in the schema-validation Cypress coverage.

I re-validated locally with:
- jest --config ./scripts/jest/jest.config.js --runInBand --watch=false src/__tests__/useForm.test.tsx
- eslint src/logic/createFormControl.ts src/__tests__/useForm.test.tsx
- prettier --check src/logic/createFormControl.ts src/__tests__/useForm.test.tsx

The new push is at 6583a6d6 and CI is running again.

louzhedong:
Update: the latest push is green again, including Cypress.

What changed in this round:
- the resolver fallback is now limited to fields declared in the current field's deps
- that keeps dependent-field errors visible without broadening the fallback to unrelated resolver errors

I re-ran the local validation while iterating, and the new CI run came back green on all checks.

louzhedong:
Rebased this branch onto the latest master and pushed the updated head (16ab0675c3c023a72b440d6e83459430675aff6a). CI has rerun on the refreshed branch; the functional behavior and local validation from the last patch remain unchanged.

louzhedong:
The rebase onto the latest `master` is now fully green again, including Cypress. The branch head is `16ab0675c3c023a72b440d6e83459430675aff6a`, and the functional change is unchanged from the last patch.

PR #13518, opened 2026-06-11
🐞 fix: restore promise return from controller onChange

louzhedong:
## What
Restore the return value from `field.onChange` in `useController`.

## Why
After the 7.78.0 rewrite, `field.onChange` stopped returning the promise-like result from the underlying register handler, which breaks callers that chain `then(...)` or `await` it.

## Validation
- `jest --config ./scripts/jest/jest.config.js --runInBand --watch=false src/__tests__/useController.test.tsx`
- `eslint src/useController.ts src/__tests__/useController.test.tsx`
- `prettier --check --single-quote --end-of-line lf src/useController.ts src/__tests__/useController.test.tsx`

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -37,6 +37,43 @@ describe('useController', () => {
     render(<Component />);
   });
 
+  it('should return a promise-like value from field.onChange', async () => {
+    let onChangeResult: any;
+
+    const Component = () => {
+      const { control } = useForm<{ test: string }>();
+      const { field } = useController({
+        control,
+        name: 'test',
+        defaultValue: '',
+      });
+
+      return (
+        <button
+          type="button"
+          onClick={() => {
+            onChangeResult = field.onChange({
+              target: {
+                name: 'test',
+                value: 'next',
+              },
+              type: 'change',
+            });
+          }}
+        >
+          change
+        </button>
+      );
+    };
+
+    render(<Component />);
+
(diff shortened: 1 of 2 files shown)


louzhedong:
Update: the latest CI run is fully green, including Cypress.

The branch is now back to the same review-ready state as the other two PRs: the functional fix is in place, the regression test is in, and the remaining blocker is review/merge flow.

Validation on this push:
- `gh run watch 27318532907 --repo react-hook-form/react-hook-form --exit-status`

src/logic/createFormControl.ts
```diff
@@ -1085,12 +1085,23 @@ export function createFormControl<
...
+        if (!errorLookupResult.error && field._f.deps) {
+          for (const dep of convertToArrayPayload(field._f.deps)) {
+            const lookupResult = schemaErrorLookup(errors, _fields, dep);
+
+            if (lookupResult.error) {
+              errorLookupResult = lookupResult;
+              break;
+            }
+          }
+        }
```
bluebill1049: can we revert this plz not related to this fix? thanks

bluebill1049 (changes requested):

dsparkplug:
thanks @louzhedong for posting a pr so quickly - but why closed without merging - the reviewer only pointed to an unrelated change in a separate file?

bluebill1049:
u only need to revert the part which i pointed out. thanks 🙏

bluebill1049 (approved):

PR #13519, opened 2026-06-11
🚷 feat: add `disabled` option to useFieldArray

louzhedong:
## What
Add a `disabled` option to `useFieldArray` to skip mount-time field-array initialization.

## Why
`useFieldArray` currently materializes missing nested arrays on mount, which mutates form values for discriminated unions and similar conditional shapes. With `disabled: true`, callers can opt out of that initialization and keep absent nested arrays absent.

## Validation
- `node_modules\\.bin\\jest.CMD --config ./scripts/jest/jest.config.js --runInBand --watch=false src/__tests__/useFieldArray.test.tsx`
- `node_modules\\.bin\\eslint.CMD src/useFieldArray.ts src/types/fieldArray.ts src/__tests__/useFieldArray.test.tsx`
- `node_modules\\.bin\\prettier.CMD --check --single-quote --end-of-line lf src/useFieldArray.ts src/types/fieldArray.ts src/__tests__/useFieldArray.test.tsx`

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -675,6 +675,7 @@ export type UseFieldArrayProps<TFieldValues extends FieldValues = FieldValues, T
     rules?: {
         validate?: Validate<FieldArray<TFieldValues, TFieldArrayName>[], TFieldValues> | Record<string, Validate<FieldArray<TFieldValues, TFieldArrayName>[], TFieldValues>>;
     } & Pick<RegisterOptions<TFieldValues>, 'maxLength' | 'minLength' | 'required'>;
+    disabled?: boolean;
     shouldUnregister?: boolean;
 };
 
--- a/src/__tests__/useFieldArray.test.tsx
+++ b/src/__tests__/useFieldArray.test.tsx
@@ -83,6 +83,48 @@ describe('useFieldArray', () => {
       ]);
     });
 
+    it('should not initialize missing nested field array values when disabled', () => {
+      type FormValues = {
+        name: string;
+        union:
+          | {
+              type: 'empty';
+            }
+          | {
+              type: 'array';
+              values: { value: string }[];
+            };
+      };
+
+      const { result } = renderHook(() => {
+        const methods = useForm<FormValues>({
+          defaultValues: {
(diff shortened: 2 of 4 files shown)


bluebill1049:
nice one, u beat me on to it.

bluebill1049 (approved):

louzhedong:
Update: the latest CI run is fully green, including Cypress.

The disabled option is now implemented, the regression test is in place, and the API report has been refreshed to match the exported surface.

BrendanC23:
Thanks for working on this PR! It might also be worth adding a test that the `disabled` prop is reactive (as it's included in the `useEffect` dependency array). If `disabled: false` is later passed to `useFieldArray()`, the value will be initialized with an empty array.

Here's a test case (passing) that demonstrates this:

```ts
it('should initialize field array when no longer disabled', async () => {
  type FormValues = {
    name: string;
    union:
      | {
          type: 'empty';
        }
      | {
          type: 'array';
          values: { value: string }[];
        };
  };

  const { result } = renderHook(() => {
    const [disabled, setDisabled] = useState(true);

    const methods = useForm<FormValues>({
      defaultValues: {
        name: '',
        union: {
          type: 'empty',
        },
      },
    });

    return {
      ...methods,
      fieldArray: useFieldArray({
        control: methods.control,
        name: 'union.values',
        disabled,
      }),
      setDisabled,
    };
  });

  expect(result.current.fieldArray.fields).toEqual([]);
  expect(result.current.getValues()).toEqual({
    name: '',
    union: {
      type: 'empty',
    },
  });

  act(() => {
    result.current.setValue('union.type', 'array');
    result.current.setDisabled(false);
  });

  expect(result.current.fieldArray.fields).toEqual([]);
  expect(result.current.getValues()).toEqual({
    name: '',
    union: {
      type: 'array',
      values: [],
    },
  });
});
```

There might be a better way to structure the `setDisabled` call without needing to return it. I'm not sure if this is better off as a separate test case or if it makes more sense to combine it with the first because most of the code is duplicated.

bluebill1049:
I will get this in first and follow up with the remaining.

PR #13532, opened 2026-06-16
fix: restore isValid after clearErrors() is called without arguments

kamja44:
clearErrors() without arguments cleared errors but did not revalidate isValid, leaving it stuck as false.

Fixed by calling _setValid() after clearing errors so isValid is recalculated correctly.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1255,9 +1255,11 @@ export function createFormControl<
       });
     } else {
       // Clear all errors - emit without name to notify all subscribers
+      _formState.errors = {};
       _subjects.state.next({
         errors: {},
       });
+      _setValid();
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/logic/createFormControl.ts
```diff
@@ -1255,9 +1255,11 @@ export function createFormControl<
       });
     } else {
       // Clear all errors - emit without name to notify all subscribers
+      _formState.errors = {};
       _subjects.state.next({
         errors: {},
       });
+      _setValid();
```
bluebill1049, line 1262: i think can set isValid to true, don't need to re-run the valdiation.

kamja44: The reason I originally used _setValid() (which re-runs the validation) was to handle cases where other registered fields might still be invalid. I wanted to ensure isValid accurately reflects the entire form's state rather than blindly setting it to true.

However, I agree that React Hook Form's lightweight philosophy of minimizing validation runs and re-renders makes directly setting isValid: true a more consistent approach

I'll close this MR. Thank you for your feedback!

PR #13555, opened 2026-06-27
fix(flatten): preserve Date values as leaf nodes

### LINKED ISSUE #13554: bug: flatten() silently drops Date values when building FormData
### Version Number

7.80.0

### Steps to reproduce

The `flatten()` utility in `src/utils/flatten.ts` is used by the `<Form>` component to convert form values into `FormData` for server submission (`src/form.tsx:65`).

When a form field's value is a `Date` object, `flatten()` silently drops it:

```js
import { flatten } from './src/utils/flatten';

const result = flatten({
  name: 'Alice',
  createdAt: new Date('2024-01-01'),
  age: 30,
});

console.log(result);
// { name: 'Alice', age: 30 }
//   ^ createdAt is silently dropped!
```

### Expected behaviour

Date values should be treated as leaf values (like `null`, strings, or numbers) and preserved:

```js
// Expected:
{ name: 'Alice', createdAt: new Date('2024-01-01'), age: 30 }
```

### Root cause

In `src/utils/flatten.ts`:

```ts
if (isObjectType(obj[key]) && obj[key] !== null) {
  const nested = flatten(obj[key]);          // flatten(someDate) → {}
  for (const nestedKey of Object.keys(nested)) { // loop never executes
    output[`${key}.${nestedKey}`] = nested[nestedKey];
  }
  // Date value is never written to output → silently dropped
}
```

`isObjectType` is `typeof value === 'object'`, which is `true` for Date objects. But `Object.keys(new Date())` returns `[]` (no own enumerable keys), so the inner loop never runs and the date disappears from the output.

This means any form using the `<Form>` component's `action` prop with a date field will silently omit that field from the submitted `FormData`.

All (Node.js repro confirmed)

### Code of Conduct

JSap0914:
## Bug

`flatten()` in `src/utils/flatten.ts` silently drops `Date` values from the flattened output. The `<Form>` component uses `flatten()` to build the `FormData` sent on form submission (`src/form.tsx:65`), so any form field holding a `Date` value is omitted from the server payload entirely.

Closes #13554

## Root cause

```ts
// Before fix — in src/utils/flatten.ts
if (isObjectType(obj[key]) && obj[key] !== null) {
```

`isObjectType(value)` is `typeof value === 'object'`, which is `true` for `Date` instances. Because `Object.keys(new Date())` returns `[]`, the inner loop is a no-op and the date is silently discarded.

## Fix

Add `!isDateObject(obj[key])` to the guard so that `Date` values fall through to the leaf branch and are preserved:

```ts
// After fix
if (isObjectType(obj[key]) && obj[key] !== null && !isDateObject(obj[key])) {
```

## Verification

```
pnpm test -- --testPathPatterns="src/__tests__/utils/flatten" --no-coverage
```

- 2 new regression tests go **RED → GREEN** (Date dropped / preserved).
- Existing snapshot test remains **GREEN** and unmodified.
- Full suite: **1184/1184 tests pass** across all 120 test suites.

> AI-assisted contribution.

### DIFF
--- a/src/__tests__/utils/flatten.test.ts
+++ b/src/__tests__/utils/flatten.test.ts
@@ -22,4 +22,25 @@ describe('flatten', () => {
       }),
     ).toMatchSnapshot();
   });
+
+  it('should preserve Date values as leaf nodes and not drop them', () => {
+    const date = new Date('2024-01-01T00:00:00.000Z');
+
+    expect(flatten({ name: 'Alice', createdAt: date, age: 30 })).toEqual({
+      name: 'Alice',
+      createdAt: date,
+      age: 30,
+    });
+  });
+
+  it('should preserve nested Date values as leaf nodes', () => {
+    const start = new Date('2024-01-01');
+    const end = new Date('2024-12-31');
+
+    expect(flatten({ range: { start, end }, label: 'year' })).toEqual({
+      'range.start': start,
+      'range.end': end,
+      label: 'year',
+    });
+  });
 });
--- a/src/utils/flatten.ts
+++ b/src/utils/flatten.ts
@@ -1,12 +1,13 @@
 import type { FieldValues } from '../types';
 
+import isDateObject from './isDateObject';
 import { isObjectType } from './isObject';
 
 export const flatten = (obj: FieldValues) => {
   const output: FieldValues = {};
 
   for (const key of Object.keys(obj)) {
-    if (isObjectType(obj[key]) && obj[key] !== null) {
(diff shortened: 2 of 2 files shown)


bluebill1049:
could u fix the lint error?

bluebill1049:
ty another pr already merged. thank you for your fix tho <3

bluebill1049:
nvm the pr was pointed to the wrong branch https://github.com/react-hook-form/react-hook-form/pull/13562

bluebill1049:
if u could fix lint, i will merge yours

bluebill1049:
all good, fix merged.

PR #13558, opened 2026-06-28
fix: prevent Date values from being dropped by flatten()

### LINKED ISSUE #13554: bug: flatten() silently drops Date values when building FormData

DakshSinghDhami:
### Problem

Date objects are silently dropped by `flatten()` because they pass the `isObjectType` check but have no enumerable own properties. This causes form Date values to disappear when building FormData.

### Changes

- Added `isDateObject` check to treat Date values as leaf nodes in flatten
- Added test verifying Date values are preserved

Fixes #13554


bluebill1049:
can u send a seperate pr? this one is not working.

Issue #13563, opened 2026-07-02
issue: formState.isSubmitting stuck true when the form is hidden by <Activity> during submit


KATT:
### Version Number

7.80.0

### Codesandbox/Expo snack

CodeSandbox link: https://codesandbox.io/p/sandbox/beautiful-einstein-42qmxz

1. Open page
2. Click submit
3. Go back

#### Expected

Button says "Review again"

#### Actual

Button says "loading"

### PR stack of repro / fixes

1. #13570 — repro: Vitest browser-mode harness with a deterministic failing test
2. #13571 — fix, stacked on #13570 (always-on internal observer keeps `_formState` authoritative + resync on `<Activity>` resume)

Tertiary: #13572 — showcases that the bug *cannot* be reproduced in Jest/jsdom or Cypress (both mask it); includes an example-app route (`/activityIsSubmitting`) for manual verification

A minimal repro is inline below (needs React 19.2+ for `<Activity>`).

### Steps to reproduce

With React's `<Activity>`, if a parent flips a `useForm` subtree to `mode="hidden"` from inside the submit handler, `formState.isSubmitting` never resets to `false`. On re-show, a submit button driven by `isSubmitting` is stuck disabled.

```tsx
import { Activity, useState } from 'react';
import { useForm } from 'react-hook-form';

function Review(props: { onBack: () => void }) {
  return <button onClick={props.onBack}>Back</button>;
}

function Form(props: { onSubmit: () => void }) {
  const form = useForm();
  return (
    <form onSubmit={form.handleSubmit(props.onSubmit)}>
      <input {...form.register('amount')} />
      <button type="submit" disabled={form.formState.isSubmitting}>Review</button>
    </form>
  );
}

function App() {
  const [reviewing, setReviewing] = useState(false);
  return (
    <>
      <Activity mode={reviewing ? 'hidden' : 'visible'}>
        <Form onSubmit={() => setReviewing(true)} />
      </Activity>
      {reviewing && <Review onBack={() => setReviewing(false)} />}
    </>
  );
}
```

1. Click **Review** → submit fires → parent flips the form to `<Activity mode="hidden">`.
2. Click **Back** → the form is shown again.
3. **Bug:** the **Review** button stays `disabled` (`isSubmitting === true`).

[...]

PR #13572, opened 2026-07-03
test: show Cypress and Jest/jsdom cannot reproduce the <Activity> isSubmitting bug

KATT:
> [!NOTE]
> **Stack** (for #13563):
>
> 1. #13570 — repro: Vitest browser-mode harness with failing test
> 2. #13571 — fix: reset `isSubmitting` (stacked on #13570)
>
> Tertiary: #13572 — showcase that the bug *cannot* be reproduced in Jest/jsdom or Cypress **← this PR**

## Background

Tertiary companion to #13563 — not required for the fix to land. It showcases that this bug **cannot be reproduced in Jest/jsdom or Cypress**. Before landing the Vitest browser-mode harness (#13570), we attempted to reproduce the `<Activity>` `isSubmitting` bug in the repo's existing test harnesses. This PR documents that **neither Cypress nor Jest/jsdom can observe the bug**, even against a build where it is present (plain Playwright against the same URL reproduces it 100% of the time, even with a 500ms delay before revealing the form).

Why they can't:

- **Jest/jsdom**: RTL wraps events in `act`, which flushes React's deferred offscreen `<Activity>` work before assertions run.
- **Cypress**: the harness yields/flushes between commands. All three assertion styles pass on a buggy build — retrying `.should`, immediate `.then`, and even clicking Back and reading the DOM synchronously in the same tick (the preceding `cy.contains` retry loop already gave React the flush opportunity).

The specs are kept as normal-path regression guards, with comments explaining the limitation so nobody mistakes them for a faithful repro.

## Changes

- example app: `/activityIsSubmitting` route (React bumped to 19.2 for `<Activity>`) for manual browser verification — the bug **is** visible when clicking through by hand
- `cypress/e2e/activityIsSubmitting.cy.ts` — three attempt variants, all documented as unable to repro
- `src/__tests__/useForm/isSubmittingActivity.test.tsx` — jsdom guard, same caveat

## Related

- Issue: #13563
- Faithful repro harness (Vitest browser mode): #13570
- Fix: #13571

## Test plan

- [ ] `pnpm start` then visit `/activityIsSubmitting`: click Review → Back; on master the button is stuck disabled (manual repro works, automated doesn't)
- [ ] `pnpm e2e` with the app running → passes (even on master, demonstrating the limitation)
- [ ] `pnpm test` → passes

Made with [Cursor](https://cursor.com)


bluebill1049:
cypress removed

PR #13583, opened 2026-07-07
fix: do not override custom date type of data in cloneObject

JesusChrisse:
Custom Date types should preserve their type and not be reset back to a basic Date type.


bluebill1049:
Ideally, we should avoid supplying custom classes or objects to RHF in the first place, as we currently don’t have strong support or sufficient test coverage for those cases. It would be better to stick to the supported patterns and avoid passing custom objects into RHF unless there is a clear need and proper coverage in place.

JesusChrisse:
I understand that. Just so you know, I have been using this code as a patch pnpm since v7.63.0.
The problem I'm facing is that the company I am part of is using DateOnly objects in the C# backend and parsing it with `new Date('2026-07-15')`. This cause a date shift because Javascript interpret this as a UTC date.
image
Well it's not a problem when you know the issue. It's just that my colleges often forget that... So I made a custom date type DateOnly on the frontend too and simply told them to use that because I was tired to repeat myself. But react-hook-form is overriding it to a simple Date since my custom Date is extending Date.
RHF is already supporting custom class object in the cloneObject well you are detecting it and returning the original like this:
image
Could we at least do the same thing of returning the original when cloneObject receive a custom Date type?

PR #13587, opened 2026-07-09
chore: upgrade pnpm to v11

abnud11:
`pnpm` 11 comes with many security policies to protect against supply chain attacks, which have recently become a threat to all libraries.

I think as library maintainers, we should migrate our project to `pnpm` 11. Fortunately, we don't have a `pnpm` entry in `package.json` and no configs to migrate(the `auto-install-peers=true` in `.npmrc` is already the default for pnpm).

This PR currently migrates the GitHub Actions. Migrating `pnpm` on devs' machines should be done by the devs themselves, because this project doesn't specify a ' pnpm version in `package.json`. I updated the contribution docs to include this information for possible contributors.


bluebill1049:
looks like [pnpm-workspace.yaml](https://github.com/react-hook-form/react-hook-form/pull/13587/files/83b91549f41a41a3c0f8581b3556cf67cdb2a5c8#diff-18ae0a0fab29a7db7aded913fd05f30a2c8f6c104fadae86c9d217091709794c) is required, can u added it back and enable to those to true

abnud11:
> looks like [pnpm-workspace.yaml](https://github.com/react-hook-form/react-hook-form/pull/13587/files/83b91549f41a41a3c0f8581b3556cf67cdb2a5c8#diff-18ae0a0fab29a7db7aded913fd05f30a2c8f6c104fadae86c9d217091709794c) is required, can u added it back and enable to those to true

I restored `pnpm-workspace.yaml`, but I don't think we should allow all builds. We should only allow a package's build (aka postinstall scripts) to run if something fails without it, in my opinion.

bluebill1049 (approved):
nice, lgtm

bluebill1049 (approved):

PR #13595, opened 2026-07-14
fix new FieldArrayComponent and add export to index

EdwardEB:
New FieldArray component was not added to src/index.ts

When adding export, FieldArray clashes with type FieldArray.
Renamed FieldArray to FieldArrayComponent and added to export. Open to other name suggestions?


bluebill1049:
oh no... :(

bluebill1049:
i think this works: https://github.com/react-hook-form/react-hook-form/pull/13596/changes

EdwardEB:
@bluebill1049 I like your solution more, I did not like the different component name.

bluebill1049:
thanks for the fix and report. my bad completely forgot to export the component...

EdwardEB:
No problem, happens to the best of us :)

PR #13599, opened 2026-07-16
🐞 fix: prevent stale _setValid() calls from overwriting isValid/isValidating

a-y-ibrahim:
## Proposed Changes

`_setValid()` is called fire-and-forget from 9 different call sites (mount, setValue, unregister, trigger, resetField, disabled-state changes, etc.) and is never awaited by its callers, with no guard against out-of-order completion. If an earlier call is still awaiting async validation (either a resolver or a field's own `validate`) when a later call starts and finishes first, the earlier call can still commit its now-stale result afterward, silently overwriting the correct, current one. This affects both `formState.isValid` and, for resolver-based forms specifically, `formState.isValidating`.

This gives each `_setValid()` invocation a monotonic call id and skips committing its result if a newer call has started since, mirroring the staleness check `onChange` already does per-field via `_updateIsFieldValueUpdated`. For any normal (non-overlapping) call, this is a no-op: the id always still matches, so behavior is unchanged.

Two commits, found and fixed in sequence:
1. The core `isValid` staleness fix, with a test reproducing a form with an async-validated field and a rule-less field: two rapid changes race, and without the fix the earlier (correct) result gets overwritten by the later, stale one.
2. While reviewing that fix, found `_updateIsValidating()` has the exact same exposure for resolver-based forms (it unconditionally clears `validatingFields` and reports `isValidating: false`), reachable independently of `onChange` (which takes its own separate resolver path, not `_setValid`, when a resolver is configured) via e.g. `resetField()` overlapping a still-pending mount-time validation. Fixed the same way, with its own regression test.

Verified both regression tests fail against the code before each respective fix and pass after. Also verified end-to-end against the actual built `dist/index.cjs.js` (not just source + jest), confirming both fixes hold through the real build pipeline.

Fixes # (no existing issue; found via direct source review and empirical reproduction)

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Checklist:

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -234,7 +234,11 @@ export function createFormControl<
       let isValid: boolean;
       if (_options.resolver) {
         isValid = isEmptyObject((await _runSchema()).errors);
-        _updateIsValidating();
+        // Same staleness concern as the isValid commit below: a superseded
+        // call finishing here would otherwise unconditionally clear
+        // validatingFields and report isValidating: false, even while a
+        // newer call is still genuinely in flight.
+        callId === _setValidCallId && _updateIsValidating();
       } else {
         isValid = await executeBuiltInValidation({
           fields: _fields,
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -224,12 +224,6 @@ export function createFormControl<
         _proxySubscribeFormState.isValid ||
         shouldUpdateValid)
     ) {
-      // _setValid is fired-and-forget from many call sites and never
-      // cancelled, so an earlier call can still be awaiting validation when
-      // a later one starts. Track which call is the most recent so a call
-      // that's been superseded by the time its validation resolves doesn't
-      // commit a stale result over a newer one, mirroring the staleness
-      // check onChange already does per-field via _updateIsFieldValueUpdated.
       const callId = ++_setValidCallId;
       let isValid: boolean;
       if (_options.resolver) {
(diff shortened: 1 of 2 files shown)


src/logic/createFormControl.ts
```diff
@@ -222,18 +224,29 @@ export function createFormControl<
         _proxySubscribeFormState.isValid ||
         shouldUpdateValid)
     ) {
+      // _setValid is fired-and-forget from many call sites and never
+      // cancelled, so an earlier call can still be awaiting validation when
+      // a later one starts. Track which call is the most recent so a call
+      // that's been superseded by the time its validation resolves doesn't
+      // commit a stale result over a newer one, mirroring the staleness
+      // check onChange already does per-field via _updateIsFieldValueUpdated.
```
```suggestion
```

src/logic/createFormControl.ts
```diff
@@ -222,18 +224,29 @@ export function createFormControl<
...
+      const callId = ++_setValidCallId;
       let isValid: boolean;
       if (_options.resolver) {
         isValid = isEmptyObject((await _runSchema()).errors);
-        _updateIsValidating();
+        // Same staleness concern as the isValid commit below: a superseded
+        // call finishing here would otherwise unconditionally clear
+        // validatingFields and report isValidating: false, even while a
+        // newer call is still genuinely in flight.
```
```suggestion
```

bluebill1049 (approved):

PR #13603, opened 2026-07-18
🐞 fix(useController): resubscribe onChange/onBlur when control changes

### LINKED ISSUE #13163: issue: useController does not resubscribe when a new control is passed

bluebill1049:
_registerProps was only ever set once via useRef, so switching the control prop kept writing field updates to the original control even though useWatch had already re-subscribed and reads reflected the new one. Store the return value of the existing control.register() call inside the effect that already re-runs on control/name change so writes stay in sync with reads.

Fixes #13163

### DIFF
--- a/src/__tests__/useController.test.tsx
+++ b/src/__tests__/useController.test.tsx
@@ -1751,7 +1751,7 @@ describe('useController', () => {
     expect(result.current.field.value).toBe('form1-value');
   });
 
-  it('should write onChange/onBlur updates to the newly-passed control (#13163)', async () => {
+  it('should write onChange updates to the newly-passed control (#13163)', async () => {
     type FormValues = {
       name: string;
     };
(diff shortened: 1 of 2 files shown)


bluebill1049:
close #13177

PR #13611, opened 2026-07-22
fix: preserve dirtyFields reference during recomputation

ManasMishraAKaThunder:
## Summary

This PR preserves the `dirtyFields` object reference during dirty field recomputation instead of replacing it with a newly allocated object.

## Problem

Since the dirty field recomputation replaces `formState.dirtyFields` with a new object, repeated calls to `setValue(..., { shouldDirty: true })` change the object reference even when the set of dirty fields remains unchanged.

This can cause unnecessary React re-renders and break code relying on referential stability.

## Solution

Instead of assigning a new object returned by `getDirtyFields()`, this change reconciles the computed values into the existing `dirtyFields` object.

This preserves the object reference while keeping the dirty field contents synchronized.

## Notes

* No changes to the current dirty-field semantics.
* Existing tests pass successfully.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -268,14 +268,23 @@ export function createFormControl<
   };
 
   const _updateDirtyFields = () => {
-    _formState.dirtyFields = getDirtyFields(
+    const nextDirtyFields = getDirtyFields(
       _defaultValues,
       _formValues,
       undefined,
       _fields,
     );
-  };
 
+    _formState.dirtyFields ??= {};
+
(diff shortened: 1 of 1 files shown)


bluebill1049:
nice fix, can we include some test coverage as well?

bluebill1049 (approved):

bluebill1049:
ty for the fix.

ManasMishraAKaThunder:
> nice fix, can we include some test coverage as well?

yes we can

bluebill1049:
thanks already merged with another pr who had fix and unit tests. thank you again tho

PR #13615, opened 2026-07-25
♻️ refactor(types): extract shared ErrorNamespacePath

candymask0712:
Hello, Bill 😁

While working on something else, I noticed that `UseFormSetError` and
`UseFormClearErrors` repeat the same `root` and `form` error path types, so I
would like to make this small type-only refactor.

### Issue

`UseFormSetError` and `UseFormClearErrors` independently declare the same
additional error paths:

```ts
'root' | `root.${string}` | 'form' | `form.${string}`;
```

### Solution

- Extract the existing path union into an internal `ErrorNamespacePath` type.

### Documentation Question

Form-level validation stores its errors under `errors.form`, but I could not
find documentation explaining whether `form` and `form.*` should be avoided as
registered field names. Would it be helpful to document this potential naming
conflict?


bluebill1049 (approved):
LGTM! 👍

PR #13624, opened 2026-07-29
🧪 test: cover Blob in FieldErrors type tests

candymask0712:
Hello, Bill 😁

The existing `FieldErrors` type test explicitly includes `Blob` in its description:

```ts
/** it should not treat Date, File, FileList or Blob as record fields */
```

However, `Blob` was missing from the actual test cases.

This PR adds top-level and nested `Blob` assertions so the test matches its description.


bluebill1049 (approved):
LGTM, ty!

PR #13626, opened 2026-07-30
fix(useWatch): sync useWatch values on restore and updates

### LINKED ISSUE #13625: issue: `useWatch` remains stale when `reset` is called while a React `Activity` is hidden
### Version Number

7.81.0

### Codesandbox/Expo snack

https://codesandbox.io/s/r949gs

### Steps to reproduce

```jsx
import { Activity, useState } from "react"
import { createRoot } from "react-dom/client"
import { useForm, useWatch } from "react-hook-form"

function WatchedName({ control }) {
  const name = useWatch({ control, name: "name" })

  return <p>useWatch: {JSON.stringify(name)}</p>
}

function App() {
  const form = useForm({ defaultValues: { name: "initial" } })
  const [mode, setMode] = useState("visible")

  return (
    <>
      <button
        onClick={() =>
          setMode((currentMode) =>
            currentMode === "visible" ? "hidden" : "visible",
          )
        }
      >
        {mode === "visible" ? "hide" : "show"}
      </button>
      <button onClick={() => form.reset({ name: "updated" })}>reset</button>
      <p>getValues: {JSON.stringify(form.getValues("name"))}</p>
      <Activity mode={mode}>
        <WatchedName control={form.control} />
      </Activity>
    </>
  )
}

createRoot(document.getElementById("root")).render(<App />)
```

1. Render the example.
2. Click **hide**.
3. Click **reset**.
4. Confirm that `form.getValues("name")` returns `"updated"`.
5. Click **show**.
6. Observe the controlled input and the value returned by `useWatch`.

### Expected behaviour

When the `Activity` becomes visible again, `useWatch` should reflect the current form value, `"updated"`.

`reset` updates the form's internal values while the activity is hidden. Re-establishing the `useWatch` subscription should synchronize its local snapshot with those current values.

Chrome

### Relevant log output

```shell
No runtime error is emitted. The rendered watched value is stale while `form.getValues()` contains the value supplied to `reset`.
```

### Code of Conduct

locphamnice:
Fixes #13625

## Summary

Fix `useWatch` returning stale values when a component is restored from a hidden React `Activity` boundary.

React Activity preserves component state and DOM while hidden, but cleans up Effects and subscriptions. If `reset()` updates form values while `useWatch` is disconnected, the previous implementation re-established the subscription without reconciling the preserved hook state. This could leave both `useWatch` and consumers such as `useController` displaying the pre-reset value.

## Root cause

`useWatch` subscribes through `useIsomorphicLayoutEffect`.

Before this change, Effect-time refreshes were limited to changes in `control` or `name`. Activity restoration recreates the Effect while preserving those values and the hook’s local state, so no refresh occurred after the subscription was restored.

As a result:

1. Activity hid the component and removed its subscription.
2. `reset()` updated `control._formValues`.
3. Activity restored the component with its previous hook state.
4. The subscription reconnected, but the watched value remained stale until another form notification occurred.

## Implementation

- Reconcile the latest form output whenever the subscription Effect is established.
- Establish the subscription before reading the snapshot so updates cannot be missed between reconciliation and subscription setup.
- Continue using `useIsomorphicLayoutEffect` so corrected values are committed before restored content is painted.
- Track a detached Effect snapshot to detect form data changed while the watcher was disconnected, including values that can be mutated in place.
- Use a functional state update with `deepEqual` to:
  - preserve the previous state when output is unchanged;
  - avoid additional initial renders;
  - prevent Effect loops for inline array names, whole-form watches, and object-valued outputs.
- Track the form output that produced the current local state so changed `compute` callbacks can be reapplied without overwriting inline `defaultValue` behavior.
- Synchronize `_computeFormValues` with the reconciled output so the next live notification does not cause a redundant render.
- Defer compute reconciliation while the watcher is disabled and apply it when the subscription becomes active again.
- Preserve the existing live-emission callback and synchronous render-time handling for `control` and `name` changes.
- Unsubscribe during cleanup and preserve the latest connected snapshot for subsequent Activity or Strict Mode reconnections.

## Tests

Added integration coverage for:

- Restoring a visible Activity after:
  - hiding the watcher;
  - calling `reset({ name: 'updated' })`;
  - confirming `getValues('name')` is current;
  - synchronously asserting both `useWatch` output and the controlled input contain `"updated"`.
- Updating a field with `setValue` while Activity is hidden and reconciling it on restoration.
- Initially hidden Activity content whose Effects have never mounted:
  - render with `"initial"`;
  - reset while hidden;
  - reveal and assert `"updated"` on the first subscription setup.
- Changing a `compute` callback while Activity is hidden.
- Preventing an unchanged notification after compute reconciliation from causing another render.
- Preserving inline array-name defaults across subscription reconnections.
- Existing compute render-count behavior.

## Validation

```bash
pnpm test src/__tests__/useWatch.test.tsx --runInBand --watchman=false
pnpm type
pnpm lint


bluebill1049 (commented):
I’m going to hold off on supporting this for now. I think the current implementation is more complicated than the value it provides, so it may be worth revisiting with a simpler approach later on.

bluebill1049:
Thanks for the PR btw. 🙏

locphamnice:
Thanks for taking the time to review it.

bluebill1049:
Thanks for understanding @locphamnice i may borrow your pr/commit once if i figure out how to simplify this.

locphamnice:
Sounds good! Feel free to reuse anything from my PR or commit. I’d be happy to review the simplified approach once you have it.

Issue #12581, opened 2025-02-06
issue: Reset API is inconsistent with array fields and keepDirtyValues option


qelso:
### Version Number

7.54.2

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/react-hook-form-usefieldarray-rules-forked-ghlpnn?file=%2Fsrc%2Findex.js%3A35%2C28

### Steps to reproduce

Go to https://codesandbox.io/p/sandbox/react-hook-form-usefieldarray-rules-forked-ghlpnn?file=%2Fsrc%2Findex.js%3A35%2C28

1. Click on append, this will add a new element to the array.
2. Click on reset, this will reset the form with a new default value that has an array with a single element and with the KeepDirtyValues option. 
3. The resulting array will contain 4 elements, where 2 of them have values set to undefined. (checked with getValues, see console logs)

Another inconsistency
1. Remove 2 elements from the array
2. Reset
3. Append an element
4. Reset
5. Append an element
6. The resulting array contains 4 elements 

### Expected behaviour

In the first issue I expected to have 2 elements as I reset the array with a single element and the previously added element should be the only dirty value.

In the second one I expected to have 3 elements: 
The first reset set the array with one element and the dirty elements are the 2 that have been removed.
Then one new element is added and reset is called, the array should contain 2 elements, the one set by reset and the added one that is a dirty value. 
However with getValues we can see that a third element with undefined values has also been added, and the last append renders the UI with 4 elements.

### What browsers are you seeing the problem on?

Chrome, Firefox

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bluebill1049:
This is the fastest way to mark dirty fields for field array, without delete/remove node inside of each array.

Issue #12099, opened 2024-07-11
issue: Duplicate fields after deleting. useFieldArray and react-query


xiaohai-huang:
### Version Number

7.52.1

### Codesandbox/Expo snack

https://stackblitz.com/edit/vitejs-vite-sjvkpq?file=src%2FApp.jsx

### Steps to reproduce

1. Go to https://stackblitz.com/edit/vitejs-vite-sjvkpq?file=src%2FApp.jsx
2. Click on the "New" button, enter some text at the input field then click the "Save" button.
3. Repeat the above step to create another todo item.
4. Click the "Delete" button at the first todo item.
5. The first todo item's text is changed to the second todo's text

[useFieldArray-bug.webm](https://github.com/react-hook-form/react-hook-form/assets/56380096/3a747ed7-01ed-41e0-8194-79c3c1f38097)

### Expected behaviour

The first todo item should be deleted instead of using the text from the next todo item.

### What browsers are you seeing the problem on?

Chrome, Edge

### Relevant log output

```shell
Sorry for not using the __codesandbox__ template. The react version of that template is low that doesn't support `react-query`, that's why I used __stackblitz__.
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

abnud11:
On second thought, it's the usage of an array field with dirty fields state I don't understand, when inspecting `dirtyFields` of `formState` I get two array items when adding and saving two items, it looks like this:

```json
{
  "todos": [
    {
      "uid": false,
      "text": false,
      "checked": false
    },
    {
      "uid": true,
      "text": true,
      "checked": true
    }
  ]
}
```
While `values` is like this:
```json
{
  "values": {
    "todos": [
      {
        "uid": "0.36061416878791586",
        "text": "234234",
        "checked": false
      },
      {
        "uid": "0.32548578068248113",
        "text": "2342626",
        "checked": false
      }
    ]
  }
}
```
See my updated example which shows multiple inspections of `values`, `watch`, `dirtyFields` and `fields` from `useFieldsArray`
https://stackblitz.com/edit/vitejs-vite-qu6gpp?file=src%2FApp.jsx

xiaohai-huang:
@abnud11 Thank you.
> when inspecting `dirtyFields` of `formState` I get two array items when adding and saving two items, it looks like this

[...]

Issue #12097, opened 2024-07-10
issue: reset() and keepDirtyValues on dirty fields containing an array always resets the value


tibbe:
### Version Number

7.52.1

### Codesandbox/Expo snack

https://codesandbox.io/s/react-keepdirtyvalues-forked-l2h95t?file=/src/App.js

### Steps to reproduce

1. Go to https://codesandbox.io/s/react-keepdirtyvalues-forked-l2h95t?file=/src/App.js
2. Click on "Add row"
3. Click on "Start fetch"
4. Observe that after 3s the "users" field is reset to its default value (i.e. an empty array), despite being dirty (i.e. it contains one row) and have `keepDirtyValues` set.

### Expected behaviour

Fields containing array values should work with `keepDirtyValues`.

Note that this is a simplified example and I can't use `useFieldArray` in my real code as the field name is dynamic. (I'm building a form builder so all field names are dynamic and only known at runtime).

### What browsers are you seeing the problem on?

Chrome

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

artptr:
The issue is reproduced since 7.0.0

PR #13630, opened 2026-07-31
🐞 fix #13627 keepDirtyValues: reconcile field array rows by id so dirty/touched/error state follows reordered rows

### LINKED ISSUE #13627: `values` + `keepDirtyValues` silently corrupts arrays and freezes objects on refetch — resync should just work by default
## Summary

`useForm({ values, resetOptions: { keepDirtyValues: true } })` is the documented way to keep a form in sync with server data that can refetch while the user edits. Today it is only safe for flat scalar fields. For the two most common structured shapes it silently produces wrong state:

### 1. Arrays: index-based merge corrupts the list

Dirty fields are merged into the incoming `values` by field path, and an array row's path is its index. When the server returns the same rows in a different order (for us: a tournament participant list the server re-sorts by seed after every save), still-dirty rows are kept at indices that now hold *different* rows — entries duplicate and others vanish. Minimal repro: rows `[Alpha, Bravo, Charlie, Delta]` with two dirty rows resync into `[Alpha, Bravo, Alpha, Bravo]`.

A second-order effect makes it latch: the corrupting reset updates `_defaultValues` to the corrupted result, so later *identical* server payloads deep-equal the defaults and the resync is skipped — the corrupted list sticks until the server data changes again.

### 2. Objects: one dirty leaf freezes the whole container

With a nested object value, dirtying a single child causes the entire container to be kept at its old value on resync — clean sibling fields stop adopting server changes. Unlike the array case there is no identity ambiguity here: the correct behavior (merge at leaf granularity, keep only the dirty leaves) is well-defined. This looks like a plain bug.

Versions checked: 7.66.1 and 7.83.0 (identical behavior). We know #8491 frames array index-reconciliation as by-design with "restructure to object/flat keys" as the escape hatch, and the `keepDirtyValues` RFC acknowledged the refetch-while-editing array case as unsolved. But restructuring naturally list-shaped domain data into flat keyed fields costs schema/typing ergonomics everywhere, and the current behavior isn't a safe degradation — it's silent corruption.

## Proposal: correct by default, no new required API

- **Objects:** merge at leaf granularity. Dirty leaves are kept, clean leaves adopt the incoming values. Pure fix, no API change.
- **Arrays:** reconciliation needs row identity — that's inherent (it's why React requires `key` on lists). But it can still be a zero-config default by convention, the way table/data libraries default `getRowId`: **if array rows have an `id` property, match rows by it** — dirty fields follow their row to its new index, server-added/removed rows are inserted/dropped. Rows without an `id` keep today's index behavior. An option (e.g. `resetOptions.keyName`) is needed only to *override* the convention for differently-named keys.

With that, `useForm({ values, resetOptions: { keepDirtyValues: true } })` just works for scalars, objects, and the overwhelmingly common id-carrying arrays alike — no restructuring, no per-form configuration.

If there's appetite for this, we're happy to contribute the implementation (we've contributed here before). If the default-by-convention is considered too breaking for a minor, we'd still take it behind an option — but the default is where the value is: nobody expects a documented sync mode to silently corrupt their list.

Our current workaround is to skip resync entirely while any field is dirty, which loses per-field granularity. Happy to put together a CodeSandbox repro of both cases on request.

jtomaszewski:
Follow-up to #13628 / closes the array half of #13627.

## Problem

`reset(values, { keepDirtyValues: true })` (and the `values` prop resync it powers) merges dirty fields into the incoming values **by path**, and an array row's path is its index. When a refetch returns the same rows in a different order, still-dirty rows are kept at indices that now hold *different* rows — entries duplicate and others vanish (`[Alpha, Bravo, Charlie, Delta]` with two dirty rows resyncs into `[Alpha, Bravo, Alpha, Bravo]`). #13628 fixed the object half; this PR fixes arrays, nested arrays included.

## Approach

Before the existing per-path dirty merge runs, a new `reconcileFieldArraysById` module realigns the previous form values **and** `dirtyFields`, `touchedFields`, `errors` from the old row order to the incoming row order, matching rows by their `id` property (the zero-config convention proposed in #13627, same as table libraries' `getRowId` default). The existing merge loop then works unchanged — every path refers to the same row on both sides, so dirty leaves stay with their row, clean leaves adopt server values, server-added rows appear and server-removed rows drop.

Addressing the concerns raised in the issue discussion:

- **“touchedFields, errors and validation refs don't follow the row”** — touched and error trees are remapped alongside values (including field array `root` errors). Validation refs are positional by design and self-heal: `useFieldArray` regenerates its keys on reset, so rows remount and re-register against the reconciled values. For fixed-position `register` inputs without `useFieldArray`, a kept-dirty value whose row moved is pushed into the refs explicitly — guarded by a `deepEqual` so an input being typed in during a background refetch is never rewritten (cursor untouched).
- **“only covers the single-key, single-array case”** / **“conflicts with useFieldArray's keyName”** — a new `resetOptions.getRowId?: (row, arrayPath) => string | number | undefined` covers renamed keys (`(row) => row.uuid`), per-array choices via `arrayPath`, composite keys, and per-array opt-out (`undefined` → index behavior). It's deliberately not named `keyName` because it means something different.
- **Safety fallbacks** — an array keeps today's index behavior whenever any row on either side lacks a unique id, and (under the default matcher) when a row's `id` field is itself dirty — an edited id no longer identifies its row, so falling back beats guessing. Ids match across `1`/`"1"` type flips; same-array duplicates fall back.
- **Second-order “latch” from the issue** — gone by construction: `_defaultValues` now stays the pristine server payload instead of a corrupted merge, so identical refetches deep-equal correctly.

Non-plain objects (`File`, `Blob`, class instances) pass through reconciliation by reference — never spread-copied.

## Tests

35 tests target this change (20 unit for the reconcile module, 15 integration in `reset.test.tsx`): reorder repro from the issue, insert/remove with a moved dirty row, nested arrays at both levels (with and without id-carrying parents), errors + touched following rows under `keepErrors`/`keepTouched`, the `values`-prop refetch flow incl. identical-payload stability, idempotent repeated reset, pristine reorder adoption, `Controller` rows, fixed-position `register` repaint, dirty `File` preservation, edited-id fallback, custom `getRowId` (global, per-array, id-flip, duplicate/missing ids), and error-`root`/state-leaf preservation.

The tests were mutation-checked: disabling reconciliation fails exactly the 10 array tests, removing the ref repaint fails exactly its test, and reverting each hardening fix fails its regression test.

Known limitation (unchanged from master): transient async-validation state (`validatingFields`, delayed-error timers) resolving mid-reset still targets old indices; it self-corrects on the next validation cycle.

Docs: JSDoc added for `getRowId` and the `reset` example; happy to send a companion PR to `react-hook-form/documentation` for the `reset`/`values` pages once this lands.

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -437,6 +437,7 @@ export type KeepStateOptions = Partial<{
     keepIsValid: boolean;
     keepSubmitCount: boolean;
     keepFieldsRef: boolean;
+    getRowId: (row: unknown, arrayPath: string) => string | number | undefined;
 }>;
 
 // @public (undocumented)
@@ -1024,8 +1025,8 @@ export type WatchValue<TFieldName, TFieldValues extends FieldValues = FieldValue
 
 // Warnings were encountered during analysis:
 //
-// src/types/form.ts:513:3 - (ae-forgotten-export) The symbol "Subscription" needs to be exported by the entry point index.d.ts
-// src/types/form.ts:887:3 - (ae-forgotten-export) The symbol "FormState_2" needs to be exported by the entry point index.d.ts
+// src/types/form.ts:514:3 - (ae-forgotten-export) The symbol "Subscription" needs to be exported by the entry point index.d.ts
+// src/types/form.ts:888:3 - (ae-forgotten-export) The symbol "FormState_2" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
 
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -1025,8 +1025,8 @@ export type WatchValue<TFieldName, TFieldValues extends FieldValues = FieldValue
 
 // Warnings were encountered during analysis:
 //
-// src/types/form.ts:514:3 - (ae-forgotten-export) The symbol "Subscription" needs to be exported by the entry point index.d.ts
-// src/types/form.ts:888:3 - (ae-forgotten-export) The symbol "FormState_2" needs to be exported by the entry point index.d.ts
+// src/types/form.ts:527:3 - (ae-forgotten-export) The symbol "Subscription" needs to be exported by the entry point index.d.ts
+// src/types/form.ts:909:3 - (ae-forgotten-export) The symbol "FormState_2" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
 
--- a/src/__tests__/logic/reconcileFieldArraysById.test.ts
+++ b/src/__tests__/logic/reconcileFieldArraysById.test.ts
@@ -195,6 +195,224 @@ describe('reconcileFieldArraysById', () => {
     expect(result.rows.root).toEqual({ type: 'min', message: 'too short' });
   });
 
+  it('should match rows with a custom getRowId', () => {
+    const previousValues = {
+      rows: [
+        { uuid: 'a', name: 'Alpha' },
+        { uuid: 'b', name: 'Bravo' },
+      ],
+    };
+    const nextValues = {
+      rows: [
+        { uuid: 'b', name: 'Bravo' },
+        { uuid: 'a', name: 'Alpha' },
+      ],
+    };
+
+    expect(
+      reconcileFieldArraysById(
+        { rows: [{ name: true }, undefined] },
+        previousValues,
+        nextValues,
+        (row) => (row as { uuid: string }).uuid,
+      ),
+    ).toEqual({
+      rows: [undefined, { name: true }],
+    });
+  });
+
+  it('should let getRowId opt arrays in or out per array path', () => {
+    const previousValues = {
+      keyed: [
+        { key: 'a', name: 'Alpha' },
+        { key: 'b', name: 'Bravo' },
+      ],
+      positional: [
+        { key: 'a', name: 'One' },
+        { key: 'b', name: 'Two' },
+      ],
+    };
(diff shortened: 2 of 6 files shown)


bluebill1049:
Thanks for the PR! We appreciate the effort and the thought behind this. After reviewing it, I don’t think we’ll move forward with this direction, but thanks again for putting it together.

jtomaszewski:
Understood, no problem — thanks for taking a look.

Would you be open to saying which part you’d rather not take? I ask because the underlying behavior keeps resurfacing (#12489, #12099, #11748, #12581, #12097), and I’d rather aim at whatever you’d accept than guess:

- the id-matching concept itself, and arrays should stay index-identified?
- the concept is fine but not as a default — an opt-in resetOptions.getRowId only?
- the size cost in the reset path?

If it’s “none of the above, this stays as-is,” that’s a fine answer too. In that case I’d like to send a docs-only PR to react-hook-form/documentation describing the current index-matching behavior on the reset page — strictly descriptive, no code change implied.

The reasoning is the principle of least surprise. Arrays work fine until identity matters: with keepDirtyValues (and the values prop that routes through it), a row is identified by its index, so a refetch that returns the same rows in a different order keeps the edits at the index rather than with the row. It fails silently and produces plausible-looking data, which is what makes it costly to discover. The docs don’t mention it today — #11748 was closed as a docs problem, and the page still doesn’t say this. A warning would let developers, and the coding agents that read these docs on their behalf, know to look for it.

On our side we’ll either carry this PR as a patch or avoid field arrays where a refetch can reorder rows, so that forms just work.

Issue #11927, opened 2024-05-23
issue: getFieldState().error type broken on nested FieldValues


elecdeer:
### Version Number

v7.51.5

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/issue-getfieldstate-error-type-broken-on-nested-fieldvalues-z3m5qy?file=%2Fsrc%2FApp.tsx%3A98%2C38

### Steps to reproduce

Due to the changes in #11831, the type of `getFieldState().error` for nested FieldValues has been broken. 

Please check Codesandbox.
In v7.51.4, the return type of `getFieldState("users.0.firstName", formState).error` was `FieldError | undefined`, but in v7.51.5, it has become `unknown`.
In runtime, when an error occurs, the value obtained from `getFieldState` is of type `FieldError`, so I think this type should be `FieldError`.

### Expected behaviour

There should not be a type error.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

bluebill1049:
reverted will release the patch soon: https://github.com/react-hook-form/react-hook-form/commit/927535bf87412b2baa91faefb4c7ded327126c2b

PR #11831, opened 2024-04-28
fix: correct type of `error` field in `getFieldState` return object

cj-young:
To fix #11814.

I changed the type of the `error` field inside the return object of a form's `getFieldState` to be more general. The current implementation doesn't account for nested errors inside objects since it uses the basic `FieldError` type. I changed it to an indexed access type that uses the type of the actual errors object present in the form state, which itself accounts for nested objects.

PR #13632, opened 2026-08-01
⌨️ fix(types): resolve getFieldState error from field path

candymask0712:
Hello, Bill 😁

While looking into the type of `getFieldState(name).error`, I noticed that it does not always match the error node returned at runtime.

### History

#11831: Direct `FieldErrors` indexing  
↓  
#11927: Regression reported  
↓  
#11831: Reverted

### Issue

`getFieldState(name).error` is always typed as `FieldError | undefined`, although parent paths can return nested error nodes from `formState.errors`.

```ts
type FormValues = {
  users: Array<{
    firstName: string;
  }>;
};

const form = useForm<FormValues>();
const error = form.getFieldState('users').error;

error?.[0]?.firstName?.message; // TypeScript error
```

If `users.0.firstName` has an error, `getFieldState('users').error` can return the nested errors for the `users` array at runtime. However, the current type treats it as a single `FieldError`, so accessing `error[0].firstName` causes a TypeScript error.

### Cause

`UseFormGetFieldState` always returns `FieldError` without considering the selected field path.

The previous direct `FieldErrors` indexing handled the parent path `users`, but the deep path `users.0.firstName` was not a direct key of `FieldErrors<FormValues>` and could resolve to `unknown`.

### Solution

- Add a public `FieldPathError` type that resolves the error type from the selected field path.
- Use it for `getFieldState(name).error`.
- Keep `FieldError` for leaf and generic paths while including nested errors for parent paths.

### Compatibility

This is a type-only change. Leaf and generic paths keep the existing `FieldError | undefined` type. Parent paths now include nested errors, so code that explicitly expects `FieldError | undefined` for a parent path may need to be updated.


bluebill1049 (approved):
Nice! 👍

PR #13635, opened 2026-08-03
fix(useWatch): prefer form defaultValues over the hook's own defaultValue

### LINKED ISSUE #13634: issue: `defaultValue` on `useWatch` overwrites the `defaultValue` on `useForm`
### Version Number

7.84.0

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/usefieldarray-values-not-set-rhf-7-56-4-forked-ry8rmc?workspaceId=ws_FUAv95EyDW2c1zyTsFrAwy

### Steps to reproduce

1. load sandbox
2. observe printed values. values obtained using `useWatch` are using the default values from `useWatch` as opposed to the `defaultValues` from `useForm`.
3. make changes to the corresponding field
4. value returned from `useWatch` is now correct

### Expected behaviour

As described in the docs: Fallback value returned before the form has mounted and no current value exists yet. Once the form is mounted, the actual current form value takes precedence over this fallback.

_No response_

### Relevant log output

```shell

```

### Code of Conduct

EduardF1:
## Proposed Changes

Fixes #13634

Before the form mounts, `_getWatch` built its value source from `{ [name]: defaultValue }` whenever a string `name` and a `defaultValue` were both supplied. That threw away `useForm({ defaultValues })` for that render, so `useWatch` returned its own fallback even when the form had a real default for the path.

The documented contract is the opposite:

> Fallback value returned before the form has mounted and no current value exists yet. Once the form is mounted, the actual current form value takes precedence over this fallback.

The fix reads from `_defaultValues` for a string name and lets `generateWatchOutput` apply the fallback through `get(values, name, defaultValue)`, which already returns `defaultValue` when the path is genuinely absent. Behaviour for an array of names, and for `useWatch` with no `defaultValue`, is unchanged.

```tsx
const { control } = useForm({ defaultValues: { test: 'form default' } });
const value = useWatch({ control, name: 'test', defaultValue: 'inline fallback' });
// before: 'inline fallback'
// after:  'form default'
```

### Snapshot update

`useWatch › fieldArray with shouldUnregister true › should watch correct input update with single field array input` is the one existing test whose recorded output changes. It passes `fields` from `useFieldArray` as the `defaultValue`, so its first entry used to carry the internal `id` keys; it now reads the form's defaults, which do not. Every later entry in that same snapshot was already `id`-free, so this makes the first render consistent with the rest.

The snapshot also gains one entry. The update that reconciles the pre-mount value with `_formValues` is now a no-op state write (the two are the same reference), and React still renders once before bailing out. The rendered DOM is unchanged — the test's `getByText` assertions pass untouched.

## Type of change

## Checklist:

`pnpm test` 1219 passed / 120 suites, `pnpm test:type`, `pnpm type` and `pnpm lint` all clean locally.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -788,13 +788,14 @@ export function createFormControl<
       names,
       _names,
       {
+        // Before mount the form's own default values are the source of truth.
+        // For a single name the caller supplied `defaultValue` is only a
+        // fallback, applied by generateWatchOutput when the path is absent.
         ...(_state.mount
           ? _formValues
-          : isUndefined(defaultValue)
+          : isUndefined(defaultValue) || isString(names)
             ? _defaultValues
-            : isString(names)
-              ? { [names]: defaultValue }
-              : defaultValue),
+            : defaultValue),
       },
       isGlobal,
       defaultValue,
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -788,9 +788,6 @@ export function createFormControl<
       names,
       _names,
       {
-        // Before mount the form's own default values are the source of truth.
-        // For a single name the caller supplied `defaultValue` is only a
-        // fallback, applied by generateWatchOutput when the path is absent.
         ...(_state.mount
           ? _formValues
           : isUndefined(defaultValue) || isString(names)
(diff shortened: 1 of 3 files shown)


src/logic/createFormControl.ts
```diff
@@ -788,13 +788,14 @@ export function createFormControl<
       names,
       _names,
       {
+        // Before mount the form's own default values are the source of truth.
+        // For a single name the caller supplied `defaultValue` is only a
+        // fallback, applied by generateWatchOutput when the path is absent.
```
```suggestion
```

bluebill1049 (approved):

hongkiulam:
Thanks for the quick fix :) does it also patch the behaviour for field arrays?

PR #13637, opened 2026-08-04
🐞 fix #13636 emit values state only once per setValue

### LINKED ISSUE #13636: issue: The `subscribe` callback is invoked twice for a single `setValue`
### Version Number

7.84.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/empty-shadow-z85s4v?file=%2Fsrc%2FApp.tsx%3A19%2C1

### Steps to reproduce

  1. Open the reproduction link.
  2. Click `setValue weight once`.
  3. Check the console output.
  4. The subscription callback is invoked twice:

```text
{ "name": "weight", "weight": number }
{ "name": "weight", "weight": number }
```

 The `subscribe` callback is invoked twice for a single `setValue` call.  The reproduction does not use custom components, third-party components, or custom refs.

### Expected behaviour

 One `setValue` call should produce one subscription callback.

Chrome

### Relevant log output

```shell

```

### Code of Conduct

anshusaurav:
Fixes #13636

## Problem

`subscribe({ formState: { values: true } })` invokes its callback **twice** for a single `setValue` call when the target field's ref has no native input `type` — i.e. controlled fields (`Controller` / `useController`, whose ref is a synthetic proxy) or a field registered without a rendered input. A plain typed `<input>` is unaffected (fires once).

## Root cause

For such a field, two `values` state notifications are emitted per `setValue` in `createFormControl.ts`:

1. `setFieldValue` emits `{ name, values }` in its `!fieldReference.ref.type` branch.
2. `_setValue` then emits the canonical `values` state update at the end.

Both pass the `values` subscription filter, so the callback fires twice. Native typed inputs skip (1) because their `ref.type` is set, which is why the bug only surfaces for controlled / input-less fields.

## Fix

Thread a `skipValueRender` flag so that when `_setValue` is going to emit the `values` update itself (value changed and state emit not skipped), the redundant field-level emit in `setFieldValue` / `setFieldValues` is suppressed. Touch/dirty rendering is untouched — only the duplicate `values` emit is skipped.

## Tests

- Added a regression test in `subscribe.test.tsx`: a `values` subscription is invoked exactly once per `setValue`. It fails on `master` (2 calls) and passes with this change.
- Full suite green: **1219 tests / 120 suites**, plus `tsc --noEmit`, eslint, and prettier all clean.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -815,6 +815,7 @@ export function createFormControl<
     options: SetValueConfig = {},
     skipClone = false,
     skipRender = false,
+    skipValueRender = false,
   ) => {
     const field: Field = get(_fields, name);
     let fieldValue: unknown = value;
@@ -863,7 +864,7 @@ export function createFormControl<
         } else {
           fieldReference.ref.value = fieldValue;
 
-          if (!fieldReference.ref.type && !skipRender) {
+          if (!fieldReference.ref.type && !skipRender && !skipValueRender) {
             _subjects.state.next({
               name,
               values: skipClone ? _formValues : cloneObject(_formValues),
@@ -901,6 +902,7 @@ export function createFormControl<
     options: U,
     skipClone = false,
     skipRender = false,
+    skipValueRender = false,
   ) => {
     if (_names.array.has(name)) {
       _subjects.array.next({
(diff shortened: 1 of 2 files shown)


src/logic/createFormControl.ts
```diff
@@ -973,10 +989,27 @@ export function createFormControl<
         (Array.isArray(cloneValue) && !cloneValue.length) ||
         isEmptyObject(cloneValue);
 
+      // When the value changed, `_setValue` emits the `values` state update
+      // below, so the field-level emit would be a duplicate notification.
```
```suggestion
```

bluebill1049 (approved):

bluebill1049 (approved):
LGTM

PR #13642, opened 2026-08-06
🐞 fix #13641 stale render re-creating field array path vacated by array action

### LINKED ISSUE #13641: issue: move() on a parent field array grows a nested array of a sibling item
### Version Number

7.81.0 - master (7.80.0 is fine)

### Codesandbox/Expo snack

https://codesandbox.io/p/devbox/rhf-move-nested-array-repro-forked-9kfy3x

### Steps to reproduce

`move()` on a parent field array grows a nested array of a sibling item.

`move()` reorders `_formValues` and `_fields` synchronously, but the components that
own the field names update through subscriptions. Any render that happens in between
still uses the pre-move arrangement, so a leaf renders a path the move has just
vacated. `useController` evaluates
[`control.register(name, { value })`](https://github.com/react-hook-form/react-hook-form/blob/v7.84.0/src/useController.ts#L98-L104)
on every render, which re-creates that path in `_fields` as a new field, and
[`updateValidAndValue`](https://github.com/react-hook-form/react-hook-form/blob/v7.84.0/src/logic/createFormControl.ts#L389-L417)
then writes the stale value into `_formValues`. The array grows.

Repeating the reorder keeps growing it: `[1, 2, 1]` -> `[1, 2, 2]` -> `[2, 2, 2]`
(entries per item).

#### To Reproduce

Two failing tests, runnable in this repository's jest setup. Shared fixture:

```tsx
import React from 'react';
import { act, fireEvent, render, screen } from '@testing-library/react';

import { Controller, useFieldArray, useForm, useWatch } from 'react-hook-form';
import type { Control, UseFormReturn } from 'react-hook-form';

type FormValues = {
  steps: { approvers: { value: string }[] }[];
};

const defaultValues: FormValues = {
  steps: [
    { approvers: [{ value: 'a1' }] },
    { approvers: [{ value: 'b1' }, { value: 'b2' }] },
    { approvers: [{ value: 'c1' }] },
  ],
};

const Step = ({
  index,
  control,
}: {
  index: number;
  control: Control<FormValues>;
}) => {
  const step = useWatch({ control, name: `steps.${index}` as const });

  return (
    <div>
      {(step?.approvers ?? []).map((_: unknown, j: number) => (
        <Controller
          key={j}
          control={control}
          name={`steps.${index}.approvers.${j}.value` as const}
          render={({ field, fieldState }) => (
            <span>
              <input {...field} />
              {fieldState.error ? 'err' : ''}
            </span>
          )}
        />
      ))}
    </div>
  );
};
```

The first test is `move()` inside a transition, with a synchronous broadcast carrying
the parent path in the same event. It passes on 7.80.0 and fails from 7.81.0:

```tsx
it('regression (7.81.0+): transition move + sync parent-named broadcast', async () => {
  let methods!: UseFormReturn<FormValues>;

  const App = () => {
    methods = useForm<FormValues>({ defaultValues });
    const { fields, move } = useFieldArray({
      control: methods.control,
      name: 'steps',
    });

    return (
      <div>
        {fields.map((field, i) => (
          <Step key={field.id} index={i} control={methods.control} />
        ))}
        <button
          type="button"
          onClick={() => {
            React.startTransition(() => move(1, 2));
            methods.setError('steps', { type: 'server', message: 'x' });
          }}
        >
          move
        </button>
      </div>
    );
  };

  render(<App />);

  fireEvent.click(screen.getByText('move'));
  await act(async () => {});

  expect(
    methods.getValues().steps.map((s) => (s.approvers ?? []).length),
  ).toEqual([1, 1, 2]);
});
```

The `setError` is what opens the window: 7.81.0 added ancestor matching to
`shouldSubscribeByName` (#13553), so a notification named `steps` now reaches leaves
subscribed with their own exact name. `useWatch` ignores a broadcast that carries no
values, so the component that owns the names does not re-render, and a synchronous
pass renders only the leaves of the old tree.

#### Observed sequence

From an app on 7.84.0, with `set` instrumented to record every write that changes the
nested array lengths. One `move(1, 2)`:

```
_formValues  [1,2,1] -> [1,1,2]   steps                              set < _setFieldArray < move
_fields      [1,1,2] -> [1,2,2]   steps.1.approvers.1.assignment.a   set < register < useController < <leaf>
_formValues  [1,1,2] -> [1,2,2]   steps.1.approvers.1.assignment.a   set < updateValidAndValue < register < useController
```

No commit runs between the first and the last write: the whole thing happens inside one
render pass. `_setFieldArray` applies `moveArrayAt` to `_fields` as well, so the stale
`register` takes the new-field path and reaches `updateValidAndValue`.
`get(_defaultValues, 'steps.1.approvers.1...')` is `undefined` throughout, so the
resurrected value comes from the stale render, not from defaults.

#### Why a render in that gap sees pre-move values

`useWatch` gets a new value through exactly two paths:

1. Subscription, post-commit only. The broadcast inside `_setFieldArray` carries
   `{ name, isDirty, dirtyFields, errors, isValid }` and no values
   ([createFormControl.ts#L339-L345](https://github.com/react-hook-form/react-hook-form/blob/v7.84.0/src/logic/createFormControl.ts#L339-L345)),
   so [`shouldRenderFormState`](https://github.com/react-hook-form/react-hook-form/blob/v7.84.0/src/logic/shouldRenderFormState.ts#L9-L30)
   drops it for `{ values: true }` subscribers. The broadcast that does carry values
   comes from `useFieldArray`'s passive effect, which runs after the commit.
2. A synchronous refresh during render, only when the hook's `name` or `control`
   changed in that render.

So any component whose `name` is unchanged across the move renders pre-move values in
every render before the commit. That is the window.

`flushSync(() => move(1, 2))` does not close it. In the app it makes the owning
component itself render the vacated path:

```
step idx1   useWatch: 2 entries   control._formValues: [1,1,2]
```

The name is unchanged there because the component's index comes from a separate display
array that `flushSync` does not update, so path 2 never fires. Removing `exact: true`
from that `useWatch` changes nothing, as expected — the drop happens on the values flag,
not on name matching.

#### Not only an exposure change

The write-back itself predates 7.81.0. Two variants reproduce it on 7.80.0 as well:

- `setError` sent directly at a leaf path
- keying the field array map by index instead of by `field.id`, then calling a plain
  `move(1, 2)` — with an index key the owning component's `name` does not change, so
  path 2 above never fires and the pre-move render is guaranteed. Using the same
  `Step` as above, this fails on every version tested:

```tsx
it('mechanism (all versions): plain move with index-keyed steps', async () => {
  let methods!: UseFormReturn<FormValues>;

  const App = () => {
    methods = useForm<FormValues>({ defaultValues });
    const { fields, move } = useFieldArray({
      control: methods.control,
      name: 'steps',
    });

    return (
      <div>
        {fields.map((_, i) => (
          <Step key={i} index={i} control={methods.control} />
        ))}
        <button type="button" onClick={() => move(1, 2)}>
          move
        </button>
      </div>
    );
  };

  render(<App />);

  fireEvent.click(screen.getByText('move'));
  await act(async () => {});

  expect(
    methods.getValues().steps.map((s) => (s.approvers ?? []).length),
  ).toEqual([1, 1, 2]);
});
```

Keying by index is discouraged in the docs, so it is included here as evidence about the
mechanism rather than as the reported case. What 7.81.0 changed is that an
ancestor-named broadcast now reaches leaves, which makes the window easy to hit through
ordinary code. Reverting #13553 alone would narrow the exposure without removing the
cause.

### Expected behaviour

Reordering moves items without changing the contents of any item.

`register` during render should not re-create a path in `_formValues` that a field
array operation has just removed.

Chrome (the mechanism is scheduling-level, not browser-specific)

### Relevant log output

```shell
_formValues  [1,2,1] -> [1,1,2]   steps                              set < _setFieldArray < move
_fields      [1,1,2] -> [1,2,2]   steps.1.approvers.1.assignment.a   set < register < useController < <leaf>
_formValues  [1,1,2] -> [1,2,2]   steps.1.approvers.1.assignment.a   set < updateValidAndValue < register < useController
```

### Code of Conduct

hairihou:
## Proposed Changes

Fixes #13641

A component that still holds a pre-action field array arrangement can `register` a
path the array operation has just removed. `register` runs during render via
`useController` (the `useRef(control.register(name, { value }))` argument is evaluated
on every render), re-creates the path in `_fields` as a new field, and
`updateValidAndValue` materializes the stale value into `_formValues` — resizing a
nested array or resurrecting a removed item. The issue above has the full mechanism
and an instrumented write sequence from a real app.

This change records each acted-on array together with its pre-action registered length
(`_state.actionNames: Map<InternalFieldName, number>`, captured from `_fields` before
the method applies) and makes `updateValidAndValue` skip the materialization for a
path under an acted array when its out-of-range crossing sits:

- below the acted array (a nested array inside a moved/removed item — always a stale
  arrangement), or
- at the acted array itself on an index inside its pre-action bounds (an index the
  action vacated).

Indices beyond the pre-action bounds still materialize, so staging a controlled field
past the end of the acted array in the same commit keeps working. Arrays untouched by
the handler are never affected: one handler acting on several arrays keeps a scope per
array. Each entry clears where the action flag resets, when the owning `useFieldArray`
unmounts or becomes disabled, and on `reset`, so a leaked action flag cannot suppress
registrations indefinitely.

Regression tests cover: the reported shape (`move` with `useWatch`-derived leaves,
both a transition-interleaved render and an index-keyed render), `remove` with a
`useWatch`-driven list (item resurrection), a controlled field mounted in the same
commit as an append to another array, a controlled field staged at the end of the
appended array itself, and a two-array handler (`move` + `append`). Each test was
verified to fail against the guard-less baseline or an earlier revision of the guard.

Two boundaries of the heuristic are pinned by tests instead of being hidden. A
controlled field staged at an index the action just vacated is indistinguishable from
the stale re-registration and is treated as stale — `insert` / `update` express that
intent, and before this change such a registration materialized the removed item's old
value rather than the staged one whenever form defaults covered it. And the pre-action
bound counts registered fields, so a row that never registered an input (conditionally
rendered rows) cannot raise it.

Bundle size cost is +164 bytes gzip (13224 -> 13388 on `dist/index.cjs.js`).

## Type of change

## Checklist:

### DIFF
--- a/reports/api-extractor.md.api.md
+++ b/reports/api-extractor.md.api.md
@@ -45,6 +45,7 @@ export type Control<TFieldValues extends FieldValues = FieldValues, TContext = a
     _state: {
         mount: boolean;
         action: boolean;
+        actionNames: Map<InternalFieldName, number>;
         watch: boolean;
     };
     _reset: UseFormReset<TFieldValues>;
@@ -1030,7 +1031,7 @@ export type WatchValue<TFieldName, TFieldValues extends FieldValues = FieldValue
 // Warnings were encountered during analysis:
 //
 // src/types/form.ts:513:3 - (ae-forgotten-export) The symbol "Subscription" needs to be exported by the entry point index.d.ts
-// src/types/form.ts:887:3 - (ae-forgotten-export) The symbol "FormState_2" needs to be exported by the entry point index.d.ts
+// src/types/form.ts:888:3 - (ae-forgotten-export) The symbol "FormState_2" needs to be exported by the entry point index.d.ts
 
 // (No @packageDocumentation comment for this package)
 
--- a/src/__tests__/useFieldArray/append.test.tsx
+++ b/src/__tests__/useFieldArray/append.test.tsx
@@ -680,4 +680,120 @@ describe('append', () => {
       await screen.findByText('{"test":[{"id":"whatever","test":"1234"}]}'),
     ).toBeVisible();
   });
+
+  it('should keep the value of a controlled field mounted in the same commit as an append to another field array', async () => {
+    let getValues: () => unknown = () => ({});
+
+    const OtherInput = ({ control }: { control: Control<FormValues> }) => {
+      const { field } = useController({
+        control,
+        name: 'b.1.value',
+        defaultValue: 'mounted-with-append',
+      });
+      return <input {...field} />;
+    };
+
+    type FormValues = {
+      a: { value: string }[];
+      b: { value: string }[];
+    };
+
+    const App = () => {
+      const [extra, setExtra] = React.useState(false);
+      const methods = useForm<FormValues>({
+        defaultValues: { a: [{ value: 'a0' }], b: [{ value: 'b0' }] },
+      });
+      getValues = methods.getValues;
+      const { fields: aFields, append } = useFieldArray({
+        control: methods.control,
+        name: 'a',
+      });
+      useFieldArray({ control: methods.control, name: 'b' });
+
+      return (
+        <div>
+          <p>{aFields.length}</p>
+          {extra && <OtherInput control={methods.control} />}
+          <button
+            type="button"
+            onClick={() => {
+              append({ value: 'a1' });
+              setExtra(true);
+            }}
+          >
+            append
+          </button>
+        </div>
+      );
+    };
+
+    render(<App />);
+
+    fireEvent.click(screen.getByRole('button'));
+    await act(async () => {});
+
+    expect(getValues()).toEqual({
+      a: [{ value: 'a0' }, { value: 'a1' }],
+      b: [{ value: 'b0' }, { value: 'mounted-with-append' }],
+    });
+  });
+
+  it('should materialize a controlled field staged at the end of the array appended to in the same commit', async () => {
+    type FormValues = { test: { value: string }[] };
+
+    let getValues: () => unknown = () => ({});
+
+    const StagedInput = ({
+      control,
+      index,
+    }: {
+      control: Control<FormValues>;
+      index: number;
+    }) => {
+      const { field } = useController({
+        control,
+        name: `test.${index}.value` as const,
+        defaultValue: 'staged-by-user',
+      });
+      return <input {...field} />;
+    };
+
+    const App = () => {
+      const methods = useForm<FormValues>({
+        defaultValues: { test: [{ value: 't0' }] },
+      });
+      getValues = methods.getValues;
+      const { fields, append } = useFieldArray({
+        control: methods.control,
+        name: 'test',
+      });
+      const [staged, setStaged] = React.useState(false);
+
+      return (
+        <div>
+          {staged && (
+            <StagedInput control={methods.control} index={fields.length} />
+          )}
+          <button
+            type="button"
+            onClick={() => {
+              append({ value: 't1' });
+              setStaged(true);
+            }}
+          >
+            append
+          </button>
+        </div>
+      );
+    };
+
+    render(<App />);
+
+    fireEvent.click(screen.getByRole('button'));
+    await act(async () => {});
+
+    expect(getValues()).toEqual({
+      test: [{ value: 't0' }, { value: 't1' }, { value: 'staged-by-user' }],
+    });
+  });
 });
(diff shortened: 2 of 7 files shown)


bluebill1049:
This PR feels a bit complex for the value of the edge case it’s trying to address. I’ll take a look to see if there’s a simpler alternative or workaround that could solve this without adding too much.

PR #13643, opened 2026-08-06
🧹 chore: remove dead code and stale api-extractor report

### LINKED ISSUE #7109: issue: Populating form with reset and keepDefaultValues as true breaks fieldarray actions
### Version Number

7.20.2

### Codesandbox/Expo snack

https://codesandbox.io/s/react-hook-form-usefieldarray-forked-fxzgz?file=/src/index.js

### Steps to reproduce

1. Go to https://codesandbox.io/s/react-hook-form-usefieldarray-forked-fxzgz?file=/src/index.js
2. Click 'Load items'
3. Click 'append' or 'delete'

### Expected behaviour

FieldArray actions (append, delete etc.) should function normally after resetting the form with new values and keepDefaultValues as true.

Chrome

### Relevant log output

```shell
Uncaught TypeError: dirtyFields.slice is not a function
    at setFieldArrayDirtyFields (index.esm.js:945)
    at updateFieldArrayDirty (index.esm.js:1334)
    at Object._updateFieldArray (index.esm.js:1277)
    at Object.append$1 [as append] (index.esm.js:481)
    ...
```

### Code of Conduct

saulo-silva:
## Proposed Changes

This PR removes two dead internal modules and a stale API report file, and aligns the api-extractor config with the report it actually generates.

**Dead code (only referenced by their own test files):**

- `src/utils/deepMerge.ts` — orphaned since 9d22c8b7 (fix #7109, Nov 2021) removed its last production call site. No imports anywhere in `src/` outside its own test since then.
- `src/logic/getNodeParentName.ts` — orphaned after f29f546e (#13328) consolidated its call sites; the logic it used to serve now lives in `isNameInFieldArray`, which handles nested field arrays correctly. Also updated a comment in `isNameInFieldArray.test.ts` that referenced the removed function.

Both modules are already tree-shaken from the published bundle, so there is no runtime or bundle-size impact — this is purely source hygiene (−107 lines of unused code and tests).

**Stale API report:**

- Deleted `reports/api-extractor.md` — api-extractor enforces the `.api.md` extension, so the report is actually generated and validated as `reports/api-extractor.md.api.md`. The old file stopped being rewritten (last touched Mar 2025) and has drifted ~500 lines behind: it is missing `Watch`, `FormStateSubscribe`, `setValues`, the current `Controller` signature, and more, while still being tracked in git as if it were the public API report.
- Set `reportFileName` to `api-extractor.api.md` in `api-extractor.json` (matching the extension requirement documented in the config itself) and renamed the generated report accordingly: `reports/api-extractor.md.api.md` → `reports/api-extractor.api.md`. Re-running `pnpm api-extractor:build` reproduces the renamed file with no diff, and `pnpm api-extractor:ci` passes.

## Type of change

- [x] None of the listed options — internal cleanup with no runtime, public API, or bundle changes

## Checklist:

- [x] New and existing tests pass locally with my changes (`pnpm test`: 118 suites / 1215 tests, `pnpm test:type`, `pnpm lint`, `pnpm build`, `pnpm api-extractor:ci`)


saulo-silva:
Heads-up on the failing `playwright-run` check: the `useFieldArrayUnregister.spec.ts` failure appears to be a pre-existing flake on `master`, unrelated to this PR (which only removes files with no production importers — `dist/index.esm.mjs` builds byte-identical to master's).

Evidence from local runs of `pnpm e2e useFieldArrayUnregister --repeat-each=3`:

- this branch: 3/3 failed
- `master` (9fde5f38): one invocation passed 3/3, two subsequent invocations failed — same assertion as CI (the `dirtyFields` poll after `move(4, 2)`, missing `conditional: true` at index 2)

The Playwright page snapshot at the failure point shows the conditional input at index 2 **is** mounted and `touchedFields` moved `{ name, conditional }` to index 2 correctly — but `dirtyFields` lost the `conditional` flag. So it looks like a timing race between the `move()` dirty-state recomputation and the conditional field's `unregister({ keepDirty: true })` / re-`register` effect cycle, rather than anything in this diff.

I've re-pushed to retrigger CI. Happy to open a separate issue with the full repro details if that's useful.

bluebill1049 (approved):
LGTM 🙏

src/__tests__/logic/getNodeParentName.test.ts
```diff
@@ -1,17 +0,0 @@
-import getNodeParentName from '../../logic/getNodeParentName';
```
bluebill1049, line 1: this file still used tho

src/__tests__/utils/deepMerge.test.ts
```diff
@@ -1,77 +0,0 @@
-import deepMerge from '../../utils/deepMerge';
```
bluebill1049, line 1: this still used

src/__tests__/logic/getNodeParentName.test.ts
saulo-silva: `getNodeParentName` lost its last production consumer in 302d1608 (#13318), where `isNameInFieldArray` was rewritten to check all ancestor paths without it. Since then, the only remaining importer is this test file — which is why the test is removed together with the source. `git grep getNodeParentName` on both `master` and `react-complier` finds no other references, and the full test suite + `api-extractor:ci` pass green with the removal.

src/__tests__/utils/deepMerge.test.ts
saulo-silva: Same story for `deepMerge` — its last production usage was removed back in 9d22c8b7 (#7119, 2021). Since then, the only remaining importer is this test file, so it goes together with the source. `git grep deepMerge` on both `master` and `react-complier` finds no other references, and CI stays green with both removals.

PR #13645, opened 2026-08-07
Fix: field array root error being lost on append/prepend/insert/remove

zigzagdev:
### what i have done
- `useFieldArray`'s array-level error (`errors.<name>.root`, set via `rules.validate` / `rules.minLength` / `rules.maxLength`) was silently dropped whenever `append`/`prepend`/`insert`/`remove` ran.
- Root cause: `_updateFieldArray` rebuilds the errors array by calling the same array-mutation helper used for `_fields`/`values` (`appendAt`/`insertAt`/`removeArrayAt`/`prependAt`), all of which use spread/`slice` and therefore only copy indexed elements — the non-index `root` property is lost. `swap`/`move` mutate the array in place, so they were unaffected, leaving behavior inconsistent across field-array operations.
- Fix: `_updateFieldArray` now re-attaches the previous `root` error onto the rebuilt array, and `unsetEmptyArray` no longer clears the whole error entry when a `root` error is the only thing left (previously it would immediately wipe the just-restored `root` error when there were no per-item errors).

## Before fixed

```tsx
const { control, trigger } = useForm({ defaultValues: { items: [{ name: 'a' }] } });
const { append } = useFieldArray({
  control,
  name: 'items',
  rules: { validate: (v) => v.length >= 3 || 'Need at least 3 items' },
});

await trigger('items'); // errors.items.root.message === 'Need at least 3 items'
append({ name: 'b' });  // still only 2 items, but errors.items.root becomes undefined
```

### Test plans

- [x] Added a regression test in `src/__tests__/useFieldArray.test.tsx` that fails on the old code and passes with the fix.
- [x] `pnpm test -- src/__tests__/useFieldArray.test.tsx src/__tests__/useFieldArray src/__tests__/logic` — 31 suites / 259 tests passed.
- [x] Full suite: `pnpm test` — 95 suites / 802 tests passed.
- [x] `tsc --noEmit` — no errors.
- [x] `eslint` — clean.

### Test Results
スクリーンショット 2026-08-07 10 40 37


src/logic/createFormControl.ts
```diff
@@ -309,11 +309,12 @@ export function createFormControl<
...
-        const errors = method(
-          get(_formState.errors, name),
-          args.argA,
-          args.argB,
-        );
+        const fieldArrayErrors = get(_formState.errors, name);
+        const errors = method(fieldArrayErrors, args.argA, args.argB);
+        const rootError = (fieldArrayErrors as { root?: FieldError }).root;
+        if (rootError) {
+          (errors as { root?: FieldError }).root = rootError;
```
bluebill1049: i believe this would crash

eg

`Cannot set properties of undefined (setting 'root')`

bluebill1049 (changes requested):
thanks for the PR! We probably need some better test coverage on this change as well and a regression test covering swap (and ideally move) with a root error present.

src/logic/createFormControl.ts
zigzagdev: @bluebill1049 

> i believe this would crash

You're right, this crashes. swap returns void, [fixed it](https://github.com/react-hook-form/react-hook-form/pull/13645/commits/464af4113dba141290c508109f55e00b847b96fe) — added tests for swap/move too.

src/logic/createFormControl.ts
```diff
@@ -309,11 +309,15 @@ export function createFormControl<
...
         Array.isArray(get(_formState.errors, name))
       ) {
-        const errors = method(
-          get(_formState.errors, name),
-          args.argA,
-          args.argB,
-        );
+        const fieldArrayErrors = get(_formState.errors, name);
+        // swap mutates the array in place and returns void, so fall back to
+        // the (now mutated) original array instead of overwriting with undefined
```
```suggestion
```

src/logic/createFormControl.ts
```diff
@@ -309,11 +309,13 @@ export function createFormControl<
...
-          args.argA,
-          args.argB,
-        );
+        const fieldArrayErrors = get(_formState.errors, name);
+        const errors =
+          method(fieldArrayErrors, args.argA, args.argB) ?? fieldArrayErrors;
+        const rootError = (fieldArrayErrors as { root?: FieldError }).root;
+        if (rootError && !(errors as { root?: FieldError }).root) {
+          (errors as { root?: FieldError }).root = rootError;
+        }
```
bluebill1049: Thanks for the update, the code looks rather complicated, i will adjust.

bluebill1049 (approved):

PR #13646, opened 2026-08-07
Fix: min/max validation being skipped for `valueAsDate` fields

zigzagdev:
### What I have done

Fixed `validateField` so `min`/`max` validation is no longer silently skipped on fields registered with `valueAsDate: true`.

### Motivation

When `register(name, { valueAsDate: true })` is used, `getFieldValueAs` converts the field's value to a `Date` instance before it's stored in `_formValues`. `validateField`'s `min`/`max` branch decided which comparison path to take based on `!isNaN(inputValue as number)`. 
Since `isNaN` coerces its argument with `Number(...)`, and `Number(validDate)` is a valid timestamp, a `Date` instance always took the numeric branch instead of the intended date branch. 
That numeric branch then compared a millisecond timestamp against a date-formatted string (e.g. `'2010-01-01'`), which is always a `NaN` comparison — so `exceedMax`/`exceedMin` were always `false` and `min`/`max` on `valueAsDate` fields silently never fired.

**Code**

   ```tsx
   register('birthDate', { valueAsDate: true, max: '2010-01-01' });
   // entering 2025-01-01 into <input type="date" /> never raises a max error
   ```

Fixed by explicitly excluding `Date` instances from the numeric-branch condition so they're routed to the date-comparison branch, where `new Date(existingDateInstance)` already clones correctly at runtime.

### Test Plans
 - Added a regression test in `src/__tests__/logic/validateField.test.tsx` that passes a real `Date` instance as the field value (mirroring what `getFieldValueAs` produces for `valueAsDate` fields) and asserts `min`/`max` errors are still raised.
 - Verified the new test fails on the old code and passes with the fix.
 - Ran the full suite, lint, and type-check in a Docker container matching CI (Node 22 + pnpm 11.7.0).

### Test Results
```bash
react-hook-form:/workspace -w /workspace rhf-dev-v2 sh -c "pnpm test -- src/__tests__/logic/validateField.test.tsx"

$ jest --config ./scripts/jest/jest.config.js -- src/__tests__/logic/validateField.test.tsx
PASS Web src/__tests__/logic/validateField.test.tsx
  validateField
    ✓ should return required true when input not filled with required (16 ms)
    ✓ should return max error (3 ms)
    ✓ should return min error (2 ms)
    ✓ should return min error when the field value is already a Date object (valueAsDate) (1 ms)
    ✓ should return min and max error for custom input (1 ms)
    ✓ should return max
    ✓ should return min
    ✓ should return pattg (3 ms)
    ✓ should validate fo
    ✓ should return erroed (1 ms)
    ✓ should return result or empty string when validate has error
    ✓ if undefined returned from validate, no error is reported (1 ms)
    ✓ should do nothing ject nor function (1 ms)
    ✓ should return all
    ✓ should handle patt
    ✓ should return all (12 ms)
    ✓ should validate fitribute (1 ms)
    with Browser native
      ✓ should invoke sed input (1 ms)
      ✓ should invoke sed input with its message
(1 ms)
      ✓ should invoke se string for a valid input
(1 ms)
      ✓ should invoke ses for a required radio
group (1 ms)
      ✓ should clear setCustomValidity on all refs for a valid radio group
      ✓ should abort vals disabled (1 ms)

Test Suites: 1 passed, 1 total
Tests:       23 passed,
Snapshots:   0 total
Time:        21.236 s
Ran all test suites matching src/__tests__/logic/validateField.test.tsx.
```


bluebill1049 (approved):
Wow! This has been there forever... Great stuff! LGTM

PR #13662, opened 2026-08-15
🚚 perf: avoid cloning values in `unregister` without subscribers

candymask0712:
Hello, Bill 😁

`unregister` always cloned all form values, even when no values subscriber was active.

This PR uses the existing `_valuesSubscriberCount` guard, matching `onChange` and `setValues`. It avoids repeated clones when `_removeUnmounted` unregisters multiple fields.

Only the values update is skipped. The normal form-state update remains unchanged.

The unsubscribe test now uses the existing `changeEmits` helper, removing duplicate code and its tuple type error.


bluebill1049 (approved):

bluebill1049:
Good stuff 👍

PR #13668, opened 2026-08-21
fix: useWatch returns stale value on name change when new value is null

zigzagdev:
### Problem
- useWatch's synchronous name/control-change fast path used null as a sentinel meaning "no immediate value computed yet," then fell back to the possibly-stale state value whenever the computed output was null.

### Root cause
- A watched field's actual value can legitimately be null (or a compute function can return null). When name changed to such a field, the sentinel collided with that legitimate value, so the hook returned the previous field's stale value for one render — until a layout effect corrected it on the next tick. 
- This breaks the documented guarantee that callers like `useController` see the correct value immediately on the same render.

### What I Fixed
- Replaced the sentinel with an explicit boolean flag (shouldReturnImmediate), so a genuinely null/undefined computed value is never mistaken for not computed yet.

### Testing
- Added a regression test that renders a component watching field a, then re-renders it watching field c (whose value is null), asserting the render triggered by the name change already returns null.
- Verified the test fails on the pre-fix code (Received: "x") and passes after the fix.

### DIFF
--- a/src/__tests__/useWatch.test.tsx
+++ b/src/__tests__/useWatch.test.tsx
@@ -989,6 +989,41 @@ describe('useWatch', () => {
       expect(result.current).toBe('value1');
     });
 
+    it('should return the new field value synchronously on the same render even when that value is null', () => {
+      type FormValues = {
+        a: string;
+        c: string | null;
+      };
+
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
Great fix! Thanks

PR #13671, opened 2026-08-22
feat: add `shouldTouch` option to `trigger()`

zigzagdev:
## Summary

Adds a shouldTouch option to `trigger()`, mirroring the option that setValue already has. So callers can mark validated field(s) as touched without a setValue-based workaround.

### Motivation

`setValue` already supports `shouldTouch` to mark a field as touched, but `trigger()` has no equivalent. 
Anyone who wants to run validation and mark the validated field(s) as touched (e.g. validate-on-mount, multi-step form navigation, "show all errors on demand") currently has to reimplement this themselves by looping setValue(name, value, { shouldTouch: true }) over every field, which also re-writes values and triggers extra dirty/render work they don't actually want.

Maybe Related discussion
- https://github.com/react-hook-form/react-hook-form/discussions/9830

### What I have done

- Added `shouldTouch` to the `TriggerConfig` type.
- `trigger()` now marks the targeted field(s) as touched when shouldTouch: true is passed, following the same name / no-name scoping shouldFocus already uses:

```ts
trigger('name', { shouldTouch: true });          // marks `name` as touched
trigger(['a', 'b'], { shouldTouch: true });       // marks `a` and `b` as touched
trigger(undefined, { shouldTouch: true });        // marks every mounted field as touched
trigger('name');                                  // unchanged: no touched-state side effect
```

- Only `touchedFields` is written — `_formValues` is untouched, therefore there's no extra dirty-checking or value re-render cost compared to the setValue workaround.

### Test Plans

- Single field name marked as touched.
- Array of field names all marked as touched, others left untouched.
- No name (whole form) marks every mounted field as touched.
- Default behavior (no `shouldTouch`) stays a no-op for `touchedFields`, to guard against regressions.

### Test Results

```bash
Test Suites: 120 passed, 120 total
Tests:       1271 passed, 1271 total
Time:        22.455 s
Ran all test suites in 2 projects.
=== tsc ===
TSC_OK
=== eslint ===
ESLINT_OK
=== DONE ===
```

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1356,7 +1356,8 @@ export function createFormControl<
         ? {}
         : { name }),
       ...(_options.resolver || !name ? { isValid } : {}),
-      ...(options.shouldTouch
+      ...(options.shouldTouch &&
+      (_proxyFormState.touchedFields || _proxySubscribeFormState.touchedFields)
         ? { touchedFields: _formState.touchedFields }
         : {}),
       errors: _formState.errors,
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1339,11 +1339,6 @@ export function createFormControl<
 
     if (options.shouldTouch) {
       for (const fieldName of name ? fieldNames : _names.mount) {
-        // A field array's own name (e.g. `items`, registered for its own
-        // `rules`) is a container, not a leaf -- its touchedFields entry has
-        // to stay an array so `_setFieldArray` can keep shifting it on
-        // append/remove/etc. Marking it `true` here would overwrite that
-        // array and corrupt the field array's touched state.
         !_names.array.has(fieldName) &&
           set(_formState.touchedFields, fieldName, true);
       }
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -1339,7 +1339,13 @@ export function createFormControl<
 
     if (options.shouldTouch) {
       for (const fieldName of name ? fieldNames : _names.mount) {
-        set(_formState.touchedFields, fieldName, true);
+        // A field array's own name (e.g. `items`, registered for its own
+        // `rules`) is a container, not a leaf -- its touchedFields entry has
+        // to stay an array so `_setFieldArray` can keep shifting it on
+        // append/remove/etc. Marking it `true` here would overwrite that
+        // array and corrupt the field array's touched state.
+        !_names.array.has(fieldName) &&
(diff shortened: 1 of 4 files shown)


src/logic/createFormControl.ts
```diff
@@ -1337,13 +1337,22 @@ export function createFormControl<
...
     _subjects.state.next({
       ...(!isString(name) ||
       ((_proxyFormState.isValid || _proxySubscribeFormState.isValid) &&
         isValid !== _formState.isValid)
         ? {}
         : { name }),
       ...(_options.resolver || !name ? { isValid } : {}),
+      ...(options.shouldTouch
+        ? { touchedFields: _formState.touchedFields }
+        : {}),
```
bluebill1049, line 1357: thanks for the PR, i think this feature will corrupt field array data.

zigzagdev: @bluebill1049 

Fixed it. [changes](https://github.com/react-hook-form/react-hook-form/pull/13671/commits/18cd98f9382f3c423eacffd51a3915c75b61a67e) 

Skip field array names, only touch leaves now.

bluebill1049:
nice one, i think one more feedback it's missing the proxy check ` _proxyFormState.touchedFields || _proxySubscribeFormState.touchedFields`

zigzagdev:
@bluebill1049 

Fixed it. [changed](https://github.com/react-hook-form/react-hook-form/pull/13671/commits/c1fecebad0915e8f941a50f0fbefa3bf77fb9ace)

Only include `touchedFields` in the notification when both `shouldTouch` was requested and a component is actually reading `touchedFields`.

src/logic/createFormControl.ts
```diff
@@ -1337,13 +1337,29 @@ export function createFormControl<
       }
     }
 
+    if (options.shouldTouch) {
+      for (const fieldName of name ? fieldNames : _names.mount) {
+        // A field array's own name (e.g. `items`, registered for its own
+        // `rules`) is a container, not a leaf -- its touchedFields entry has
+        // to stay an array so `_setFieldArray` can keep shifting it on
+        // append/remove/etc. Marking it `true` here would overwrite that
+        // array and corrupt the field array's touched state.
```
```suggestion
```

all good we cover those comments in the test case.

bluebill1049 (dismissed):

bluebill1049:
nice! thanks for the update.

bluebill1049 (approved):

src/logic/createFormControl.ts
zigzagdev: @bluebill1049 

Thanks for deleting comments !

PR #13676, opened 2026-08-24
🪅 feat: add OpaqueTypes registry for opaque leaf types

### LINKED ISSUE #13677: allow opaque types
(this is really #12542 which is not possible to comment on now)

we have ~200 fields in our app which assign `Dayjs` objects directly. some debugging showed us that `tsc` spends ~25% of its time parsing these when we include lots of plugins and eventually this led to a `tsc` crash. 

we'd like to be able to mark certain types as opaque so that their properties aren't iterated when assigning instances to form values. 

pr with our solution here: https://github.com/react-hook-form/react-hook-form/pull/13676

modosc:
this pr allows you to mark certain types as opaque so that `react-hook-form` doesn't iterate through all the properties when present in a form.

this is a different implementation than #12308 - that one required wrapping each individual value in `Opaque<T>` but this one exposes a registry so that a client can register types to treat as opaque, eg
```ts
import type { Dayjs } from 'dayjs';

declare module 'react-hook-form' {
  interface OpaqueTypes {
    dayjs: Dayjs;
  }
}
```

we have ~200 `Dayjs` fields in our app so this is much simpler to implement.

closes #13677

### DIFF
--- a/reports/api-extractor.api.md
+++ b/reports/api-extractor.api.md
@@ -132,26 +132,26 @@ export type CustomElement<TFieldValues extends FieldValues> = Partial<HTMLElemen
 };
 
 // @public (undocumented)
-export type DeepMap<T, TValue> = IsAny<T> extends true ? any : T extends BrowserNativeObject | NestedValue | OpaqueTypes[keyof OpaqueTypes] ? TValue : T extends ReadonlyArray<infer U> ? Array<DeepMap<NonUndefined<U>, TValue> | undefined> : T extends object ? {
+export type DeepMap<T, TValue> = IsAny<T> extends true ? any : T extends BrowserNativeObject | NestedValue | OpaqueType ? TValue : T extends ReadonlyArray<infer U> ? Array<DeepMap<NonUndefined<U>, TValue> | undefined> : T extends object ? {
     [K in keyof T]: DeepMap<NonUndefined<T[K]>, TValue>;
 } : TValue;
 
 // Warning: (ae-forgotten-export) The symbol "IsPrimitiveLike" needs to be exported by the entry point index.d.ts
 //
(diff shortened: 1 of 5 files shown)


bluebill1049 (changes requested):
Nice PR 👍 

Still needs the same | OpaqueTypes[keyof OpaqueTypes] branch added to ArrayPathImpl's two checks and to DeepMap before this is complete/consistent.

modosc:
> Still needs the same | OpaqueTypes[keyof OpaqueTypes] branch added to ArrayPathImpl's two checks and to DeepMap before this is complete/consistent.

done, lmk if i missed anything else.

bluebill1049:
1. **Correctness bug** (`src/types/utils.ts:154`): `IsFlatObject` wasn't updated alongside the other type helpers to treat `OpaqueTypes[keyof OpaqueTypes]` (e.g. `Dayjs`) as a leaf type, like `Date` and `FileList`. As a result, registering any opaque type silently widens `FieldName<TFieldValues>` from a literal key union to plain `string`, removing autocomplete and type-safety for field names across the entire form—even for fields unrelated to the opaque type.

2. **Duplication** (`src/types/path/eager.ts:32` and 7 other sites across `utils.ts`, `errors.ts`, and `eager.ts`): The expression `OpaqueTypes[keyof OpaqueTypes]` is repeated verbatim in eight places instead of being factored into a shared alias. This duplication likely contributed to the `IsFlatObject` case being missed.

modosc:
i updated `IsFlatObject` and added `export type OpaqueType = OpaqueTypes[keyof OpaqueTypes];`

bluebill1049 (approved):

bluebill1049:
Nice one.

PR #13679, opened 2026-08-24
feat: return a cleanup function from `register()`'s ref for React 19

zigzagdev:
## Summary

`register()`'s ref callback now returns a cleanup function so it can take advantage of React 19's ref cleanup function support, instead of relying solely on the older called again with null convention.

### Motivation

React 19 introduced [cleanup functions for refs](https://react.dev/blog/2024/12/05/react-19#cleanup-functions-for-refs) a ref callback can return a function, and React calls that function on detach instead of calling the ref callback again with null. `register()`'s ref callback predates this and only implemented the null-argument style, so it never benefited from it.

Referred discussion #13084.

### What I have done

- Extracted the existing ref `===` null branch's unmount logic into a cleanup closure.
- Return cleanup from every return point inside the ref(elm) branch.
- Left the else (`ref === null`) branch calling the same`cleanup()` directly, unchanged, for React <19 — which ignores a ref callback's return value and still calls it again with null on unmount.
- Updated RefCallBack's type to `(instance: any) => void | (() => void)`.

### Test Plans

- `ref(elm)` returns a function (pins the new contract; fails against the previous implementation, which returned void).
- Full existing suite re-run to confirm no behavior change, including the `shouldUnregister` + real-unmount tests that already exercise `register()`'s ref teardown path end-to-end under the project's installed React 19.

### DIFF
--- a/reports/api-extractor.api.md
+++ b/reports/api-extractor.api.md
@@ -536,7 +536,7 @@ export type ReadFormState = {
 export type Ref = FieldElement;
 
 // @public (undocumented)
-export type RefCallBack = (instance: any) => void;
+export type RefCallBack = (instance: any) => void | (() => void);
 
 // @public (undocumented)
 export type RegisterOptions<TFieldValues extends FieldValues = FieldValues, TFieldName extends FieldPath<TFieldValues> = FieldPath<TFieldValues>> = Partial<{
--- a/src/__tests__/useForm/register.test.tsx
+++ b/src/__tests__/useForm/register.test.tsx
@@ -54,6 +54,24 @@ describe('register', () => {
     });
   });
 
+  it('should return a cleanup function from ref for React 19 ref cleanup support', () => {
+    const { result } = renderHook(() =>
+      useForm<{ test: string }>({
+        defaultValues: {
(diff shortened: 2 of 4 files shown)


bluebill1049:
nice pr 👍  i think this feature will need to target react-complier branch (v8) as it could be a breaking change

zigzagdev:
@bluebill1049 

Thank for reviewing ! 🙌

I changed branch to `react-complier` and rebuilt the branch on top of its current tip, so it applies cleanly now.

bluebill1049:
Thanks for the fix. I may hold off on this for a bit due to concerns about backwards compatibility, as well as a few other issues:

- `register()` and `useController` now use two divergent mechanisms (ref cleanup vs. effect cleanup) for the same “field detached” behavior. This could lead to double teardown if someone later “fixes” `useController` for symmetry.

-  The mount-flag toggle logic is now duplicated across `createFormControl.ts`, `useController.ts`, and `useFieldArray.ts`, with slightly different guards in each implementation.

-  The cleanup closure is allocated on every ref invocation, including the no-op fast path where we previously just returned.

PR #13687, opened 2026-08-29
🐞 fix(useFormState): never reconciles when an Activity subtree is hidden on its first render

giaBaoJS:
## Proposed Changes

`useFormState` never reconciles when its subtree is first rendered inside a hidden `<Activity>`. Form
state that changes while the subtree is hidden is missing once it becomes visible, and it stays
missing until something else pushes an update.

`useResyncOnReconnect` only marks itself initialized when it is handed an initializer:

```ts
// useResyncOnReconnect.ts:14
if (!_initialized.current && getInitialValue) {
  _initialized.current = true;
  _prevValue.current = cloneObject(getInitialValue());
}
```

`useFormState.ts:74-75` called it with no argument, so `_initialized` stayed `false` and the
first-connect branch added in #13683 could never fire:

```ts
// useResyncOnReconnect.ts:27-28
_connected.current || (_initialized.current && _renderCount.current > 1)
```

`useWatch.ts:331` does pass an initializer, which is why #13683 landed for `useWatch` and left the
sibling case behind. `Controller` inherits it through `useController.ts:92`, which builds `fieldState`
out of `useFormState`.

The fix hoists the `useState` initializer into `getCurrentFormState` and hands it to
`useResyncOnReconnect`, the same shape as `useWatch`. The layout effect now reuses that function
instead of redefining an identical one.

Two tests in the existing `with Activity` block, both failing on `master`:

- `useFormState` reading `isDirty`, with `setValue(..., { shouldDirty: true })` called while hidden.
  Reports `pristine` on master.
- `Controller` reading `fieldState.error`, mounted hidden while an error is set, then `clearErrors()`
  while still hidden. Keeps the stale message on master.

The `Controller` case has to go through an error object that gets replaced rather than mutated.
`setError` mutates `_formState.errors` in place, and the `useState` snapshot holds that same
reference, so a mutation is visible to the hidden subtree without any resync and hides the defect.
`clearErrors()` assigns a fresh object, which is what actually leaves the snapshot stale.

Full suite: 1284 passed / 120 suites (1282 before). `tsc --noEmit`, `tsc -p src/__typetest__` and
eslint clean.

### DIFF
--- a/src/__tests__/useFormState.test.tsx
+++ b/src/__tests__/useFormState.test.tsx
@@ -939,5 +939,136 @@ describe('useFormState', () => {
         expect(screen.getByTestId('submitting')).toHaveTextContent('no');
       },
     );
+
+    itWithActivity(
+      'should synchronize form state when an Activity subtree becomes visible for the first time',
+      () => {
+        type FormValues = { name: string };
+
+        const ActivityContent = ({
+          control,
+        }: {
+          control: Control<FormValues>;
+        }) => {
+          const { isDirty } = useFormState({ control });
+
+          return (
+            <span data-testid="dirty">{isDirty ? 'dirty' : 'pristine'}</span>
+          );
+        };
+
+        const Component = () => {
+          const { control, setValue } = useForm<FormValues>({
+            defaultValues: { name: 'initial' },
+          });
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
LGTM! thanks

PR #13688, opened 2026-08-30
🐞 fix(useFieldArray): reconcile stale fields after an Activity subtree reconnects

giaBaoJS:
## Proposed Changes

`useFieldArray` is the last value-holding hook that is not wired into the `<Activity>` reconnect path that #13633, #13683 and #13687 added for `useWatch` and `useFormState`.

A hidden `Activity` subtree keeps its state but destroys its Effects, so the `control._subjects.array` subscription in `useFieldArray` is torn down while the `fields` state is preserved. A `reset()` or `setValue()` on the array while the subtree is hidden is therefore never delivered, and when the subtree becomes visible again the hook re-subscribes without reconciling. `fields` stays stale for the rest of the form's life.

This reuses the existing `useResyncOnReconnect` helper: snapshot the array on unsubscribe, and compare against the live array before re-subscribing. It covers both the hidden-on-first-render case and the hide/change/show case.

The resync is guarded on `control` and `name` being unchanged, mirroring the same guard in `useWatch`. A new `control`/`name` is a different subscription rather than a reconnect, so resyncing there would compare the previous name's snapshot against the new name's array, always find them different, and needlessly call `setFields` plus regenerate every `id`. That matters for nested arrays with a dynamic name such as `test.${index}.keyValue`, where a parent `remove()` shifts the name of every surviving child.

Two regression tests are added, guarded by the same `Activity` feature detection already used in `useWatch.test.tsx` and `useFormState.test.tsx`.

### Performance

The subscription callback and every array method are unchanged.

Per render this adds one closure and the helper's refs. Per effect run, which is mount plus `control`/`name`/`disabled` changes only, it adds one `deepEqual`, and one `cloneObject` on cleanup.

I measured renders of a nested `useFieldArray` with a dynamic name across two parent `remove()` calls, counting renders of the nested component:

| | mount | after remove 1 | after remove 2 |
| --- | --- | --- | --- |
| master | 12 | 16 | 18 |
| this PR | 12 | 16 | 18 |

Bundle impact is +57 B gzipped on `index.cjs.js` (13903 -> 13960). `bundlewatch` still passes, though that leaves roughly 0.4KB of headroom against the 14KB budget, so it seemed worth flagging.

## Type of change

## Checklist:

### DIFF
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -144,9 +144,6 @@ export function useFieldArray<
       return;
     }
 
-    // A hidden `Activity` subtree keeps this state but destroys the
-    // subscription below, so reconcile against the live array first. A new
-    // `control`/`name` is a different subscription, not a reconnect.
     if (_prevControl.current === control && _prevName.current === name) {
       resyncIfNeeded(true, getCurrentFieldArray, (fieldValues) => {
         setFields(fieldValues);
--- a/src/useFieldArray.ts
+++ b/src/useFieldArray.ts
@@ -43,6 +43,7 @@ import type {
 } from './types';
 import { useFormControlContext } from './useFormControlContext';
 import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
+import { useResyncOnReconnect } from './useResyncOnReconnect';
 
 /**
  * A custom hook that exposes convenient methods to perform operations with a list of dynamic inputs that need to be appended, updated, removed etc. • [Demo](https://codesandbox.io/s/react-hook-form-usefieldarray-ssugn) • [Video](https://youtu.be/4MrbfGSFY2A)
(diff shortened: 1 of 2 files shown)


src/useFieldArray.ts
```diff
@@ -136,7 +144,20 @@ export function useFieldArray<
       return;
     }
 
-    return control._subjects.array.subscribe({
+    // A hidden `Activity` subtree keeps this state but destroys the
+    // subscription below, so reconcile against the live array first. A new
+    // `control`/`name` is a different subscription, not a reconnect.
```
bluebill1049:
```suggestion
```

bluebill1049 (dismissed):
LGTM

bluebill1049 (approved):

PR #13690, opened 2026-08-31
refactor: extract `_isTracked` helper for proxy/subscribe form state checks (partial)

zigzagdev:
### Motivation

`_proxyFormState.X || _proxySubscribeFormState.X` — checking whether either the render-proxy or the subscribe-proxy has a given form-state key tracked — is repeated throughout `createFormControl.ts` with only the key varying.

### What I have done

- Added a one-line `_isTracked(key: keyof ReadFormState)` helper.
- Applied it at the single-key, single-line call sites: `dirtyFields`, `isDirty`, `isValid`, `touchedFields` (6 occurrences).
- Left the remaining call sites as-is for this PR — several combine multiple keys or both proxies across multi-line `&&` chains (e.g. `isValidating`/`validatingFields` together, `isDirty`/`dirtyFields` together), and deserve closer review on their own.

#### Notes
**This is intentionally a partial/first pass.** If it lands cleanly, I'll follow up with a second PR extending `_isTracked` to the remaining multi-key/multi-line sites.

### Test Plans

- `yarn test` — full suite passes (120 suites / 1286 tests).
- Pure extraction of an existing OR-of-two-reads expression — no behaviour change.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -472,7 +472,7 @@ export function createFormControl<
         if (
           wasUnsetInFormValues &&
           _formState.isDirty &&
-          (_proxyFormState.isDirty || _proxySubscribeFormState.isDirty)
+          _isTracked('isDirty')
         ) {
           const isDirty = _getDirty();
           if (!isDirty) {
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -214,6 +214,8 @@ export function createFormControl<
   let _proxySubscribeFormState = {
     ..._proxyFormState,
   };
+  const _isTracked = (key: keyof ReadFormState) =>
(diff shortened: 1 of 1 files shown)


bluebill1049 (commented):
I think the same _proxyFormState.isDirty || _proxySubscribeFormState.isDirty pattern still appears untouched at src/logic/createFormControl.ts:473, good improvement overall.

zigzagdev:
@bluebill1049 

Thank you for reviewing !

I [fixe](https://github.com/react-hook-form/react-hook-form/pull/13690/commits/3b7ffe0ae5ae1bfadd1679de2b4e6e344444c1a7)d it.

bluebill1049 (approved):
nice, thanks for the update.

PR #13691, opened 2026-09-01
refactor: extend `_isTracked` to the remaining multi-key proxy state checks

zigzagdev:
### Motivation

Follow-up to #13690. 
That PR added an `_isTracked(key)` helper for the `_proxyFormState.X || _proxySubscribeFormState.X` pattern but only applied it at single-key, single-line call sites, leaving occurrences that combine multiple keys or span multi-line `&&` chains for later.

### What I have done

- Applied `_isTracked` to the remaining occurrences: `isValid` (with `shouldUpdateValid`), `isValidating`/`validatingFields`, `isDirty`/`dirtyFields`, and `touchedFields` — composing with `||` where more than one key is involved.
- Narrowed `_isTracked`'s key param from `keyof ReadFormState` to `keyof FormStateProxy`. 
The excluded keys (`values`, `defaultValues`, `isSubmitted`, `submitCount`) are optional, which was widening the helper's return type to include `undefined` and breaking type checking at boolean-typed call sites.
- Left the standalone `!_proxyFormState.isValid` check in `reset()` untouched — it only reads the render-proxy, not the subscribe-proxy, so it isn't the same pattern.

### Test Plans

- `yarn test` — full suite passes (120 suites / 1286 tests).
- Pure boolean-OR composition of already-existing reads — no behaviour change.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -20,6 +20,7 @@ import type {
   FieldRefs,
   FieldValues,
   FormState,
+  FormStateProxy,
   FromSubscribe,
   GetIsDirty,
   GetValuesConfig,
@@ -214,7 +215,7 @@ export function createFormControl<
   let _proxySubscribeFormState = {
     ..._proxyFormState,
   };
-  const _isTracked = (key: keyof ReadFormState) =>
+  const _isTracked = (key: keyof FormStateProxy) =>
     _proxyFormState[key] || _proxySubscribeFormState[key];
   const _subjects: Subjects<TFieldValues> = {
     array: createSubject(),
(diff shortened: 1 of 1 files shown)


bluebill1049 (commented):
_isTracked() remains single-key, so three call sites (~265, ~738, ~1003) have to compose two calls with || instead of using a variadic form. Two of those three sites check the same pair of keys but with the argument order swapped (isValidating/validatingFields vs. reversed). Not a functional bug — || is commutative — but it's an easy-to-miss inconsistency for future maintenance, and since the PR was already touching _isTracked's signature, a variadic _isTracked(...keys) using .some() was a natural, low-cost improvement.

bluebill1049:
all good, i have address comments here: https://github.com/react-hook-form/react-hook-form/pull/13699

Issue #6186, opened 2021-08-09
[Bug]: delayError shows obsoluted errors when value is change while the error is delayed


vimutti77:
### Version Number

7.12.2

### Codesandbox/Expo snack

https://codesandbox.io/s/react-hook-form-get-started-forked-z056q?file=/src/index.js

### Steps to reproduce

1. Go to https://codesandbox.io/s/react-hook-form-get-started-forked-z056q?file=/src/index.js
2. Type 'x' in the input
3. Remove 'x'
4. Type 'x' again before 500ms
5. Get error `This field is required`

### Expected behaviour

No error `This field is required`

### What browsers are you seeing the problem on?

Edge

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #13695, opened 2026-09-02
fix(delayError): cleared errors come back once the pending timer fires

Alioth-91:
## Proposed Changes

Clearing an error while its `delayError` timer is still pending puts the error back once the delay
elapses. `clearErrors`, `reset` and `unregister` all empty `_formState.errors` without touching the
scheduled callback.

`shouldRenderByError` schedules the display through `debounce` at `createFormControl.ts:590-596`:

```ts
if (_options.delayError && error) {
  delayErrorCallbacks[name] = debounce(name, () => updateErrors(name, error));
  delayErrorCallbacks[name](_options.delayError);
} else {
  clearTimeout(timers[name]);
  delete delayErrorCallbacks[name];
```

The cancellation lives in the `else` branch, so it only runs when the same field is revalidated and
comes back clean. Nothing else that empties `errors` goes through here:

```tsx
const { register, reset, formState: { errors } } = useForm<{ email: string }>({
  mode: 'onChange',
  delayError: 500,
});

// type an invalid value  -> timer scheduled for +500ms
// reset()                -> errors is {}, timer untouched
// +500ms                 -> errors.email is back, on an input that is now empty
```

`reset()` clears the input through the native `form.reset()` at `:1886-1901`, so the field ends up
empty while still carrying a validation error for the value the user just discarded.

`unregister` is the worse one. With `shouldUnregister` the field is gone by the time the callback
runs, so the error belongs to an input that is no longer on screen. `isValid` reads `true` alongside
it, since `_setValid` recomputes over `_fields` and `unregister` has already removed the field from
there, and nothing the user can do will clear the error.

The change adds `cancelDelayedError` next to `debounce` and calls it from the three paths, gated on
the `keepError` and `keepErrors` options they already honour:

```ts
const cancelDelayedError = (name: InternalFieldName) => {
  clearTimeout(timers[name]);
  delete timers[name];
  delete delayErrorCallbacks[name];
};
```

Dropping `delayErrorCallbacks[name]` is doing real work here, not tidying up. The blur path at
`:1162` reads that map and flushes whatever it finds:

```ts
const pendingDelayError = delayErrorCallbacks[name];
pendingDelayError && pendingDelayError(0);
```

With only the `clearTimeout`, a `clearErrors()` followed by the user tabbing out of the field brings
the error straight back through that flush. I checked by cutting the two `delete` lines and rerunning:
the error is gone right after `clearErrors` and back after the blur.

Existing coverage is closer than it looks. `formState.test.tsx:1150` and `:1268` do exercise
cancellation, through revalidation and through `setValue`, which is why I am reading this as three
paths that were missed rather than a deliberate choice. #6186 reported the same class for the
revalidation path in 2021 and was fixed by adding the `clearTimeout` that is still sitting in that
`else` branch today.

The blur flush and the `else` branch are untouched, and `keepError: true` / `keepErrors: true` keep
their timers along with their errors. Forms that do not pass `delayError` never populate
`delayErrorCallbacks`, so the added calls iterate an empty object.

No linked issue — found while reading `createFormControl`.

## Type of change

## Checklist:

The three new tests fail on unpatched source, one per path. The `clearErrors` and `reset` ones go on
to raise a fresh delayed error after the assertion, since "no error is rendered" would also hold if
the test never managed to schedule a timer. Full suite is 1289 passed / 120 suites. `pnpm test:type`,
`pnpm lint` and `pnpm build` pass on Node 22 / pnpm 11, with the same 562 lint warnings as master and
none in the touched files. Gzipped ESM goes from 27,706 to 27,775 bytes.

### DIFF
--- a/src/__tests__/useForm/clearErrors.test.tsx
+++ b/src/__tests__/useForm/clearErrors.test.tsx
@@ -372,4 +372,66 @@ describe('clearErrors', () => {
     // FieldB should NOT have re-rendered for clearErrors('a') - this is the bug fix!
     expect(afterClearErrorRenderB).toBe(afterSetErrorRenderB);
   });
+
+  it('should cancel a pending delayError timer so the cleared error does not come back', async () => {
+    jest.useFakeTimers();
+
+    const message = 'too long.';
+
+    const App = () => {
+      const {
+        register,
+        clearErrors,
+        formState: { errors },
+      } = useForm<{ test: string }>({
+        delayError: 500,
+        mode: 'onChange',
+      });
+
+      return (
+        <div>
+          <input {...register('test', { maxLength: 4 })} />
+          <button type="button" onClick={() => clearErrors('test')}>
+            clear
+          </button>
+          {errors.test && <p>{message}</p>}
+        </div>
+      );
+    };
+
+    render(<App />);
+
+    // Schedule a delayed error, then clear it before the delay elapses.
+    await act(async () => {
+      fireEvent.change(screen.getByRole('textbox'), {
+        target: { value: '123456' },
+      });
+    });
+
+    fireEvent.click(screen.getByRole('button'));
+
+    await act(async () => {
+      jest.advanceTimersByTime(500);
+    });
+
+    expect(screen.queryByText(message)).not.toBeInTheDocument();
+
+    // The delay itself must still work, otherwise the assertion above would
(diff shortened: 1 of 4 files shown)


bluebill1049:
thanks for the pr, i have address the issue and merged with your commits.

PR #13702, opened 2026-09-03
fix(useFormState): re-subscribe when control changes

zigzagdev:
### Motivation

The `useIsomorphicLayoutEffect` that subscribes via `control._subscribe(...)` reads `control._formState` and `control._defaultValues` too, but `control` was missing from its dependency array. 
The sibling hook `useWatch` has the equivalent effect with `control` correctly included in its deps (`[control, exact, name, disabled, resyncIfNeeded, snapshot]`), so this was an asymmetry rather than an intentional omission.

### What I have done

- Added `control` to the effect's dependency array, so the hook re-subscribes when the `control` prop's identity changes instead of continuing to read from a stale `control` reference.

### Test Plans

- Added a regression test switching `useFormState`'s `control` prop mid-test and asserting the newly-active control's changes are observed. 
Confirmed it fails without the fix (reverted `control` from the deps array locally) and passes with it restored.
- `yarn test` — full suite passes (120 suites / 1293 tests).

### DIFF
--- a/src/__tests__/useFormState.test.tsx
+++ b/src/__tests__/useFormState.test.tsx
@@ -855,6 +855,42 @@ describe('useFormState', () => {
     });
   });
 
+  it('should re-subscribe when the control prop changes identity', async () => {
+    function Consumer({ control }: { control: Control<{ test: string }> }) {
+      const { isDirty } = useFormState({ control });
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
Nice one.

PR #13711, opened 2026-09-06
🐞 fix(reset): clear isValidating and validatingFields

giaBaoJS:
## Proposed Changes

`reset()` restores every form state slice except `validatingFields` / `isValidating`, and the `keepIsValidating` flag that `KeepStateOptions` advertises for `reset` is never read (only `unregister` honours it).

Most of the time this only shows up as a brief wrong reading, but with an async resolver it becomes permanent:

1. A form with an async `resolver` subscribes to `isValid` and `isValidating`.
2. The user types, which starts a schema run and marks the field as validating.
3. Before the run settles, the form is reset and the field goes away (dialog closed, wizard step changed, section hidden).
4. `_reset` replaces `_names.mount` with an empty set, so when the in-flight run finishes, `_updateIsValidating()` has no names left to unset.

`formState.isValidating` stays `true` and `formState.validatingFields` keeps the stale entry for the rest of the form's life, so anything gated on `isValidating` (a disabled submit button, a spinner) never recovers.

Root cause: `_reset`'s final state notification lists `touchedFields`, `errors`, `isSubmitting` and friends but omits `validatingFields` and `isValidating`, so nothing clears them and `keepIsValidating` has no effect.

The fix emits both from `_reset`, gated on `keepIsValidating` in the same shape as the neighbouring `keepTouched` / `keepErrors` entries.

## Tests

Two tests in `src/__tests__/useForm/reset.test.tsx`:

- `should clear isValidating when reset is called while a validation is pending` reproduces the scenario above and asserts the status paragraph reads `status:idle` and `validatingFields` is empty both right after the reset and after the in-flight resolver settles. Without the source change it fails with `Expected: "status:idle" / Received: "status:validating"`.
- `should keep isValidating when reset is called with keepIsValidating option` covers the other branch: the flag and the tracked field survive the reset.

Full suite (1300 tests), `tsc --noEmit`, eslint and prettier all pass.

## Type of change

## Checklist:

### DIFF
--- a/src/__tests__/useForm/reset.test.tsx
+++ b/src/__tests__/useForm/reset.test.tsx
@@ -2288,4 +2288,135 @@ describe('reset', () => {
 
     jest.useRealTimers();
   });
+
+  it('should clear isValidating when reset is called while a validation is pending', async () => {
+    let resolveResolver: (() => void) | undefined;
+
+    const App = () => {
+      const [visible, setVisible] = React.useState(true);
+      const {
+        register,
+        reset,
+        formState: { isValid, isValidating, validatingFields },
+      } = useForm<{ test: string }>({
+        defaultValues: { test: '' },
+        resolver: async (values) => {
+          await new Promise<void>((resolve) => {
+            resolveResolver = resolve;
+          });
+          return { values, errors: {} };
+        },
+      });
+
+      return (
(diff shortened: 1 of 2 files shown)


bluebill1049:
ty fix merged

PR #13715, opened 2026-09-07
fix(reset): clear stale `isValidating/validatingFields`, honor `keepIsValidating`

zigzagdev:
### Motivation

`_reset`'s final state notification omits `validatingFields`/`isValidating`, so `reset()` never clears them and `keepIsValidating` has no effect.

`_reset` also empties `_names.mount`. When an in-flight async resolver's completion callback finally runs, it has nothing left to unset.
`isValidating` stays stuck `true` forever, leaving things like a submit button permanently disabled.

### What I have done

- Emit `validatingFields`/`isValidating` from `_reset`'s final state notification, gated on `keepIsValidating`, in the same shape as the neighbouring `keepTouched`/`keepErrors` entries.

### Test Plans

- Added two tests to `src/__tests__/useForm/reset.test.tsx`:
  - `should clear isValidating when reset is called while a validation is pending` — reproduces the scenario above; confirmed it fails without the fix (status never leaves `"validating"`) and passes with it restored.
  - `should keep isValidating when reset is called with keepIsValidating option`.

### DIFF
--- a/src/logic/createFormControl.ts
+++ b/src/logic/createFormControl.ts
@@ -2036,6 +2036,12 @@ export function createFormControl<
         ? _formState.isSubmitSuccessful
         : false,
       isSubmitting: false,
+      validatingFields: keepStateOptions.keepIsValidating
+        ? _formState.validatingFields
+        : {},
+      isValidating: keepStateOptions.keepIsValidating
+        ? _formState.isValidating
+        : false,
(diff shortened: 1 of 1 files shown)


bluebill1049:
could u resolve the conflicts? ty

zigzagdev:
@bluebill1049 

Sorry, I close this PR close.

Because fully superseded by #13714 and #13722.
Rebased onto master to confirm and find zero diff remains.

PR #13723, opened 2026-09-09
fix(unregister): recompute isValidating after clearing `validatingFields`

zigzagdev:
### Motivation

`unregister` clears `validatingFields` but never recomputed `isValidating`. 
Unregistering a field mid-validation could leave `isValidating` stuck  type`true` forever.

For example, 

```
a dynamic form (`useFieldArray`, or a field conditionally shown/hidden) has a field with an async `validate` (e.g. a server-sideduplicate-email check). 
The user types, validation starts (`isValidating: true`), then the row is removed before the response comes back. `isValidating` stays stuck `true` forever — a submit button gated on it never re-enables, or a loading spinner never goes away.
```

### What I have done

After clearing `validatingFields`, recompute `isValidating` in the same call, gated on `keepIsValidating`. 
Same pattern as the existing `keepDirty`/`isDirty` override right above it.

### Test Plans

- Added two tests: one reproduces the stuck `isValidating` (fails without the fix, passes with it), one covers `keepIsValidating`. 
- Full suite passes (1306 tests).

### DIFF
--- a/src/__tests__/useForm/unregister.test.tsx
+++ b/src/__tests__/useForm/unregister.test.tsx
@@ -296,4 +296,106 @@ describe('unregister', () => {
       expect.any(Object),
     );
   });
+
+  it('should recompute isValidating after unregistering a field with a pending validation', async () => {
+    let resolveValidate: (value: boolean) => void;
+
+    const App = () => {
+      const {
+        register,
+        unregister,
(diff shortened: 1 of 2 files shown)


bluebill1049 (approved):
LGTM!

PR #13729, opened 2026-09-10
fix: remove() leaves an empty touchedFields entry after its last row is removed

soroush5:
Blur an input inside a field array, then remove the row(s) so no touched row is left.
`formState.touchedFields` keeps a `{ items: [] }` ghost and `getFieldState('items').isTouched`
stays `true`, even though nothing is touched anymore.

Repro: two-row array, blur row 0, `remove(0)` (or blur both rows and `remove()`).
Before: `touchedFields` is `{ test: [] }`. After: `{}`, matching what `errors` already
does via `unsetEmptyArray` and what a fresh form reports.

The fix is one line in `_setFieldArray`: run the same `unsetEmptyArray` cleanup for
`touchedFields` that already exists for `errors` right above it.

Verification: full jest suite passes (122 suites / 1347 tests), `tsc` typetests pass,
eslint and prettier clean. Added a regression test in
`src/__tests__/useFieldArray/remove.test.tsx` and updated the old `{ test: [] }`
expectation to `{}`.

### DIFF
--- a/src/__tests__/useFieldArray/remove.test.tsx
+++ b/src/__tests__/useFieldArray/remove.test.tsx
@@ -427,7 +427,51 @@ describe('remove', () => {
 
     fireEvent.click(screen.getByRole('button', { name: 'remove all' }));
 
-    expect(touched).toEqual({ test: [] });
+    expect(touched).toEqual({});
+  });
+
+  it('should drop the touched entry once its last touched row is removed', () => {
+    let touched: any;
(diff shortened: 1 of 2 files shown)


soroush5:
Withdrawing this one — the Playwright run caught a real regression on my change: `useFieldArrayNested` expects empty `keyValue: []` touched markers to survive remove-all, so these markers are intentional and my blanket cleanup was wrong. Sorry for the noise; closing.

PR #13745, opened 2026-09-15
fix(onChange): ignore stale built-in validation result after reset

giaBaoJS:
## Proposed Changes

`onChange` has two validation branches. The resolver branch captures
`_resetCallId` before awaiting and bails when `reset()` bumped it (#13744).
The built-in validation branch does the same kind of async work and had no
guard, so a `validate` promise that settles after `reset()` writes its
discarded result into the form that `reset()` already cleaned.

Minimal repro (`mode: 'onChange'`, no resolver):

```tsx
const { register, reset, formState: { errors } } = useForm({
  mode: 'onChange',
  defaultValues: { test: '' },
});

<input {...register('test', { validate: () => slowPromise })} />
<button onClick={() => reset(undefined, { keepValues: true })}>reset</button>
```

1. Type into the input. `validate` starts and stays pending.
2. Click reset. `errors` is empty, as expected.
3. `validate` resolves with a message. `errors.test` comes back.

Two distinct states are affected:

- `errors`: `src/logic/createFormControl.ts:1299-1308` awaits `validateField`,
  then `shouldRenderByError` at line 1335 sets the error. `keepValues` (or a
  `reset()` that restores the same value) is what exposes it, because
  `_updateIsFieldValueUpdated` otherwise masks the write by bailing when the
  in-flight value no longer matches the current one.
- `isValid`: `src/logic/createFormControl.ts:1317` awaits
  `executeBuiltInValidation` for the `isValid` verdict. That is a second
  suspension point, and a `reset()` landing there leaves `isValid` set from a
  result that was meant to be discarded.

The fix captures `_resetCallId` once at the top of the branch and returns at
both await points, matching the existing guards in
`executeSchemaAndUpdateState`, `trigger`, `handleSubmit` and the resolver
branch of `onChange`.

Tests, in `src/__tests__/useForm/reset.test.tsx`:

- a `validate` promise settling after `reset({ keepValues: true })` no longer
  resurrects the error,
- an `isValid` pass spanning `reset()` no longer overwrites `isValid`,
- control: with no `reset()`, a slow `validate` result is still applied.

Reverting only the source change turns the first two red and leaves the
control green.

Scope note: `props.validate` (the form-level `validate` option) has the same
shape of hole through `validateForm`, which calls `setError(FORM_ERROR_TYPE)`
after awaiting `props.validate`. A form-level validate that resolves to a
message after `reset()` will resurrect the root error the same way. That one
is shared by `trigger`, `handleSubmit` and `executeBuiltInValidation`, so
guarding it is a wider change than this PR and I left it untouched.

No issue filed for this; found by reading the two branches side by side.

## Type of change

## Checklist:


src/logic/createFormControl.ts
```diff
@@ -1306,6 +1307,13 @@ export function createFormControl<
             _options.shouldUseNativeValidation,
           )
         )[name];
+
+        // A reset() that lands while this validation is pending has already
+        // cleared the state this result describes.
+        if (resetCallId !== _resetCallId) {
```
```suggestion
        if (resetCallId !== _resetCallId) {
```

bluebill1049 (approved):

PR #13750, opened 2026-09-17
fix(useFieldArray): do not mark form dirty when array default is null

### LINKED ISSUE #13749: useFieldArray over a null default marks the form dirty on mount
Hey! Found that mounting useFieldArray over an array field whose default is null marks the form dirty right away, even though the user touched nothing. Any form fed by an API that returns null for an empty list shows unsaved-changes state on load. Happy to PR a fix that only initializes the field array when the value is actually undefined if that helps.

soroush5:
Mounting useFieldArray over an array field whose default is null marked the form dirty immediately, so API-fed forms showed unsaved-changes state on load with zero user input. The mount guard now only initializes when the value is actually undefined.

Verified: new regression test fails before and passes after, field-array neighbors green.

Fixes #13749


bluebill1049 (approved):

soroush5:
Thanks for the review and merge.

PR #13752, opened 2026-09-17
fix(setValues): deep merge nested partials instead of dropping siblings

### LINKED ISSUE #13751: setValues with nested partials wipes untouched sibling values
Hey! Found that setValues({ user: { profile: { firstName: 'John' } } }) wipes untouched nested siblings (lastName, the whole address subtree) from form state, while the inputs still display them — so submit silently loses data. Bulk-updating one section from server data hits this every time. Happy to PR a fix that deep-merges plain-object subtrees if that helps.

soroush5:
setValues with nested partials replaced whole subtrees in form state, silently wiping untouched values while the inputs still showed them — submit then lost data. Plain-object subtrees now merge recursively; arrays, Dates and other values still replace by reference.

Verified: new tests fail before (2 failed) and pass after (23 green), full suite green.

Fixes #13751


soroush5:
You are right, nested partials are outside the setValues contract. Closing this, thanks for explaining.

PR #13758, opened 2026-09-18
fix: clear validation rules removed from register options at runtime

### LINKED ISSUE #13757: Validation rule removed at runtime is still enforced
Hey! Found that removing a validation rule at runtime has no effect — `register('c', cond ? {required: '...'} : {})` (or Controller `rules`) keeps enforcing the stale rule after `cond` turns false, so the form stays unsubmittable with an error on a visibly optional field. Tightening works, removal never takes effect. Happy to PR a fix if that helps.

soroush5:
Removing a validation rule at runtime had no effect — re-registration only added keys, never cleared them. Rules absent from the incoming options are now cleared, so optional fields stop blocking submit.

Verified: 3 new tests fail before and pass after, full suite green.

Fixes #13757


bluebill1049 (approved):

soroush5:
Thanks for merging.

PR #13761, opened 2026-09-18
fix: ignore stale built-in validation result after reset in handleSubmit

### LINKED ISSUE #13760: handleSubmit ignores reset() during async validation
Hey! Found that if reset() runs while handleSubmit is awaiting an async field-level validate, the late result still lands — either a stale error in formState.errors (plus a stale onInvalid) or onValid called with the discarded pre-reset values. Happy to PR a fix if that helps.

soroush5:
If reset() runs while handleSubmit awaits an async field validate, the late result still landed as a stale error or a stale onValid call. The in-flight result is now discarded once a reset happened, same guard style as #13733.

Verified: 2 new tests fail before and pass after, full suite green.

Fixes #13760


bluebill1049:
I think there’s still a stale validation case here with form-level `props.validate`.

`executeBuiltInValidation` calls `validateForm()` before the new `shouldAbort` check, and `validateForm()` can call `setError` / `clearErrors` after async validation resolves.

So if `reset()` happens while `props.validate` is pending, the validation result could still write stale errors after the reset, even though `onValid` / `onInvalid` are correctly skipped.

Could we also cover the form-level async `validate` case and make sure we abort before applying its result?

soroush5:
Good catch, you are right. Pushed a follow-up: shouldAbort is now threaded into validateForm too, so a late props.validate result after reset no longer touches errors. Added a form-level regression test (fails before, passes after), full suite green.

bluebill1049:
nice, could you resolve the conflicts?

soroush5:
Rebased and reworked on the new architecture: the guard now lives inside executeBuiltInValidation right after each validateField, so a late field result after reset never reaches formState.errors either. Verified the gap first: the 3 regression tests fail on current master (stale onInvalid still fires), all green with this fix, full suite 1396 passed.

bluebill1049 (approved):

soroush5:
Thanks for merging.

PR #13765, opened 2026-09-19
fix: revalidate deps dependents on setValue with shouldValidate

### LINKED ISSUE #13764: setValue with shouldValidate does not revalidate deps dependents
Hey! Found that setValue(src, v, { shouldValidate: true }) never revalidates deps dependents — typing into src clears a confirm-password mismatch, but a programmatic set (like a generate-password button) leaves a ghost error and a stuck isValid-gated save button. Happy to PR a fix if that helps.

soroush5:
Typing into a field revalidated its deps dependents, but a programmatic setValue with shouldValidate left ghost errors and a stuck isValid state. Deps are now revalidated on the setValue path too, mirroring the onChange behavior.

Verified: 2 new tests fail before and pass after, full suite green, plus a real Chromium check.

Fixes #13764


bluebill1049 (approved):

soroush5:
Thanks for merging.

PR #13767, opened 2026-09-19
fix: clear stale field array root error once an op satisfies the rule

### LINKED ISSUE #13766: Field array root error stays after append satisfies the rule
Hey! Found that an array-level useFieldArray rule (like minLength) leaves a ghost errors.<name>.root message after append satisfies the rule — isValid heals to true but the stale message stays visible. Everyday case is an at-least-N-items list where the user adds rows to fix the error. Happy to PR a fix if that helps.

soroush5:
Array-level rules left a ghost root error after an op satisfied them, while isValid healed — a visible broken invariant. The now-passing result now clears just the stale root, row errors survive.

Verified: 3 new tests fail before and pass after, full suite green, real Chromium both directions.

Fixes #13766


bluebill1049 (approved):

bluebill1049:
Thanks for all the PRs and contributions lately! Really appreciate the time and effort you've put into tracking down and fixing these issues. 🙏

soroush5:
Thanks for merging.

PR #13769, opened 2026-09-19
fix: cancel nested delayError timers on resetField parent reset

### LINKED ISSUE #13768: resetField on parent leaves nested delayError timers alive
Hey! Found that resetField('parent') leaves nested delayError timers running — a stale error pops back after a per-section reset and blocks submit while isValid reads true. Clearing or unregistering the whole tree cancels them, but resetField on a parent does not. Happy to PR a one-line fix if that helps.

soroush5:
resetField on a parent left nested delayError timers alive, so a stale error popped back after a per-section reset and blocked submit while isValid read true. The tree-cancel used by clear/unregister now runs here too.

Verified: new test fails before and passes after, full suite green, real Chromium both directions.

Fixes #13768


bluebill1049 (approved):

soroush5:
Thanks for merging.

Issue #12931, opened 2025-07-03
feature: Built-in form data extraction strategies


aress31:
**Is your feature request related to a problem? Please describe.**

Currently, when working with forms that need different data extraction strategies (e.g., only `dirtyFields` for PATCH operation, or all fields for POST operations), developers have to manually implement the logic to extract the relevant data from `formState`.

**Describe the solution you'd like**

Add built-in methods to extract form data based on different strategies:

```js
const {
  control,
  formState: { isDirty, isValid },
  handleSubmit,
  reset,
  getFormData, // New method with strategy support
} = useForm({ values: { records: values } });

// Usage examples:
const onSubmit = async (data) => {
  // Get only dirty fields
  const dirtyData = getFormData('dirty');
  
  // Get only touched fields  
  const touchedData = getFormData('touched');
  
  // Get all fields (default behavior)
  const allData = getFormData('all');
  
  // Get only valid fields
  const validData = getFormData('valid');
  
  // Custom strategy with callback
  const customData = getFormData((fieldName, fieldState) => {
    return fieldState.isDirty && !fieldState.error;
  });
};
```

**Describe alternatives you've considered**

Currently we have to manually extract data like this:

```js
// Manual dirty fields extraction
const allDirtyFields = Object.values(dirtyFields.records || {}).flatMap(
  (fieldStates) =>
    Object.keys(fieldStates).filter((key) => fieldStates[key])
);

const valuesToUpdate = Object.fromEntries(
  allDirtyFields.map((field) => [field, updatedRecords[0][field]])
);
```

**Additional context**

This feature would be particularly useful for:

- Bulk edit scenarios where you only want to update changed fields
- Partial updates to avoid validation errors on unchanged required fields
- Performance optimisation by sending minimal data
- Different API requirements (some APIs expect only changed data, others expect all data)

bluebill1049:
Something we can consider with `getValues` API

```tsx
const {
  getValues,
} = useForm()

// flags: more aligned with existing API
getValues('records', {
  touchedFields: true,
  dirtyFields: true,
})

getValues({
  touchedFields: true,
  dirtyFields: true,
})

// variants
getValues('records', {
  returnTypes: 'touchedFields', // 'all' default 
})
```

aress31:
@bluebill1049, you marked this as completed but I do not see an associated PR. What is the current status for this ticket?

bluebill1049:
refer here:
[...]

PR #13773, opened 2026-09-20
fix(getValues): extract only the marked entries of a field array

giaBaoJS:
## Proposed Changes

`getValues(names, { dirtyFields: true })` returns every row of a field array, not just the changed ones.

Repro: a `records` array of `{ name, note }` rows, edit only `records.1.name`, then call `getValues(undefined, { dirtyFields: true })`. Both rows come back with both keys, so a PATCH payload built from it sends the whole array. The docs say the option "Returns only dirty fields" (https://react-hook-form.com/docs/useform/getvalues), and #12931, the request this API came from, asks for it specifically to build partial updates out of `dirtyFields.records`.

Cause: `src/utils/extractFormValues.ts:14` gates the recursive branch on `isObject(fieldState)`, and `isObject` is false for arrays (`src/utils/isObject.ts:8`). An array marker falls into `else if (fieldsState[key])` and the value is copied through whole. Nested plain objects are filtered correctly, which the existing unit test in that file already pins down.

The fix recurses into array markers and builds the container as an array, so row indexes line up with the marker map. `dirtyFields` keeps holes for clean rows on purpose, see `clearDirtyField` in `src/logic/getDirtyFields.ts`. The `isObject(nestedFieldsState)` guard goes with it: it was always true before and would now reject arrays.

Boundary: `getValues('records.0.name', { dirtyFields: true })` now returns `undefined` for a clean row instead of the value, which is what already happens for a clean key under a plain object. Not touched here: `getValues` dispatches with `config.dirtyFields ? dirtyFields : touchedFields` (`createFormControl.ts:1486`), so `{ dirtyFields: false }`, which `GetValuesConfig` accepts, silently reads touched fields. That is a flaw in the flag dispatch, not in the extraction.

Tests: a unit case in `src/__tests__/utils/extractFormValues.test.ts` and a `useFieldArray` case in `src/__tests__/useForm/getValues.test.tsx`. Both fail on master.

## Type of change

## Checklist:


src/utils/extractFormValues.ts
```diff
@@ -4,22 +4,23 @@ export default function extractFormValues<
   T extends object,
   K extends Record<string, unknown>,
 >(fieldsState: T, formValues: K) {
-  const values: Record<string, unknown> = {};
+  // keep array indexes so the caller can tell which row a value came from
```
bluebill1049:
```suggestion
```

bluebill1049 (dismissed):

bluebill1049 (approved):

PR #13777, opened 2026-09-22
fix(validate): run the latest form level validate function

### LINKED ISSUE #12235: issue: stale closure of register option onChange.
### Version Number

7.53.0

### Codesandbox/Expo snack

https://codesandbox.io/p/sandbox/condescending-mopsa-66dxll?file=%2Fsrc%2FApp.tsx%3A31%2C22

### Steps to reproduce

1. Go to [code sandbox](https://codesandbox.io/p/sandbox/condescending-mopsa-66dxll?file=%2Fsrc%2FApp.tsx%3A31%2C22)
2. Click on radio A, B, C, D
3. Regardless of what you clicked, D was printed on the console.

https://github.com/user-attachments/assets/5a7f754a-9b57-4173-9008-1d903d93c672

### Expected behaviour

If you click A, A is displayed on the console.
If you click B, B is displayed on the console.
...

It looks like there is a stale closure. I wonder if it is the intended action. 
I know that you can approach and solve it with an `event`.

Chrome

### Relevant log output

_No response_

### Code of Conduct

dfedoryshchev:
## Proposed Changes

The form-level `validate` option is read off the props object `createFormControl` was constructed with, so the function that actually runs is the one from the first render and never the current one.

`useForm` refreshes `control._options = props` on every render (`src/useForm.ts:81`) and the `_options` setter merges (`src/logic/createFormControl.ts:2300`), which is how every other option stays current. `validate` is the one option that bypasses it: `src/logic/createFormControl.ts:711` and `:713` test and await `props.validate`, and `:779`, `:1226` and `:1286` gate on it. You closed #12235, a stale closure in a `register` option, with "use e.target.value instead"; that answer does not carry here, because a form-level validator reading component state has no equivalent value to read off the event, and because this one is an internal inconsistency rather than a request to change how closures work.

The visible consequence is a stale closure. A validator written inline, which is the shape in the docs and in all four existing tests, captures component state on the first render and keeps reading those values forever; the `formValues` argument is fresh, so only what the closure holds goes wrong. It reaches the user through `setError(FORM_ERROR_TYPE)` / `clearErrors(FORM_ERROR_TYPE)` at `:727`-`:750`, through `context.valid` into `formState.isValid`, and through the `onValid` / `onInvalid` branch of `handleSubmit`. The gate at `:1226` is the harder edge: a `validate` supplied after the first render never un-gates `onChange`, so it never runs at all.

Existing coverage misses it because all four form-level `validate` tests define the validator inline but read only the `formValues` argument, so a frozen closure and a fresh one behave identically.

The fix reads `_options.validate` at all five sites, which is the same value on the first render and the current one afterwards.

- `should run the latest validate function after a re-render` fails on master with `Expected: "foo is banned" / Received: ""` and passes after.
- `should run a validate function supplied only after the first render` fails on master with `Expected: "form is invalid" / Received: ""` and passes after; `validate` is never called at all.
- `should keep running a stable validate function on every change` is a control for the common case and passes both before and after.

One behaviour change worth flagging: `useForm({ formControl, validate })`, where the control was built by a `createFormControl()` call that did not itself receive `validate`, currently ignores the validator and will now run it. That follows from `_options` being the merged view, and it is the same path every other option already takes, but it is a real change rather than a no-op and I am happy to narrow it if you would rather keep that case as it is.

Related but deliberately not bundled: `props.shouldUnregister` at `:514` and `props.shouldUseNativeValidation` at `:852` read the frozen object the same way while their siblings elsewhere in the file read `_options`. Happy to follow up separately.

Full suite green (124 suites, 1412 tests), `tsc --noEmit`, `tsc -p src/__typetest__/tsconfig.json` and eslint all clean.


bluebill1049 (commented):
Looks good overall, but I think we need coverage for removing `validate` after it has been set.

`_options` is merged with the latest props, so if a later render omits `validate` (e.g. `...(enabled ? { validate } : {})`), the previous `_options.validate` remains and will continue running.

Could we add the reverse of the new conditional-validator test and ensure disabling/removing `validate` stops form-level validation?

dfedoryshchev:
@bluebill1049 resolved, have a look.
Btw, do you mind connecting on [LinedkIn](https://www.linkedin.com/in/dife/) ? I wanted to ask you couple more questions, but I did not find any contact details of yours.

bluebill1049:
> @bluebill1049 resolved, have a look. Btw, do you mind connecting on [LinedkIn](https://www.linkedin.com/in/dife/) ? I wanted to ask you couple more questions, but I did not find any contact details of yours.

Ah, I haven’t used LinkedIn in ages! Feel free to drop any questions here instead. And thanks again for all the awesome contributions — really appreciate it!

src/useForm.ts
```diff
@@ -78,7 +80,8 @@ export function useForm<
   }
 
   const control = _formControl.current.control;
-  control._options = props;
+  // the setter merges, so an omitted `validate` would keep the previous one running
```
```suggestion
```

bluebill1049 (approved):

PR #13779, opened 2026-09-23
fix(setValue): clear inputs dropped by a parent-path value

### LINKED ISSUE #13778: setValue on a parent path leaves removed child inputs showing stale values
Calling setValue with an object on a parent path replaces that subtree in form state, but inputs for children missing from the new value keep showing their old text.

Repro: register a.b and a.c (defaults one/two), then setValue('a', {}). State becomes { a: {} } yet both inputs still display the old values. Same with a partial value: setValue('a', { b: 'CHANGED' }) drops c from state while its input still shows two. What gets submitted no longer matches what the user sees.

Expected: inputs for children absent from the new value clear, the same as calling setValue on each of those leaves with undefined.

soroush5:
## Proposed Changes

setValue with an object on a parent path replaces that subtree in form state, but inputs for children missing from the new value kept showing stale text, so submitted data no longer matched the screen. After the change those inputs clear, the same as setting each dropped leaf to undefined (dirty, touch, and validation options keep flowing through the leaf path).

Fixes #13778

## Type of change

## Checklist:


soroush5:
On reflection this one needs a maintainer semantics decision first (group-sync behavior for parent-path setValue), so I am withdrawing it rather than push something the team may not want. Happy to rework if you see a direction worth taking.

